PR #52, opened 2020-06-01
Feature/add active storage

Rhian:
documents

### Description of change

Added ActiveStorage integration, updated Planning Application overview page and added tests including system tests for new Manage Documents page.

### Story Link

https://www.pivotaltracker.com/n/projects/2441805/stories/172844513

### Decisions 

This uses local storage of images for development and will use S3 for production.

### Known issues 

Will need update to AWS infrastructure and to Terraform to allow for S3 integration, and also when we receive the RIPA API, a download image method. It has been tested on my Unboxed AWS account downloading from one S3 bucket and uploading to another and works fine.


spec/system/drawings/index_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      page.driver.browser.switch_to.window(page.driver.browser.window_handles.last)
+      expect(current_url).to include("/rails/active_storage/")
+    end
+
+    scenario "User can log out from drawings page" do
+      click_button "Log out"
+
+      expect(page).to have_current_path(/sign_in/)
+      expect(page).to have_content("You need to sign in or sign up before continuing.")
+    end
```
NeilvB, line 53: I like these specs 👍 

I wondered if we might want to include a simple spec for checking our different, simpler layout for this page. Maybe that could be done using the CSS classes of those elements, or the text content of one of their accordions.

app/models/document.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Document < ApplicationRecord
+  belongs_to :planning_application
+
+  has_one_attached :plan
```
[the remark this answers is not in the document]
Rhian: We can implement categories as part of the tagging story. Not in scope for pt-172844513

Dockerfile
```diff
@@ -26,6 +26,9 @@ RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
     echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
     apt-get update && apt-get install -y yarn
 
+# Install MuPdf for PDF previews
+RUN apt-get install mupdf mupdf-tools
```
EGiataganas, line 30: Is that the dev version? Also, you will need to update the [Dockerfile.production][1] to include it as well 

[1]: https://github.com/unboxed/bops/blob/master/Dockerfile.production

Gemfile
```diff
@@ -15,6 +15,9 @@ gem "puma", "~> 4.1"
 gem "pundit"
 gem "rails", "~> 6.0.2"
 gem "webpacker", "~> 4.0"
+gem "aws-sdk-s3", require: false
+gem "mini_magick"
+gem "image_processing", "~> 1.2"
```
EGiataganas, line 20: can you please sort them by alphabetical order?

db/schema.rb
```diff
@@ -68,6 +89,14 @@
     t.index ["user_id"], name: "index_decisions_on_user_id"
   end
 
+  create_table "documents", force: :cascade do |t|
+    t.string "name", default: "f", null: false
```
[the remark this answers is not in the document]
Rhian: We can do this when we implement tagging etc

spec/system/drawings/index_spec.rb
Rhian: I've added a spec that checks the planning application info accordion is not present

Dockerfile
Rhian: Thanks for the reminder - will do that now!

PR #72, opened 2020-06-12
Feature/pt 172451141/planning applications api

EGiataganas:
### Description of change

Simple API to pull out planning applications decisions

### Story Link

https://www.pivotaltracker.com/story/show/172451141

### DIFF
--- a/app/controllers/api/v1/planning_applications_controller.rb
+++ b/app/controllers/api/v1/planning_applications_controller.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class Api::V1::PlanningApplicationsController < Api::V1::ApplicationController
+  before_action :set_cors_headers, only: %i[index], if: :json_request?
+
+  def index
+    @planning_applications = PlanningApplication.determined
+
+    respond_to(:json)
+  end
+end
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -2,4 +2,8 @@
 
 class Site < ApplicationRecord
   has_many :planning_applications, dependent: :restrict_with_exception
+
+  def full_address
+    "#{address_1}, #{town}, #{postcode}"
+  end
 end
--- a/app/views/api/v1/planning_applications/index.json.jbuilder
+++ b/app/views/api/v1/planning_applications/index.json.jbuilder
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+json.data @planning_applications.each do |planning_application|
+  json.partial! "show.json.jbuilder", planning_application: planning_application
+end
--- a/spec/requests/planning_application_list_spec.rb
+++ b/spec/requests/planning_application_list_spec.rb
@@ -1,0 +1,98 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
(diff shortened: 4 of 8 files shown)


spec/requests/planning_application_list_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+  let(:access_control_allow_origin) { response.headers['Access-Control-Allow-Origin'] }
+  let(:access_control_allow_methods) { response.headers['Access-Control-Allow-Methods'] }
+  let(:access_control_allow_headers) { response.headers['Access-Control-Allow-Headers'] }
+  let(:reviewer) { create :user, :reviewer }
+
+  before do
+    create(:decision, :granted, user: reviewer)
+  end
+
+  describe "format" do
```
NeilvB, line 15: We could move the header variables under this `describe` block to make it a bit clearer where they're used. Having them at the top level makes it look a bit like they're relevant for all of the specs.

spec/requests/planning_application_list_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    let(:data) { json["data"] }
+
+    it "returns an empty response if no petitions are public" do
+      get "/api/v1/planning_applications.json"
+
+      expect(response).to be_successful
+      expect(data).to be_empty
+    end
+
+    it "returns a list of serialized petitions in the expected order" do
```
NeilvB, line 41: We should update this and the above spec to refer to `planning applications` instead.

spec/requests/planning_application_list_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      planning_application_1 = create(:planning_application, :determined)
+      planning_application_2 = create(:planning_application, :determined)
```
NeilvB, line 43: It would be good to also include the creation of a planning application which is NOT in the determined status in our test setup, because we want to be certain that our response only contains determined ones.

app/models/site.rb
```diff
@@ -2,4 +2,8 @@
 
 class Site < ApplicationRecord
   has_many :planning_applications, dependent: :restrict_with_exception
+
+  def full_address
+    "#{address_1}, #{town}, #{postcode}"
+  end
```
NeilvB, line 8: Did we try using a view helper method for this? I'd prefer this string formatting to stay in our view layer.

app/views/api/v1/planning_applications/index.json.jbuilder
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+json.data @planning_applications.each do |planning_application|
+  json.partial! "show.json.jbuilder", planning_application: planning_application
```
NeilvB, line 4: Nice use of a partial 👍

app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class Api::V1::PlanningApplicationsController < Api::V1::ApplicationController
```
NeilvB, line 3: I'm glad we have versioning 👍

PR #76, opened 2020-06-22
Refactor comment box

Rhian:
comment

### Description of change

Small suggested refactor of comment box on Assessment view using Stimulus to conditionally display only one div instead of two and to attach the correct IDs when yes and no are clicked.


EGiataganas:
This looks good, by refactoring it to use stimulus it would be great if we have one comment field in the db! What do you think?

EGiataganas:
Closing in favour of #106

PR #81, opened 2020-06-23
Notify Integration part I of II

NeilvB:
### Description of change

This PR adds authorisation for users being able to set application statuses.

This is a bit of a detour from the main point of the story, which is to integrate with Notify. But since that story involves changing how we detect determined status from using the decision to the planning application, and the user submitting a 'Determined' application, this felt like a reasonable time to do this.

### Story Link

https://www.pivotaltracker.com/n/projects/2441805/stories/172451136

### DIFF
--- a/app/policies/planning_application_policy.rb
+++ b/app/policies/planning_application_policy.rb
@@ -18,4 +18,18 @@ class PlanningApplicationPolicy < ApplicationPolicy
   def update?
     super || signed_in_editor?
   end
+
+  def unpermitted_statuses
+    PlanningApplication.statuses.keys - permitted_statuses
+  end
+
+  def permitted_statuses
+    if @user.assessor?
+      [ "awaiting_determination" ]
+    elsif @user.reviewer? || @user.admin?
+      [ "determined" ]
+    else
+      []
+    end
+  end
 end
--- a/app/views/api/v1/planning_applications/_show.json.jbuilder
+++ b/app/views/api/v1/planning_applications/_show.json.jbuilder
@@ -6,5 +6,5 @@ json.site_address planning_application.site.full_address
 json.application_type t(planning_application.application_type)
 json.summary_of_proposal planning_application.description
 json.received_date planning_application.created_at.iso8601
-json.decided_at planning_application.reviewer_decision.decided_at&.iso8601
+json.determined_at planning_application.determined_at.iso8601
 json.status planning_application.reviewer_decision.status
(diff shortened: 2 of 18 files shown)


app/policies/planning_application_policy.rb
```diff
@@ -18,4 +18,18 @@ def edit?
...
   end
+
+  def unpermitted_statuses
+    PlanningApplication.statuses.keys - permitted_statuses
+  end
+
+  def permitted_statuses
+    if @user.assessor?
+      [ "awaiting_determination" ]
+    elsif @user.reviewer? || @user.admin?
```
Rhian, line 29: Are we sure that the reviewer should not be permitted to do this? We would normally consider a manager to have a super-set of permissions rather than a different set of permissions? (I'm aware that this is more of a product question than a code question).

NeilvB: I guess the way I was thinking about this was that we don't currently have a UI for the manager to push something from in assessment to awaiting determination. So I wanted to block things until we know otherwise.

We do have an upcoming story for a manager to put something back into 'in_assessment' (asking the officer to re-evaluate), so we'll definitely need to revisit this then if not before.

app/views/api/v1/planning_applications/_show.json.jbuilder
```diff
@@ -6,5 +6,5 @@ json.site_address planning_application.site.full_address
 json.application_type t(planning_application.application_type)
 json.summary_of_proposal planning_application.description
 json.received_date planning_application.created_at.iso8601
-json.decided_at planning_application.reviewer_decision.decided_at&.iso8601
+json.determined_at planning_application.determined_at&.iso8601
```
EGiataganas, line 9: Do we still want to have the ampersand operator? At this point `determined_at` shouldn't not be `nil` so I think we can delete `&`. What do you think

db/migrate/20200623100000_rename_completed_at_to_determined_at_on_planning_applications.rb
```diff
@@ -0,0 +1,5 @@
+class RenameCompletedAtToDeterminedAtOnPlanningApplications < ActiveRecord::Migration[6.0]
+  def change
+    rename_column :planning_applications, :completed_at, :determined_at
```
EGiataganas, line 3: 👍

spec/requests/planning_application_list_spec.rb
```diff
@@ -40,8 +40,8 @@
     end
 
     it "returns a list of serialized planning application" do
-      planning_application_1 = create(:planning_application, :determined)
-      planning_application_2 = create(:planning_application, :determined)
+      planning_application_1 = create(:planning_application).tap { |pa| pa.update_and_timestamp_status(:determined) }
+      planning_application_2 = create(:planning_application).tap { |pa| pa.update_and_timestamp_status(:determined) }
```
EGiataganas, line 44: How about updating factories to update `determined_at` on the `after_create`?

NeilvB: Thanks. Yes I'm not sure why I went this long way round. Updated.

PR #83, opened 2020-07-01
Feature/pt 173110690/multistep archive form 

Rhian:
archive
confirm
0-144b31f9c2b0.png">
index

### Description of change

This implements the document archive journey.

1. A filter is applied to the Planning Application overview page to show only documents that are not archived
2. The document display has been updated to use divs instead of a table
3. The filter is applied to the Drawings index page and an archive table has been added to display archived documents
4. An Archive page has been added which uses a new model (DrawingWizard) that temporarily stores the reason for archive
5. A Confirm page has been added which adds the new attribute to the Drawing model on saving
6. Flash messages have been suppressed on the drawing page to allow the display of a single-use custom flash message confirming the document has been archived

### Story Link

### Known issues 

Things you know need further follow on work 
1. Tags are displayed as a placeholder until the migration has been added

### DIFF
--- a/app/helpers/drawing_helper.rb
+++ b/app/helpers/drawing_helper.rb
@@ -1,0 +1,12 @@
+
+# frozen_string_literal: true
+
+module DrawingHelper
+  def filter_archived(drawings)
+    drawings.select { |plan| plan.is_archived? == true }
+  end
+
+  def filter_current(drawings)
+    drawings.select { |plan| plan.is_archived? == false }.sort_by(&:archived_at)
+  end
+end
--- a/app/helpers/drawing_helper.rb
+++ b/app/helpers/drawing_helper.rb
@@ -1,12 +1,11 @@
-
 # frozen_string_literal: true
 
 module DrawingHelper
   def filter_archived(drawings)
-    drawings.select { |plan| plan.is_archived? == true }
+    drawings.select { |plan| plan.archived? == true }
   end
 
   def filter_current(drawings)
-    drawings.select { |plan| plan.is_archived? == false }.sort_by(&:archived_at)
+    drawings.select { |plan| plan.archived? == false }.sort_by(&:archived_at)
   end
 end
--- a/app/javascript/stylesheets/application.scss
+++ b/app/javascript/stylesheets/application.scss
@@ -16,10 +16,12 @@ $govuk-assets-path: "~govuk-frontend/govuk/assets/";
   float:left;
   padding-bottom: 5%;
   width: 40%;
+  margin-left: 2%;
 }
 
 .thumbnail-right {
   float:right;
   padding-bottom: 5%;
   width: 40%;
+  margin-right: 10%;
 }
--- a/app/javascript/stylesheets/application.scss
+++ b/app/javascript/stylesheets/application.scss
@@ -20,8 +20,71 @@ $govuk-assets-path: "~govuk-frontend/govuk/assets/";
 }
 
 .thumbnail-right {
-  float:right;
-  padding-bottom: 5%;
-  width: 40%;
-  margin-right: 10%;
+   float:right;
+   padding-bottom: 5%;
+   width: 40%;
+   margin-right: 10%;
+ }
+
+.flash-archive {
+  border: 5px solid govuk-colour("green");
+  @include govuk-font($size: 19);
+  color: govuk-colour("black");
+  display: block;
+  overflow: hidden;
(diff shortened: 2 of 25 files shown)


app/models/drawing.rb
```diff
@@ -4,4 +4,15 @@ class Drawing < ApplicationRecord
   belongs_to :planning_application
 
   has_one_attached :plan
+
+  enum archive_reason: { scale: 0, design: 1,
+                         dimensions: 2, other: 3 }
+
+  def is_archived?
+     archived_at == nil ? false : true
```
EGiataganas, line 12: How about `archived_at.present?`?

As I understand the method `is_archived?` is a helper method to check whether the `archived_at` timestamp is there

EGiataganas:
I prefer to use `Time.current` rather than `Time.zone.now`. There is no difference except cleaner/shorter syntax if you are using Rails as is, plus we will be consistent cause we use `.current` atm.

app/models/drawing.rb
EGiataganas: Also how about rename it to `archived?`

spec/system/drawings/archive_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+      expect(page).to have_content("You need to sign in or sign up before continuing.")
+    end
+
+    scenario "Assessor sees flash warning if no radio button selected" do
+      click_button "Save"
+      expect(page).to have_text("Please select valid reason for archiving")
+    end
+
+    scenario "Assessor can proceed to confirmation page" do
+      page.find(id: "revise-dimensions").click
```
NeilvB, line 96: I wonder if we can use something like `choose "Revise Dimensions"` rather than looking up by the id..

app/models/drawing_wizard/archive_form.rb
```diff
@@ -6,8 +6,8 @@ class ArchiveForm < BaseForm
     validates :updated_at, presence: true
 
     validates :archive_reason,
-              inclusion: {in: %w(scale design dimensions other),
+              inclusion: { in: %w(scale design dimensions other),
```
NeilvB, line 9: We might be able to avoid some duplication here. We should be able to ask the `Drawing` for its list of reasons. E.g. `Drawing.archived_reasons.keys`

spec/system/drawings/archive_spec.rb
Rhian: I'm having some issues with this... working on it.

app/helpers/drawing_helper.rb
```diff
@@ -0,0 +1,12 @@
+
```
EGiataganas, line 1: Can you remove this empty line please?

app/views/application/_success_flash.html.erb
```diff
@@ -0,0 +1,7 @@
+<% flash.each do |name, msg| %>
+  <div class="flash-archive">
+    <svg class="moj-banner__icon" color="green" role="presentation" focusable="false" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 25 25" height="25" width="25">
+      <path d="M25,6.2L8.7,23.2L0,14.1l4-4.2l4.7,4.9L21,2L25,6.2z"></path></svg>
+    <%= msg.to_s %>
+  </div>
+<% end %>
```
EGiataganas, line 7: There is already a [flash partial][1] can we reuse it?

[1]: https://github.com/unboxed/bops/blob/master/app/views/application/_flashes.html.erb

app/javascript/stylesheets/application.scss
```diff
@@ -11,3 +11,80 @@ $govuk-assets-path: "~govuk-frontend/govuk/assets/";
...
+  border-color: govuk-colour("green");
+  color: govuk-colour("green");
+}
+
+
+.moj-banner--warning {
+  border-color: govuk-colour("red");
+  color: govuk-colour("red");
+}
+
```
EGiataganas, line 90: Have a look the [flashes component][1] I think some of them if not all of the are covered there 😄 

[1]: https://github.com/unboxed/bops/blob/master/app/javascript/stylesheets/components/_flashes.scss

app/views/application/_success_flash.html.erb
Rhian: The styling is completely different for this particular message - it needs to look like the MoJ banner. I thought it would be not ideal to have two completely different sets of css in the same partial, but if you can think of a way to do it then please suggest

PR #88, opened 2020-07-08
refactored archive form to use radio button collection

Rhian:
### Description of change

Radio buttons in Archive form have been refactored to generate automatically using the native Rails collection_radio_buttons syntax.

### Story Link

### Decisions 

The Confirm form was not refactored in the same way as it has only two buttons which behave differently, so it was more straightforward and readable to leave it as HTML.

### DIFF
--- a/app/helpers/drawing_helper.rb
+++ b/app/helpers/drawing_helper.rb
@@ -8,4 +8,8 @@ module DrawingHelper
   def filter_current(drawings)
     drawings.select { |plan| plan.archived? == false }.sort_by(&:archived_at)
   end
+
+  def archive_reason_collection_for_radio_buttons
+    Drawing.archive_reasons.keys.map { |k| [k, I18n.t(k)] }
+  end
 end
--- a/app/views/drawings/_archive_form.html.erb
+++ b/app/views/drawings/_archive_form.html.erb
@@ -3,33 +3,25 @@
     <%= hidden_field_tag :current_step, 'archive' %>
     <fieldset class="govuk-fieldset">
       <div class="govuk-form-group <%= form.object.errors.any? ? 'govuk-form-group--error' : '' %>">
-      <% if form.object.errors.any? %>
-        <% form.object.errors.each do |attribute, error| %>
-            <span id="status-error" class="govuk-error-message">
-              <span class="govuk-visually-hidden">Error:</span><%= error %></span>
-        <% end %>
-      <% end %>
-      <div class="govuk-radios govuk-radios--conditional" data-module="govuk-radios">
-        <div class="govuk-radios govuk-radios--small">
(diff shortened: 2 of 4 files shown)


app/helpers/drawing_helper.rb
```diff
@@ -8,4 +8,8 @@ def filter_archived(drawings)
   def filter_current(drawings)
     drawings.select { |plan| plan.archived? == false }.sort_by(&:archived_at)
   end
+
+  def archive_reason_collection_for_radio_buttons
+    Drawing.archive_reasons.keys.map { |k| [k, I18n.t(k)] }
```
EGiataganas, line 13: I know that in the project we generally use `I18n.t` where we could use `t`. It's on my list of things we can improve. But since I bump into this again I think it would be great to start using just `t` which is a Rails helper method and also adds some extra features, which are not provided by the `I18n.t` method. Rails `t` helper method also adds security awareness through its built in XSS protection, which the `t` method has to deal with because it adds strings to the view.

Feel free to have a look at [this][1]

[1]: https://oananao.wordpress.com/2014/07/12/whats-the-difference-between-using-i18n-t-and-simply-t/

app/views/drawings/_archive_form.html.erb
```diff
@@ -3,33 +3,25 @@
...
-          <div class="govuk-radios__item">
-            <%= form.radio_button :archive_reason, "other", class: "govuk-radios__input", "aria-controls": "archive-reason-other" %>
-            <%= form.label :archive_reason, "Other", class: "govuk-label govuk-radios__label", value: "other" %>
-          </div>
+        <div class="govuk-radios govuk-radios--conditional" data-module="govuk-radios">
+          <%= form.collection_radio_buttons :archive_reason, archive_reason_collection_for_radio_buttons, :first, :last do |button| %>
+            <div class='govuk-radios govuk-radios--conditional'>
+              <div class="govuk-radios govuk-radios--small">
+                <div class="govuk-radios__item">
+                  <%= button.radio_button(class: "govuk-radios__input") %>
```
EGiataganas, line 17: Can you also add the `aria-controls` please?

spec/system/drawings/archive_spec.rb
```diff
@@ -115,7 +115,7 @@
     scenario "Assessor is returned to archive page if 'No' is selected" do
       choose "scale"
       click_button "Save"
-      page.find(id: "archive-no").click
+      choose "No"
```
EGiataganas, line 118: 👍

app/helpers/drawing_helper.rb
Rhian: Good idea - I have incorporated this into the Pivotal story pt-173627179.

PR #90, opened 2020-07-10
[#173744456] Breadcrumb Navigation

NeilvB:
### Description of change

Small change to how we manage breadcrumbs, to allow us to nest pages with more than two parent pages, such as those for document archiving, which require the following breadcrumbs, all linked except for the last:

Home > Application > Documents > Archive Document

The downside of this approach is that it means that some views make multiple calls to add parent pages, with some duplication. I wasn't sure how to address this. One option could be to add a nested layout for drawings, which makes the relevant parent breadcrumbs.

### Story Link

https://www.pivotaltracker.com/n/projects/2441805/stories/173744456

### DIFF
--- a/app/views/application/_breadcrumb_navigation.html.erb
+++ b/app/views/application/_breadcrumb_navigation.html.erb
@@ -1,5 +1,5 @@
 <% if current_user %>
-  <% unless nav.length > 1 %>
+  <% if current_page?(planning_applications_path(@planning_application)) %>
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-full">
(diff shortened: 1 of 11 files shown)


app/views/application/_breadcrumb_navigation.html.erb
```diff
@@ -1,6 +1,6 @@
 <% if current_user %>
-  <% unless nav.length > 1 %>
-    <div class="govuk-grid-row">
+  <% if current_page?(planning_applications_path(@planning_application)) %>
+      <div class="govuk-grid-row">
       <div class="govuk-grid-column-full">
```
EGiataganas, line 4: Can you indent this `div` please?

PR #94, opened 2020-07-13
Fix/pt 173110690/add confirmation error

Rhian:
confirm_error

### Description of change

This adds the missing error message when neither of the radio buttons is clicked on the second step of the document archiving story.

### Story Link

### Decisions 

This is a quick fix to add error messaging. In the longer term, we should add a separate form controller, as previously discussed.

### DIFF
--- a/app/views/drawings/_confirm_form.html.erb
+++ b/app/views/drawings/_confirm_form.html.erb
@@ -1,21 +1,30 @@
 <%= form_for(:confirmation, url: planning_application_drawing_validate_step_path) do |form| %>
-  <%= hidden_field_tag :current_step, 'confirm' %>
-  <fieldset class="govuk-fieldset">
-  <div class="govuk-radios govuk-radios--conditional" data-module="govuk-radios" style="margin-bottom: 20px;">
-    <div class="govuk-radios govuk-radios--small">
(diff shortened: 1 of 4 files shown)


app/views/drawings/_confirm_form.html.erb
```diff
@@ -1,21 +1,32 @@
...
-    <div class="govuk-radios govuk-radios--small">
-      <div class="govuk-radios__item">
-        <%= hidden_field_tag :archive_reason, @drawing_form.archive_reason %>
-        <%= form.radio_button :confirm, "yes", class: "govuk-radios__input", id:"archive-yes", "aria-controls":"archive-yes" %>
-        <%= form.label :confirm, "Yes", class: "govuk-label govuk-radios__label", for: "archive-yes" %>
+  <div class="govuk-form-group">
+    <%= hidden_field_tag :current_step, 'confirm' %>
+    <fieldset class="govuk-fieldset">
+      <div class="govuk-form-group <%= @drawing_form.errors.any? ? 'govuk-form-group--error' : '' %>">
+        <% if @drawing_form.errors.any? %>
```
NeilvB, line 6: I think we can remove this conditional block

PR #99, opened 2020-07-22
Move proposal docs to right hand side menu

melzreal:
### Description of change

Move proposal docs to right hand side menu

### Story Link

https://www.pivotaltracker.com/story/show/173769622

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.

image

### DIFF
--- a/app/javascript/stylesheets/application.scss
+++ b/app/javascript/stylesheets/application.scss
@@ -3,6 +3,7 @@ $govuk-assets-path: "~govuk-frontend/govuk/assets/";
 @import "components/buttons";
 @import "components/flashes";
 @import "components/task_list";
+@import "proposal";
 
 .noborder {
   border-bottom: none !important;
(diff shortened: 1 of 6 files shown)


melzreal:
Screenshot 2020-07-22 at 15 28 26
Modified version now with scrollbar and capitalised file names

Also note I had to add a few line breaks between proposal files because otherwise the name shows up like this:

Screenshot 2020-07-22 at 14 59 10


instead of having spacing (this is what it currently looks like):
Screenshot 2020-07-22 at 14 58 06

package.json
```diff
@@ -6,6 +6,7 @@
     "@rails/ujs": "^6.0.0",
     "@rails/webpacker": "4.2.2",
     "govuk-frontend": "^3.6.0",
+    "node-sass": "^4.14.1",
```
Rhian, line 9: Just checking the reason for the package.json update?

melzreal: ohh this was accidental. I had to uninstall and re-install node-sass 3 times throughout the day because the scss wasn't rendering - I'll untrack it!

PR #100, opened 2020-07-23
Feature/pt 173125684/add reviewer comment

Rhian:
officer1
officer2
officer3
officer4
officer5
officer6

### Description of change

This changes the journey where the manager disagrees with the officer's decision. Once the manager has clicked 'No' and provided the mandatory correction, the application moves to a new tab on the index (Corrections Requested) and an alert is shown to the officer. Once the application returns to the manager, they are shown an alert and are able to view the officer's comments and their own correction before determining the application.

### Story Link

https://www.pivotaltracker.com/n/projects/2441805/stories/173338338
https://www.pivotaltracker.com/story/show/173125684

### Decisions 

I decided against setting entirely new statuses for the applications where corrections were requested and provided, and instead use two methods that look for the presence of a correction and the role that last updated. In my opinion, the correction statuses are subsets of the awaiting_determination status rather than statuses in themselves. I have also removed the specs for the saved state when the reviewer sees the review corrections/review recommendation layout, as the new journey means that the reviewer is now no longer allowed to access the story once they have reviewed it - the link becomes non-clickable.

### DIFF
--- a/app/models/decision.rb
+++ b/app/models/decision.rb
@@ -20,8 +20,7 @@ class Decision < ApplicationRecord
   def ensure_correction
     return unless planning_application.awaiting_determination?
     return unless user.reviewer?
-    return if planning_application.reviewer_decision.status ==
-        planning_application.assessor_decision.status
+    return unless planning_application.reviewer_disagreed_with_assessor?
 
     if correction.blank?
       errors.add(:correction, "Please enter a reason in the box")
--- a/app/models/decision.rb
+++ b/app/models/decision.rb
@@ -9,7 +9,22 @@ class Decision < ApplicationRecord
   validates :status, inclusion: { in: ["granted", "refused"],
     message: "Please select Yes or No" }
 
+  # validate :ensure_correction, on: :update
+
   def comment_made?
     granted? && comment_met.present? || refused? && comment_unmet.present?
+  end
+
+  private
+
+  def ensure_correction
+    return unless planning_application.correction?
+    return unless user.reviewer?
+    return if planning_application.reviewer_decision.status ==
+        planning_application.assessor_decision.status
+
+    if decision_params[correction].blank?
+      errors.add(:correction, "Please enter a reason into the box")
+    end
   end
 end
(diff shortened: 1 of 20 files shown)


app/models/decision.rb
```diff
@@ -9,7 +9,22 @@ class Decision < ApplicationRecord
...
+
   def comment_made?
     granted? && comment_met.present? || refused? && comment_unmet.present?
   end
+
+  private
+
+  def ensure_correction
+    return unless planning_application.awaiting_determination?
+    return unless user.reviewer?
```
melzreal, line 22: should this return on line 22 not be controlled by pundit? if a user isn't a reviewer are they allowed to access the decision model?

Rhian: We use the same decision model for both assessor and reviewer. You're right that Pundit controls their access to the UI where they would be able to make this change, but without the validation there would be nothing stopping a correction being applied to the assessor's decision if a REST call was made directly, so it's just another layer of security.

app/models/decision.rb
```diff
@@ -9,7 +9,22 @@ class Decision < ApplicationRecord
...
+    return if planning_application.reviewer_decision.status ==
+        planning_application.assessor_decision.status
```
EGiataganas, line 24: can we extract them into separate methods please?

app/views/planning_applications/_assess_proposal.html.erb
```diff
@@ -2,34 +2,46 @@
...
+        <% end %>
       </h2>
       <ul class="app-task-list__items">
         <li class="app-task-list__item">
-          <% step_name = "Assess the proposal" %>
+          <% if @planning_application.correction?  %>
+            <% step_name = "Reassess the proposal" %>
+          <% else %>
+            <% step_name = "Assess the proposal" %>
+          <% end %>
```
EGiataganas, line 17: I think it would be great if we could we extract this functionality into a service and here we just initialise the name `step_name`. What do you think?

NeilvB: I've added a few helper methods https://github.com/unboxed/bops/pull/100/commits/9985df90cee6fb5cd4c41c21dc356d7427c2f08d

app/models/decision.rb
NeilvB: I've added a method on the planning application https://github.com/unboxed/bops/pull/100/commits/6e57d9b41f1371acaf5cdbda665ee72cc1e4f174

EGiataganas:
Closed in favour of #113

PR #102, opened 2020-07-24
WIP control user list display

melzreal:
### Description of change

On index, ensure that a user with role of type "assessor" only sees both their own planning applications as well as applications that have not yet been assigned to an assessor.

Also adds a "View all applications" so the assessor can toggle the view to see all existing applications

### Story Link

https://www.pivotaltracker.com/story/show/173508545

### DIFF
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -14,6 +14,10 @@ class PlanningApplicationsController < AuthenticationController
     @planning_applications = policy_scope(PlanningApplication.all)
   end
 
+  def show_all
+    @planning_applications = authorize(PlanningApplication.all)
+  end
+
   def show
   end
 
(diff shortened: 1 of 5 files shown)


app/controllers/planning_applications_controller.rb
```diff
@@ -11,7 +11,16 @@ class PlanningApplicationsController < AuthenticationController
     with: :decision_notice_mail_error
 
   def index
-    @planning_applications = policy_scope(PlanningApplication.all)
+    all_applications = policy_scope(PlanningApplication.all)
+    if current_user.assessor?
+      @planning_applications = PlanningApplication.where(
```
EGiataganas, line 16: That looks great! Have you thought of moving that logic into a scope in the `planning_application.rb`?

melzreal: ah thanks Evangelos! Yes will do - I'm going to try and place this logic inside a scope now and hopefully it won't break the associaiton - this first WIP was just to try and get the logic to work :)

EGiataganas: Yeah that's not a problem at all! 😄

melzreal:
image
and 
image

vs #103 (both images are "view my applications" first and "view all applications" second
image
and 
image

melzreal:
closed in favour of #103

PR #103, opened 2020-07-24
Use params to amend user views

melzreal:
### Description of change
Display different views to an assessor - upon login assessor only views its own and unassigned applications. Adds option to toggle that view.

### Story Link

https://www.pivotaltracker.com/story/show/173508545

### DIFF
--- a/spec/system/planning_applications/index_spec.rb
+++ b/spec/system/planning_applications/index_spec.rb
@@ -11,63 +11,116 @@ RSpec.feature "Planning Application index page", type: :system do
   context "as an assessor" do
     before do
       sign_in users(:assessor)
-      visit planning_applications_path
+      visit root_path
     end
 
-    scenario "Planning Application status bar is present" do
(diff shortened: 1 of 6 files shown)


spec/system/planning_applications/index_spec.rb
```diff
@@ -69,6 +71,28 @@
...
+    scenario "On login, assessor gets redirected to a view with its own and unassigned Planning Applications" do
+      visit planning_applications_path(assessor: "assessor")
+
+      within("#in_assessment") do
+        expect(page).to have_link(planning_application_1.reference)
+        expect(page).to have_link(planning_application_2.reference)
+        expect(page).not_to have_link(planning_application_3.reference)
+      end
+    end
+
```
melzreal, line 84: given application_3 belongs to assessor_2 that's the reason it shouldn't be displayed here - perhaps its best to create assessor 2 and application3 inside the block  instead of using "let!" at the top, so the test is more explicit?

PR #106, opened 2020-07-29
Feature/pt 173960508/decision comments

EGiataganas:
### Description of change

- Drop `comment_met` from decisions
- Rename `comment_unmet` to `public_comment` in decisions
- Show `public_comment` to decision_notice
- Add validation on `public_comment` when assessor user choose no to write a comment
- Add `private_comment` on decision table
- Update with latest content design changes

### Story Link

https://www.pivotaltracker.com/story/show/173960508

### Screenshots

Screenshot 2020-07-29 at 13 21 03
Screenshot 2020-07-29 at 13 21 38
Screenshot 2020-07-29 at 13 22 20

### DIFF
--- a/app/models/decision.rb
+++ b/app/models/decision.rb
@@ -10,6 +10,6 @@ class Decision < ApplicationRecord
     message: "Please select Yes or No" }
 
   def comment_made?
-    granted? && comment_met.present? || refused? && comment_unmet.present?
+    refused? && comment_unmet.present?
   end
 end
--- a/app/models/decision.rb
+++ b/app/models/decision.rb
@@ -9,7 +9,18 @@ class Decision < ApplicationRecord
   validates :status, inclusion: { in: ["granted", "refused"],
     message: "Please select Yes or No" }
 
+  validate :validate_public_comment
+
   def refused_with_public_comment?
     refused? && public_comment.present?
+  end
+
+  def validate_public_comment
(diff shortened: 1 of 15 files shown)


app/models/decision.rb
```diff
@@ -10,6 +10,6 @@ class Decision < ApplicationRecord
     message: "Please select Yes or No" }
 
   def comment_made?
```
NeilvB, line 14: Please could we include `public` in the method name here to differentiate it from the private comments.

spec/models/decision_spec.rb
```diff
@@ -26,6 +26,15 @@
     end
```
NeilvB, line 27: It would be good to include a test for the model being invalid. E.g. when the user is an assessor, it's refused, and there's an empty string as the comment.

spec/system/planning_applications/assessing_spec.rb
```diff
@@ -154,4 +154,6 @@
   end
 
   include_examples "assessor decision error message"
+
+  include_examples "assessor refused decision without public_comment error message"
```
NeilvB, line 164: Is this the only place where we're using this new shared example? If so I think it's best to remove the shared example and keep that example here in this spec. My reasoning is that with the single use shared example the reader has to jump to another file to read it, but we're not getting the benefit of reducing duplication across our specs.

PR #110, opened 2020-08-04
Add partials for read only view

siapankina:
### Description of change

Added partials for read only view.
And updating titles and breadcrumbs on each stage (awaiting determination , determined)

### Story Link

https://www.pivotaltracker.com/story/show/173061267

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. Evangelos is working on a story that will affect this story. Need to be updated as soon as Evangelos finishes his story

none

### DIFF
--- a/app/views/planning_applications/edit.html.erb
+++ b/app/views/planning_applications/edit.html.erb
@@ -3,21 +3,27 @@
     <% content_for :title, t("#{@planning_application.status}.recommendation_step") %>
 
     <h2 class="govuk-heading-m"><%= t("#{@planning_application.status}.recommendation_step") %></h2>
-    <p class="govuk-body">The following decision notice has been created based on your answers.</p>
+    <% if @planning_application.determined? %>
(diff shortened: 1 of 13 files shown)


app/views/planning_applications/edit.html.erb
```diff
@@ -3,21 +3,27 @@
...
-      <strong>
-        If you agree with the decision notice, please submit it to your manager.
-      </strong>
-      If your manager disagrees with your recommendation they will send it back to you to make changes.
-    </p>
-
-    <%= form_with model: @planning_application, local: true do |form| %>
-      <%= form.hidden_field :status, value: :awaiting_determination %>
-      <%= form.submit "Submit to manager", class: "govuk-button", data: { module: "govuk-button" } %>
+    <% if @planning_application.in_assessment? || @planning_application.awaiting_correction?  %>
```
EGiataganas, line 13: Do yo think we can extract this into a method in `planning_application.rb`?

PR #111, opened 2020-08-05
Feature/pt 173125684/yml file

EGiataganas:
### Description of change

We're moving all the related text for the task-list into a yml and adding a logic for all users share the same task-list depending on the planning application status

### DIFF
--- a/app/views/planning_applications/_assess_proposal.html.erb
+++ b/app/views/planning_applications/_assess_proposal.html.erb
@@ -33,11 +33,7 @@
           <span class="app-task-list__task-name"><%= step_name %></span>
         <% end %>
 
-        <% if @planning_application.in_assessment? && @planning_application.assessor_decision&.valid? %>
-          <%= tag.strong "Completed", id: "#{step_name.parameterize}-completed", class: "govuk-tag app-task-list__task-completed" %>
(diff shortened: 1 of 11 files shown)


app/views/planning_applications/_assess_proposal.html.erb
```diff
@@ -1,96 +1,62 @@
...
+      <li class="app-task-list__item">
+        <% step_name = t("#{@planning_application.status}.recommendation_step") %>
+
+        <% if (current_user.assessor? || current_user.admin?) && @planning_application.assessor_decision %>
+          <% aria_attributes = @planning_application.awaiting_determination? ? { describedby: "#{step_name.parameterize}-completed" } : {} %>
+
+          <span class="app-task-list__task-name"><%= link_to step_name, edit_planning_application_path(@planning_application), aria: aria_attributes %></span>
+        <% elsif current_user.reviewer? && @planning_application.reviewer_decision %>
+          <% aria_attributes = @planning_application.awaiting_determination? ? { describedby: "#{step_name.parameterize}-completed" } : {} %>
+
```
NeilvB, line 50: It looks like we can reduce some duplication by setting `aria_attributes` before the `if`

PR #119, opened 2020-08-11
Updated create_sample_data task to use real addresses

Rhian:
multiple

### Description of change

Added sample data YAML to allow multiple sample data creation using genuine Southwark addresses.


james:
I'm assuming we should close this?

Rhian:
This is now redundant following further development of the sample data and the ability to submit via API

PR #123, opened 2020-11-18
Add subdomains [WIP]

melzreal:
### Description of change

image

image

This WIP plays with the ability to add subdomains using native rails methods, without depending on gems.
To see this locally, start your rails server normally then head to 

```http://lambeth-council.lvh.me:3000/```
```http://southwark-council.lvh.me:3000/```

The idea would be to scope users per council, re-routing users depending on which council they belong to. Still thinking about the best practice to scope this.

Maybe a join table for council users instead of doing the dependency directly users - to - council would be best, to allow for multiple councils per user should the need occur.

### Story Link

https://trello.com/c/5Oujf3pn/6-support-multiple-councils

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -170,7 +170,7 @@ GEM
       byebug (~> 11.0)
       pry (~> 0.13.0)
     public_suffix (4.0.4)
-    puma (4.3.5)
+    puma (4.3.6)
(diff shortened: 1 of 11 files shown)


melzreal:
Closed in favour of using the acts_as_tenant gem instead

PR #124, opened 2020-11-18
Replace fixtures with factories in specs

melzreal:
### Description of change
Change fixtures to use factories. Reasons:
- Fixtures were causing issues when testing subdomains
- Consistency: factories allow for flexibility, so why would we have two different sets of tests when factories can do exactly the same thing
- Visibility: With factories, we override the values that we are testing in the same file at the top of the file. So this way we don't need to check two different files to amend values being tested (the two files being fixtures and the test values)


james:
This all looks good so far! But I don't think it fulfils the full statement of "Replace fixtures with factories in specs" - as the tests are still reliant on the users.yml fixtures. I'd like an extra commit on this PR that removes them as well, and then we can merge them together.

james:
Another side request for best practice: I'd like to ensure that we include the reasoning behind a change in the git commit message as well as the Github Pull Request. This ensures that we always include our thinking behind code changes within the git repository itself, meaning we're not relying on this github repository always be available in the future, and makes it easier to find in the git log.

spec/policies/planning_application_policy_spec.rb
```diff
@@ -22,23 +21,26 @@
 
   describe "#permitted_statuses" do
     context "an assessor" do
-      let(:user) { users(:assessor) }
+      let(:user) {  create :user, :assessor }
+      let(:policy) { described_class.new(user, record) }
```
james, line 25: Why did we have to move the policy inside each context? I tried putting it back to where it was, and everything still passed for me. Maybe it was something you tried, but didn't undo?

spec/rails_helper.rb
```diff
@@ -28,7 +28,6 @@
   config.file_fixture_path = "#{::Rails.root}/spec/fixtures/files"
 
   config.use_transactional_fixtures = true
-  config.global_fixtures = :users
```
james, line 31: Yay! Not having these user fixtures loaded for all tests will make it a lot easier to manage users for different subdomains :-)

spec/policies/planning_application_policy_spec.rb
melzreal: it had to be moved inside context because the block above iterates through the users then creates blocks for them. It will work for you if you copy paste this specific user and move it outside the context because then that user will be acessible to all the contexts. If you try removing the policy from the block just above this one youll see it break

spec/system/drawings/archive_spec.rb
```diff
@@ -3,9 +3,9 @@
 require "rails_helper"
 
 RSpec.feature "Drawings index page", type: :system do
-  fixtures :sites
-
-  let!(:site) { sites(:elm_grove) }
+  let!(:site) { create :site, address_1: "Elm Grove" }
+  let!(:assessor) { create :user, :assessor }
+  let!(:reviewer) { create :user, :reviewer }
```
james, line 8: This is the tiniest comment in the world, but this is a good use-case of using `let` instead of `let!` for `assessor` and `reviewer`. That way, the specs that don't use the reviewer, won't create it in the database.

Of course in this situation, that will save us about 0.3 milliseconds (complete guess) from our test suite time, so it's really not a big deal, but I thought I'd mention it anyway.

spec/system/drawings/index_spec.rb
```diff
@@ -3,9 +3,9 @@
 require "rails_helper"
 
 RSpec.feature "Drawings index page", type: :system do
-  fixtures :sites
-
-  let!(:site) { sites(:elm_grove) }
+  let!(:site) { create :site, address_1: "7 Elm Grove" }
+  let(:assessor) { create :user, :assessor }
```
james, line 7: Ah, just like you've used `let` here :-)

spec/system/drawings/archive_spec.rb
melzreal: yes I wasn't sure what to go for.. since I've seen let! used in many files then let in others, I kinda just tried to replace everything that was let! with let! and so on

james (approved):
🎉

PR #125, opened 2020-11-19
Add subdomains

melzreal:
### Description of change

- Allows for different subdomains to be used on the application. It seeds the database with the two first ones, Lambeth and Southwark.
- Removes all fixtures in favour of factories, to standardise testing and also because fixtures caused issues when testing subdomains
- Modifies pundit policies to scope views by subdomain
- Increases overtly restrictive rubocop line limit

### Story Link

https://trello.com/c/5Oujf3pn/6-support-multiple-councils

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -76,7 +76,7 @@ Layout/LeadingCommentSpace:
   Enabled: true
 
 Layout/LineLength:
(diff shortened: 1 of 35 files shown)


james (approved):
🎉

PR #126, opened 2020-11-23
Spike/create input api

Rhian:
### Description of change

This pull request is for a new POST endpoint for the BoPS API, which allows external users to send planning applications directly to the system. It also adds a new field (UPRN) to the Site model and contains logic for either attaching an existing site to the planning application (if the UPRN exists) or creating a new site and attaching it. Applicant and Agent tables have been deleted and the contact details added to the PlanningApplication table. Front end and tests have been refactored to reflect this.

New methods and dictionary have been added to dynamically display constraints information on the front end, and an API user is now added in the create_sample_data task.

Further work will be needed to add authentication and improve error handling.

### Known issues 

- Drawings upload method needs to be added
- There is a Swagger issue where authorization in api-docs does not work - currently being investigated.

swagger


db/migrate/20201119144047_drop_agents_table.rb
```diff
@@ -0,0 +1,8 @@
+class DropAgentsTable < ActiveRecord::Migration[6.0]
+  def up
+    drop_table :agents
+  end
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
melzreal, line 6: just curiosity - I get the irreversible migration is there since we are dropping a table forever.. what are the benefits of raising an error to prevent the rollback of this? 
As in, isn't the action of dropping a table always not going to be rolled back? Just wondering if there are some rails mysteries you could enlighten me on :D

app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -11,38 +11,31 @@ def index
   end
 
   def create
-    @planning_application = PlanningApplication.create(full_planning_application_params)
-    if @planning_application.save!
-      @planning_application.create_policy_evaluation
-      render json: {"id": "#{@planning_application.reference}",
-                    "message": "Application created: id was #{@planning_application.reference}"}, status: 200
+    @plan_app = PlanningApplication.create(full_planning_application_params)
```
james: Is there a reason you've abbreviated this variable name?

app/views/planning_applications/_decision_notice.html.erb
```diff
@@ -5,7 +5,7 @@
   <p class="govuk-body">Certificate of lawfulness of proposed use or development: <strong><%= "#{decision.status}." %></strong></p>
   <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
   <p class="govuk-body">
-    <strong>Applicant:</strong>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;<%= planning_application.applicant.full_name %><br>
```
james, line 8: Outside of this PR, but I'm going to have to check out what's going on with this `&nbsp`s at some point!

db/migrate/20201119144047_drop_agents_table.rb
james: There's a reasonable assumption that rolling back a migration may be destructive on the data within the database, so we don't need to `raise` to warn of that. But in this situation, the migration wouldn't even know how to rollback the scheme, because it doesn't know what structure of the `agent`s table was when it existed.

You can actually make a `drop_table` reversible, by giving it the state of the table when you dropped it, like:

```
drop_table :agents do |t|
  t.string :name
end
```

In which case the migration would be reversible. But we're unlikely to need to be able to do that, so I guess Rhian though this would suffice, but the `raise` makes it clear that this work hasn't been done, which I'm fine with.

james (approved):
All looks good to me!

app/controllers/api/v1/planning_applications_controller.rb
Rhian: Not a particularly logical reason... I will go back to it and reinstate the full name

app/views/planning_applications/_decision_notice.html.erb
Rhian: Yes, there has to be a better way to do it. I'll make a note and try to find time to raise a separate ticket for it

james:
[BoPS receives permitted dev. applications](https://trello.com/c/P8s9Yklg/7-bops-receives-permitted-dev-applications)

app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -13,9 +13,11 @@ def index
   def create
     @planning_application = PlanningApplication.create(full_planning_application_params)
     if @planning_application.save!
-      render json: {"message": "Application created: id was #{@planning_application.reference}"}, status: 200
+      @planning_application.create_policy_evaluation
```
james: This is a complete aside, but why do we have policy_evaluation as a `has_one`?

app/views/shared/_supporting_information.html.erb
```diff
@@ -65,16 +65,22 @@
       </h2>
     </div>
     <div id="accordion-default-content-3" class="govuk-accordion__section-content" aria-labelledby="accordion-default-heading-3">
-      <% if @agent %>
+      <% if @planning_application.agent_first_name %>
```
james, line 64: Might be a bit more accurate if we have an `agent?` and `applicant?` method on the PlanningApplication model that returns true if first_name, last_name, phone *or* email is present?

app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -11,15 +11,28 @@ def index
   end
 
   def create
-    @plan_app = PlanningApplication.create(full_planning_application_params)
-    if @plan_app.save!
-      @plan_app.create_policy_evaluation
-      render json: { "id": "#{@plan_app.reference}",
-                    "message": "Application created" }, status: 200
+    site_id = create_site
```
james, line 14: 🏄  `create_site` isn't quite an accurate method name here, as it's a find and create. In fact, I'd say that it doesn't need to be it's own method, as the code within the method is more descriptive than any method name I can think of.

app/views/shared/_supporting_information.html.erb
Rhian: Good idea 👍

app/controllers/api/v1/planning_applications_controller.rb
Rhian: From memory, I think it was the necessity to model the real-world application - ie, the requirement that a planning application is assessed against only one set of policy considerations.

app/views/shared/_supporting_information.html.erb
Rhian: Fixed on last commit.

PR #131, opened 2020-11-25
Prevent login from different subdomain

melzreal:
### Description of change
- Prevent users that belong to subdomain southwark from logging in to lambeth subdomain and vice versa
- This is achieved by using the routes file to pass the subdomain to the params used by the devise authentication method "find_for_authentication"

### Story Link

https://trello.com/c/04CvzZFU/18-prevent-login-to-users-belonging-to-a-different-council-then-the-subdomain

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -6,9 +6,14 @@ class User < ApplicationRecord
   enum role: { assessor: 0, reviewer: 1, admin: 2 }
 
   devise :database_authenticatable, :recoverable,
-         :rememberable, :validatable
+         :rememberable, :validatable, request_keys: [:subdomains]
 
   has_many :decisions, dependent: :restrict_with_exception
   has_many :planning_applications, through: :decisions
   belongs_to :local_authority
+
+  def self.find_first_by_auth_conditions(conditions)
+    local_authority = LocalAuthority.find_by(subdomain: conditions[:subdomains].first)
+    find_by(email: conditions[:email], local_authority: local_authority.id)
+  end
 end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -47,4 +47,24 @@ RSpec.describe User, type: :model do
     email = build(:user, email: nil)
     expect(email).to_not be_valid
   end
+
+  it "does not allow for duplicate users within the same domain" do
+    domain = create(:local_authority)
+    user_one = create(:user, email: "pompom@pom.com", local_authority: domain)
+    user_two = build(:user, email: "pompom@pom.com", local_authority: domain)
+
+    expect(user_two.save).to be false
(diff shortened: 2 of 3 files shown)


config/routes.rb
```diff
@@ -1,31 +1,44 @@
 # frozen_string_literal: true
 
 Rails.application.routes.draw do
-  root to: "planning_applications#index", defaults: { q: "exclude_others" }
+  class SubdomainConstraint
```
james, line 4: For some reason, I'm a bit nervous about having this logic in the routes and would prefer to have it as a before_action in ApplicationController alongside `find_current_local_authority_from_subdomain`.

I think that's a) keeping the two bits of logic together will make it easier to understand what's going on and b) doing database lookups within routes feels like it confuses the routing a bit too much.

Interested in your thoughts.

melzreal: so I've placed the login here after the conversation I had in the dev channel, it was the suggestion from Andrew - https://ubxd.slack.com/archives/C026M627E/p1605891757006700

Putting this method inside routes makes sure it runs before the application controller does without another before action - it also allows us to define further "match" limitation constraints should we want them in the future, and those match limitations would be in routes as far as I understand? (i.e if route southwark login to x index page if route lambeth go to x index page)

Since this method ensures the params is being passed from the URL into the authentication controller it made sense to me that it gets called before. I'd also really like to debate the option of overusing ```before_action``` - the developer I worked on the last project didn't want to use before_action for anything, preferring an alternative approach to load data using blocks, like this https://blog.ragnarson.com/2015/09/30/problems-with-typical-rails-controllers-and-before-actions.html

I understand from this latest reading that before_action is indeed best used for authentication and authorization, which is what we are doing - but we'd still need to chain them right? So we'd have to have the before action to pass the params to authentication before the before action to lookup the local authority 🤔 

I can give it a spin although I'd also like your thoughts on the overuse of before_action, which seems to be used a lot around the app, and per what I'm reading shouldn't be used to load records - just for authentication actions

melzreal: actually I could just do the lookup inside the same method.. will try that now

james: I think that article about `before_action` is addressing a different problem to the one we have here (as an aside I'm on the fence about whether I agree with that article anyway, I think it's a style choice and I'm happy with either style)

In our case, the `find_current_local_authority_from_subdomain` is both creating a global variable for the whole application, and is doing authentication, so I think is a good use of before_action. I like what you've come up with :-)

james (approved):
🎉

james:
[Prevent login to users belonging to a different council then the subdomain](https://trello.com/c/04CvzZFU/18-prevent-login-to-users-belonging-to-a-different-council-then-the-subdomain)

PR #133, opened 2020-11-26
Add SSH access to fargate web tasks

james:
This ensures that an SSH deamon is installed and running on the web instances.
We add authorised keys from the AWS secrets.


james:
[Developers able to log in to production via SSH](https://trello.com/c/BuwiyXJA/24-developers-able-to-log-in-to-production-via-ssh)

PR #134, opened 2020-11-27
Add bucks to sample data

melzreal:


james:
create_sample_data.rake is getting very large and unruly. I was thinking this might be a good opportunity to refactor it a bit. Is there any way of DRYing it up a bit? Can we wrap more of it in the `local_authorities.each` block?

james:
If we need a bit of customisation, maybe we can create an array of hashes, eg:

`[{local_authority: southwark, ward: 'Dulwich Wood', address_1: '90A College Road'}].each do |authority_information| `

melzreal:
> If we need a bit of customisation, maybe we can create an array of hashes, eg:
> 
> `[{local_authority: southwark, ward: 'Dulwich Wood', address_1: '90A College Road'}].each do |authority_information| `

Each local authority needs specific sample data so the local authority would not be the point where we can refactor. We could probably refactor with the planning_applications instead, but again each of them has addresses specific to each council.

We should instead split the data by council into separate files, and have create_sample_data call those files. So have this task call the separate files.

I don't see how a hash could uncomplicate things to be honest, the data does need to be all different - what would be the benefit of the hash? Its not a matter of speed but readability right?

james (approved):
🎉

PR #135, opened 2020-11-27
Update the drawings created in sample data task

Rhian:
The drawings need to be better-quality than the current images and PDFs so the sample data has been changed to attach genuine planning drawings as PDFs. The factory is still using a .png, for speed purposes.


james (approved):
🎉

PR #136, opened 2020-11-27
Add apt get update to build file

melzreal:
- Ensures package versions are pulled from the repository so they get correctly installed when running the CI


james (approved):
🎉

PR #138, opened 2020-11-30
Remove seeds and add unique index to subdomain

melzreal:
- Currently our deployment runs db:reset, meaning the seeds file runs before create_sample_data. This would not be a problem if find_or_create_by was working properly but instead, duplicate subdomains are being created, and users cannot be found on login because two subdomains with the same name exist in the db


Rhian (approved):
This looks fine to me.

PR #139, opened 2020-12-01
Fix loading of rswag libraries

james:
As described in the rswag README, there are two ways of loading the rswag library. We were doing it both ways.

This removes the loading of the whole rswag library for all environments, and updates the requires in the initializers to reflect this.


melzreal:
🚀

melzreal (approved):

PR #141, opened 2020-12-03
Allow plans via api

Rhian:
### Description of change

This adds a method for uploading and tagging plans that are attached to planning applications submitted via the API. Currently, this represents the happy path only and we need to add some more specific validation and error handling further down the line.


app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -36,14 +37,30 @@ def attach_question_flow(questions)
...
+  def upload_drawings(drawing_params)
+    unless drawing_params.nil?
+    begin
+    drawing_params.each do |param|
+      drawing = @planning_application.drawings.create(tags: Array(param[:tags]))
+      drawing.plan.attach(io: File.open(open(param[:filename])), filename: "#{new_plan_filename(param[:filename])}")
+    rescue
+    end
+    end
+    end
```
james, line 49: This is interesting indentation. Surprised this is how rubocop wants it

Rhian: Running Rubocop without -a meant that it still passed AND looks better :) But yes, really odd that it autocorrected to this!

james (approved):

PR #142, opened 2020-12-07
Switch to deploying on EC2 instances

pixeltrix:
This uses the same deployment pattern as UK Petitions/Welsh Petitions.


james (approved):

james:
[Developers able to log in to production via SSH](https://trello.com/c/BuwiyXJA/24-developers-able-to-log-in-to-production-via-ssh)

PR #146, opened 2020-12-09
Enforce local authority relationship to user

Rhian:
We needed to prevent the creation of users who do not belong to a local authority. Spec also added, and so was the creation of admin users when the sample data task is run.


james (changes requested):
We have decided to remove the admin user instead

james (approved):
🎉

james:
[BUG: 500 error logged in as admin when attempting to assess proposal](https://trello.com/c/5DNpfC8d/28-bug-500-error-logged-in-as-admin-when-attempting-to-assess-proposal)

PR #149, opened 2020-12-15
Github action to deploy to preview

james:
On push to master, run `rake deploy:preview`

----

I can't find an easy way to try running this without it being merged into master, which is a bit annoying.

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,0 +1,34 @@
+name: Deploy to Preview
+
+on:
+  push:
+    branches: [master]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Setup ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 2.6
+
+    - name: Install gems
+      run: |
+        gem install bundler
+        bundle install --jobs 4 --retry 3 --path vendor/bundle
+
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        aws-region: eu-west-2
+
+    - name: Deploy application
+      run: |
+        bundle exec rake deploy:preview


.github/workflows/preview.yml
```diff
@@ -0,0 +1,35 @@
+name: Deploy to Preview
+
+on:
+  push:
+    branches: [main]
```
pixeltrix, line 5: Did you copy this from the Welsh Petitions repo? If so the default branch name is `main` there so this will need to be `master` to work.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,35 @@
...
+    needs: [build]
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Setup ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 2.6.6
```
pixeltrix, line 19: I think if you use `2.6` it should use the latest patch version

.github/workflows/preview.yml
```diff
@@ -0,0 +1,35 @@
+name: Deploy to Preview
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    needs: [build]
```
pixeltrix, line 10: There's no `build` job in this file (or any other workflow file - build.yml has a job called `test`) - unfortunately I don't think GitHub Actions currently has a way to include a job from another file so you'll need to include the job in this file. On Welsh Petitions the GHA file for building pull requests uses this to restrict it to just PRs that target the main branch:

```
on:
  pull_request:
    branches: [main]
```

Obviously you'll need to change `main` to `master` if you want to use a similar config in build.yml.

pixeltrix (changes requested):

.github/workflows/preview.yml
james: Aha, I thought [main] might be a github keyword for the primary branch. Updated.

.github/workflows/preview.yml
james: 👍

.github/workflows/preview.yml
james: Given we have the tests run on all commits and PRs (which can't be merged into master unless they are passing), I thought I'd speed up deploys by not running the test suite again (and duplicate all the config for running those tests).

Missed this `needs` though, so I've removed it. Sound reasonable?

pixeltrix: That's fine but merging a PR counts as a push to the master branch so it will run tests anyway - is that what you intended?

james: I don't mind the tests running, I just don't think we need the deploy task to depend on the tests passing (as they will already have done on the branch)

pixeltrix (approved):

PR #150, opened 2020-12-15
Use 'master' branch in package-builder

james:
I believe the below commit message is accurate, but let me know if I've misunderstood something.

This code was copied from another project, who's production branch was called 'main', whereas ours is 'master'.

This was causing the [deploy to skip in Gtihub actions](https://github.com/unboxed/bops/runs/1557472572?check_suite_focus=true), as it didn't think it was on the correct branch.

### DIFF
--- a/lib/package_builder.rb
+++ b/lib/package_builder.rb
@@ -159,8 +159,8 @@ class PackageBuilder
     end
   end
 
-  def main_ref?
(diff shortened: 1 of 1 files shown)


pixeltrix:
Good catch

pixeltrix (approved):

PR #151, opened 2020-12-15
Allow editing of application drawings

james:
- New actions on drawings controller: edit and update
- spec to test valid application, invalid file format and unauthorised access
- moved some shared components between edit and new forms into partials

I have decided not to use the DrawingWizard conventions here, as this is not a multi-stage form.

### DIFF
--- a/app/views/drawings/index.html.erb
+++ b/app/views/drawings/index.html.erb
@@ -60,6 +60,7 @@
 
       <div class="govuk-grid-column-one-quarter">
         <p style="text-align: right" class="govuk-body">
+          <%= link_to "Edit", edit_planning_application_drawing_path(@planning_application, drawing) %><br />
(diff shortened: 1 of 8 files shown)


james:
[Edit document](https://trello.com/c/U6B2CH14/53-edit-document)

app/views/drawings/index.html.erb
```diff
@@ -60,6 +60,7 @@
 
       <div class="govuk-grid-column-one-quarter">
         <p style="text-align: right" class="govuk-body">
+          <%= link_to "Edit", edit_planning_application_drawing_path(@planning_application, drawing) %><br />
```
Rhian, line 63: Should the reviewer be able to edit this? If not, we should probably reinstate `if current_user.assessor? `

Rhian (approved):
Everything looks good to me.

PR #153, opened 2020-12-17
Refactor statuses

Rhian:
### Description of change

Four further statuses have been added: Not started, Invalidated, Returned and Withdrawn. Schema has been updated to use strings instead of an enum.

The AASM gem has been added so that we can set rules for transitions. Guards will be added in further commits.

In the UI, the tabs on the index page have been refactored to add Not started/invalidated applications and to rename the Determined tab to Closed.

### Story Link

https://trello.com/c/PnFbvAhB/41-refactor-statuses-and-update-tabs


app/models/planning_application.rb
```diff
@@ -41,6 +44,37 @@ class PlanningApplication < ApplicationRecord
     scope status_string.to_sym, -> { where(status: status_string) }
   end
```
james, line 48: I think adding AASM means that my metaprogramming hacks to add the getters, setters and scopes are redundant now? What happens if we remove them?

app/controllers/planning_applications_controller.rb
```diff
@@ -26,24 +26,53 @@ def show
   def edit
   end
 
-  # rubocop:disable Metrics/MethodLength
   def update
```
james, line 29: If this method isn't doing anything, and isn't being used, can we delete it and remove it from the routes?

app/policies/planning_application_policy.rb
```diff
@@ -27,7 +30,7 @@ def update?
   end
 
   def unpermitted_statuses
```
james, line 32: I think this manual checking of `permitted_statuses` is now handled by AASM, unless I'm misreading what it is doing.

app/views/planning_applications/_planning_application_table.html.erb
```diff
@@ -36,14 +36,14 @@
...
-        <% elsif id == "determined" %>
+        <% elsif id == "closed" %>
           <td class="govuk-table__cell"><%= planning_application.determined_at.strftime("%e %b") %></td>
         <% end %>
         <td class="govuk-table__cell">
           <% if planning_application.user %>
             <%= planning_application.user.name %>
           <% else %>
-            <strong class="govuk-tag govuk-tag--purple">Not started</strong>
+            <strong class="govuk-tag govuk-tag--purple">Not assigned</strong>
```
james, line 46: Can we ask Martyn about making this not a purple tag? I think it suggests it's a state, which it isn't really.

app/views/planning_applications/_planning_application_tabs.erb
```diff
@@ -1,22 +1,23 @@
...
-    <% end %>
-    <li class="govuk-tabs__list-item" role="presentation">
-      <%= link_to "Awaiting manager's determination", "#awaiting_determination", class: "govuk-tabs__tab", role: "tab", "aria-controls":"awaiting-determination", "aria-selected": false, tabindex: 1 %>
-    </li>
-    <li class="govuk-tabs__list-item" role="presentation">
-      <% if current_user.assessor?%>
-        <%= link_to "Corrections requested (#{@planning_applications.awaiting_correction.count})", "#awaiting_correction", class: "govuk-tabs__tab", role: "tab", "aria-controls":"awaiting-determination", "aria-selected": false, tabindex: 2 %>
-      <% else %>
-        <%= link_to "Corrections requested", "#awaiting_correction", class: "govuk-tabs__tab", role: "tab", "aria-controls":"awaiting-determination", "aria-selected": false, tabindex: 2 %>
+        <% unless current_user.reviewer? && exclude_others? %>
```
james, line 3: I'm a bit confused about what `exclude_others?` does here?

Rhian: I was too! I looked it up and it's essentially a filter that allows either the assessor or the reviewer to see all applications that are currently sitting in assessment (the default view is for the assessor to see only their own, and for the reviewer to not see any in assessment planning applications at all). I think we need to refactor this because the new statuses mean it makes less sense.

spec/factories/planning_application.rb
```diff
@@ -57,6 +65,11 @@
     end
   end
 
+  trait :invalidated do
+    status        { :invalidated }
+    determined_at { Time.current }
```
james, line 70: Wouldn't an invalidated document have a nil determined_at timestamp?

spec/models/planning_application_spec.rb
```diff
@@ -39,31 +39,235 @@
...
+
+        subject.return
+      end
+
+      it "sets the status to returned" do
+        expect(subject.status).to eq "returned"
+      end
+
+      it "sets the timestamp for returned_at to now" do
+        expect(subject.send("returned_at")).to be_within(1.second).of(Time.current)
```
james, line 83: I would much prefer to use [Timecop](https://github.com/travisjeffery/timecop) or [ActiveSupport::Testing::TimeHelpers](https://api.rubyonrails.org/v5.2.4/classes/ActiveSupport/Testing/TimeHelpers.html) to fix the time during these tests, rather than assuming they will run within a second.

spec/requests/planning_application_request_spec.rb
```diff
@@ -1,103 +1,104 @@
...
+#             planning_application.reload.status
+#           }
+#
+#           expect(response.code).to eq "302"
+#           expect(response).to redirect_to root_path
+#         end
+#       end
+#     end
+#   end
+# end
```
james, line 104: Can we delete this file?

james:
[Refactor statuses and update tabs](https://trello.com/c/PnFbvAhB/41-refactor-statuses-and-update-tabs)

app/models/planning_application.rb
Rhian: Good suggestion - I needed to update some of the methods in the specs to accommodate this, but there is much less duplication now.

app/policies/planning_application_policy.rb
Rhian: I've removed this - but left the permitted_status method for now as this relates to user privilege.

app/views/planning_applications/_planning_application_table.html.erb
Rhian: Martyn agreed, so this is now an empty string and the spec has been removed.

app/views/planning_applications/_planning_application_tabs.erb
Rhian: We can refactor at a later date.

spec/factories/planning_application.rb
Rhian: This was a bad copy-paste - thanks for spotting.

spec/models/planning_application_spec.rb
Rhian: I've updated to use TimeHelpers to avoid third-party dependencies.

spec/requests/planning_application_request_spec.rb
Rhian: Deleted.

james (approved):
🎉

PR #154, opened 2020-12-18
Add payment_reference to planning applications

james:
New attribute on PlanningApplication for payment_references
Show this value in information accordion
Accept this value via the API

Note: there isn't really a good spec for the API part of this, as there are currently no good specs for the API.

### DIFF
--- a/app/controllers/api/v1/planning_applications_controller.rb
+++ b/app/controllers/api/v1/planning_applications_controller.rb
@@ -78,7 +78,8 @@ class Api::V1::PlanningApplicationsController < Api::V1::ApplicationController
                       :applicant_first_name, :applicant_last_name,
                       :applicant_phone, :applicant_email,
                       :agent_first_name, :agent_last_name,
-                      :agent_phone, :agent_email, :questions, :plans]
+                      :agent_phone, :agent_email, :questions, :plans,
+                      :payment_reference]
     params.permit permitted_keys
   end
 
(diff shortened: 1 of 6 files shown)


james:
[Payment reference](https://trello.com/c/IBuC48UA/3-payment-reference)

app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -78,7 +78,8 @@ def planning_application_params
                       :applicant_first_name, :applicant_last_name,
                       :applicant_phone, :applicant_email,
                       :agent_first_name, :agent_last_name,
-                      :agent_phone, :agent_email, :questions, :plans]
+                      :agent_phone, :agent_email, :questions, :plans,
+                      :payment_reference]
```
Rhian, line 82: I wonder if, longer term, there is a possibility that there will be more than one payment reference? I think for now this is sufficient and we can go ahead and merge, but I wanted to raise the question. Also, do we have a requirement for making this a mandatory field and validating against it? I couldn't see it in the A/Cs but again, just asking.

Rhian (approved):
Looks good to me.

app/controllers/api/v1/planning_applications_controller.rb
james: These are two good product questions, but for MVP, I think we should assume that there is only one payment reference, and that it is optional for now.

PR #155, opened 2020-12-21
Swagger refactor

james:
A pretty major overhaul of how we do swagger (now [OAS3](https://swagger.io/blog/news/whats-new-in-openapi-3-0/)) documentation. I went on quite a journey, and some of the commits in this PR are redundant by the end of the PR, but I think it's useful to show how I got to where I ended up.

I'd like to add a Full example to the oas3 document, but I wanted to get your thoughts before I continued on this path.


Rhian (approved):
This is a much better and clearer way to present these and I don't have any suggestions for improvement!

PR #156, opened 2020-12-22
Add form for withdrawing and returning applications

Rhian:
cancel_index
cancel

### Description of change

Created cancel and cancel_confirmation routes which allow all types of user to withdraw or return an application and add a comment. Also added logic for displaying status on the index UI and reinstated user creation to create_sample_data rake task, which was removed earlier in error
### Story Link

https://trello.com/c/Md5k7OGz/2-withdraw-return-applications

### Known issues [OPTIONAL]

We will need to add a ticket for showing on the application page that it has been cancelled, and possibly for displaying the comment.

### DIFF
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -3,7 +3,8 @@
 class PlanningApplicationsController < AuthenticationController
   include PlanningApplicationDashboardVariables
 
-  before_action :set_planning_application, only: [ :show, :edit, :assess, :determine, :request_correction, :cancel_confirmation, :cancel ]
+  before_action :set_planning_application, only: [ :show, :edit, :assess, :determine, :request_correction,
+                                                   :cancel_confirmation, :cancel ]
   before_action :set_planning_application_dashboard_variables,
                 only: [ :show, :edit, :assess, :determine, :request_correction, :cancel_confirmation, :cancel ]
 
@@ -76,21 +77,17 @@ class PlanningApplicationsController < AuthenticationController
   end
 
   def cancel
-    status = authorize_user_can_update_status(
-        planning_application_params[:status])
+    status = authorize_user_can_update_status(params[:planning_application][:status])
     apply_cancellation(status)
-    @planning_application.update!(cancellation_comment: planning_application_params[:cancellation_comment])
-    if @planning_application.withdrawn? || @planning_application.returned?
-      flash[:notice] = "Application has been cancelled"
-    end
+    flash[:notice] = "Application has been cancelled"
     redirect_to @planning_application
   end
 
   def apply_cancellation(status)
     if status == "withdrawn"
-      @planning_application.withdraw!
+      @planning_application.withdraw!(:withdrawn, params[:planning_application][:cancellation_comment])
     elsif status == "returned"
-      @planning_application.return!
+      @planning_application.return!(:returned, params[:planning_application][:cancellation_comment])
     end
   end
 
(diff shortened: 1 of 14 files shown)


app/controllers/planning_applications_controller.rb
```diff
@@ -71,14 +71,37 @@ def withdraw
...
 
+  def cancel_confirmation
+    render :cancel_confirmation
+  end
+
+  def cancel
+    status = authorize_user_can_update_status(
+        planning_application_params[:status])
+    apply_cancellation(status)
+    @planning_application.update!(cancellation_comment: planning_application_params[:cancellation_comment])
```
james, line 82: Could we add the cancellation_comment as an optional parameter to the `withdrawn` and `returned` aasm events? Something like this? https://github.com/aasm/aasm/issues/506#issuecomment-532434609

app/controllers/planning_applications_controller.rb
```diff
@@ -71,14 +71,37 @@ def withdraw
...
+    if @planning_application.withdrawn? || @planning_application.returned?
```
james, line 83: Is there a situation where this wouldn't be true? Do we have a test for the case where it wouldn't be true? If not, I'd remove this `if` statement, as we always expect it to be true.

app/controllers/planning_applications_controller.rb
```diff
@@ -71,14 +71,37 @@ def withdraw
...
+    end
+  end
+
   private
 
   def set_planning_application
     @planning_application = authorize(PlanningApplication.find(params[:id]))
   end
 
   def planning_application_params
```
james, line 103: This has made me realize that we're doing any bulk attribute updating in this controller at all? If that's the case, we can get get rid of this method and it's use of strong parameters.

app/views/planning_applications/_assessment_dashboard.html.erb
```diff
@@ -29,6 +29,11 @@
 </div>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds application">
+    <% unless @planning_application.determined? || @planning_application.returned? || @planning_application.withdrawn? %>
```
james, line 32: I think we should have a method on PlanningApplication called `cancelable?` or similar for this.

In fact, further to that, I _think_ AASM provides us `@planning_application.may_return?`

app/views/planning_applications/_planning_application_table.html.erb
```diff
@@ -28,16 +32,22 @@
...
         <td class="govuk-table__cell"><%= planning_application.target_date.strftime("%e %b") %></td>
         <td class="govuk-table__cell">
-          <strong class="govuk-tag govuk-tag--<%= days_color(planning_application.days_left) %>">
-            <%= planning_application.days_left %>
-          </strong>
+          <% if id == "closed" %>
+            <strong class="govuk-tag govuk-tag--<%= display_status(planning_application)[:color] %>">
+              <%= display_status(planning_application)[:decision] %>
+            </strong>
+            <% else %>
```
james, line 39: The indentation is a bit off here

spec/system/planning_applications/cancelling_spec.rb
```diff
@@ -0,0 +1,286 @@
...
+  let!(:policy_evaluation) do
+    create :policy_evaluation,
+            planning_application: planning_application,
+            policy_considerations: [policy_consideration_1, policy_consideration_2]
+  end
+
+  let!(:drawing) do
+    create :drawing, :with_plan, :proposed_tags,
+           planning_application: planning_application
+  end
```
james, line 44: I think drawing, policy_evaluation, policy_consideration_2 and policy_consideration_1 aren't used or needed for these tests, so we can remove them?

james (approved):
Apart from all my niggles, functionally this is all good! I'd appreciate the bits of tidying up I've commented, but none are a blocker to merging.

app/views/planning_applications/_assessment_dashboard.html.erb
Rhian: I need to investigate further why `.may_return?` and `.may_withdraw?` didn't work as expected. For the moment I've gone with   `true unless determined? || returned? || withdrawn?`

PR #157, opened 2020-12-22
Add 'Full' example to API for application posting

james:
Add an example to the swagger doc that includes values for every parameter we currently support.

Add a spec that uses this example and tests that it gets saved accurately.


Rhian (approved):
This is much better!

PR #158, opened 2020-12-23
Update planning applications list in API

james:
Add more attributes that we expose through the API, add a new show endpoint, and a few other tweaks explained in the commit messages.


james:
[Make application data available in API](https://trello.com/c/TLSgEclJ/8-make-application-data-available-in-api)

app/views/api/v1/planning_applications/_show.json.jbuilder
```diff
@@ -1,10 +1,35 @@
...
-json.application_type t(planning_application.application_type)
-json.summary_of_proposal planning_application.description
-json.received_date planning_application.created_at.iso8601
-json.determined_at planning_application.determined_at.iso8601
-json.status planning_application.reviewer_decision.status
+json.site do |site_json|
+  site_json.partial! "site.json.jbuilder", site: planning_application.site
+end
+json.received_date planning_application.created_at
+json.decision planning_application.reviewer_decision.status if planning_application.reviewer_decision
```
Rhian, line 33: Are we deliberately not including assessor_decision? I can see the logic if this is a choice, but wanted to be sure

spec/requests/planning_application_list_spec.rb
```diff
@@ -32,68 +28,99 @@
...
-      )
+    def json_time_format(time)
+      time.utc.iso8601(3) if time.present?
     end
 
-    it "returns an empty response if the planning application is in assessment" do
-      planning_application_1 = create(:planning_application)
-      decision_granted = create(:decision, :granted, user: reviewer, planning_application: planning_application_1)
-
+    it "returns an empty response if no planning application" do
```
Rhian, line 37: I'm confused about this - should this be expected to be a 404?

spec/requests/planning_application_list_spec.rb
```diff
@@ -32,68 +28,99 @@
...
+        expect(planning_application_json["site"]["uprn"]).to eq(planning_application.site.uprn)
+        expect(planning_application_json["questions"]).to eq(JSON.parse(planning_application.questions))
+        expect(planning_application_json["constraints"]).to eq(JSON.parse(planning_application.constraints))
+      end
+
+      context "for a granted planning application" do
+        let!(:planning_application) { create(:planning_application, :determined) }
+        let!(:decision) { create(:decision, :granted, user: reviewer, planning_application: planning_application) }
+
+        it "returns the accurate data" do
```
Rhian, line 87: I'm pondering whether these specs might be split into several different ones so we can more easily see where any failures are: for example, one spec for the site info, one for contact details of agent/applicant, or whatever? That way if we have a second failure after the test has stopped, it would be more quickly identified

Rhian (approved):
This is a much better way of creating and documenting the API!

app/views/api/v1/planning_applications/_show.json.jbuilder
james: This is how I found it, and I didn't want to tweak it too much before we do our decision refactor

spec/requests/planning_application_list_spec.rb
james: No, it would be a 404 if it was the show action, this is the index action

spec/requests/planning_application_list_spec.rb
james: I definitely thought about that, but realised that unless I did a before(:all), which I try and avoid, it would be making dozens of setups and teardowns every test run.

PR #159, opened 2020-12-30
Add document validation step

Rhian:
validate3
validate2
validate1

This adds a form for the assessor to validate the documents. The application cannot transition to in_assessment without the documents_validated_at field being populated. The assessor also has the option to invalidate the application, and in each case the planning application appears correctly flagged in the index table.

### Story Link

https://trello.com/c/wNJMfBnm/5-indicate-documents-are-valid-invalid

### Decisions 

We made a decision to use the GovUK Formbuilder gem. An initial attempt to use the Ministry of Justice `gov_uk_date_fields` gem was abandoned after discovering that its date fields do not work when nested under Formbuilder fields.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,8 +7,6 @@ class ApplicationController < ActionController::Base
   attr_reader :current_local_authority
   helper_method :current_local_authority
 
-  default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
-
   private
 
   def find_current_local_authority_from_subdomain
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,6 +7,8 @@ class ApplicationController < ActionController::Base
   attr_reader :current_local_authority
   helper_method :current_local_authority
 
+  default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
+
   private
 
   def find_current_local_authority_from_subdomain
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -112,7 +112,7 @@ class PlanningApplicationsController < AuthenticationController
   def apply_validation(status)
     if status == "invalidated"
       @planning_application.invalidate!
-    elsif status == "in_assessment" && document_validation_params.values.join("-").match(/\d{2}-\d{2}-\d{4}/)
+    elsif status == "in_assessment"
       update_validation_and_start
     else
       @planning_application.errors.add(:planning_application, "Please choose Yes or No")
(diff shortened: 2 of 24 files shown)


app/controllers/planning_applications_controller.rb
```diff
@@ -109,6 +148,12 @@ def unpermitted_status_for_user?(status)
     status == :awaiting_determination if current_user.assessor?
   end
 
+  def document_validation_params
+    params.require(:planning_application).permit(:status,  :'documents_validated_at(3i)',
+                                                 :'documents_validated_at(2i)',
+                                                 :'documents_validated_at(1i)')
```
james, line 154: We're not doing any bulk attributes, so don't need to use strong params, and I'm a bit confused by the inclusion of `status` in here. Won't `document_validation_params.values.join("-")` = 'not_started-31-12-2020'? Is that why the `match` is used further above?

Rhian: `:status` was a regression that I left in by accident. I guess maybe a better approach would be to define and join the three date field params in a method other than using strong params, as we need to just define them rather than explicitly permit them? Unless you can think of a better way, which I'm sure there is.

app/controllers/application_controller.rb
```diff
@@ -7,6 +7,8 @@ class ApplicationController < ActionController::Base
   attr_reader :current_local_authority
   helper_method :current_local_authority
 
+  default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
```
james, line 10: I note that we have this, but also have explicilty set the builder on the `form_with` in `app/views/drawings/index.html.erb`. Is one of these redundant?

db/schema.rb
```diff
@@ -38,8 +38,8 @@
   end
 
   create_table "api_users", force: :cascade do |t|
-    t.string "name"
-    t.string "token"
+    t.string "name", default: "", null: false
+    t.string "token", default: "", null: false
```
james, line 42: Where did this change come from?

spec/system/planning_applications/validating_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+RSpec.describe "Planning Application Assessment", type: :system do
+  let(:local_authority) {
+    create :local_authority,
+    name: "Cookie authority",
+    signatory_name: "Mr. Biscuit",
+    signatory_job_title: "Lord of BiscuitTown",
+    enquiries_paragraph: "reach us on postcode SW50",
+    email_address: "biscuit@somuchbiscuit.com"
+    }
```
james, line 13: Tiniest nitpick, but if would be good to be consistent between using `do...end` for multiline blocks or `{}` in this file.

app/controllers/planning_applications_controller.rb
james: Yup, I agree, I think a method along the lines of `date_string_from_params` would probably be neater :-)

james (changes requested):
All looks great, apart from planning_applications_controller.rb, which as I have commented think could be a bit clearer to read. Happy to let you have another stab at it, or chat about it directly :-)

db/schema.rb
Rhian: That's a good question, as I don't remember doing this, and it's certainly not part of this story. Am assuming a bad merge somewhere down the line - I will revert it, thanks

james (approved):
🎉

PR #160, opened 2020-12-30
Rename Drawings to Documents everywhere

james:
We want to be more generic when talking about documents. Documents could be drawings, but they also could be invoices or other kinds of documents. At the moment, we don't have a way of distinguishing between these different types of documents, which we may do in the future, but for now, we will call them all documents.


james:
[Drawings/Documents naming](https://trello.com/c/LyHR6ZTj/62-drawings-documents-naming)

Rhian (approved):
Looks good to me

PR #162, opened 2020-12-30
Reveal documents in API

james:
Adds documents array to show and index actions of the API. This means that all documents are revealed to the public.

We probably want to be able to limit access to some documents in the future, so rather than changing the storage configuration to make all files public, I have created a new action that redirects to the temporary URLs that ActiveStorage creates. This will make it easier to add conditionals to it later.


james:
[Make documents available via API](https://trello.com/c/hl7AGv8a/54-make-documents-available-via-api)

Rhian (approved):
All looks good to me.

PR #163, opened 2020-12-31
Add new workplace_status field

Rhian:
### Description of change

We need a new field to indicate whether the works for which permission is being applied for are Proposed or Existing. I've renamed this field `work_status` as I felt that `completed` could be too easily confused with application overall status. In addition to displaying this as a Yes/No field on the application information template, we also need to display Proposed/Existing status interpolated into the text in the Decision Notice and mail template.

### Story Link

https://trello.com/c/1Yucb479/1-proposed-or-existing-development

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -40,8 +40,6 @@ en:
     recommendation_step: "View the decision notice"
   full: "Full Householder Application"
   lawfulness_certificate: "Certificate of Lawfulness"
-  Proposed: "proposed"
-  Existing: "existing"
   user_not_authorized: "You are not authorized to perform this action."
   scale: "Missing scale bar or north arrow"
   design: "Revise design"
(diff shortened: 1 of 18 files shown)


config/locales/en.yml
```diff
@@ -40,6 +40,8 @@ en:
     recommendation_step: "View the decision notice"
   full: "Full Householder Application"
   lawfulness_certificate: "Certificate of Lawfulness"
+  Proposed: "proposed"
+  Existing: "existing"
```
james, line 44: This all looks great, with one minor thing: we store everything else like this as lowercase in the DB and API, and use i18n to `humanize` if needed when presenting it, and for this, we're doing the opposite. I think it would be best if we were consistent and did the same here?

james:
[Proposed or existing development](https://trello.com/c/1Yucb479/1-proposed-or-existing-development)

james:
[Proposed or existing development](https://trello.com/c/1Yucb479/1-proposed-or-existing-development)

james (approved):

PR #166, opened 2021-01-04
Tell rubocop package_builder doesn't use Rails

james:
Rubocop assumed that package_builder.rb had Rails loaded, and therefore autocorrected Time.now to Time.zone.now. Undo that, and exclude Rails rules from this file.


Rhian (approved):
All looks good to me.

PR #168, opened 2021-01-04
Add work_status to API output fields

Rhian:
The GET request needs to include work_status so this has been added to jsonbuilder and relevant specs


spec/factories/planning_application.rb
```diff
@@ -8,6 +8,8 @@
     status           { :in_assessment }
     in_assessment_at { Time.zone.now }
     ward             { Faker::Address.city }
+    documents_validated_at { Time.zone.today }
```
melzreal, line 11: this one doesn't need to be part of the exposed api field too?

melzreal (approved):
💃 LGTM

spec/factories/planning_application.rb
Rhian: that's a good question - I don't think it's specified at the moment, so I didn't add it, but I need to check!

PR #173, opened 2021-01-08
Scope planning applications in API to local authority

james:
Previously we were showing all planning applications in the system through the API. This was wrong and we only want to show the applications relevant to the current local authority (as specified by subdomain)

I've also disabled the RSpec/InstanceVariable cop, as this is the cleanest way I can think of managing the subdomain for the specs.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -22,5 +22,8 @@ RSpec/ContextWording:
 RSpec/LetSetup:
   Enabled: false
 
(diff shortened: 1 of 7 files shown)


james:
[Make application data available in API](https://trello.com/c/TLSgEclJ/8-make-application-data-available-in-api)

Rhian (approved):
Looks good to me.

PR #174, opened 2021-01-11
Update target_date when documents validated

james:
Before validation, the target date is 8 week from the creation of the application, but this might change when the documents are validated. This updates the target_date.

Implementation note: I am using both `before_create` and `before_update` here, as `before_save`, in the current version of Rails, happens before the created_at timestamp is created, which results in trying to do maths on nil.

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -25,6 +25,7 @@ class PlanningApplication < ApplicationRecord
   belongs_to :local_authority
 
   before_create :set_target_date
+  before_update :set_target_date
 
   validate :assessor_decision_associated_with_assessor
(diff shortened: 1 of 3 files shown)


Rhian (approved):
I'm probably overthinking this, but maybe we could add a couple of specs to ensure nothing untoward is happening when the assessor sets an unexpected validation date, possibly by accident - for example, before the creation date or several months into the future. This could be more of a product question actually - do we need some kind of warning or logic if such a validation date is selected?

james:
It's an interesting thought, but I think is outside this ticket. I also don't think it breaks anything for the app, it would just be a weird thing to do, and I'm hesitant to put hard limits into the application for things that don't cause actual application errors.

PR #175, opened 2021-01-11
Remove validation form from closed applications

Rhian:
### Description of change

This means the form for validating documents will not display on applications that are determined, returned or withdrawn.

### Story Link

https://trello.com/c/M7nZkxgu/79-when-application-determined-withdrawn-returned-remove-validation-control-in-manage-docs

### DIFF
--- a/app/views/documents/index.html.erb
+++ b/app/views/documents/index.html.erb
@@ -18,7 +18,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <% if current_user.assessor? && (!@planning_application.determined? || !@planning_application.withdrawn? || !@planning_application.returned?) %>
+    <% if @planning_application.cancellable? && current_user.assessor? %>
       <%= form_with model: @planning_application, url: validate_documents_planning_application_path(@planning_application), local: true, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |form| %>
         <fieldset class="govuk-fieldset">
           <% if form.object.errors[:planning_application].any? %>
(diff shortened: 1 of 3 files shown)


app/views/documents/index.html.erb
```diff
@@ -18,7 +18,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <% if current_user.assessor? && (!@planning_application.determined? || !@planning_application.withdrawn? || !@planning_application.returned?) %>
+    <% if @planning_application.cancellable? && current_user.assessor? %>
```
james, line 21: So from my reading, this commit doesn't actually change the functionality? planning_application.cancellable? is the same as what was there before. This is a much nicer refactor though.

app/views/documents/index.html.erb
```diff
@@ -38,9 +38,9 @@
                 <%= form.submit "Save", class: "govuk-button", data: { module: "govuk-button" } %>
                 <%= link_to 'Cancel', planning_application_path(@planning_application), class: "govuk-button govuk-button--secondary" %>
               </p>
-           </div>
-      <% end%>
-    <% end %>
+        <% end%>
+     <% end %>
+    </div>
```
james, line 43: I think the indentation is wrong here, and making it hard for me to see if everything is closed correctly.

app/views/documents/index.html.erb
Rhian: Yes, but it's totally weird... this works as expected and for some reason the previous iteration didn't seem to.

app/views/documents/index.html.erb
Rhian: You were right: the indentation was wrong. I've force-pushed a fix for it.

james (approved):

app/views/documents/index.html.erb
james: Hmm, I think maybe we're getting confused with logical arguments. This is why we write tests :-D

PR #176, opened 2021-01-12
Remove edit document numbers

Rhian:
This removes not only the link from the task list, but the Edit Numbers view itself and changes the logic around which documents are displayed on the decision notice (now all numbered documents are shown, not only proposed). The constraint for all proposed documents being numbered before application can be submitted is also removed.

The Document List Form and its spec have also been removed.

### Story Link

https://trello.com/c/0G6svB8O/37-remove-add-drawing-numbers-step-and-change-logic-for-how-they-are-displayed-on-decision-notices


spec/system/documents/edit_documents_spec.rb
```diff
@@ -30,6 +30,8 @@
     it "with valid data" do
       click_link "Edit"
 
+      expect(page).to have_content("If you add one or more numbers to this document, it will be published in the decision notice")
```
james, line 33: This is the tiniest thing, but are we actually testing anything with this assertion? There's no logic to this text being there, it will always be there.

spec/system/documents/edit_numbers_spec.rb
```diff
@@ -12,30 +12,12 @@
   let(:assessor) { create :user, :assessor, local_authority: local_authority }
   let(:reviewer) { create :user, :reviewer, local_authority: local_authority }
 
-  context "as a user who is not logged in" do
```
james, line 15: Maybe this file just needs renaming, or maybe it needs deleting altogether? But seeing as "Edit document numbers page" doesn't exist any more, what is this testing, and is it tested elsewhere?

james (approved):
Great work!

Just a couple of tiny testing related comments, but neither are blockers.

spec/system/documents/edit_documents_spec.rb
Rhian: That's true. I figured as it was specified in the A/Cs that we need this text on the Edit page, I should have a test for it. But then it's not likely to be something that intermittently disappears, so maybe a waste of an assertion. I will remove

spec/system/documents/edit_numbers_spec.rb
Rhian: I think renaming is the way to go. This tests an area of functionality that is not tested elsewhere (verifying that numbers can be added to documents at varying points in the journey and that those numbers then appear on the decision notice) so we need to keep it BUT the current filename sounds like we are testing the now-deleted route. I will change it to `numbering_documents.spec` for clarity.

PR #179, opened 2021-01-14
Send email on validation

Rhian:
### Description of change

This sends a notification via Notify to the applicant's email once their application has been validated and makes the "email not sent" error message for Determination more generic

### Story Link

https://trello.com/c/VK9OMXtT/32-notify-applicant-when-application-is-marked-as-valid

### DIFF
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -114,7 +114,8 @@ class PlanningApplicationsController < AuthenticationController
       else
         @planning_application.documents_validated_at = date_from_params
         @planning_application.start!
-        flash[:notice] = "Application is ready for assessment"
+        validation_notice_mail
+        flash[:notice] = "Application is ready for assessment and applicant has been notified"
         redirect_to @planning_application
       end
     elsif status == "invalidated"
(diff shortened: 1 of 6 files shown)


app/controllers/planning_applications_controller.rb
```diff
@@ -127,8 +128,6 @@ def validate_documents
     end
   end
 
-private
```
james, line 130: Why did we un-`private` these methods?

spec/system/planning_applications/reviewing_spec.rb
```diff
@@ -297,7 +297,7 @@
           click_link "Publish the recommendation"
           click_button "Determine application"
 
-          expect(page).to have_content("The Decision Notice cannot be sent. Please try again later.")
+          expect(page).to have_content("The email cannot be sent. Please try again later.")
```
james, line 300: Should we also have a feature spec that checks we send the email when validating?

Rhian: I've added a line into the validation spec to verify that the delivered mails count increments by 1 on validation, but if you can think of a more complete way to do it, please let me know. I figured we didn't need to validate the content as we are already doing this in the mailer spec.

app/controllers/planning_applications_controller.rb
Rhian: ..entirely by accident 😩 Thank you very much for picking up on this!

james (approved):
Looks good to me once you've fixed rubocop :-)

PR #180, opened 2021-01-14
Add helper text to document edit view

Rhian:
### Description of change

This addition is to replace the helper text in the  document numbers view, which was deleted.

doc_numbers

### Story Link

https://trello.com/c/NZ2sMxgm/89-helper-text-for-document-numbers

### DIFF
--- a/app/views/documents/edit.html.erb
+++ b/app/views/documents/edit.html.erb
@@ -19,9 +19,6 @@
     <p class="govuk-body" style="line-height: 1.6;">
       If you add one or more numbers to this document, it will be published in the decision notice. Only documents with numbers will appear on the decision notice.
     </p>
-    <p class="govuk-body" style="line-height: 1.6;">
-      If there are multiple numbers, please provide a comma-separated list.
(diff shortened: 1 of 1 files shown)


app/views/documents/edit.html.erb
```diff
@@ -17,7 +17,10 @@
     <h2 class="govuk-heading-m"><%= @document.name %></h2>
 
     <p class="govuk-body" style="line-height: 1.6;">
-      If you add one or more numbers to this document, it will be published in the decision notice.
```
james, line 20: I guess this is more of a design question, but should this helper text not be "helper text" for the numbers field? https://design-system.service.gov.uk/components/text-input/#hint-text

Rhian: It means it will be repeated multiple times on the page, but I guess this is correct in terms of design principles. I'll update the PR and submit it on the basis that it is the 'right' thing to do and we can discuss with design/product at UAT.

james (approved):

PR #181, opened 2021-01-15
Allow viewing of decision notices on complete applications

james:


james:
[Planners and managers should be able to view a decision notice (but not make any changes to it) after it has been determined.](https://trello.com/c/223F1w9X/76-planners-and-managers-should-be-able-to-view-a-decision-notice-but-not-make-any-changes-to-it-after-it-has-been-determined)

Rhian (approved):
Looks good to me.

PR #182, opened 2021-01-19
Add step to before_install to remove node_modules

Rhian:
Deployments were failing because disk space was filling up with node_modules. This step in before_install deletes them.


james (approved):
Thanks for sorting this out!

PR #186, opened 2021-01-20
Decision notice pdf

james:
A refactoring of the decision notice HTML, then using that to generate a PDF using PDFKit, and use that PDF in decision notice email.


james:
[Improving decision notices and generating accessible PDF versions of them and sharing them with applicants](https://trello.com/c/a8ib1g0l/93-improving-decision-notices-and-generating-accessible-pdf-versions-of-them-and-sharing-them-with-applicants)

app/views/planning_applications/_decision_notice.html.erb
```diff
@@ -63,8 +86,13 @@
        <% end %>
       </tbody>
     </table>
+  <% else %>
+    <p class="govuk-body">
+      No drawings specified
```
Rhian, line 91: Tiny question... am assuming "drawings" is OK here rather than "plans"?

app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -19,6 +20,11 @@ def show
     end
   end
 
+  def decision_notice
+    @planning_application = current_local_authority.planning_applications.where(id: params[:id]).first
+    @disable_headers = true
```
Rhian, line 25: It could be just me, but `disable_headers` momentarily made me think something was being done to the HTTP headers rather than the layout

Rhian (approved):
All looks great!

PR #187, opened 2021-01-20
Fix host for URLs in emails

james:
We can't set the default url host for emails, as we support subdomains. However currently all emails are triggered from a HTTP request, so we can borrow the host domain from that request and pass it into the mailer.


Rhian (approved):
Looks good to me.

PR #188, opened 2021-01-20
Change tagging system and add new tags

Rhian:
### Description of change

Removed the logic that combines proposed and existing tags with orientation, and got rid of some unused logic around tagging with proposed tags. Introduced a new array of tags and removed null=false from the database field to allow documents to be submitted without tags

### Story Link

https://trello.com/c/pkH3v0Cn/36-add-further-drawings-tags

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -9,33 +9,26 @@ class Document < ApplicationRecord
                          design: 1,
                          dimensions: 2,
                          other: 3 }
-
-  ORIENTATION_TAGS = [
-    "front elevation",
-    "side elevation",
-    "floor plan",
-    "section",
+  TAGS = %w[
+    Front
+    Rear
+    Side
+    Roof
+    Garden
+    Floor
+    Site
+    Plan
+    Elevation
+    Section
+    Proposed
+    Existing
   ].freeze
-
-  PROPOSED_TAGS = ORIENTATION_TAGS.product(%w[proposed]).map { |tags|
-    [tags.first, tags.second].join(" - ")
-  }.freeze
-
-  EXISTING_TAGS = ORIENTATION_TAGS.product(%w[existing]).map { |tags|
-    [tags.first, tags.second].join(" - ")
-  }.freeze
-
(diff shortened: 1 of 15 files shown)


app/models/document.rb
```diff
@@ -9,33 +9,26 @@ class Document < ApplicationRecord
...
-
   PERMITTED_CONTENT_TYPES = ["application/pdf", "image/png", "image/jpeg"].freeze
 
   validate :tag_values_permitted
   validate :file_content_type_permitted
 
-  scope :has_proposed_tag, lambda {
-    where("tags ?| array[:proposed_tag_array]",
-          proposed_tag_array: Document::PROPOSED_TAGS)
-  }
```
james, line 38: 🎉

public/api-docs/v1/swagger_doc.yaml
```diff
@@ -67,7 +67,7 @@ paths:
                         archived_at: '2020-05-16T05:18:17.540Z'
                         archive_reason: other
                         tags:
-                          - front elevation - proposed
+                          ["Side", "Elevation", "Proposed"]
```
james, line 70: Is there a reason we are not using YAML style arrays here?

spec/system/documents/number_documents_spec.rb
```diff
@@ -19,26 +19,21 @@
...
       let!(:proposed_document_1) do
-        create :document, :with_file, tags: [proposed_tag],
-                                      planning_application: planning_application
+        create :document, :with_file, :with_tags,
+               planning_application: planning_application
       end
 
       let!(:proposed_document_2) do
-        create :document, :with_file, tags: [proposed_tag],
+        create :document, :with_file, tags: [Document::TAGS.first],
```
james, line 30: you've updated `let!(:tag)`, but then seem to have stopped using it here? Could this not be `tags: [tag]`

james: in fact: I'm not sure any of these need to have tags at all?

james (approved):
Just a couple of 🏄  comments, and it needs rebasing, but this all looks great! And yet another net negative LOC!

PR #192, opened 2021-01-26
Display application status

Rhian:
### Description of change

This commit extends the existing planning application helper method which creates coloured tags according to status. Status, together with due date (if not closed) is now displayed on the assessment dashboard and target date tag is removed from the application page.

### Story Link

https://trello.com/c/r13g3U0c/103-display-an-applications-status-on-application-detail-page

### Decisions [OPTIONAL]

I used multiple elsifs in the helper rather than a case statement as I think the logic isn't straightforward enough for the latter, but there may be a better way to do it.

### DIFF
--- a/app/views/planning_applications/_assessment_dashboard.html.erb
+++ b/app/views/planning_applications/_assessment_dashboard.html.erb
@@ -19,7 +19,7 @@
 </div>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds application">
-    <% if @planning_application.cancellable? %>
+    <% if @planning_application.in_progress? %>
       <p class="govuk-body">
         <%= link_to "Cancel application", cancel_confirmation_planning_application_path(@planning_application) %>
       </p>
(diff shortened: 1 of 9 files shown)


app/views/shared/_status_panel.html.erb
```diff
@@ -0,0 +1,19 @@
+<p class="govuk-body">
+  <% if @planning_application.cancellable? %>
```
james, line 2: This is a tiny thing, but what do you think about creating a new method for something like this? Something like `open?`, `active?` or `in_progress?`. I realise the method would probably just be `!cancellable?`, but it feels a little bit more explanatory in the logic here?

Rhian: Great idea! Will add to this PR.

app/views/planning_applications/_assessment_dashboard.html.erb
```diff
@@ -14,17 +14,7 @@
     <h2 class="govuk-heading-m" style="margin-bottom: 75px;">
```
james, line 14: Just looking at this locally, I think we can remove this manual margin-bottom now?

james (approved):
Looks good!

PR #193, opened 2021-01-26
Only show documents that are for publication in the API

james:
Previously we were revealing all documents through the API. We only want to show documents that are "for publication" (i.e. have been given a number and have not been archived). We are removing the archived_at and archive_reason from the API too, as they will never not be nil now.


james:
[Don't include documents that don't have any numbers in API](https://trello.com/c/qNYW94YJ/4-dont-include-documents-that-dont-have-any-numbers-in-api)

Rhian (approved):
From a code point of view, all looks good. However, this raises a product-type question that somehow (helper text?) we need to make planners aware that simply archiving a document will not remove it from public view if they later need to redact something, and that removing something from the API means removing the drawing number rather than archiving.

james:
I think you've got the logic backwards? (I did go round it a few times). If you archive the document it *does* get removed from the API.

Rhian:
> I think you've got the logic backwards? 

Yes, I did 😩 I had to pull it and run it for myself to prove that it was the reverse of what I thought! Sorry

PR #194, opened 2021-01-26
Add status column to all tabs on applications index

Rhian:
### Description of change

This commit tidies up the planning applications table so that we have consistent column display and naming and the status tag is always shown

### Story Link

https://trello.com/c/Vs5pjRyw/77-status-field-to-be-added-to-in-assessment-and-awaiting-determination-tables

### DIFF
--- a/app/views/planning_applications/_planning_application_table.html.erb
+++ b/app/views/planning_applications/_planning_application_table.html.erb
@@ -7,14 +7,11 @@
       <th scope="col" class="govuk-table__header">Site address</th>
       <th scope="col" class="govuk-table__header">Application type</th>
       <th scope="col" class="govuk-table__header">Target date</th>
-      <% if id == "closed" %>
-        <th scope="col" class="govuk-table__header">Decision</th>
-      <% else %>
+      <% if id != "closed" %>
         <th scope="col" class="govuk-table__header">Days left</th>
       <% end %>
-      <% if (id == "in_assessment" || id == "not_started_and_invalid") %>
-        <th scope="col" class="govuk-table__header">Status</th>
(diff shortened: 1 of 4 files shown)


app/views/planning_applications/_planning_application_table.html.erb
```diff
@@ -7,14 +7,11 @@
...
       <th scope="col" class="govuk-table__header">Application type</th>
       <th scope="col" class="govuk-table__header">Target date</th>
-      <% if id == "closed" %>
-        <th scope="col" class="govuk-table__header">Decision</th>
-      <% else %>
+      <% if id != "closed" %>
         <th scope="col" class="govuk-table__header">Days left</th>
       <% end %>
-      <% if (id == "in_assessment" || id == "not_started_and_invalid") %>
         <th scope="col" class="govuk-table__header">Status</th>
```
james, line 13: This need to be outdented by 1 level, as it looks like it's within the if statement to me

app/views/planning_applications/_planning_application_table.html.erb
```diff
@@ -31,26 +28,19 @@
         <td class="govuk-table__cell"><%= planning_application.site.full_address %></td>
         <td class="govuk-table__cell"><%= t(planning_application.application_type) %></td>
         <td class="govuk-table__cell"><%= planning_application.target_date.strftime("%e %b") %></td>
+        <% if id != "closed" %>
         <td class="govuk-table__cell">
```
james, line 32: This needs to be indented by one so it's clear it's inside an if statement.

james (approved):
This is an improvement! I still think this is a little more complicated than it should be, but can't think of anything to simplify it more than you have!

PR #195, opened 2021-01-28
Improve tags

Rhian:
### Description of change

This PR contains various improvements around document tagging and uploads:
- The null constraint for tags has been removed, and the form updated to allow a blank tag submission in a hidden field
- The tags are now displayed inline on the documents index page, rather than stacked 
- The document upload process has been reduced from two steps to one, removing the confirmation page. This means we are able to remove two of the documents wizard models
- The document wizard itself has not been removed as it contains the models for the archive two-step process

### Story Link

https://trello.com/c/pkH3v0Cn/36-add-further-drawings-tags

### Decisions [OPTIONAL]

Validation for the file attachment is done in the drawings controller and not as a validates presence in the model, as we do not always want to attach a file to every document (for example, for tests). We need to validate the attachment for the purposes of the form upload only.


app/controllers/documents_controller.rb
```diff
@@ -36,50 +35,34 @@ def archive
...
-    @document_form = DocumentWizard::UploadForm.new
+    @document = @planning_application.documents.build
   end
 
   def create
-    @document_form = DocumentWizard::ConfirmUploadForm.new(
-      document_upload_confirmation_params,
-    )
+    @document = @planning_application.documents.build
+    unless document_params[:file].nil?
```
james, line 43: Rather than checking to see if the file is present in the controller, could we not add presence validation for file in the document model? I don't think there is any situation we would want a document without a file attached? And that would simplify this controller action even more!

app/controllers/documents_controller.rb
```diff
@@ -36,50 +35,34 @@ def archive
...
+      document_params.each do |_param|
+        @blob = ActiveStorage::Blob.find_signed(document_params[:file])
```
james, line 45: Do we need to do this @blob and then .attach now? Can we not just pass the file param to the model and activestorage does magic?

app/views/documents/form_partials/_tags.html.erb
```diff
@@ -14,7 +14,7 @@
     <% end %>
 
     <div class="govuk-checkboxes govuk-checkboxes--small">
-      <%= form.collection_check_boxes :tags, Document::TAGS, :itself, :itself, include_hidden: false do |b| %>
```
james, line 17: oh hey, look at that. good catch.

james:
Ah that's so satisfying!

From my comments, I think it can be simplified a tiny bit more though! I know, I'm greedy.

james:
Oh, I realise you'd already addressed my comment in the PR description!

Hmmm. I'm not a fan of compromising the main code too much for the advantage of the tests. I'd maybe lean towards adding files to all the tests, and if we want to make them run faster, then we should do some mocking and stubbing magic.

Rhian:
>  I know, I'm greedy.

But also entirely right ;-) I will look at refactoring the tests to reflect this, and it should solve the other issues in the process!

james (approved):
🎉

PR #197, opened 2021-01-28
Improve status panel layout

Rhian:
### Description of change

After design review, implemented minor layout changes to the application show page

### Story Link

https://trello.com/c/r13g3U0c/103-display-an-applications-status-on-application-detail-page

### DIFF
--- a/app/views/shared/_status_panel.html.erb
+++ b/app/views/shared/_status_panel.html.erb
@@ -3,17 +3,19 @@
     <strong class="govuk-tag govuk-tag--<%= display_status(@planning_application)[:color] %>">
     <%= display_status(@planning_application)[:decision] %>
     </strong>
-    Due: <%= @planning_application.target_date.strftime("%d %B") %> - <%= @planning_application.days_left %>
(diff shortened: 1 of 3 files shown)


app/views/shared/_status_panel.html.erb
```diff
@@ -3,17 +3,19 @@
     <strong class="govuk-tag govuk-tag--<%= display_status(@planning_application)[:color] %>">
     <%= display_status(@planning_application)[:decision] %>
     </strong>
-    Due: <%= @planning_application.target_date.strftime("%d %B") %> - <%= @planning_application.days_left %>
+    &nbsp;Due: <%= @planning_application.target_date.strftime("%d %B %Y") %> - <%= @planning_application.days_left %>
```
james, line 6: Could we not achieve this by adding a 'planning_application_status' class to the wrapping `p` tag, and then adding something like the following CSS:

```
.planning_application_status . govuk-tag {
  margin-right: 1em;
}
```

I'd like to avoid using HTML for presentation.

Rhian: Good point. I will do something sort-of-like this, but for `span` because the extra space that was requested is in between the tag and the text that appears on the same line

james (approved):

PR #200, opened 2021-02-02
Check if document is published before revealing in API

james:
Adds a check to the API call to show a document file to ensure it is published before redirecting to temp s3 URL.

Add new helper method for using within the logged in API: if the document is published, then use the public API method, else create a temporary secure s3 URL directly.

### DIFF
--- a/spec/requests/document_show_spec.rb
+++ b/spec/requests/document_show_spec.rb
@@ -2,9 +2,9 @@
 
 require "rails_helper"
 
-RSpec.describe "API request to list planning applications", type: :request, show_exceptions: true do
+RSpec.describe "API request to show document file", type: :request, show_exceptions: true do
   let!(:planning_application) { create(:planning_application, :not_started) }
-  let!(:document) { create(:document, :with_file, planning_application: planning_application) }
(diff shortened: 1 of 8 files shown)


james:
[Don't include documents that don't have any numbers in API](https://trello.com/c/qNYW94YJ/4-dont-include-documents-that-dont-have-any-numbers-in-api)

spec/requests/document_show_spec.rb
```diff
@@ -2,9 +2,9 @@
 
 require "rails_helper"
 
-RSpec.describe "API request to list planning applications", type: :request, show_exceptions: true do
+RSpec.describe "API request to show document file", type: :request, show_exceptions: true do
   let!(:planning_application) { create(:planning_application, :not_started) }
-  let!(:document) { create(:document, :with_file, planning_application: planning_application) }
+  let!(:document) { create(:document, :with_file, :numbered, planning_application: planning_application) }
```
Rhian, line 7: This is a reminder for me to create a chore to change this in the document factory: we don't need this `with_file` trait any more as the basic document factory now contains a file, I should go through and remove these instances. Not for this PR, though.

Rhian (approved):
I like having a `published?` method for documents - it's a much clearer distinction

PR #204, opened 2021-02-11
Upgrade yarn packages

james:
We had various security warnings and development issues with our yarn packages, which are resolved by upgrading them all.


melzreal (approved):
🚀 🚀 🚀

PR #205, opened 2021-02-12
Decision refactor

james:
OK, here it is. I think this branch now meets all the acceptance criteria in the following stories:

https://trello.com/c/UKxCEXGZ/91-allow-reviewers-to-take-the-same-actions-as-assessors
https://trello.com/c/AUTIF137/112-implement-new-task-list-layout-and-logic-and-new-validation-task-page
https://trello.com/c/A65PX5G4/138-allow-and-display-multiple-comments-from-assessors-and-reviewers-design-and-dev

Given I've paired with both of you on this, and how big this PR is, this will be a slightly different PR process. Feel free to read through it all again if you want, but I think the most useful thing would be for you to pull this branch to your local dev environments, `rake db:reset`, and try using it locally! Some basic regression testing locally would be helpful before it goes onto preview for it to be properly acceptance tested.


db/seeds.rb
```diff
@@ -0,0 +1,51 @@
...
+        user.password = user.password_confirmation = "password"
+      else
+        user.password = user.password_confirmation = SecureRandom.uuid
+        user.encrypted_password =
+          "$2a$11$uvtPXUB2CmO8WEYm7ajHf.XhZtBsclT/sT45ijLMIELShaZvceW5."
+      end
+      user.role = admin_role
+    end
+  end
+end
```
melzreal, line 51: I just pulled the branch and reset the db - now there are no planning applications, which we were creating before with create_sample_data?
Also I remember there used to be a dependency on create_sample_data during the build process, I guess that's been fixed now?

Gemfile.lock
```diff
@@ -332,9 +328,6 @@ GEM
     tzinfo (1.2.9)
       thread_safe (~> 0.1)
     unicode-display_width (1.7.0)
-    unicorn (5.8.0)
```
melzreal, line 335: what was unicorn doing before, and is it being removed because it was a dependency on something else or its being removed for some other reason?
Just asking as gunicorn for example is what I know we use on another project only for the production server, so I'm not sure if this was doing something similar in which case it'd be good to double check

db/seeds.rb
james: I explained my reasoning in the comment to https://github.com/unboxed/bops/pull/205/commits/f0d5556c2ea695b38e02adc1a043f57ff7eb3e8a

create_sample_data was never in the build process

Gemfile.lock
james: This is the Gemfile.lock file, not the Gemfile file, so unicorn wasn't being used. I think it's actually my mistake from the PDF rendering spike that it's in here at all.

app/controllers/documents_controller.rb
```diff
@@ -123,9 +123,7 @@ def form_params
   end
 
   def set_planning_application
-    @planning_application = authorize(PlanningApplication.find(
-                                        params[:planning_application_id],
-                                      ))
+    @planning_application = current_local_authority.planning_applications.find(params[:planning_application_id])
```
melzreal, line 126: so what we had before with pundit on here, that did the authorise on the planning application find, it would render an explicit error saying not authorised if the planning application find failed. We could do a quick if else to render the 402 not authorised, or at least catch the error? 
I guess I did mention it friday but was just mulling through it again in terms of best practice if we should not always be explicit instead of returning a not found. Hmm

db/seeds.rb
melzreal: I get we are not using it for build but I still don't have any planning applications to test locally?

melzreal (changes requested):
This needs a README update - or that or bringing back create_sample_data - since the instructions for setup will be incorrect without one or the other

app/views/planning_applications/_policy_consideration_list.html.erb
```diff
@@ -0,0 +1,16 @@
...
+    </h3>
+  </div>
+  <div id="accordion-default-content-2" class="govuk-accordion__section-content" aria-labelledby="accordion-default-heading-2">
+    <ol class="govuk-body" style="line-height: 2;">
+      <% policy_considerations.order(created_at: :asc).each do |policy_consideration| %>
+        <%= render "question_answer_pair", policy_consideration: policy_consideration %>
+      <% end %>
+    </ol>
+  </div>
+</div>
```
Rhian, line 16: Very tiny suggestion about end of line needed here

app/views/planning_applications/_policy_consideration_list.html.erb
```diff
@@ -0,0 +1,16 @@
...
+    <h3 class="govuk-accordion__section-heading">
+      <button type="button" id="accordion-default-heading-1" aria-controls="accordion-default-content-2" class="govuk-accordion__section-button" aria-expanded="false">
+        Proposal Details
+        <span class="govuk-accordion__icon" aria-hidden="true"></span></button>
+    </h3>
+  </div>
+  <div id="accordion-default-content-2" class="govuk-accordion__section-content" aria-labelledby="accordion-default-heading-2">
+    <ol class="govuk-body" style="line-height: 2;">
+      <% policy_considerations.order(created_at: :asc).each do |policy_consideration| %>
+        <%= render "question_answer_pair", policy_consideration: policy_consideration %>
```
Rhian, line 12: This may be an issue with my sample data, but I am currently seeing no proposal details displaying. I will investigate further and will delete this comment if it's just a local issue

app/controllers/documents_controller.rb
```diff
@@ -12,7 +12,7 @@ class DocumentsController < AuthenticationController
   before_action :disable_flash_header, only: :index
 
   def index
-    @documents = policy_scope(@planning_application.documents).order(:created_at)
+    @documents = @planning_application.documents.order(:created_at)
     @planning_application.documents_validated_at ||= @planning_application.created_at
```
Rhian, line 16: This line is no longer setting the validated_at date in the documents index

app/views/planning_applications/show.html.erb
```diff
@@ -1,5 +1,101 @@
...
+            </span>
+
+            <% if @planning_application.review_assessment_complete? %>
+              <%= tag.strong "Completed", id: "review_assessment-completed", class: "govuk-tag app-task-list__task-tag" %>
+            <% elsif @planning_application.can_review_assessment? && current_user.assessor? %>
+              <%= tag.strong "Waiting", id: "review_assessment-waiting", class: "govuk-tag app-task-list__task-tag" %>
+            <% end %>
+          </li>
+
+          <li class="app-task-list__item">
```
Rhian, line 79: I'm not sure this flow is quite right now I'm looking at it a second time (we can always fix later as a chore, so it doesn't have to be a blocker. We are removing this link once the application has been determined. This is kind of correct behaviour because obviously someone should not be able to publish the application twice. But it means that the application cannot be viewed once it has been determined. Probably a question for Product.

app/controllers/documents_controller.rb
james: It is a 404, as southwark.bops/planning_applications/1 is not the correct URL for that planning application, it is bucks.bops/planning_applications/1

It would be a 403 if the planning application was accessible from the same URL but logged in as a different user.

app/views/planning_applications/_policy_consideration_list.html.erb
Rhian: As an update to the point above, this is not an issue related to this PR  - it is a pre-existing issue with the Swagger data, so I'm going to resolve.

Rhian (approved):
I'll create a chore to fix the small issue where the documents date is not pre-populating, but I don't see any blockers. All looks good to me!

app/views/planning_applications/_policy_consideration_list.html.erb
james: Thanks, I would like to fix the swagger data though! Could you create a ticket?

app/controllers/documents_controller.rb
james: Good catch! This was actually prefilling the validation form, which has been moved to it's own page. I have now moved where we did this.

PR #206, opened 2021-02-15
Update planning application example in Swagger

Rhian:
### Description of change

The existing example was missing the options array which meant that the proposal details did not show

### Story Link

https://trello.com/c/cjbBsnGS/151-bug-fix-swagger-yml-to-display-proposal-details


james (approved):
Good catch :-)

PR #207, opened 2021-02-15
Fix PDF rendering of decision notice

james:
This partial rendering was trying to pass through a decision model that no longer exists, nor does the partial need it.


Rhian (approved):
Looks good to me 🚀

PR #208, opened 2021-02-15
Remove planning application edit action

james:
Found some more code that is no longer being used.


app/helpers/document_helper.rb
```diff
@@ -12,13 +12,4 @@ def filter_current(documents)
   def archive_reason_collection_for_radio_buttons
     Document.archive_reasons.keys.map { |k| [k, I18n.t(k)] }
   end
-
```
Rhian, line 15: Good spot! I was convinced we were still using that somewhere, but have checked and indeed we are not!

Rhian (approved):
All looks good.

PR #209, opened 2021-02-15
Show decision on review form

james:
As per the designs, the review form should have a heading, and state what the decision the assessor is recommending.


james:
[BUG: review form doesn't show decision status](https://trello.com/c/prqa48Jc/155-bug-review-form-doesnt-show-decision-status)

Rhian (approved):
I'm pretty sure we applied logic to ensure that a planning application can't get into this status without a decision? Based on the assumption that this form is never shown in a state where it doesn't have a decision, happy to approve.

PR #210, opened 2021-02-15
Restyle task list on planning application show page

Rhian:
tasklist

### Description of change

This has been styled using divs so it matches the prototype but we will need to have another look at it when we move to full householder to try to make it fit a task list format

### Story Link

https://trello.com/c/YvwZjtK1/153-add-styling-to-new-task-list

### Decisions 

On the prototype, these items were styled as `<dd>` but I thought the markup context would be misleading as they are not listed for this purpose so I have created them as `<div>`s.

### DIFF
--- a/app/views/planning_applications/show.html.erb
+++ b/app/views/planning_applications/show.html.erb
@@ -1,101 +1,124 @@
 <%= render "assessment_dashboard" do %>
-  <% unless @planning_application.withdrawn?  ||  @planning_application.returned?  %>
-    <ol class="app-task-list">
-      <li>
-        <h2 class="app-task-list__section">
-          <span class="app-task-list__section-number"><%= t("#{@planning_application.status}.heading") %></span>
-        </h2>
-        <ul class="app-task-list__items">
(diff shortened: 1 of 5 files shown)


app/views/planning_applications/show.html.erb
```diff
@@ -1,101 +1,124 @@
...
-    <ol class="app-task-list">
-      <li>
-        <h2 class="app-task-list__section">
-          <span class="app-task-list__section-number"><%= t("#{@planning_application.status}.heading") %></span>
-        </h2>
-        <ul class="app-task-list__items">
-          <li class="app-task-list__item">
+  <% unless @planning_application.withdrawn? || @planning_application.returned? %>
+    <h2 class="app-task-list__section">
+      <span class="app-task-list__section-number"><%= t("#{@planning_application.status}.heading") %></span>
```
james, line 4: I don't think this was made particularly clear from the designs, but I've checked with Ali, and this heading should always be "Process Application", so we can just put this in without doing the `t()` thing.

james (approved):

PR #212, opened 2021-02-15
Move 'cancel application' into own govuk-grid-row

james:
It was breaking the grid alignment as a floating p tag.


Rhian (approved):
Sorry, I caused a conflict with my last merge 😬 Approved apart from that!

PR #213, opened 2021-02-15
Remove validate documents link from applications awaiting determination

Rhian:
### Description of change

Added awaiting_determination? to the can_validate? method and updated specs to ensure we do not allow someone to click on Validate Documents when the application is in an incorrect state

### Story Link

https://trello.com/c/sm7J2pWo/158-validate-documents-should-not-be-accessible-in-awaitingdetermination


spec/system/planning_applications/reviewing_spec.rb
```diff
@@ -30,6 +30,8 @@
       expect(page).to have_content("New assessor comment")
     end
 
+    expect(page).to have_no_link("Validate documents")
```
james, line 33: I think the task_list_spec.rb covers this, so I don't think we need to put this check in here as well. (keen to keep these feature specs small and manageable)

james (approved):
Just that one nonblocker of a request, but all good otherwise!

PR #214, opened 2021-02-15
Add breadcrumb titles

melzreal:
### Description of change

Turns out all was needed was a "content_for title" in the new partials - also grabbed the correct breadcrumb from gov uk that has each item contained in a li tag instead of a div tag

### Story Link

https://trello.com/c/2RCAws7e/154-bug-fix-breadcrumbs

### DIFF
--- a/app/views/application/_create_breadcrumb_navigation.html.erb
+++ b/app/views/application/_create_breadcrumb_navigation.html.erb
@@ -1,9 +1,9 @@
 <% nav.each do |n| %>
-  <div class="govuk-breadcrumbs__list-item" aria-current="page">
+  <li class="govuk-breadcrumbs__list-item" aria-current="page">
     <%= link_to n[:title], n[:path], class: "govuk-button__link" %>
-  </div>
+  </li>
 <% end %>
 
-<div class="govuk-breadcrumbs__list-item" aria-current="page">
+<li class="govuk-breadcrumbs__list-item" aria-current="page">
   <%= yield :title %>
-</div>
+</li>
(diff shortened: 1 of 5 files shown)


app/views/application/_breadcrumb_navigation.html.erb
```diff
@@ -10,7 +10,9 @@
       <div class="govuk-grid-column-two-thirds">
         <div class="govuk-breadcrumbs">
           <ol class="govuk-breadcrumbs__list">
-            <%= render "create_breadcrumb_navigation", nav: nav %>
+            <li class="govuk-breadcrumbs__list-item">
```
james, line 13: Huh, I don't think you meant to commit this change, but it does raise a good point: why are the list items a `div`s not `li`s in create_breadcrumb_navigation?

I think this `li` should come out of this file (otherwise we're wrapping *all* of the crumbs in a single `li`), but I wouldn't say no to you changing

`<div class="govuk-breadcrumbs__list-item" aria-current="page">`

to

`<li class="govuk-breadcrumbs__list-item" aria-current="page">`

in _create_breadcrumb_navigation.html.erb in this commit.

app/views/application/_create_breadcrumb_navigation.html.erb
```diff
@@ -6,4 +6,4 @@
 
 <div class="govuk-breadcrumbs__list-item" aria-current="page">
   <%= yield :title %>
-</div>
+</div>
```
james, line 9: Can we remove this whitespace change please

james (approved):
🎉

PR #217, opened 2021-02-16
Hide task list when application is determined

Rhian:
### Description of change

Once the planning application has moved to Determined status, the task list should be hidden. Assessors and reviewers should be able to view the decision notice via a link near the top of the page

### Story Link

https://trello.com/c/qvB833hO/160-hide-task-list-and-give-link-to-decision-notice-once-determined

### Decisions 

Although it was not in the story, I have also added a conditional for not displaying Publish recommendation heading if the application is closed.

### DIFF
--- a/app/views/planning_applications/publish.html.erb
+++ b/app/views/planning_applications/publish.html.erb
@@ -1,8 +1,8 @@
 <%= render "planning_applications/assessment_dashboard" do %>
   <% content_for :title, "Publish the recommendation" %>
 
-  <h2 class="govuk-heading-m">Publish the recommendation</h2>
-  <p class="govuk-body">The following decision notice was created based on the planning officer's recommendation and comment. Please review and publish it.</p>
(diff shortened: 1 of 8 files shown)


app/views/planning_applications/publish.html.erb
```diff
@@ -1,9 +1,10 @@
 <%= render "planning_applications/assessment_dashboard" do %>
   <% content_for :title, "Publish the recommendation" %>
 
-  <h2 class="govuk-heading-m">Publish the recommendation</h2>
```
james, line 4: I'm not a fan of re-using the "publish" action as a "view decision notice" action. Feels like they should be separate?

What do you think about a new action at `decision_notice_planning_application_path`?

Rhian: Good idea - will be much cleaner 👍

james (approved):

PR #221, opened 2021-02-17
Close accordion when re-rendering page

melzreal:
-Removing aria controls has the desired effect. This is because accordion-default-content is the class being called to setState in JavaScript, thus using that state in order to maintain the accordion open or closed. 
Without that class, no state is found or stored in the browser and thus the accordions close as they should.
Adding a separate ticket as well to fix the javaScript library in the long term so we can re-add aria controls for accessibility.

### Description of change

Remove aria controls that were storing state for the accordion and forcing it to stay open.

### Story Link

https://trello.com/c/LSGIdnO4/179-accordion-should-not-save-state-between-page-reloads


james (approved):
🎉  Again, nice clean atomic commit. Thanks again.

PR #222, opened 2021-02-18
Move closing div to correct place in accordion

melzreal:
- This ensures the accordion behaves as it should, only opening the selected section instead of the 3 sections this div was encompassing.

### Story

https://trello.com/c/faRnFF6i/156-bug-main-accordion-opening-and-closing-behaviour

image


james (approved):
Nice clean, atomic PR. Thanks!

PR #224, opened 2021-02-19
Add granted comment to recommendation

melzreal:
### Description of change

Make the public comment field compulsory to all decisions regardless of granted or not, and display it on decision notice. 
After aligning with design it seems we no longer want different conditional comments meaning there is no need for a db change, we can just use the "public comment" for all instances.
Also aligned regarding the comment that should show up on the review screen - we still want the "comment to your manager" to show, not the "granted comment" (public comment) so the functionality doesn't change there, its just a text change.

image

image

image

### Story Link

https://trello.com/c/x35QLpFq/139-add-planning-officers-note-for-granted-decisions-to-the-notice

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -223,7 +223,7 @@ private
   end
 
   def public_comment_present
-    if decision == "refused" && public_comment.blank?
+    if decision_present? && public_comment.blank?
       errors.add(:planning_application, "Please fill in the GDPO policies text box.")
     end
   end
--- a/app/views/planning_applications/recommendation_form.html.erb
+++ b/app/views/planning_applications/recommendation_form.html.erb
@@ -20,7 +20,8 @@
           </span>
         <% end %>
       <% end %>
-      <div class="govuk-radios govuk-radios--conditional" data-module="govuk-radios" style="margin-bottom: 20px;">
+     <div class="govuk-form-group">
+      <div class="govuk-radios" data-module="govuk-radios" style="margin-bottom: 20px;">
         <div class="govuk-radios govuk-radios--small">
           <div class="govuk-radios__item">
             <%= form.radio_button :decision, "granted", class: "govuk-radios__input", id:"status-granted-conditional", "aria-controls": "conditional-status-granted-conditional", "aria-expanded": "false" %>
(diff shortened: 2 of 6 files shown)


app/models/planning_application.rb
```diff
@@ -223,7 +223,7 @@ def has_validation_date?
   end
 
   def public_comment_present
-    if decision == "refused" && public_comment.blank?
+    if decision && public_comment.blank?
```
james, line 226: `decision.present?` would be more defensive here. While the application currently doesn't ever create a decision of a blank string, it's not impossible and we might do that in the future, and if we did I'd assume that we'd want a blank string to mean "no decision".

app/views/planning_applications/recommendation_form.html.erb
```diff
@@ -30,26 +31,19 @@
...
+     </div>
+      <div class="govuk-form-group">
+        <p class="govuk-body">
+          <br>
+          <strong>Please provide supporting information to indicate which GDPO policy(s) have or have not been met. </strong>
+        </p>
+        <p class="govuk-body">
+          This information <strong>will</strong> appear on the decision notice.
+        </p>
+        <%= form.text_area :public_comment, class: "govuk-textarea govuk-!-margin-bottom-9", id: "public_comment", rows: "3", "aria-describedby": ":public-comment-hint" %>
```
james, line 45: We no longer provide a hint text, so I don't think we should have `"aria-describedby": ":public-comment-hint"` on this any more?

james: In fact, this text_area doesn't have a label for it, which doesn't seem right. I think it's been copied and pasted a few too many times and lost some of its semantics. Could you fix it up so it matches https://design-system.service.gov.uk/components/textarea/?

spec/helpers/planning_application_helper_spec.rb
```diff
@@ -31,7 +31,7 @@
     end
 
     context "granted" do
-      let(:planning_application) { create :planning_application, :determined, decision: "granted" }
+      let(:planning_application) { create :planning_application, :determined, decision: "granted", public_comment: "All GDPO compliant" }
```
james, line 34: Explicitly setting the `public_comment` in a `let` should only *really* be done if it's being explicitly tested in the included specs. If it's just for creating a valid object, then it should be done in the factory (`spec/factories/planning_application.rb`). Could you add that there, and then remove all these explicit sets?

james:
3 small things, but overall looks good!

app/views/planning_applications/recommendation_form.html.erb
melzreal: I took the hint out because that's how the design is asking for it to be now:

image

melzreal: on the aria-described by, I had understood it was used for accessibility purposes to help screen readers relate parts of content to the other - and because we still have aria-describedby in the "private comment" ?

james (approved):

PR #226, opened 2021-02-24
Introduce auditing to planning applications & documents

melzreal:
### Description of change

Introduce auditing. 

image

image

image

### Story Link

https://trello.com/c/USMlBnD5/122-introduce-auditing-dev

### Decisions (design)

Have coordinated with Martyn and Ali throughout the ticket, some decisions made:

- Don't save recommendation saved just save it when submitting to manager ( ignore action of when a comment is saved by an assessor/reviewer. Only track the comment made when they send the decision back and forth by hitting the "determine" action)
- Use the terms "recommendation challenged" instead of "assessment rejected" or "recommendation rejected"
- When a decision is published, add "Application granted" or "Application refused" as a comment to the audit step
- Keep track of "unassigned" actions as well as "assigned" actions when assigning an application to users or to unassigned

### Decisions (dev)
- Eliminate unused code in planning controller
- Organise planning controller methods to match routes file
- Store optional comment data as strings instead of JSON, for simplicity and because we only have 1 argument in each case
- Store optional user args as strings to pass to helper method that displays content
- To match the required design, we have default text in the index showing user "API" created the application, for every single application.  This will be replaced in the future with the appropriate api user, but we've added that as a separate story so we can discuss what that should look like [ https://trello.com/c/m1y87dfT/187-audit-user-we-will-need-a-relationship-between-the-planning-applications-table-and-the-user-table-create-a-separate-story-to-en ]

### DIFF
--- a/app/controllers/api/v1/planning_applications_controller.rb
+++ b/app/controllers/api/v1/planning_applications_controller.rb
@@ -32,6 +32,18 @@ class Api::V1::PlanningApplicationsController < Api::V1::ApplicationController
     )
     if @planning_application.valid? && @planning_application.save!
       upload_documents(params[:files])
+
+      api_user = User.find_or_create_by!(
+        name: "API",
+        email: "api_user@bops.com",
+        local_authority: @current_local_authority,
+      ) { |user| user.password = SecureRandom.base64(12) }
+
+      Audit.create!(
+        planning_application_id: @planning_application.id,
+        user: api_user,
+        activity_type: "created",
+      )
       send_success_response
     else
       send_failed_response
--- a/app/controllers/api/v1/planning_applications_controller.rb
+++ b/app/controllers/api/v1/planning_applications_controller.rb
@@ -32,18 +32,6 @@ class Api::V1::PlanningApplicationsController < Api::V1::ApplicationController
     )
     if @planning_application.valid? && @planning_application.save!
       upload_documents(params[:files])
-
-      api_user = User.find_or_create_by!(
-        name: "API",
-        email: "api_user@bops.com",
-        local_authority: @current_local_authority,
-      ) { |user| user.password = SecureRandom.base64(12) }
-
-      Audit.create!(
-        planning_application_id: @planning_application.id,
-        user: api_user,
-        activity_type: "created",
-      )
       send_success_response
     else
       send_failed_response
--- a/app/controllers/audits_controller.rb
+++ b/app/controllers/audits_controller.rb
@@ -7,6 +7,8 @@ class AuditsController < AuthenticationController
     @audits = Audit.where(planning_application_id: params[:planning_application_id])
   end
 
+private
+
   def set_planning_application
     @planning_application = current_local_authority.planning_applications.find(params[:planning_application_id])
   end
--- a/app/controllers/audits_controller.rb
+++ b/app/controllers/audits_controller.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class AuditsController < AuthenticationController
+  before_action :set_planning_application
+
+  def index
+    @audits = Audit.where(planning_application_id: params[:planning_application_id])
+  end
+
+  def set_planning_application
+    @planning_application = current_local_authority.planning_applications.find(params[:planning_application_id])
+  end
+end
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -42,6 +42,7 @@ class DocumentsController < AuthenticationController
 
     if @document.save
       flash[:notice] = "#{@document.file.filename} has been uploaded."
+      audit("uploaded", @document.file.filename)
       redirect_to planning_application_documents_path
     else
       render :new
@@ -63,6 +64,7 @@ class DocumentsController < AuthenticationController
 
   def confirm_archived
     @document.archive(params[:archive_reason])
+    audit("archived", @document.file.filename)
     flash[:notice] = "#{@document.name} has been archived"
     redirect_to planning_application_documents_path
   end
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -11,7 +11,6 @@ class PlanningApplicationsController < AuthenticationController
                                                     review
                                                     publish
                                                     determine
-                                                    request_correction
                                                     validate_documents_form
                                                     validate_documents
                                                     cancel_confirmation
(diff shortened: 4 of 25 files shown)


app/controllers/documents_controller.rb
```diff
@@ -99,6 +101,16 @@ def validate_step
 
 private
 
+  def audit(activity_type, audit_comment = nil, *activity_information)
```
james, line 104: I like this method, but I don't think a splat is what you're looking for here for activity_information. I'd just use `activity_information = nil` like you have with audit_comment

app/controllers/documents_controller.rb
```diff
@@ -99,6 +101,16 @@ def validate_step
...
+    Audit.create!(
+      planning_application_id: @planning_application.id,
+      user_id: current_user.id,
```
james, line 107: Can we set these associations with their associations rather than the IDs? Like:
```
planning_application: @planning_application,
user: current_user
```

app/controllers/documents_controller.rb
james: in fact, I think keyword arguments might be an even cleaner way of doing this: https://thoughtbot.com/blog/ruby-2-keyword-arguments

app/controllers/planning_applications_controller.rb
```diff
@@ -177,6 +168,16 @@ def validate_documents
 
 private
 
+  def audit(activity_type, audit_comment = nil, *activity_information)
```
james, line 171: I think it's OK to move this method up into application_controller, so it's available to both planning applications and documents controller, and you don't have the duplicate the code

app/views/audits/index.html.erb
```diff
@@ -0,0 +1,36 @@
...
+    </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+    <% @audits.each do |audit_item| %>
+      <tr class="govuk-table__row">
+        <td class="govuk-table__cell"><%= audit_item.created_at.strftime('%d-%m-%Y %H:%M') %></td>
+        <td class="govuk-table__cell"><%= User.find_by_id(audit_item.user_id).name %></td>
+        <td colspan="2" class="govuk-table__cell">
+          <%= activity(audit_item.activity_type, audit_item.activity_information) %> <br>
+          <% if !audit_item.audit_comment.nil? %>
```
james, line 29: using `if audit_item.audit_comment.present?` would be better here: more readable, plus returns false if it's an empty string.

app/controllers/documents_controller.rb
melzreal: I gave this a go with 
```
  def audit(activity_type, audit_comment:, activity_information:)
  ```
  
  also tried assigning to 
  
  ```  def audit(activity_type, audit_comment: nil, activity_information: nil)```
  
  and ```  def audit(activity_type, audit_comment: "nil", activity_information: "nil") ```
  
  but could not achieve the same effect that we have with the current method, so I'll keep it - open to suggestions on how it could actually be done better but I couldn't see a way to use keyword arguments with default to nil, so the current method is looking like:
  
  
  ```   def audit(activity_type, audit_comment = nil, activity_information = nil)
    Audit.create!(
      planning_application_id: @planning_application.id,
      user: current_user,
      audit_comment: audit_comment,
      activity_type: activity_type,
    )
  end
  ```

File: app/controllers/api/v1/planning_applications_controller.rb [hunk 5]
```diff
@@ -32,6 +32,18 @@ def create
     )
     if @planning_application.valid? && @planning_application.save!
       upload_documents(params[:files])
+
+      api_user = User.find_or_create_by!(
```
james, line 36: We don't want to be creating API users here! `current_api_user` is set in `Api::V1::ApplicationController`

app/controllers/planning_applications_controller.rb
```diff
@@ -42,26 +41,38 @@ def assign
                                    else
                                      current_local_authority.users.find(params[:planning_application][:user_id])
                                    end
+      @planning_application.user.nil? ? audit("assigned") : audit("assigned", "", @planning_application.user.name)
```
james, line 44: I think `audit("assigned", nil, @planning_application.user&.name)` is a neater way of doing this line

james (changes requested):
Looking good! A lot going on here, so I've got a few requested changes, and then I'll have another look through. Thanks!

app/views/audits/index.html.erb
```diff
@@ -0,0 +1,58 @@
...
+  </div>
+  <div class="govuk-grid-column-two-thirds">
+    <p class="govuk-body">
+      Assigned to:
+      <strong>
+        <%= @planning_application.user ? @planning_application.user.name : "Unassigned" %>
+      </strong>
+    </p>
+  </div>
+</div>
```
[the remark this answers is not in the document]
melzreal: So this is what it currently shows:

image

but if I wrap "assessment_dashboard" (```<%= render "/planning_applications/assessment_dashboard" do %> ```)

we get:

image

app/helpers/audit_helper.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module AuditHelper
+  def activity(type_of_activity, *args)
```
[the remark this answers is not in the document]
melzreal: went with the first option as it looked simpler and is consistent with the rest of the methods this pr introduces ! thanks :)

app/views/audits/index.html.erb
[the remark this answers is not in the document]
james: Aha! I'd missed that the designs were slightly different.

I'm still wary of duplicating code here: it's exactly the sort of thing we'd update in one place, and then miss that we need to update elsewhere. Would it be possible to move this chunk of code into it's own partial (maybe `plannting_applications/_proposal_header.html.erb` or similar) and require that from assessment_dashboard and in here?

app/controllers/audits_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AuditsController < AuthenticationController
+  before_action :set_planning_application
+
+  def index
+    @audits = Audit.where(planning_application_id: params[:planning_application_id])
+  end
+
+  def set_planning_application
```
james, line 10: This method should be `private`

db/migrate/20210222100817_create_audits.rb
```diff
@@ -0,0 +1,13 @@
+class CreateAudits < ActiveRecord::Migration[6.0]
+  def change
+    create_table :audits do |t|
+      t.references :planning_application, null: false, foreign_key: true
+      t.references :user
+      t.jsonb :activity_type, default: [], null: false
```
james, line 6: Why is this a `jsonb` field? As we're using it as an enum, I think it should be an `int`?

james (approved):
🎉

PR #228, opened 2021-02-24
New proposal details API

james:
We have agreed a new API spec for the `proposal_details` section of the API with RIPA.

This is the first PR to implement it. This will allow RIPA to start inputting data, which will allow us to refine the designs.

screencapture-southwark-lvh-me-3000-planning-applications-8-2021-02-24-18_27_51


app/views/planning_applications/_policy_consideration_list.html.erb
```diff
@@ -12,8 +12,30 @@
   </div>
   <div class="govuk-accordion__section-content" aria-labelledby="accordion-default-heading-2">
     <ol class="govuk-body" style="line-height: 2;">
-      <% policy_considerations.order(created_at: :asc).each do |policy_consideration| %>
-        <%= render "question_answer_pair", policy_consideration: policy_consideration %>
+      <% @planning_application.parsed_proposal_details.each do |proposal_detail| %>
```
melzreal, line 15: since we had the line above to order them by when they were created, should we do something similar? having 
@planning_application.parsed_proposal_details.all.order("created_at ASC") in the controller for ex would also probably be more performant than the previous view ordering we had here

app/views/shared/_application_information.html.erb
```diff
@@ -66,8 +66,6 @@
       </table>
     </div>
   </div>
-    <% if @planning_application.policy_evaluation %>
-      <%= render "planning_applications/policy_consideration_list", policy_considerations: @planning_application.policy_evaluation.policy_considerations %>
-    <% end %>
+    <%= render "planning_applications/policy_consideration_list" %>
```
melzreal, line 69: do we no longer need the if statement here? Are proposal details now compulsory to a successful api call?

app/views/planning_applications/_policy_consideration_list.html.erb
james: `parsed_proposal_details` is not an ActiveRecord association any more, but a parsed JSON hash. We don't really know what order we will want them in yet (needs more user research), so I'm sticking with the order that RIPA is giving us.

app/views/shared/_application_information.html.erb
james: It is not required, but if the proposal_details are nil, then the .each in the partial will just not render anything.

melzreal (approved):
LGTM 👍

PR #229, opened 2021-03-01
Add Google map link to application information partial

Rhian:
### Description of change

Created a helper method that takes a postcode, strips whitespace and creates a Google map link. Will return a not found string if postcode is nil

### Story Link

https://trello.com/c/wBolEDiI/170-link-to-google-maps-based-on-address

### Decisions

May change if a design is provided

### DIFF
--- a/app/helpers/planning_application_helper.rb
+++ b/app/helpers/planning_application_helper.rb
@@ -27,6 +27,10 @@ module PlanningApplicationHelper
     end
   end
 
+  def map_link(address)
+    "https://google.co.uk/maps/place/#{CGI.escape(address)}"
+  end
+
   def list_constraints(constraints)
     JSON.parse(constraints).select { |_category, value| value == true }.keys unless constraints.empty?
   end
--- a/app/views/shared/_application_information.html.erb
+++ b/app/views/shared/_application_information.html.erb
@@ -18,6 +18,10 @@
             <td class="govuk-table__cell"><%= @planning_application.site.full_address %></td>
           </tr>
           <tr class="govuk-table__row">
+            <td class="govuk-table__cell"><strong>Location:</strong></td>
+            <td class="govuk-table__cell"><%= link_to 'Map link', map_link(@planning_application.site.full_address) %></td>
+          </tr>
+          <tr class="govuk-table__row">
             <td class="govuk-table__cell"><strong>UPRN:</strong></td>
             <td class="govuk-table__cell"><%= @planning_application.site.uprn %></td>
           </tr>
--- a/spec/helpers/planning_application_helper_spec.rb
+++ b/spec/helpers/planning_application_helper_spec.rb
@@ -17,6 +17,12 @@ RSpec.describe PlanningApplicationHelper, type: :helper do
     end
   end
 
+  describe "#map_link" do
+    it "returns the correct link for a valid address" do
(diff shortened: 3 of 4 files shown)


app/helpers/planning_application_helper.rb
```diff
@@ -19,6 +19,14 @@ def proposed_or_existing(planning_application)
     planning_application.work_status == "proposed" ? "No" : "Yes"
   end
 
+  def map_link(postcode)
+    if postcode.nil?
+      "No location found"
```
james, line 24: This would result in `<a href="No location found">Map link</a>` being put on the page, which doesn't seem right. We probably want to just hide the whole `<tr>` if `postcode.blank?`

Rhian: oh yes!!

Rhian: I've removed any logic from the helper method for dealing with an incorrect postcode, because as long as there is some kind of value there, Google will try to create a link

app/helpers/planning_application_helper.rb
```diff
@@ -19,6 +19,10 @@ def proposed_or_existing(planning_application)
     planning_application.work_status == "proposed" ? "No" : "Yes"
   end
 
+  def map_link(postcode)
+    "https://www.google.co.uk/maps/place/#{postcode.gsub(/\s+/, '')}"
```
james, line 23: Sorry, should have realised this in my first review, but I think it would be best if we used `CGI::escape` here, so we a) maintain the space in case that's helpful, and b) automatically handle anything else weird that might happen here.

In fact, and just an idea that might not work, what happens if we pass in `CGI::escape(@planning_application.site.full_address)` - does that create a valid link to the exact address?

app/views/shared/_application_information.html.erb
```diff
@@ -17,6 +17,12 @@
             <td class="govuk-table__cell"><strong>Address:</strong></td>
             <td class="govuk-table__cell"><%= @planning_application.site.full_address %></td>
           </tr>
+          <% unless @planning_application.site.postcode.blank? %>
```
james, line 20: Seeing as I'm cracking this one open again, `if @planning_application.site.postcode.present?` is a *smidge* more readable, but this is such a minor thing.

melzreal, line 20: would we not want to make the site postcode required? It the site doesn't have a postcode the application shouldn't be valid?

james: The postcode field does not currently validate presence. Adding validation is possibly the right thing to do but a) we'd need to check that that's correct (I think people can ask for planning permission on plots of land that don't have post codes? But would have to check that), and b) is outside the scope of this ticket (it would change the API, as post code would then be a required field)

app/helpers/planning_application_helper.rb
Rhian: I've tried this with a variety of weird and wonderful inputs and it seems to more or less find the exact address each time

app/views/shared/_application_information.html.erb
Rhian: Now redundant, as we will always have some kind of address (and if we don't the other places where we are calling `.full_address` would give us issues, so I don't think we now need the conditional at all)

melzreal (approved):
LGTM

james (approved):

PR #231, opened 2021-03-01
Allow for multiple submissions to have the same UPRN

melzreal:
### Description of change

Remove sites db table, add address fields directly into planning applications
Support multiple submissions per UPRN

### Story Link

https://trello.com/c/EWYX4tzi/180-allow-different-addresses-to-be-submitted-with-the-same-uprn

### Decisions

- I've removed two tests that expected "county" to be present.
County wasn't expected/allowed in site_params in the api application controller so I'm not sure why the test was there, but can bring that back.


spec/requests/api/oas3_spec.rb
```diff
@@ -59,33 +59,31 @@ def example_response_hash_for(*attrs)
     expect(PlanningApplication.last.documents.first.file).to be_present
   end
 
-  it "successfullies return the listing of applications as specified" do
+  xit "successfully returns the listing of applications as specified" do
```
Rhian, line 62: Very tiny thing in such a big PR, but looks like you have a stray 'x' before 'it'

Rhian (approved):
I'm not sure about the 'county' param. I'm pretty sure once we do more testing with councils outside London we are going to need this, but as you say, for now it is not being submitted. I guess we can add it later. Other than that, all looks good to me (once conflicts resolved) and I agree with your decision re. the tests.

spec/requests/api/oas3_spec.rb
melzreal: ah no its on purpose, to skip this test

app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -26,10 +26,15 @@ def decision_notice
...
   def create
-    site_id = create_site
     @planning_application = PlanningApplication.new(
-      full_planning_params.merge!(site_id: site_id, local_authority_id: @current_local_authority.id),
+      full_planning_params.merge!(
+        local_authority_id: @current_local_authority.id,
+      ),
     )
+    if site_params
+      @planning_application.update!(site_params)
```
james, line 35: You're calling update!, which saves the record, and raises an error if not valid. Ideally we would only call save on this record on line 38.

I would propose merging the site_params into full_planning_params in the method on line 111

spec/requests/api/oas3_spec.rb
melzreal: (this is the test that asks for a strict json order in the return, so regardless of receiving a 200 and all the content being there the test fails because the content is in a different order now)

app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -67,21 +72,17 @@ def send_not_found_response
...
   def planning_application_params
     permitted_keys = %i[application_type
                         description
                         ward
-                        site_id
+                        address_1
+                        address_2
+                        town
+                        postcode
+                        uprn
```
james, line 85: As you are setting these params directly (currently on line 35), they don't need to be permitted on the top level of the hash.

db/migrate/20210301151645_drop_sites.rb
```diff
@@ -0,0 +1,9 @@
+class DropSites < ActiveRecord::Migration[6.0]
+  def up
+    drop_table "sites"
+  end
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+  remove_index :planning_applications, name: :index_planning_applications_on_site_id
```
james, line 8: This `remove_index`, if needed (I think it gets dropped by the drop_table anyway), should be inside the `up` method.

spec/fixtures/files/minimal_planning_application.json
```diff
@@ -1,4 +1,4 @@
-{
+{ "application_type": 0,
```
james, line 1: This works, due to how the `enum` works internally, but we are specifying in the swagger docs that we expect `application_type` to be "lawfulness_certificate"

spec/models/planning_application_spec.rb
```diff
@@ -320,13 +320,6 @@
   end
 
   describe "#applicant?" do
-    it "returns false if no values are given" do
-      planning_application.update!(applicant_first_name: "", applicant_last_name: "",
-                                   applicant_phone: "", applicant_email: "")
-
-      expect(planning_application.applicant?).to eq false
-    end
-
```
james, line 329: why did this spec get deleted? Is this no longer the case?

spec/requests/api/oas3_spec.rb
```diff
@@ -59,33 +59,31 @@ def example_response_hash_for(*attrs)
...
-    expected_reponse = example_response_hash_for("/api/v1/planning_applications", "get", 200, "Full")
-    expected_reponse["data"].first["documents"].first["url"] = api_v1_planning_application_document_url(planning_application, planning_application_document)
-    expect(JSON.parse(response.body)).to eq(expected_reponse)
+    expected_response = example_response_hash_for("/api/v1/planning_applications", "get", 200, "Full")
+    expected_response["data"].first["documents"].first["url"] = api_v1_planning_application_document_url(planning_application, planning_application_document)
+    expect(JSON.parse(response.body)).to eq(expected_response)
   end
 
-  it "successfullies return an application as specified" do
+  xit "successfully returns an application as specified" do
```
james, line 76: There's a rogue 'x' here too

spec/requests/api/oas3_spec.rb
```diff
@@ -59,33 +59,31 @@ def example_response_hash_for(*attrs)
...
     planning_application_hash = example_response_hash_for("/api/v1/planning_applications/{id}", "get", 200, "Full")
-    site = Site.create! planning_application_hash.fetch("site")
-    planning_application = PlanningApplication.create! planning_application_hash.except("application_number", "received_date", "documents").merge(site: site, local_authority: @default_local_authority)
+    planning_application = PlanningApplication.create! planning_application_hash.except("application_number", "received_date", "documents", "site").merge(local_authority: @default_local_authority, applicant_first_name: "Mel")
```
james, line 78: Do we need to exclude the "site" params in this hash now that you've removed it?

spec/requests/api/planning_applications_post_spec.rb
```diff
@@ -114,15 +114,13 @@
       permitted_development_json = File.read(valid_json)
 
       it "renders success message" do
-        create(:site, uprn: "100081043511")
```
james, line 117: I think we can just remove this whole context block now?

james (changes requested):
Lots of good work here, a couple of blockers I'd like fixed before it goes into production. Feel free to ask me any questions.

app/controllers/api/v1/planning_applications_controller.rb
melzreal: this site params separate call was to make sure the application didn't error when the site_params are nil - which happens in a few of the tests. I'll change update! to update so it doesn't save the record, I did think maybe there was a better way to do this (maybe with a @planning_application << method) but it can't be merged into the one above or a bunch of tests fail..

db/migrate/20210301151645_drop_sites.rb
melzreal: yes the remove index is needed explicitly because for some reason dropping the table left the index alone when I tried it in the past, causing errors on creating applications without it. I'm not sure why that is,maybe it needed to have been created with a dependent destroy 🤔

spec/requests/api/oas3_spec.rb
melzreal: I've ignored this test on purpose, see description

app/controllers/api/v1/planning_applications_controller.rb
melzreal: I've solved this by using assign_attributes instead of update - we still need the "if" check provided by the site params method hence the need to keep it separate - but if you know of a way to keep the if check and merge that into the other method in an elegant way I'm happy to do it, just don't see how its possible

james: Using `assign_attributes` fixes the "double save" and raising an application error issue, so the code now works as we want :-)

My following suggestion is only a different way to achieve the same thing, but I think is a little bit simpler?

If you replace `full_planning_params` with the following:

```ruby
def full_planning_params
  planning_application_params.merge!({
    proposal_details: (params[:proposal_details].to_json if params[:proposal_details].present?),
    constraints: (params[:constraints].to_json if params[:constraints].present?),
    audit_log: params.to_json,
    uprn: params[:site][:uprn],
    address_1: params[:site][:address_1],
    address_2: params[:site][:address_2],
    town: params[:site][:town],
    postcode: params[:site][:postcode],
  })
end
```

Then I think you can remove the `site_params` method, and lines 34-36. I think that's a bit cleaner?

spec/requests/api/oas3_spec.rb
james: Aha, I see. I think this test is a useful test, although agree that the order of the hash doesn't matter. But I don't think the matcher pays attention to the order anyway?

`irb(main):001:0> {a: 'b', c: 'd'} == {c: 'd', a: 'b'}
=> true
`

What particularly isn't working here?

melzreal: This test somehow thinks the fields are nil (i.e uprn, town, see screenshot) but they are displayed just fine, so the test must be wrong. Whatever the "example response is" It can't be right, but I can't see the "example response" it is using? Its quite an obscure test.

image
image


ex JSON of an application:

image

james (approved):

PR #232, opened 2021-03-02
Add logic for display and publish documents

Rhian:
publishable

### Description of change

We previously displayed document references on the decision notice and made them public depending on whether they had a number. The new logic means we can separate these two functions and we have two different scopes (for_publication and for_display) to reflect this. Radio buttons have been added to Edit and New (upload) views for documents.

Note that we have not updated the API controller as we have retained the for_publication method name which dictates which documents are publicly viewable.

### Story Link

https://trello.com/c/qQ8nOXNy/133-indicating-which-documents-to-include-on-decision-notice-and-which-one-will-be-public-or-private

### Decisions


app/views/documents/new.html.erb
```diff
@@ -16,9 +16,42 @@
...
+        <div class="publish">
+          <%= form.govuk_radio_buttons_fieldset(:status, legend: { size: 's', text: 'Should this document be made publicly available?' }) do %>
+            <div id="event-name-hint" class="govuk-hint">
+              Please indicate whether this document should be made available for public viewing on the planning register or via the public API. Select "No" if the document contains any sensitive personal information
+            </div>
+            <%= form.govuk_radio_button :publishable, 'true', label: { text: 'Yes' } %>
+            <%= form.govuk_radio_button :publishable, 'false', label: { text: 'No' } %>
+          <% end %>
+        </div>
+      </div>
```
james, line 54: The first A/C on the ticket is "When a new document is uploaded or replaced the user should see the same options as when they edit a document including add drawing number and the two questions to be added in this story." which if I recall was our way of saying that the edit and new actions can just use the same form partial?

public/api-docs/v1/swagger_doc.yaml
```diff
@@ -354,4 +354,4 @@ paths:
                   files:
                     - filename: https://bops-test.s3.eu-west-2.amazonaws.com/proposed-first-floor-plan.pdf
                       tags:
-                        - Side
\ No newline at end of file
+                        - Side
```
james, line 357: whitespace churn ;-)

spec/system/documents/display_and_publish_documents_spec.rb
```diff
@@ -77,6 +75,79 @@
...
+          choose "No"
+        end
+
+        click_button "Save and return"
+
+        within(all(".app-task-list__item").first) do
+          click_link "Edit"
+        end
+        # the submitted values are re-presented in the form
+        expect(find_field("numbers")).to have_content "new_number_1, new_number_2"
```
james, line 99: Does `expect(page).to have_field("numbers", with: "new_number_1, new_number_2")` work here?

james (commented):
Overall, looks great, and no blockers! I've just suggested a couple of things to tidy it up even further.

spec/system/documents/display_and_publish_documents_spec.rb
Rhian: Yes! 🚀

james (approved):

app/views/documents/form_partials/_privacy.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  <%= form.govuk_radio_buttons_fieldset(:status, legend: { size: 's', text: 'Do you want to list this document on the decision notice?' }) do %>
+    <%= form.govuk_radio_button :referenced_in_decision_notice, 'true', label: { text: 'Yes' } %>
+    <%= form.govuk_radio_button :referenced_in_decision_notice, 'false', label: { text: 'No' } %>
+  <% end %>
+</div>
+
+<div class="publish">
+  <%= form.govuk_radio_buttons_fieldset(:status, legend: { size: 's', text: 'Should this document be made publicly available?' }) do %>
+    <div id="event-name-hint" class="govuk-hint">
+      Please indicate whether this document should be made available for public viewing on the planning register or via the public API. Select "No" if the document contains any sensitive personal information
```
melzreal, line 20: image
I think we need a little space here

melzreal (changes requested):

melzreal (approved):
🚢 🚢  :shipit:

PR #234, opened 2021-03-02
Configure PDFKit to use XVFB

Rhian:
### Description of change

wkhtmltopdf does not work out of the box on Ubuntu because it needs a display. We are adding XVFB as a dependency separately in our Terraform  but we also need to tell PDFKit to use xvfb-run to execute wkhtmltopdf if this executable is present

### DIFF
--- a/config/initializers/pdfkit.rb
+++ b/config/initializers/pdfkit.rb
@@ -1,6 +1,6 @@
-require "mkmf"
+# frozen_string_literal: true
 
-if find_executable("xvfb-run")
+if system("which xvfb-run > /dev/null 2>&1")
   PDFKit.configure do |config|
     config.use_xvfb = true
   end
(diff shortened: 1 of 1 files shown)


james:
Just reading up on `mkmf` a bit more, it seems it's not advised to include this in your ruby projects so willingly, because of reasons I don't fully understand, but enough for me to pause.

Some more googling has given me this: `system("which xvfb-run > /dev/null 2>&1")` which is a) more minimal but also b) a little more brittle (will also return false if `which` isn't installed, or anything else goes wrong with this command), but given I don't think we have to worry about BOPS being installed on a windows system any time soon, I think it's a better trade off. What do you think?

Rhian:
> Just reading up on `mkmf` a bit more, it seems it's not advised to include this in your ruby projects so willingly, because of reasons I don't fully understand

That's good to know! I think your proposed solution seems sensible

james (approved):

PR #236, opened 2021-03-08
Improve/document publish and display

Rhian:
### Description of change

Two small improvements have been made to the logic that displays documents on the decision notice and/or makes them public:

- A helper method has been added to the document view which ensures that the numbers field does not display as a blank array in the input field if someone tries to save it when empty
- The remaining helper methods that belonged to the two-step document upload wizard have been removed, and specs added to ensure that document privacy preferences and numbers can be added at the time of creation.

### Story Link

https://trello.com/c/qQ8nOXNy/133-indicating-which-documents-to-include-on-decision-notice-and-which-one-will-be-public-or-private


app/controllers/documents_controller.rb
```diff
@@ -38,7 +38,11 @@ def new
   end
 
   def create
-    @document = @planning_application.documents.build(tags: document_params[:tags], file: document_params[:file])
+    @document = @planning_application.documents.build(tags: document_params[:tags], file: document_params[:file],
+                                                      numbers: document_params[:numbers],
+                                                      publishable: document_params[:publishable],
+                                                      referenced_in_decision_notice:
+                                                          document_params[:referenced_in_decision_notice])
```
james, line 45: `params` are a bit of a mess in this controller.

We have direct setting of params here, `documents_list_params`, `document_params`, `document_upload_params` and `document_form_params`! Surely we have simplified this controller now to the point we can just have one `document_params` and use it for `create` and `update`? Unless I am missing something.

I realise this is totally outside the scope of this ticket, but seems like a good opportunity to tidy it up?

james: I missed `form_params`!

james: I also think we have unused actions (I think all the confirm actions are not used any more?), and `set_document` is being used inconsistently (edit and update should all be using `set_document`, right?)

Again, all outside the ticket, but would be nice :-)

app/helpers/document_helper.rb
```diff
@@ -12,4 +12,8 @@ def filter_current(documents)
   def archive_reason_collection_for_radio_buttons
     Document.archive_reasons.keys.map { |k| [k, I18n.t("archive_reasons.#{k}")] }
   end
+
+  def override_numbers(document)
```
james, line 16: I don't quite understand what this method is doing or for? Maybe it just needs a clearer name?

app/controllers/documents_controller.rb
Rhian: I'm really pleased we were able to do this - it's much simpler :)

app/helpers/document_helper.rb
Rhian: By changing the numbers column type to a string, we are able to get rid of various helper methods (not just this one, but also the methods that split the string etc)

app/controllers/documents_controller.rb
```diff
@@ -4,11 +4,7 @@ class DocumentsController < AuthenticationController
   include ActiveStorage::SetCurrent
 
   before_action :set_planning_application
-  before_action :set_document, except: %i[index
-                                          new
-                                          create
-                                          edit
-                                          update]
+  before_action :set_document, only: %i[archive confirm_archive]
```
james, line 7: As lines 15 and 19 are doing this, can we add edit and update to this list, and remove those lines?

app/views/documents/index.html.erb
```diff
@@ -56,17 +56,17 @@
           <p class="govuk-body govuk-!-margin-bottom-1">
             Date received: <%= document.created_at.strftime("%e %B %Y") %>
           </p>
-          <% if document.numbers.present? %>
+          <% unless document.numbers.blank? %>
```
james, line 59: Smallest of quibbles, but this change feels like making it into a double negative? Really don't mind though.

spec/system/documents/archive_spec.rb
```diff
@@ -61,9 +61,8 @@
       click_link "Archive document"
     end
 
-    it "Archive page contains radio buttons and image" do
-      expect(page).to have_text "Missing scale bar or north arrow"
-      expect(page).to have_css(".govuk-radios__item")
+    it "Archive page contains expected text" do
+      expect(page).to have_text "Why do you want to archive this document?"
```
james, line 65: Not sure this is really testing anything here, and we can probably just delete it?

james (approved):
This looks great! And so many deletions :-)

All my comments are more just notes that actual change requests. Feel free to ignore them.

PR #237, opened 2021-03-08
Change logic for archive document function

Rhian:
### Description of change

We need to allow access to archive, edit, upload in all states except when application is closed or awaiting determination. Access to the routes has been restricted and the buttons and links hidden when the planning application is in any state other than: awaiting determination, determined, withdrawn or returned. We can reuse the `can_validate?` method for this.

### Story Link

https://trello.com/c/joo32duV/86-archive-documents-function-currently-only-visible-once-documents-have-been-validated-should-always-be-visible

### DIFF
--- a/app/views/documents/archive.html.erb
+++ b/app/views/documents/archive.html.erb
@@ -1,34 +1,32 @@
-<% if @planning_application.can_validate? %>
-  <% add_parent_breadcrumb_link "Home", planning_applications_path %>
-  <% add_parent_breadcrumb_link "Application", planning_application_path(@planning_application) %>
-  <% add_parent_breadcrumb_link "Documents", planning_application_documents_path(@planning_application) %>
+<% add_parent_breadcrumb_link "Home", planning_applications_path %>
+<% add_parent_breadcrumb_link "Application", planning_application_path(@planning_application) %>
+<% add_parent_breadcrumb_link "Documents", planning_application_documents_path(@planning_application) %>
 
-  <% content_for :title, "Archive document" %>
+<% content_for :title, "Archive document" %>
 
-  <div class="govuk-grid-row">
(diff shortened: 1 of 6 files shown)


app/views/documents/archive.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if current_user.assessor? %>
+<% if @planning_application.can_validate? %>
```
james, line 1: We can remove this `if` altogether, seeing as we forbid seeing this page if this is false.

james (approved):
Just one unnecessary logic statement, but all good!

spec/system/planning_applications/validating_spec.rb
```diff
@@ -55,6 +55,19 @@
...
+    click_link "Validate documents"
+
+    choose "No"
+
+    click_button "Save"
+
+    expect(page).to have_content("Application has been invalidated")
+
+    planning_application.reload
+  end
```
melzreal, line 70: not sure this test does what the description says? it just seems to be invalidating a planning application, not "allowing documents to be edited, archived and uploaded"?

melzreal (commented):
LGTM!

melzreal (approved):
:shipit: :shipit:

app/views/documents/index.html.erb
```diff
@@ -68,52 +68,53 @@
...
+            <%= link_to "Archive document", planning_application_document_archive_path(document_id: document.id) %>
+          </p>
+        </div>
+      <% end %>
     </li>
     <% end %>
   <% end %>
 </ul>
 
 <% unless @documents.nil? %>
```
melzreal, line 84: blank? or present? are usually safer then nil? as they cover more cases - we also seem to use them everywhere else so perhaps going with blank? here would be best

PR #238, opened 2021-03-09
Upgrade rails

melzreal:
### Description of change

Upgrade rails to 6.1.3

### Story Link

https://trello.com/c/XK2rojew/222-upgrade-app-to-rails-613

### Decisions

- Errors no longer accept two arguments, this commit fixes this deprecation warning:
```
You are passing a block expecting two parameters, so the old hash behavior is simulated. As this is deprecated,
this will result in an ArgumentError in Rails 6.2. (called from views/planning_applications/recommendation_form.html.erb:16)
.DEPRECATION WARNING: Enumerating ActiveModel::Errors as a hash has been deprecated.
```

- Rendering actions with '.' in the name is deprecated so all partials with .html or .json now have the simplified naming convention
- raise_on_missing_translations has been replaced with i18n_raise_on_missing_translations due to deprecation
- host! has been deprecated with no replacement - the only place where we don't get any deprecation warnings is in the "request" setting in rails helper, where we do need it or all api requests fail.
I could not find an alternative to host! that would force the GET URI to not have www and since no warnings appear it seemed ok to keep.


config.ru
```diff
@@ -5,3 +5,4 @@
 require_relative "config/environment"
 
 run Rails.application
+Rails.application.load_server
```
melzreal, line 8: the rails app update would replace what we had with this, so I'm not sure if these statements could conflict or not. To review?

config/puma.rb
```diff
@@ -1,5 +1,49 @@
...
+# Use the `preload_app!` method when specifying a `workers` number.
+# This directive tells Puma to first boot the application and load code
+# before forking the application. This takes advantage of Copy On Write
+# process behavior so workers use less memory.
+#
+# preload_app!
+
+# Allow puma to be restarted by `rails restart` command.
+plugin :tmp_restart
+
```
melzreal, line 46: I've added the new puma config on top of our existing one. It brings in all these changes which I guess only affect development so should be ok to keep, but maybe they need to be wrapped in a if ENV == development ?

app/views/documents/index.html.erb
```diff
@@ -34,40 +34,31 @@
...
-        </div>
+        </p>
+
+        <p class="govuk-body">
+          <% document.tags.each do |tag| %>
+            <strong class="govuk-tag govuk-tag--turquoise"><%= tag %></strong>
+          <% end %>
+        </p>
+      </div>
+
```
james, line 61: Are the changes to this file relevant to the Rails upgrade?

config/puma.rb
james: Does it work if you don't add all of this? These feel like a template of default values that we don't need?

james:
Good work so far. I think you need to go through new_framework_defaults_6_1.rb and see if you can painlessly flick them all on?

config/puma.rb
melzreal: it does - I wasn't sure how important the options were to left them - will take them out

config/initializers/new_framework_defaults_6_1.rb
```diff
@@ -0,0 +1,67 @@
+# Be sure to restart your server when you modify this file.
+#
+# This file contains migration options to ease your Rails 6.1 upgrade.
+#
+# Once upgraded flip defaults one by one to migrate to the new default.
```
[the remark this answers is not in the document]
melzreal: First time I've noticed the comments were meant to do something - going through it now. There are a few that say ```# It's best enabled when your entire app is migrated and stable on 6.1.``` so I'll go for all the other ones that don't specify that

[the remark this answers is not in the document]
melzreal: ah on a side note: I also had to run "bundle update" to make the app work. But while that changed the Gemfile.lock it didn't change the actual Gemfile, so maybe I need to manually add the latest versions of the gems manually to the gemfile?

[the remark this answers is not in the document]
james: This is how Gemfiles work. The Gemfile defines the gems, and can limit the versions (either to a specific version, or a range https://bundler.io/gemfile.html) and then the lockfile sets it to a specific version. When you run "bundle update" it will check to see if there are any newer versions of those gems that match the Gemfile.

james (approved):

PR #239, opened 2021-03-10
Edit and create application using front-end

melzreal:
### Description of change

- Allow for manual creation of applications, and allow for these to be edited. 
- Ensure all update actions are tracked in audit log
- Detail what exact actions were taken and what elements were updated in audit log
- Small erb changes were also added to match the designs in the ticket (https://trello-attachments.s3.amazonaws.com/5faad383fa04d62473073c19/60425473625e9355ab2be669/b33122ad31616427ad5d87b236e4691b/Screenshot_2021-03-05_at_16.00.16.png)

### Story Link

https://trello.com/c/6nFn0oFQ/221-edit-documents-add-new-application

### Decisions
I spoke to Ali about some design choices:
- Its ok to keep First name and Full name in separate fields despite the design having a "Full name" field
- The "date received" field will be manipulated to change the "created_at" date which is what we are currently using in the api - so I've kept the field there for assessors to be able to manually amend the "created at"


app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -38,6 +38,12 @@ def create
 
     if @planning_application.valid? && @planning_application.save!
       upload_documents(params[:files])
+      Audit.create!(
+        planning_application_id: @planning_application.id,
+        api_user_id: current_api_user.id,
```
james, line 43: Can we have `api_user: current_api_user` here?

app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -38,6 +38,12 @@ def create
 
     if @planning_application.valid? && @planning_application.save!
       upload_documents(params[:files])
+      Audit.create!(
+        planning_application_id: @planning_application.id,
```
james, line 42: and `planning_application: @planning_application` here

app/controllers/planning_applications_controller.rb
```diff
@@ -155,6 +190,43 @@ def decision_notice
...
+                        created_at(1i)
+                        description
+                        proposal_details
+                        payment_reference
+                        postcode
+                        town
+                        uprn
+                        ward
+                        work_status]
+    params.fetch(:planning_application, {}).permit permitted_keys
```
james, line 217: why are we use `fetch(:planning_application)` instead of `require(:planning_application)` here? Is there any situation that `params[:planning_application]` wouldn't be set?

app/controllers/planning_applications_controller.rb
```diff
@@ -155,6 +190,43 @@ def decision_notice
...
+                        applicant_phone
+                        applicant_email
+                        agent_first_name
+                        agent_last_name
+                        agent_phone
+                        agent_email
+                        county
+                        created_at(3i)
+                        created_at(2i)
+                        created_at(1i)
```
james, line 208: as we are manually setting the created_at here, I think we can remove them from this mass assignment?

app/controllers/planning_applications_controller.rb
melzreal: I went with the same convention we have in document_params right now:

```document_params = params.fetch(:document, {}).permit(:archive_reason, :name, :archived_at,
                                                         :numbers, :publishable, :referenced_in_decision_notice, :file, tags: [])```
   and because we are not actually requiring them, they are all optional - so it seemed to make sense?

File: app/controllers/planning_applications_controller.rb (hunk 4 above)
Comment (melzreal) [reply app/controllers/planning_applications_controller.rb:208]: but then they won't be registered as allowed params when coming in from the form and updating as planning_application_params?

File: app/controllers/planning_applications_controller.rb (hunk 3 above)
Comment (melzreal) [reply app/controllers/planning_applications_controller.rb]: changed it

File: app/controllers/planning_applications_controller.rb (hunk 4 above)
Comment (melzreal) [reply app/controllers/planning_applications_controller.rb:208]: yep I can't remove these or the update action stops working for the date

File: app/views/audits/index.html.erb [hunk 5]
```diff
@@ -15,18 +15,17 @@
...
+        <% if audit_item.user %>
+          <td class="govuk-table__cell"><%= audit_item.user.name %></td>
+        <% elsif audit_item.api_user %>
+          <td class="govuk-table__cell"><%= audit_item.api_user.name %></td>
+        <% end %>
         <td colspan="2" class="govuk-table__cell">
-          <%= activity(audit_item.activity_type, audit_item.activity_information) %> <br>
-          <i> <%= audit_item.audit_comment if audit_item.audit_comment.present? %> </i>
+          <%= activity(audit_item.activity_type, audit_item.activity_information) %>
+          <i> <%= simple_format(audit_item.audit_comment, {}, wrapper_tag: "div") if audit_item.audit_comment.present? %> </i>
```
james, line 28: Two things here: I think the i tag is frowned upon, and em is preferred? I might be out of date here though. But also what you end up doing here is wrapping a p tag (created by the simple_format) in an i tag, when it should be the other way around. I can't easily find a way of specifying a class name on the p tag simple_format creates, so maybe the best solution is the wrap this all in a div with a class name and set the CSS on that div to italicize the text?

app/controllers/planning_applications_controller.rb
james: You aren't setting these params in the mass assignment on line 46. You are manually setting it on line 47, so it doesn't need to be allowed via strong params

app/views/planning_applications/_form.html.erb
```diff
@@ -0,0 +1,94 @@
+<%= form_with model: @planning_application, local: true, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |form| %>
+  <% if @planning_application.errors.any? %>
+  <span id="status-error" class="govuk-error-message">
```
james, line 3: The contents of this `if` would be more readable if indented

app/views/planning_applications/_form.html.erb
```diff
@@ -0,0 +1,94 @@
...
+    <span class="govuk-visually-hidden">
+      <h2>
+        <%= pluralize(@planning_application.errors.count, "error") %>
+        prohibited this @planning_application from being saved:
+      </h2>
+    </span>
+    <ul>
+      <% @planning_application.errors.messages.each do |error| %>
+      <li><%= error %></li>
```
james, line 12: another lack of indentation

app/controllers/planning_applications_controller.rb
```diff
@@ -155,6 +190,43 @@ def decision_notice
...
+                        description
+                        proposal_details
+                        payment_reference
+                        postcode
+                        town
+                        uprn
+                        ward
```
james, line 215: Ward isn't being set in the form, so we can remove it from here.

app/views/audits/index.html.erb
melzreal: I can add a class to audit_item to make it italicize with the gov uk library if we don't like i tags (although I don't see how they are different from other tags (?) 

doing ```wrapper_tag div ``` ensures we get the content inside a div instead of inside a paragraph, because the paragraph was adding extra line breaks in between content which we didn't want. so just 

```
simple_format(audit_item.audit_comment

```
produced this:

image

but what we currently have produces this:

image
image

james: I was suggesting manually wrapping the simple_format in a div instead of `i`, so:

`<div class="audit_details"><%= simple_format(audit_item.audit_comment, {}, wrapper_tag: "div") if audit_item.audit_comment.present? %></div>`

and then having

```
.audit_details {
  font-style: italic;
}
```

in the CSS

james: context: `i` used to mean `italic`, which is presentational not semantic, so is frowned upon, with a preference for `em`. I see that they've changed the meaning of `i` to mean "[Idiomatic Text](Idiomatic Text)", which is news to me, but I see govuk still suggests `em`, so I'd stick with that.

But in this context, either `i` or `em` is an inline element, and `p` is a block element, and you shouldn't wrap a block element in an inline element.

app/views/audits/index.html.erb
```diff
@@ -15,18 +15,17 @@
...
-      <td class="govuk-table__cell"> <%= @planning_application.created_at.strftime("%d-%m-%Y %H:%M") %></td>
-      <td class="govuk-table__cell"> API </td>
-      <td class="govuk-table__cell"> Application created </td>
-    </tr>
     <% @audits.each do |audit_item| %>
       <tr class="govuk-table__row">
         <td class="govuk-table__cell"><%= audit_item.created_at.strftime("%d-%m-%Y %H:%M") %></td>
-        <td class="govuk-table__cell"><%= audit_item.user.name %></td>
+        <% if audit_item.user %>
+          <td class="govuk-table__cell"><%= audit_item.user.name %></td>
```
james, line 22: One last thing: what happens is a user or api_user gets deleted? I think it would cause an error on the audit log page? Two ways of resolving this:

1) we set `has_many :audits, dependent: :destroy` on User and ApiUser, which will ensure that if we delete a user, the audit logs get deleted to
2) we set `has_many :audits, dependent: :nullify` on User and ApiUser, and ensure that on this line we catch if `audit_item.user` is nil, we say "User deleted" instead.

I think the second option is better?

app/controllers/planning_applications_controller.rb
melzreal: I'm not setting them on the "create" action, but in the update action I don't explicitly assign those date attributes. I simply call
```
@planning_application.update(planning_application_params)
```
and the date gets updated. However, if I remove those 3 from strong params, the date update doesn't happen at all, because the created_at is no longer accepted as a param. I would need to instead of just calling
@planning_application.update(planning_application_params) also do some extra work on update to assign time attributes that now get updated just because of that params.

melzreal: if you give a branch a pull try deleting those params and then running spec/system/planning_applications/update_spec.rb

app/views/audits/index.html.erb
melzreal: that makes sense thanks - I'll change it

app/views/audits/index.html.erb
melzreal: I tried deleting it and nothing much happened:

image

will see if dependent nullify keeps the record whilst having an empty user, probably that's indeed the best way to go

spec/system/planning_applications/create_spec.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
```
Rhian, line 1: Not sure if this is spec overload, but maybe we should also include a spec to ensure the reviewer can also create an application? And that an invalid user cannot access the form page?

app/views/audits/index.html.erb
melzreal: so because I did not enforce any has_many (we only have a belongs_to on the api, and it has "optional true") it was enough to just add an "else" statement to have the user as deleted

image

```
      <td class="govuk-table__cell"><%= audit_item.created_at.strftime("%d-%m-%Y %H:%M") %></td>
        <% if audit_item.user %>
          <td class="govuk-table__cell"><%= audit_item.user.name %></td>
        <% elsif audit_item.api_user %>
          <td class="govuk-table__cell"><%= audit_item.api_user.name %></td>
        <% else %>
        <% end %>
       ``` 
       
       this seemed like the "minimum effort" approach. Since the users are always marked as optional on the model I didn't see any possible problems with just doing this little catch clause.. sounds ok?

File: app/controllers/planning_applications_controller.rb (hunk 4 above)
Comment (james) [reply app/controllers/planning_applications_controller.rb:208]: Huh, that's weird. Why is it working via mass assignment on update, but not create? I can't look at it right now, but will try and take a look in a short while.

File: app/views/audits/index.html.erb (hunk 9 above)
Comment (james) [reply app/views/audits/index.html.erb:22]: That works well :-)

I'd still like to have the has_many dependent nullify, as that way we won't be left with records that have a user_id/api_user_id that points to a deleted record, if that makes sense.

Comment (james) [reply app/views/audits/index.html.erb:22]: Not that that will actually prevent any errors, but might help us out if we're trying to debug something in a hurry later on.

File: spec/system/planning_applications/create_spec.rb (hunk 10 above)
Comment (melzreal) [reply spec/system/planning_applications/create_spec.rb:1]: on the invalid user, that would be covered by the logins/ local authority specs - since a user belonging to a different local authority cannot login to this one or access anything - and we removed pundit so there were no specific policies added for this use case that would require spec coverage regarding access?


on the reviewer one will it now  :)

Comment (james) [reply spec/system/planning_applications/create_spec.rb:1]: I don't think we need to duplicate all of them for reviewer, but checking that we don't allow logged out users accessing the form is good

File: app/controllers/planning_applications_controller.rb (hunk 4 above)
Comment (melzreal) [reply app/controllers/planning_applications_controller.rb:208]: it does also work on create, I just coped the time zone_parse method I'd seen as I assumed it was necessary.. but it wasn't! Deleted the extra method - my bad

File: spec/system/planning_applications/create_spec.rb (hunk 10 above)
Comment (melzreal) [reply spec/system/planning_applications/create_spec.rb:1]: I've added it in - although the test spec looks exactly like any "logged out" action so it feels redundant

File: app/controllers/planning_applications_controller.rb [hunk 11]
```diff
@@ -34,6 +36,39 @@ def index
...
+
+  def update
+    if @planning_application.update(planning_application_params)
+      planning_application_params.keys.map do |p|
+        if @planning_application.saved_change_to_attribute?(p)
+          audit("updated", "Changed from: #{@planning_application.saved_changes[p][0]} \r\n Changed to: #{@planning_application.saved_changes[p][1]}", p.split("_").join(" ").capitalize)
+        end
+      end
+      flash[:notice] = "Planning application was successfully updated."
+      redirect_to planning_application_url
```
james, line 66: huh, how is this working without passing the @planning_application to this method? I'd have thought this gave an error, but I guess it's not? Shouldn't it be `redirect_to planning_application_url(@planning_application)`

db/migrate/20210311152412_add_api_user_to_audit.rb
```diff
@@ -0,0 +1,6 @@
+class AddApiUserToAudit < ActiveRecord::Migration[6.0]
+  def change
+    add_column :audits, :api_user_id, :integer
+    add_index :audits, :api_user_id
```
james, line 4: This is absolutely fine, but just as a tip, you can achieve the same thing as these two lines with `add_reference(: audits, :api_user, foreign_key: true)`

app/controllers/planning_applications_controller.rb
melzreal: this one not only works  - it also doesn't create that problem with brakeman unlike what happens everytime we use @planning_application - brakeman doesn't like the @planning_application id params for some reason.

it seems that rails routes has a neat little feature where the id is filled from the action:

https://stackoverflow.com/questions/40785322/why-redirecting-to-show-action-without-passing-the-id-param-work-in-rails

db/migrate/20210311152412_add_api_user_to_audit.rb
melzreal: ah yes, I actually did that last time I added an index - but this time I was lazy and copy pasted it from another one of our files hahahaaa will change it, its the last migration so best be neat

app/controllers/planning_applications_controller.rb
james: huh! I learned a thing.

james (approved):

PR #244, opened 2021-03-19
Rescue notify errors and send them to AppSignal

melzreal:
### Description of change

Catching errors that notify provides to us so we can rescue these instead of hitting a 500 page on the front end.

### Story Link

https://trello.com/c/0k4XVeKD/98-catch-more-notify-errors

### Decisions 

Moved the rescue clauses to the application controller which should by default inherit from ActiveSupport::Rescuable, and because we didn't seem to be hitting the single rescue_from on the planning application controller
Consolidated flashes success with the flashes erb file we already had
Changed config.action_mailer.raise_delivery_errors = true so we can raise errors for failed emails

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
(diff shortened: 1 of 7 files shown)


james (approved):
No comments, looks good to me!

PR #245, opened 2021-03-22
Upgrade webpacker to fix security dependency

melzreal:
### Description of change
Upgrading webpacker due to a security warning affecting an svg package that the old webpacker version depended on:
https://github.com/unboxed/bops/security/dependabot/yarn.lock/is-svg/open

- Also update github actions file inline with the comment on the GithubActions for Yarn:
https://github.com/marketplace/actions/github-action-for-yarn

```
Please keep in mind that this Action was originally written for GitHub Actions beta (when Docker was the only way of doing things).
Consider using actions/setup-node to work with Yarn.
This repository will be mostly supporting the existing flows.
```

### DIFF
--- a/config/webpacker.yml
+++ b/config/webpacker.yml
@@ -6,7 +6,6 @@ default: &default
   public_root_path: public
   public_output_path: packs
   cache_path: tmp/cache/webpacker
-  check_yarn_integrity: false
   webpack_compile_output: true
 
   # Additional paths webpack should lookup modules
@@ -54,7 +53,7 @@ development:
   extract_css: true
 
   # Verifies that correct packages and versions are installed by inspecting package.json, yarn.lock, and node_modules
-  check_yarn_integrity: false
+  check_yarn_integrity: true
 
   # Reference: https://webpack.js.org/configuration/dev-server/
   dev_server:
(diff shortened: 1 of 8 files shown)


config/webpacker.yml
```diff
@@ -51,10 +50,6 @@ default: &default
 development:
   <<: *default
   compile: true
-  extract_css: true
```
james, line 54: So I had to turn this on previously because:

a) having the CSS loaded in via Javascript in development, but not production just makes me very nervous. Probably because I'm being freaked out by all the modern day FE pipeline stuff.
b) I was validated in me being weirded out! The PDF renderer can't handle this approach to loading CSS, so in development, the PDFs lose their CSS

And I can see that webpacker still behaves like this if we remove this line: the PDFs don't have styling in development mode.

I have a proposed solution, based off of skimming this very detailed discussion: https://github.com/rails/webpacker/issues/2605

I created a new file in `app/javascript/packs/application.scss` which just included the following line:

`@import "stylesheets/application";`

I then turned `extract_css` back on for development. This appears to have fixed everything for me locally.

james (approved):

PR #247, opened 2021-03-22
Amend flash and dev config for delivery errors

melzreal:
### Description of change
Bug fixes: 
- Make sure error warning only shows on failed emails, not on all the other flash messages within the documents page (currently error box shows when an upload is successful)
- Set delivery errors in dev environment to false

### DIFF
--- a/app/views/application/_flash_content.html.erb
+++ b/app/views/application/_flash_content.html.erb
@@ -1,15 +1,5 @@
 <% flash.each do |name, msg| %>
-  <% if msg.include?("success") %>
-  <div class="flash-archive">
-    <div class="govuk-error-summary__body govuk-!-padding-2">
-      <svg class="moj-banner__icon" color="green" role="presentation" focusable="false" xmlns="http://www.w3.org/2000/svg"
-        viewBox="0 0 25 25" height="25" width="25">
-        <path d="M25,6.2L8.7,23.2L0,14.1l4-4.2l4.7,4.9L21,2L25,6.2z"></path>
-      </svg>
-      <%= msg %>
-    </div>
-  </div>
-  <% else %>
+  <% if msg.include?("unable") %>
   <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1">
     <svg class="alert__icon" fill="red" role="presentation" focusable="false" xmlns="http://www.w3.org/2000/svg"
       viewBox="0 0 25 25" height="25" width="25">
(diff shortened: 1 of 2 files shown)


app/views/application/_flash_content.html.erb
```diff
@@ -1,15 +1,5 @@
...
-    <div class="govuk-error-summary__body govuk-!-padding-2">
-      <svg class="moj-banner__icon" color="green" role="presentation" focusable="false" xmlns="http://www.w3.org/2000/svg"
-        viewBox="0 0 25 25" height="25" width="25">
-        <path d="M25,6.2L8.7,23.2L0,14.1l4-4.2l4.7,4.9L21,2L25,6.2z"></path>
-      </svg>
-      <%= msg %>
-    </div>
-  </div>
-  <% else %>
+  <% if msg.include?("unable") %>
```
james, line 2: I've only just noticed the way that you're handling flashes. Flash is a hash, which means we can scope each type of message based on what the key of the message is. In our application, we currently have two types of flash: `flash[:error]` and `flash[:notice]`.

I think it would be better to scope this to those two flash types, so for example:

```erb
<% if flash[:error].present? %>
  ...
<% end %>
<% if flash[:notice].present? %>
  ...
<% end %>
```

rather than iterating on everything that is put into flash, and scoping it based on the contents of the flash message.

I think this would be more explicit, but also means that if we want to create a new the flash type for a message somewhere else in the design, it won't accidentally turn up here as well.

melzreal: ah also this file is exclusive to - app/views/documents/index.html.erb - I'm not amending the global flashes file, since we don't want the flash warning that shows at the top of the page to be changed. where we had just a flash_success in documents before now I render this file which has the success and warning stylings.

```
<%= render "flash_content" %>
```

james: I'm saying I don't think we should be iterating over the flash messages with `<%flash.each do |name, msg| %>` at all

melzreal: but that's what we are already doing in all our defaults:
```
app/views/application/_flashes.html.erb
```

has 
```
<% flash.each do |name, msg| %>
  <div class="govuk-width-container">
    <div class="flash <%= name %>">
      <p class="govuk-!-margin-top-1 govuk-!-margin-bottom-1">
        <%= msg %>
      </p>
    </div>
  </div>
<% end %>
```

so I'm only following that convention we used in the rest of the app here - I can change it but I'm not understanding what is the suggested alternative?

melzreal: either way this is a small fix that just changes the flash order - if we want to rethink flashes across the app maybe we need a separate ticket instead of doing it on a non-pointed thing I was trying to do before anyone realised the "bug" of showing success messages with a red box on the documents upload page...

james (approved):

PR #250, opened 2021-03-24
Add email reset functionality using notify

melzreal:
### Description of change
https://trello.com/c/ZHJQhOMN/240-fix-forgotten-password-error


.rubocop.yml
```diff
@@ -11,6 +11,7 @@ inherit_mode:
 AllCops:
   Exclude:
     - "db/migrate/*"
+    - "app/mailers/devise_mailer.rb"
```
james, line 14: what was failing in this file? If there's a specific rule that doesn't make sense, I'd prefer we exclude that particular cop, either through this file, or with `# rubocop:disable` comment

app/controllers/passwords_controller.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class PasswordsController < Devise::PasswordsController
+  skip_before_action :find_current_local_authority_from_subdomain
```
james, line 4: Why would we not have a local authority?

app/mailers/application_mailer.rb
```diff
@@ -1,6 +1,4 @@
 # frozen_string_literal: true
 
-class ApplicationMailer < ActionMailer::Base
-  default from: "from@example.com"
-  layout "mailer"
```
james, line 5: nice to clean this up :-)

app/views/application/_header.html.erb
```diff
@@ -1,7 +1,11 @@
 <header class="govuk-header  with-proposition " role="banner" data-module="govuk-header">
   <div class="govuk-header__container govuk-width-container">
     <div class="govuk-header__content">
-      <%= link_to "Back-office Planning System: #{current_local_authority.name}", root_path, class: "govuk-header__link govuk-header__link--service-name" %>
+      <% if current_local_authority %>
+        <%= link_to "Back-office Planning System: #{current_local_authority.name}", root_path, class: "govuk-header__link govuk-header__link--service-name" %>
+      <% else %>
+        <%= link_to "Back-office Planning System", root_path, class: "govuk-header__link govuk-header__link--service-name" %>
+      <% end %>
```
james, line 8: Again, why would the local authority not be present?

config/environments/production.rb
```diff
@@ -65,6 +65,7 @@
   config.action_mailer.notify_settings = {
     api_key: ENV["NOTIFY_API_KEY"],
   }
+  config.action_mailer.default_url_options = { host: (ENV["DOMAIN"] || "bops.services"), port: 3000 }
```
james, line 68: whoops, we don't want to be setting port 3000 here!

.rubocop.yml
melzreal: rubocop complained about this entire line:
https://github.com/heartcombo/devise/blob/45b831c4ea5a35914037bd27fe88b76d7b3683a4/lib/devise/mailers/helpers.rb#L17

because my devise mailer overrides the call made to method above, we need to get rubocop to ignore the problems it has with the line to make sure the override works properly - but I didn't want to make a rule for the whole app because it must complain for a reason (it doesn't like the fact that opts = {} is not in the end of the method call and it doesn't like &block - it prefers _block 

```
class DeviseMailer < Devise::Mailer
  def devise_mail(record, action, opts = {}, &block)
    initialize_from_record(record)
    view_mail("c978a30f-0578-4636-b07e-1e497c3b3893", headers_for(action, opts))
  end
end

```

james: Ah yes, that makes sense. But as I mentioned, I'd prefer it if we were just disabling that single cop rather than AllCops for the file, similar to how to have disabled Rails cops at https://github.com/unboxed/bops/blob/master/.rubocop.yml#L15 - what is currently here disables all of rubocop for this file.

james (approved):

PR #253, opened 2021-03-24
Update mimemagic to patched version

melzreal:
### Description of change
Mimemagic 0.3.2 is broken. Latest patched version is  0.3.6 but that is now released under a GPL-2 license

https://github.com/minad/mimemagic/issues/98#issuecomment-805766052

Rails maintainers are working on a replacement that uses MIT, but that is work in progress:
https://github.com/rails/rails/pull/41751


Rhian (approved):
Looks good to me from a technical POV, but let's wait and think about the further implications before merging.

Gemfile.lock
```diff
@@ -178,7 +178,7 @@ GEM
     marcel (0.3.3)
       mimemagic (~> 0.3.2)
```
kceb, line 179: does this one need to be updated too?

melzreal:
Closing this, as the rails team is working on the apache based solution that should be available soon - so its best to wait for that instead of patching to the 0.3.7 (this one now released under MIT but apparently has some caveats) - https://github.com/rails/rails/issues/41757

PR #254, opened 2021-03-25
Add edit constraints form

Rhian:
constraints

This commit contains controller and helper methods to parse the JSON, check it against incoming and existing constraints and remove where necessary

### Description of change

This logic needs to do several things:

- Allow the planner to select from a selection of fixed constraints, grouped into categories
- Allow the planner to enter freeform text
- Display any constraints that have been submitted via the API as checkboxes in the form. These should be displayed as checked if they match any of the existing categories, or should be added as checkboxes in the 'Other' category if they are extra
- Display any constraints in the relevant accordion
- Filter constraints submitted via the API so that those marked 'false' do not display (this is a bugfix)

### Story Link

https://trello.com/c/78OC9dEt/220-build-add-constraints-screen


app/controllers/planning_applications_controller.rb
```diff
@@ -165,6 +168,19 @@ def determine
     redirect_to @planning_application
   end
 
+  def edit_constraints_form; end
+
+  def edit_constraints
+    constraint = params.reject! { |k, v| v.blank? || k.include?("-true") }.select { |k, _v| k.include?("constraint") }
+    constraint[params[:local].to_s] = true if params[:local]
+    if @planning_application.update(constraints: constraint.to_json)
```
melzreal, line 176: this will throw a rubocop error unless we use assign_attribute (it wants the update action to be called as update!)
An alternative if you cant use update! is to use assign_attributes(constraints) and then if @planning_application.save in the next line

app/controllers/planning_applications_controller.rb
```diff
@@ -250,4 +267,8 @@ def validation_notice_mail
   def ensure_user_is_reviewer
     render plain: "forbidden", status: 403 and return unless current_user.reviewer?
   end
+
+  def ensure_constraint_edits_unlocked
```
melzreal, line 271: Would it not make more sense to have a "constraints" controller?

melzreal: I'm thinking it'd be a nice way to separate the code, having a constraints controller with "create" and "new" actions for the constraints and dedicated helper methods / model validations for later 🤔

spec/system/planning_applications/constraints_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+      click_button("Save")
+
+      expect(page).to have_no_content("protected_trees")
+      expect(page).to have_no_content("false_constraint")
+    end
+
+    it "correctly displays constraint not in the dictionary submitted via the input field" do
+      click_link "Update"
+
+      fill_in "local", with: "Bats live here"
```
melzreal, line 80: ❤️ 🦇  xD

spec/system/planning_applications/constraints_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  context "Editing constraints" do
+    it "can add and change constraints" do
+      click_link("Update")
+
+      check("Listed building")
+      check("Broads")
+      check("Flood zone 3")
+      uncheck("Conservation Area")
+      # The factory provides Conservation Areaa as a pre-existing constraint, so checking it will remove it
```
melzreal, line 26: "edit" / "update" are much easier to read / understand when declaring what we are updating at the top - factories should make it easier to create default attributes but we need to know what those attributes are in order to make sure the update is working correctly - so I'd say it needs to be part of this file. Could we declare the list of attributes at the top that we are modifying here, otherwise we need to check in 2 different places (the factory and here) to know what was updated

app/controllers/planning_applications_controller.rb
melzreal: I also feel that this bit would be simplified by having a dedicated controller that accepts all these attributes as contraints params - in the building safety platform we did that here: 
https://github.com/unboxed/southwark-bsp/blob/main/app/controllers/surveys/building_external_wall_structures_controller.rb

thus allowing us to call all the params as "booleans" except one of the params "other" as a string, so instead of having all these transformations its true that a lot more text is written but then we can just call Constraints.update(contraints_params) 🤔  it feels like we are doing a lot of legwork instead of relying on the native rails methods

melzreal: image
params[:local] is already type string so we don't need the to_s

app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -31,7 +31,7 @@ def create
         local_authority_id: @current_local_authority.id,
         boundary_geojson: (params[:boundary_geojson].to_json if params[:boundary_geojson].present?),
         proposal_details: (params[:proposal_details].to_json if params[:proposal_details].present?),
-        constraints: (params[:constraints].to_json if params[:constraints].present?),
+        constraints: (params[:constraints].present? ? params[:constraints].to_unsafe_hash.collect { |key, value| key if value }.compact : []),
```
james, line 34: maybe this should be broken into a private method on the controller called something like `constraints_array_from_param` or something?

spec/system/planning_applications/constraints_spec.rb
Rhian: Agreed - have updated this with the constraints explicitly set in the spec 👍

app/controllers/api/v1/planning_applications_controller.rb
Rhian: I've done this now and yes, much better :)

app/helpers/planning_application_helper.rb
```diff
@@ -27,12 +27,48 @@ def filter_text
...
+  def existing_constraints(constraints)
+    constraints - standard_constraints
+  end
+
+  def standard_constraints
+    [flooding_constraints, military_constraints, ecology_constraints, heritage_constraints, policy_constraints, tree_constraints, other_constraints].flatten
+  end
+
+  def check_added?(constraint, existing_constraints)
+    JSON.parse(existing_constraints).include?(constraint) ? true : false
```
james, line 67: Again, not sure on the naming of this method, not entirely clear what it is for. And does it need JSON.parse? I thought we'd got rid of JSON?

app/views/planning_applications/edit_constraints_form.html.erb
```diff
@@ -0,0 +1,127 @@
...
+<%= form_with model: @planning_application, url: edit_constraints_planning_application_path(@planning_application), local: true do |form| %>
+
+
+  <div class="govuk-grid-column-two-thirds">
+
+    <div class="govuk-grid-row govuk-!-margin-top-6">
+      <h3 class="govuk-heading-s">General Policy</h3>
+      <div class="form-group">
+        <div class="govuk-checkboxes govuk-checkboxes--small">
+          <%= form.collection_check_boxes :constraints, policy_constraints, :itself, :itself do |b| %>
```
james, line 18: oo, I didn't know about the :itself keyword! very nice.

app/views/planning_applications/edit_constraints_form.html.erb
```diff
@@ -0,0 +1,127 @@
...
+        <div class="govuk-checkboxes govuk-checkboxes--small">
+          <%= form.collection_check_boxes :constraints, heritage_constraints, :itself, :itself do |b| %>
+            <div class="govuk-checkboxes__item">
+              <%= b.check_box class: "govuk-checkboxes__input" %>
+              <%= b.label class: "govuk-label govuk-checkboxes__label tag_checkbox_label" %>
+            </div>
+          <% end %>
+        </div>
+      </div>
+    </div>
```
james, line 40: As this is repeatedly duplicated code, I'm wondering if we could DRY it up a bit. Something like:

```erb
<% {
   "Heritage & Conservation" => heritage_constraints,
...
}.each do |constraint_group_title, constraints_array| %>
...
<h3 class="govuk-heading-s"><%= title %></h3>
...
<%= form.collection_check_boxes :constraints, constraints_array, :itself, :itself do |b| %>
...
<% end %>
```

app/views/shared/_supporting_information.html.erb
```diff
@@ -14,11 +14,14 @@
     <div class="govuk-accordion__section-content" aria-labelledby="accordion-default-heading-1">
       <ul class="govuk-list govuk-list--bullet">
         <% if @planning_application.constraints %>
```
james, line 16: this should be `@planning_application.constraints.present?` if we want this check, as the default is an empty array, which is "thruthy" - but also it doesn't seem to be doing anything anyway, so we can remove it.

spec/system/planning_applications/constraints_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+
+    it "correctly displays a constraint not in the approved list supplied by the API" do
+      new_constraints = { conservation_area: true, protected_trees: false, bat_habitat: true }.to_json
+      planning_application.update!(constraints: new_constraints)
+      planning_application.reload
+      click_link("Update")
+
+      expect(page).to have_field("bat_habitat", checked: true)
+    end
```
james, line 61: How is this working? We're setting constraints directly on the model as a json hash but it's working anyway? Maybe Rails is doing something clever and making this work. Either way, this test doesn't seem quite right!

app/helpers/planning_application_helper.rb
```diff
@@ -27,12 +27,48 @@ def filter_text
...
+
+  def other_constraints
+    ["Safety hazard area", "Within 3km of the perimeter of an aerodrome"]
   end
 
-  def list_constraints(constraints)
-    JSON.parse(constraints).select { |_category, value| value == true }.keys unless constraints.empty?
+  def existing_constraints(constraints)
+    constraints - standard_constraints
+  end
```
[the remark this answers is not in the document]
Rhian: I'll change the method name. I was trying to get across that these are constraints that have already been added and attached to the planning application, but 'custom' is probably more descriptive

app/helpers/planning_application_helper.rb
Rhian: We can get rid of this method - it was left over from before. The reason it's not throwing an error is because it's not used anywhere

app/views/planning_applications/edit_constraints_form.html.erb
Rhian: Good idea - will implement this.

app/views/shared/_supporting_information.html.erb
Rhian: Oh yes, we can remove this now. Before we had a default, we needed it because it would have thrown an error if there weren't any.

spec/system/planning_applications/constraints_spec.rb
Rhian: The reason it was passing was that it was rendering the entire hash as a string. You're right - this is now an irrelevant test that should have been removed after the refactor, as we are already checking this scenario by setting the constraints on line 9 of this file. Removing.

app/helpers/planning_application_helper.rb
Rhian: Removed.

app/helpers/planning_application_helper.rb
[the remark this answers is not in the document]
Rhian: Updated.

app/views/planning_applications/edit_constraints_form.html.erb
Rhian: Done - it's much better.

app/views/shared/_supporting_information.html.erb
Rhian: Removed.

spec/system/planning_applications/constraints_spec.rb
Rhian: Removed.

james (approved):
🎉

db/schema.rb
```diff
@@ -131,6 +130,7 @@
     t.string "postcode"
     t.string "uprn"
     t.json "boundary_geojson"
+    t.text "constraints", default: [], null: false, array: true
```
melzreal, line 133: this null false: we always force constraints to exist? Won't this break existing applications if we are not adding defaults? 🤔

app/views/planning_applications/edit_constraints_form.html.erb
```diff
@@ -0,0 +1,59 @@
...
+
+    <div class="govuk-grid-row govuk-!-margin-top-6">
+      <div class="form-group">
+        <%= form.label :constraints, "Local constraints", class: 'govuk-heading-s govuk-!-margin-top-6' %>
+        <%= form.text_area :constraints, class: "govuk-textarea", rows: "3", "aria-describedby": "constraints-hint", multiple: true, value: "" %>
+      </div>
+    </div>
+
+
+    <div class="govuk-grid-row govuk-!-margin-top-6">
```
melzreal, line 52: just minor: a bit of unecessary white space (2 line returns) in 50,51 and 10,11

melzreal (approved):
Great stuff 🚀 :shipit:

PR #256, opened 2021-03-26
Map

james:
Add geojson map. See commit comments.

### DIFF
--- a/app/javascript/packs/govuk.js
+++ b/app/javascript/packs/govuk.js
@@ -1,3 +1,13 @@
-require("govuk-frontend/govuk/all").initAll()
+var govukall = require("govuk-frontend/govuk/all")
+
+govukall.Accordion.prototype.originalsetExpanded = govukall.Accordion.prototype.setExpanded
+govukall.Accordion.prototype.setExpanded = function (expanded, $section) {
+  this.originalsetExpanded(expanded, $section)
+  if (typeof accordion_expanded == 'function') { 
+    accordion_expanded(); 
+  }  
+}
+
+govukall.initAll()
 
 require.context('govuk-frontend/govuk/assets/images', true)
--- a/app/views/shared/_application_information.html.erb
+++ b/app/views/shared/_application_information.html.erb
@@ -66,6 +66,42 @@
       </table>
     </div>
   </div>
+
+  <div class="govuk-accordion__section">
+    <div class="govuk-accordion__section-header">
+      <h3 class="govuk-accordion__section-heading">
+        <!-- We have removed aria controls due to a problem with the default npm gov-uk functionality, which forces accordions to stay open by using the aria controls classes to setState in the browser, and force them open.
+        Our research suggests this makes the website harder to use as there is too much information on the screen overwhelming the user.
+        For now we are removing these until we create our own version of the gov-uk library that stops the behaviour of forcing accordions open. -->
+        <button type="button" id="accordion-default-heading-2" aria-controls="" class="govuk-accordion__section-button" aria-expanded="false">
+          Site Map
+        </button>
+          <span class="govuk-accordion__icon" aria-hidden="true"></span>
+      </h3>
+    </div>
+    <div class="govuk-accordion__section-content" aria-labelledby="accordion-default-heading-2">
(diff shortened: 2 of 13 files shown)


james:
[View site plan on a map](https://trello.com/c/fOkaha9Z/223-view-site-plan-on-a-map)

app/views/shared/_application_information.html.erb
```diff
@@ -66,6 +66,42 @@
...
+        <button type="button" id="accordion-default-heading-2" aria-controls="" class="govuk-accordion__section-button" aria-expanded="false">
+          Site Map
+        </button>
+          <span class="govuk-accordion__icon" aria-hidden="true"></span>
+      </h3>
+    </div>
+    <div class="govuk-accordion__section-content" aria-labelledby="accordion-default-heading-2">
+      <% if @planning_application.boundary_geojson.present? %>
+        <div id='map' style="width: 100%; height: 200px"></div>
+        <script>
```
Rhian, line 85: I've got 2 queries/comments about this code:
- would it be better to abstract the script into a separate JavaScript file, or do you think it is more readable/transparent to include the code here so other developers can see what's going on?
- wondering if we should at some point be checking/sanitising the json, as calling `html_safe` will just make this blow up if the input is not what's expected? On the other hand, I guess we are taking the data from trusted systems rather than relying on user-generated input, so maybe it's overkill

Rhian (approved):
Approved, but with one question/comment.

app/views/shared/_application_information.html.erb
james: 1) I actually didn't think about doing this, which I should have. But then again: I quite like it being obvious what is going on, and also it has to embed the geojson within it. I'm inclined to leave it embedded, but what do you think?

2) This is why I am running it through the JSON parser and back out again pretty unnecessarily here, to ensure that it is JSON that we've been passed before rendering it to the page.

app/javascript/packs/govuk.js
```diff
@@ -1,3 +1,13 @@
-require("govuk-frontend/govuk/all").initAll()
+var govukall = require("govuk-frontend/govuk/all")
```
melzreal, line 1: we could change this to a const instead of var assignment, for safety / JS best practice

app/views/shared/_application_information.html.erb
```diff
@@ -66,6 +66,42 @@
...
+            maxZoom: 22,
+            maxNativeZoom: 19
+          }).addTo(map);
+          var boundary = L.geoJSON(<%= JSON.parse(@planning_application.boundary_geojson).to_json.html_safe %>, {"color": "#ff0000", "weight": 2, "opacity": 0.65}).addTo(map);
+          map.fitBounds(boundary.getBounds());
+
+          function accordion_expanded() {
+            map.invalidateSize().fitBounds(boundary.getBounds());
+          }
+        </script>
```
melzreal, line 98: curious about us using JS within the erb as it breaks convention with the rest of the app - is this what was necessary to override the native JS library? otherwise we should keep it separate

melzreal (changes requested):
Unless the JS inline styles are absolutely necessary to make the JS override work, we should keep them separate in their own JS files or its very painful to maintain in the future to be looking inside ERB views for JS snippets

app/views/shared/_application_information.html.erb
Rhian: Re. 2, I just meant in case the JSON we have been passed not being correct

james: Ah! Yes, that is something we should look at on the API side, for this but also constraints etc

app/controllers/planning_applications_controller.rb
```diff
@@ -17,9 +17,11 @@ class PlanningApplicationsController < AuthenticationController
                                                     validate_documents
                                                     cancel_confirmation
                                                     cancel
-                                                    decision_notice]
+                                                    decision_notice
+                                                    map]
 
   before_action :ensure_user_is_reviewer, only: %i[review review_form]
+  protect_from_forgery except: :map
```
melzreal, line 24: why can't we protect the map from forgery?

melzreal (approved):
👍

PR #260, opened 2021-03-29
Improvement/layout edits

Rhian:
### Description of change

These are primarily text changes based on user research, with some small alignment and spacing amendments.

### Story Link

https://trello.com/c/VUOL0LKu/254-style-changes
https://trello.com/c/IxRh9Z5K/243-design-changes-layout-changes-to-documents
https://trello.com/c/CFhRE1Gl/260-assess-the-proposal-text-change-only
https://trello.com/c/mXJshTED/258-validate-application-step
https://trello.com/c/JPbcXcX1/259-changes-to-manage-documents-page-text-changes-only


app/views/documents/form_partials/_upload.html.erb
```diff
@@ -1,5 +1,9 @@
 <div class="govuk-form-group <%= form.object.errors[:file].any? ? 'govuk-form-group--error' : '' %>">
-  <%= form.label :file, "Upload a file", class: "govuk-label" %>
+  <% if current_page?(new_planning_application_document_path(@planning_application)) %>
```
james, line 2: Small thing, but I wonder if `@document.new_record?` rather than checking the request path is a more reliable and direct way of doing this?

app/views/documents/index.html.erb
```diff
@@ -61,60 +61,101 @@
...
+              <% else %>
+                <em>No reason provided</em>
+              <% end %>
+            </p>
+          </div>
+        </li>
       <% end %>
-    </table>
+      </ul>
+    <% end %>
```
james, line 159: I'm finding this quite hard to read! There are two `if/else` statements that are checking the same thing? It's tabular data that's not in a table?

app/views/planning_applications/recommendation_form.html.erb
```diff
@@ -21,38 +23,43 @@
...
+        <p class="govuk-hint">
           This information <strong>will</strong> appear on the decision notice.
         </p>
-        <%= form.text_area :public_comment, class: "govuk-textarea govuk-!-margin-bottom-9", rows: "3" %>
+        <%= form.text_area :public_comment, class: "govuk-textarea", rows: "3" %>
       </div>
     </div>
     <div class="govuk-form-group">
-      <p class="govuk-body">
+      <span class="govuk-body">
```
james, line 53: What's the reason for making these spans? `p` feels a bit more semantic?

james (approved):
No big blockers from me. Just a few questions.

melzreal (approved):
LGTM 🕺 💃

app/views/documents/form_partials/_upload.html.erb
Rhian: Thanks, yes - have implemented this. It's better.

app/views/documents/index.html.erb
Rhian: I've rationalised this so we don't have the same conditional. Essentially, we needed the whole page to align and look like current documents and archived documents are structured in the same way, so we either needed both sections to be a table or both to be a grid to ensure the lines and headings lined up, especially after we added in the thumbnails in the archive section, so I've left it as a grid.

app/views/planning_applications/recommendation_form.html.erb
Rhian: It was more to do with the spacing - Ali needed some elements to look like they were sitting with other elements rather than being spaced separately into paragraphs. Using a span meant that we didn't have to use a <p> tag with a whole lot of messy custom minus padding.

app/views/documents/index.html.erb
james: I'm splitting hairs a bit here, but would a table with custom column widths as described in https://design-system.service.gov.uk/components/table/ be an easier way of achieving this?

PR #263, opened 2021-03-31
Update packagebuilder for production

Rhian:
This should fix a bug where the build is skipping on Production because it is expecting the GITHUB_REF environment variable to be set to master


melzreal (approved):
lgtm 👍

PR #266, opened 2021-04-16
Add ability to mark documents as valid/invalid 

melzreal:
### Description of change

- Add validation to ensure comment is filled when the option for invalid is selected
- Make default value of validated be set to true for existing documents
- Add audit for when document is invalidated
- Add audit for when document is changed from invalid to valid
- Add notification in documents index noting how many documents are invalid, and showing invalid comment in bold

### Story Link

https://trello.com/c/A4tti71C/280-flag-when-a-document-is-valid-or-invalid

image
image
image
image


app/controllers/documents_controller.rb
```diff
@@ -9,6 +9,7 @@ class DocumentsController < AuthenticationController
   before_action :ensure_document_edits_unlocked, only: %i[new edit update archive]
 
   def index
+    @invalid_documents = @planning_application.documents.where(validated: false).where(archived_at: nil)
```
Rhian, line 12: Possibly move this to a scope in the model?

app/controllers/documents_controller.rb
```diff
@@ -19,6 +20,14 @@ def edit
   def update
     @document = @planning_application.documents.find(params[:id])
     if @document.update(document_params)
+      if @document.saved_change_to_attribute?(:validated)
+        # we want to track cases where the document was changed from invalid to valid, hence this statement
+        if @document.saved_changes[:validated][1] == true && @document.saved_changes[:validated][0] == false
```
Rhian, line 25: Could these be helper methods that simply return true or false?

app/views/documents/form_partials/_validate_document.html.erb
```diff
@@ -0,0 +1,16 @@
...
+    <% if form.object.errors.any? %>
+      <% form.object.errors.each do |error| %>
+        <span class="govuk-error-message">
+        <span class="govuk-visually-hidden">Error:</span><%= error.type %></span>
+      <% end %>
+    <% end %>
+    <%= form.govuk_radio_buttons_fieldset(:validated, legend: { size: 's', text: 'Is the document valid?' }) do %>
+      <%= form.govuk_radio_button :validated, 'true', label: { text: 'Yes'} %>
+      <%= form.govuk_radio_button :validated, 'false', label: { text: 'No' } do  %>
+        <%= form.govuk_text_area :invalidated_document_reason, label: { text: '' }, hint: { text: 'Describe in full why the document is invalid. This will be sent to the applicant.'}, rows: 3 %>
```
Rhian, line 12: Andrew reminded me last week that it's better to always have text for labels even if you have to attach a class to visually hide the text if you don't want it to display. This would also help with the Capybara test, as you are not then having to use a selector to identify the input field

app/controllers/documents_controller.rb
melzreal: I have always seen .where clauses used in controller methods so thought it was best practice - we have another one for ex in published? on line 57. What would be the logic for placing it in the model ( also considering this one is only needed for the index page) ?

spec/system/documents/edit_documents_spec.rb
```diff
@@ -52,6 +52,82 @@
...
+    end
+
+    it "requires a comment if the document is marked as not valid" do
+      click_link "Edit"
+
+      within("#validate-document") do
+        choose "No"
+      end
+
+      element = find(:css, "#document-invalidated-document-reason-field")
```
Rhian, line 78: I left a comment above on the form - attaching label text, even if the css hides it, is better for accessibility and will also help Capybara find the input box by text rather than css

spec/system/documents/index_spec.rb
```diff
@@ -62,4 +62,24 @@
       expect(page).to have_content("You need to sign in or sign up before continuing.")
     end
   end
+
+  context "handling invalid documents" do
+    let!(:invalid_document) do
+      create :document, :with_file,
+             planning_application: planning_application,
+             validated: false, invalidated_document_reason: "Missing a tikki bar"
```
Rhian, line 70: 😂

app/controllers/documents_controller.rb
melzreal: I am a little afraid the code can get hard to read since we are using an "audit" helper method already here. on line 23 it is just a true/false that is being returned. Then on 25 we could abstract this but then we'd need to be going up and down to see what is being audited, and is it worth abstracting given its only used once? 

Not sure that would make this more unreadable? 🤔

james, line 25: Oo, I haven't really explored ActiveRecord::AttributeMethods::Dirty much before, but I think this works well. I find this syntax a little hard to read though, not knowing what is [1] and what is [2].

I noticed that [`saved_change_to_attribute?`](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Dirty.html#method-i-saved_change_to_attribute-3F) accepts and `from` and `to` option, so would this syntax work?

`if @document.saved_change_to_attribute?(:validated, from: true, to: false)`

app/controllers/documents_controller.rb
james, line 12: I'd prefer to use named scopes for these where filters, in case we want to change how they work in the future. We already have a named scope for `active` so we can use that, and we can add a new one for `invalid`?

app/models/audit.rb
```diff
@@ -17,5 +17,7 @@ class Audit < ApplicationRecord
     uploaded: "uploaded",
     started: "started",
     withdrawn: "withdrawn",
+    document_invalidated: "document_invalid",
+    document_changed_to_validated: "document_changed_to_valid",
```
james, line 21: everything else in here the keys and values are the same, but they're different here, and I'm not sure why?

app/views/documents/form_partials/_validate_document.html.erb
```diff
@@ -0,0 +1,16 @@
+<%= form_with model: [@planning_application, @document], local: true, class: "document", builder: GOVUKDesignSystemFormBuilder::FormBuilder do |form| %>
```
james, line 1: This is creating a form nested within another form, which is incorrect. You should remove this `form_with`

app/models/audit.rb
melzreal: it was the artist writer in me 😆  I'll change them to match

app/controllers/documents_controller.rb
melzreal: ahhh named scopes is what @Rhian was referring to in her comment - I didn't know about these. Ok will change them.

app/views/documents/form_partials/_validate_document.html.erb
james, line 12: I realise it is easier to label the form input as hint text as that's closer to the style, but it's important that the label is in the label tag, for accessibility, useability and it will mean you don't have to do the `find(:css, "#document-invalidated-document-reason-field")` in your tests.

I'd put the hint text into the label text, and figure out a way to override the style if the design insists on a label that doesn't look like a label

app/controllers/documents_controller.rb
melzreal: it does! Should have read the documentation again, thanks for this

app/views/documents/form_partials/_validate_document.html.erb
melzreal: changed, thank you!

db/migrate/20210416094652_add_validated_to_documents.rb
```diff
@@ -0,0 +1,6 @@
+class AddValidToDocuments < ActiveRecord::Migration[6.1]
+  def change
+    add_column :documents, :validated, :boolean, required: true, default: true
```
james, line 3: Have we confirmed that this defaults to true? I realise we didn't capture this in the acceptance criteria, but I think it's an important product question.

Option 1) as has currently been built where new documents are valid by default
Option 2) documents are invalid by default (I suspect this would be weird, but wanted to capture it!)
Option 3) documents by default have a null validation status, where they are neither invalid or valid until they are marked as so.

I'll ping this comment to Ali.

app/views/documents/form_partials/_validate_document.html.erb
melzreal: oh, I did the other suggestion that rhian mentioned - have both and just css the text to be transparent

db/migrate/20210416094652_add_validated_to_documents.rb
AliBlake: Option 3. Documents should be neither valid or invalid until the planning officer has marked it as so!

app/views/documents/form_partials/_validate_document.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="govuk-form-group" id="validate-document">
+  <% if form.object.errors.any? %>
+    <% form.object.errors.each do |error| %>
```
james, line 3: Missed this first run-through, but just noticed this prints all errors on any field on the @document object, whereas I think you only want to print the errors related to the validated and invalidated_document_reason fields?

An example of the potential bug here: If there is an error on the `numbers` field, it will create an error message on line 4 of _edit_and_upload.html.erb *and* here.

james (approved):
🎉

PR #269, opened 2021-04-22
Add description change requests

melzreal:
### Description of change

Allow for the creation of description change requests
Only allow for invalid applications types to have description change requests created on them
Add a validation to ensure the new application description is present
Display change requests in the application validation step
Redirect to the application validation step when an application is marked as invalid,
so the user can create change requests from there
Add helpers to change tag colours depending on set conditions of time and status (state)
Ensure that an application can only be marked as valid if no open change requests exist
Add show pages for each change request and link the show pages to each description
Add business time gem and calculate response due
Add a scope to order change description requests per date, latest request displayed at the top

### Story Link

https://trello.com/c/tJ4uzfrb/290-as-an-officer-i-can-create-a-request-for-change-to-a-description

### Decisions

- Have a separate db table, model and controller for each type of description change request so we can keep things RESTFUL and clean
- We will consider how to group decision change requests on the next stage of development, when we will start supporting multiple types of CR

### DIFF
--- a/app/helpers/description_change_request_helper.rb
+++ b/app/helpers/description_change_request_helper.rb
@@ -1,0 +1,21 @@
+module DescriptionChangeRequestHelper
+  def display_request_date_state(description_change_request)
+    if description_change_request.state == "closed"
+      "grey"
+    elsif description_change_request.days_until_response_due.positive?
+      "green"
+    else
+      "red"
+    end
+  end
+
+  def display_request_status(description_change_request)
+    if description_change_request.state == "open"
+      "grey"
+    elsif description_change_request.approved == false
+      "red"
+    else
+      "green"
+    end
+  end
+end
--- a/app/views/planning_applications/validate_documents_form.html.erb
+++ b/app/views/planning_applications/validate_documents_form.html.erb
@@ -1,32 +1,82 @@
 <% content_for :title, "Validate documents" %>
-
 <%= render "planning_applications/assessment_dashboard" do %>
 
-  <h2 class="govuk-heading-m govuk-!-padding-top-3">Validate application</h2>
+<h2 class="govuk-heading-m govuk-!-padding-top-3">Validate application</h2>
 
-  <p class="govuk-body"><%= link_to "Check documents", planning_application_documents_path(@planning_application) %></p>
+<p class="govuk-body">
+  <%= link_to "Check documents", planning_application_documents_path(@planning_application) %>
+</p>
+<table class="govuk-table">
+  <thead class="govuk-table__head">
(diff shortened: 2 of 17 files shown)


app/helpers/description_change_request_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+      "green"
+    else
+      "red"
+    end
+  end
+
+  def display_request_status(description_change_request)
+    if description_change_request.state == "open"
+      "grey"
+    elsif description_change_request.approved == false
```
Rhian, line 15: I was going to suggest using `!description_change_request.approved?` but I don't know if using an `if` negatively is better than what you have?

app/views/planning_applications/validate_documents_form.html.erb
```diff
@@ -1,32 +1,87 @@
...
+        <th scope="col" class="govuk-table__header govuk-!-width-one-quarter">
+          Time remaining
+        </th>
+        <th scope="col" class="govuk-table__header govuk-!-width-one-quarter">
+          Status
+        </th>
+      </tr>
+    </thead>
+    <tbody class="govuk-table__body">
+      <% @planning_application.description_change_requests.each do |description_change_request| %>
```
Rhian, line 26: More of a product question, but do we want to order these with Open ones at the top, regardless of date? In that case, we'd need a scope

spec/system/planning_applications/validating_spec.rb
```diff
@@ -105,6 +109,22 @@
...
     include_examples "validate and invalidate"
+
+    it "shows error if trying to mark as valid when open change request exists on planning application" do
+      create :description_change_request, planning_application: planning_application, state: "open"
+      click_link planning_application.reference
+      click_link "Validate application"
+
+      choose "Yes"
+
+      fill_in "Day", with: "03"
```
Rhian, line 120: Theoretically, you shouldn't need to fill this in as it should be pre-populated

app/helpers/description_change_request_helper.rb
melzreal: because the approved can be nil if it is in pending, in which case it is still !approved, so that's why this uses the comparison to false instead

app/views/planning_applications/validate_documents_form.html.erb
melzreal: really good point, that actually reminds me they should be grouped according to the date they were sent in per one of the designs! I'll create a scope for that and double check with ali if she wants them ordered by open or by date

Rhian (approved):
LGTM!

PR #270, opened 2021-04-26
Get, post and update paths for change requests API

melzreal:
### Description of change

- Separate change requests and description  change requests into their own controllers
- Add separate change requests yaml file and README instructions on how to have separate yaml files
- Updates state of change request to close upon successful PUT
- Add swagger index spec to ensure all the requests are being listed on swagger page
- Add validation to prevent rejecting a description change request without a comment

### Story Link

https://trello.com/c/g4SHx8aA/292-as-an-applicant-i-can-accept-or-reject-a-request-to-change-a-description

### Decisions

- Separated yaml files to make the code more maintainable
- Created separate controller for change requests which inherits from the planning applications controller

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,3 +96,29 @@ Once you have the application running, you can submit planning application throu
 
 [1]: https://www.docker.com/products/docker-desktop
 [2]: http://localhost:3000/
+
+## Working with api documentation: aggregate swagger files
+
+We need a single openapi file to exist, but to keep the code easier to maintain we have multiple files that are then compiled into this single file:
+
+```public/api-docs/v1/_build/swagger_doc.yaml```.
+
+So to create a new api endpoint, create your yaml doc inside public/api-docs/v1 and reference it in
+
+``` public/api-docs/v1/swagger_doc.yaml ```
+
+like so:
+
+```
+  $ref: "./your_new_file_name.yaml"
+```
+
+Make changes to your new file, and when you're happy aggregate them into our single file by installing this package in your machine:
+
+``` npm install -g swagger-cli ```
+
+and running:
+
+```
+swagger-cli bundle public/api-docs/v1/swagger_doc.yaml --outfile public/api-docs/v1/_build/swagger_doc.yaml --type yaml
+```
--- a/app/controllers/api/v1/change_requests_controller.rb
+++ b/app/controllers/api/v1/change_requests_controller.rb
@@ -1,0 +1,16 @@
+class Api::V1::ChangeRequestsController < Api::V1::ApplicationController
+  before_action :check_token_and_set_application, only: %i[index], if: :json_request?
+
+  def index; end
+
+private
+
+  def check_token_and_set_application
+    @planning_application = current_local_authority.planning_applications.where(id: params[:planning_application_id]).first
+    if params[:change_access_id] != @planning_application.change_access_id
+      render json: {}, status: 401
+    else
+      @planning_application
+    end
+  end
+end
--- a/public/api-docs/v1/change_requests.yaml
+++ b/public/api-docs/v1/change_requests.yaml
@@ -1,0 +1,99 @@
+"/api/v1/planning_applications/{planning_application_id}/change_requests":
+  get:
+    summary: Retrieves all determined planning applications
+    security:
+      - bearerAuth: []
+    parameters:
+      - in: path
+        name: planning_application_id
+        required: true
+        schema:
+          type: integer
+          minimum: 1
+        description: The planning application ID
+      - in: query
+        name: change_access_id
+        schema:
+          type: string
+        required: true
+        description: A unique access code for this planning application
+
+    responses:
+      "200":
+        description: All change requests
+        content:
+          application/json:
+            examples:
+              Full:
+                value:
+                  data:
+                    - id: 1
+                      type: description_change_request
+                      proposed_description: A proposed new description
+                      state: open
+                      response_due: "2020-05-14T05:18:17.540Z"
+                      approved: true
+                      rejection_reason: nil
+
+
+"/api/v1/planning_applications/{planning_application_id}/description_change_requests/{description_change_request_id}":
+  patch:
+    summary: Saves the response for a description change request
+    security:
+    - bearerAuth: []
+    parameters:
+      - in: path
+        name: planning_application_id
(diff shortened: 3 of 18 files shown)


spec/requests/api/description_change_request_put_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+          headers: { "CONTENT-TYPE": "application/json", "Authorization": "Bearer #{api_user.token}" }
+
+    expect(response).to be_successful
+
+    description_change_request.reload
+    expect(description_change_request.state).to eq("closed")
+    expect(description_change_request.approved).to eq(true)
+  end
+
+  it "successfully accepts a rejection" do
```
james, line 32: We want to save the rejection reason if not approved.

app/controllers/api/v1/change_requests_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    if @description_change_request.update(site_params)
+      @description_change_request.update!(state: "closed")
+      render json: { "message": "Change request updated" }, status: :ok
+    end
+  end
+
+private
+
+  def description_change_params
```
james, line 22: This method is not used anywhere

app/controllers/api/v1/change_requests_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    permitted_chr_keys = %i[data
+                            change_access_id
+                            planning_application_id
+                            id
+                            change_request
+                            format]
+    params.permit permitted_chr_keys
+  end
+
+  def site_params
```
james, line 32: I don't think you mean "site_params"

app/views/api/v1/change_requests/show.json.jbuilder
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+json.extract! @description_change_request,
+              :id,
+              :planning_application_id,
+              :state,
+              :response_due,
+              :proposed_description,
+              :rejection_reason,
+              :approved
```
james, line 10: On further thought, I really don't think we need this action to exist, at least not yet. If we do want it to exist, then we should be writing a `spec/requests/api/change_request_show_spec.rb`, but I think this is us writing and maintaining an active API endpoint that we don't have a need for yet.

public/api-docs/v1/change_requests.yaml
```diff
@@ -0,0 +1,144 @@
...
+                  data:
+                    - id: 1
+                      type: description_change_request
+                      proposed_description: A proposed new description
+                      state: open
+                      response_due: "2020-05-14T05:18:17.540Z"
+                      approved: true
+                      rejection_reason: nil
+
+? "/api/v1/planning_applications/{planning_application_id}/change_requests/{change_request_id}"
```
james, line 38: This should be at
`/api/v1/planning_applications/{planning_application_id}/description_change_requests`
When we add different types of change request, then we need to be able to know which table we're looking up.

james: `change_request_id` doesn't need to be added to the path either, as it is already specified in the query params below

james (changes requested):

README.md
```diff
@@ -96,3 +96,29 @@ Once you have the application running, you can submit planning application throu
...
+```
+
+Make changes to your new file, and when you're happy aggregate them into our single file by installing this package in your machine:
+
+``` npm install -g swagger-cli ```
+
+and running:
+
+```
+swagger-cli bundle public/api-docs/v1/swagger_doc.yaml --outfile public/api-docs/v1/_build/swagger_doc.yaml --type yaml
```
Rhian, line 123: 🤩

james (changes requested):
Much improved! Few more things to update, and I didn't get a response about either removing the change request show API action or adding some tests for it.

public/api-docs/v1/change_requests.yaml
```diff
@@ -0,0 +1,103 @@
...
+                    - id: 1
+                      type: description_change_request
+                      proposed_description: A proposed new description
+                      state: open
+                      response_due: "2020-05-14T05:18:17.540Z"
+                      approved: true
+                      rejection_reason: nil
+
+
+"/api/v1/planning_applications/{planning_application_id}/description_change_requests":
```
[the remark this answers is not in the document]
melzreal: I removed the ID because of this comment you made about it already being part of the params? https://github.com/unboxed/bops/pull/270#discussion_r620313488

public/api-docs/v1/change_requests.yaml
melzreal: > Much improved! Few more things to update, and I didn't get a response about either removing the change request show API action or adding some tests for it.

I just removed it? I forgot to take it out of the routes but I took it out of the controller and deleted all tests as well as the show jsonb file

public/api-docs/v1/change_requests.yaml
```diff
@@ -0,0 +1,103 @@
...
+              Full:
+                value:
+                  data:
+                    - id: 1
+                      type: description_change_request
+                      proposed_description: A proposed new description
+                      state: closed
+                      response_due: "2020-05-14T05:18:17.540Z"
+                      approved: false
+                      rejection_reason: This is poorly written
```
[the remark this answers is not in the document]
melzreal: if we don't add a responses then document.valid? become false - the api speficitation requires it. I'll change the response instead.

Also note we have the wrong description summary in line 85 which should read "retrieves a single application", should I also amend that?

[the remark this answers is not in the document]
james: ah! didn't realise we *had* to. OK, let's just update this to be accurate then?

james (approved):
I've got two very small change requests in here, but once you've done them, it's good to go! :-)

public/api-docs/v1/change_requests.yaml
```diff
@@ -0,0 +1,99 @@
...
+                  approved: false
+                  rejection_reason: "Refusal reason"
+
+    responses:
+      "200":
+        description: Update change request
+        content:
+          application/json:
+            examples:
+              Full:
```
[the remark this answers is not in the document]
melzreal: sounds good, will change it

PR #271, opened 2021-04-29
Add change request email methods

Rhian:
### Description of change

This creates the secure url parameterised with the change_id token and sends out the email to the applicant. Note that it forms only a small part of the story linked below.

### Story Link

https://trello.com/c/g4SHx8aA/292-as-an-applicant-i-can-accept-or-reject-a-request-to-change-a-description-part-2

### DIFF
--- a/app/controllers/description_change_requests_controller.rb
+++ b/app/controllers/description_change_requests_controller.rb
@@ -10,6 +10,7 @@ class DescriptionChangeRequestsController < ApplicationController
     @description_change_request.user = current_user
 
     if @description_change_request.save
+      change_request_mail
       flash[:notice] = "Change request for description successfully sent."
       redirect_to validate_documents_form_planning_application_path(@planning_application)
     else
@@ -29,5 +30,13 @@ private
 
   def set_planning_application
     @planning_application = PlanningApplication.find(params[:planning_application_id])
+  end
+
+  def change_request_mail
+    PlanningApplicationMailer.change_request_mail(
+      @planning_application,
+      request.host,
+      @description_change_request,
+    ).deliver_now
   end
 end
--- a/app/controllers/description_change_requests_controller.rb
+++ b/app/controllers/description_change_requests_controller.rb
@@ -8,9 +8,10 @@ class DescriptionChangeRequestsController < ApplicationController
   def create
     @description_change_request = @planning_application.description_change_requests.new(description_change_request_params)
     @description_change_request.user = current_user
+    @current_local_authority = current_local_authority
 
     if @description_change_request.save
-      change_request_mail
+      send_change_request_email
       flash[:notice] = "Change request for description successfully sent."
       redirect_to validate_documents_form_planning_application_path(@planning_application)
     else
@@ -32,10 +33,9 @@ private
     @planning_application = PlanningApplication.find(params[:planning_application_id])
   end
 
-  def change_request_mail
+  def send_change_request_email
     PlanningApplicationMailer.change_request_mail(
       @planning_application,
-      request.host,
       @description_change_request,
     ).deliver_now
   end
--- a/app/mailers/planning_application_mailer.rb
+++ b/app/mailers/planning_application_mailer.rb
@@ -26,8 +26,7 @@ class PlanningApplicationMailer < Mail::Notify::Mailer
     )
   end
 
-  def change_request_mail(planning_application, host, change_request)
-    @host = host
+  def change_request_mail(planning_application, change_request)
     @planning_application = planning_application
     @change_request = change_request
 
--- a/app/mailers/planning_application_mailer.rb
+++ b/app/mailers/planning_application_mailer.rb
@@ -25,4 +25,16 @@ class PlanningApplicationMailer < Mail::Notify::Mailer
       to: @planning_application.applicant_email,
     )
   end
+
+  def change_request_mail(planning_application, host, change_request)
+    @host = host
+    @planning_application = planning_application
+    @change_request = change_request
+
+    view_mail(
+      NOTIFY_TEMPLATE_ID,
+      subject: "Change requested to your planning application",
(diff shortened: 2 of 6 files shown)


app/models/planning_application.rb
```diff
@@ -231,6 +235,16 @@ def documents_validated_at_date
...
 
+  def change_url
+    if ENV["RAILS_ENV"] == "production"
+      "bops-applicants.services"
+    elsif ENV["RAILS_ENV"] == "staging"
+      "bops-applicants-staging.services"
+    else
+      ""
+    end
+  end
```
james, line 246: Instead of this if, could we have an ENV['applicants_app_host']?

spec/mailer/planning_application_mailer_spec.rb
```diff
@@ -76,4 +78,24 @@
...
+      expect(change_request_mail.subject).to eq("Change requested to your planning application")
+      expect(change_request_mail.to).to eq([planning_application.applicant_email])
+    end
+
+    it "renders the body" do
+      expect(change_request_mail.body.encoded).to include("Application received: #{planning_application.created_at.strftime('%e %B %Y')}")
+      expect(change_request_mail.body.encoded).to include(change_request.user.name)
+      expect(change_request_mail.body.encoded).to include(change_request.response_due.strftime("%e %B %Y"))
+      expect(change_request_mail.body.encoded).to include(planning_application.change_access_id)
+      expect(change_request_mail.body.encoded).to include("https://default.example.com/planning_applications/#{planning_application.id}/change_requests?change_access_id=#{planning_application.change_access_id}")
```
james, line 95: This doesn't appear to be using the change_url?

spec/mailer/planning_application_mailer_spec.rb
```diff
@@ -76,4 +81,26 @@
...
+    end
+
+    it "renders the body" do
+      expect(change_request_mail.body.encoded).to include("Application received: #{planning_application.created_at.strftime('%e %B %Y')}")
+      expect(change_request_mail.body.encoded).to include(change_request.user.name)
+      expect(change_request_mail.body.encoded).to include(change_request.response_due.strftime("%e %B %Y"))
+      expect(change_request_mail.body.encoded).to include(planning_application.change_access_id)
+      expect(change_request_mail.body.encoded).to include("http://cookies.localhost/change_requests?planning_application_id=#{planning_application.id}&change_access_id=#{planning_application.change_access_id}")
+      expect(change_request_mail.body.encoded).to include("Mr. Biscuit")
+      expect(change_request_mail.body.encoded).to include("Cookie authority")
```
melzreal, line 102: 😆 🚀

melzreal (approved):
LGTM 🛴

app/controllers/description_change_requests_controller.rb
```diff
@@ -8,8 +8,10 @@ def new
   def create
     @description_change_request = @planning_application.description_change_requests.new(description_change_request_params)
     @description_change_request.user = current_user
+    @current_local_authority = current_local_authority
```
james, line 11: I don't think we need to set this as a variable. Line 41 can just call the method directly?

app/controllers/description_change_requests_controller.rb
```diff
@@ -8,8 +8,10 @@ def new
...
 
     if @description_change_request.save
+      change_request_mail
```
james, line 14: Could we call this method `send_change_request_email`?

app/mailers/planning_application_mailer.rb
```diff
@@ -25,4 +25,17 @@ def validation_notice_mail(planning_application, host)
       to: @planning_application.applicant_email,
     )
   end
+
+  def change_request_mail(planning_application, host, change_request, local_authority)
+    @host = host
```
james, line 30: I don't think @host gets used anywhere, so we can remove it?

app/models/planning_application.rb
```diff
@@ -215,6 +215,14 @@ def full_address
     "#{address_1}, #{town}, #{postcode}"
   end
 
+  def secure_change_url(local_authority, application_id, secure_token)
```
james, line 218: I think this should be a helper method, rather than a method on the model?

app/models/planning_application.rb
```diff
@@ -215,6 +215,14 @@ def full_address
...
+    if ENV["RAILS_ENV"] == "production" || ENV["RAILS_ENV"] == "preview"
+      "https://#{local_authority}.#{ENV['applicants_app_host']}/change_requests?planning_application_id=#{application_id}&change_access_id=#{secure_token}"
```
james, line 220: could this be `local_authority_name`? I think of it as the local_authority model object otherwise.

app/models/planning_application.rb
```diff
@@ -215,6 +215,14 @@ def full_address
     "#{address_1}, #{town}, #{postcode}"
   end
 
+  def secure_change_url(local_authority, application_id, secure_token)
+    if ENV["RAILS_ENV"] == "production" || ENV["RAILS_ENV"] == "preview"
```
james, line 219: We don't need this `if` statement any more, as both options are identical.

app/models/planning_application.rb
james: I've changed my mind about this actually. In the future we might want this to be set via the API, so it being on the model makes sense.

app/models/planning_application.rb
james: In fact, seeing as we're inside the planning_application model, we can just call `local_authority.subdomain`, and we don't need to pass the name through at all?

app/models/planning_application.rb
Rhian: I added it because localhost is http, not https

james: ah! I missed that.

PR #274, opened 2021-05-04
Freeze time in spec testing business days calculation

james:
Because the number of business days over a period of days changes day to day, this spec was failing some days, and not others.

It is important to therefore freeze the spec to a specific date to get consistent results.


melzreal (approved):
🚀

PR #277, opened 2021-05-10
Contact email is visible if it exists

melzreal:
### Description of change

Until now, we only displayed contact information if all fields were present for agent or applicant. However, now we started making email a required field - and it is the only required field - so it makes sense to display that email in contact information if it exists, given its the only field we are guaranteed to have.

### Story Link

https://trello.com/c/Z4DqLRhJ/312-chore-ensure-contact-email-is-displayed-if-present

image


james:
If `@planning_application.applicant?` was true, then the email address would be displayed twice?

PR #279, opened 2021-05-12
Temporarily hardcode app_host for applicants

Rhian:
### Description of change

This can be reverted later - it's a patch for an issue with the secure url


melzreal (approved):
🚀

PR #281, opened 2021-05-12
Add previous description to planning applications

melzreal:
### Description of change

Display both the new and previous descriptions on the change request show page
image


Rhian (changes requested):
This looks good, but I think we should add a quick line to the spec to show that it's updating correctly

melzreal:
closing, will submit an alternative PR to track changes directly on the CR instead on the planning application - so we have the changes pertaining to each CR correctly displayed.

PR #282, opened 2021-05-12
Display previous description & http email endpoint 

melzreal:
### Description of change

Keep track of new and current descriptions within each description change request. 

- Change requests keep track of previous descriptions of the applications
- We have decided that bops-applicants will be served over http as it is only a proof of concept,
so the email sent to the applicant should contain the http app link instead of https

image

### DIFF
--- a/app/controllers/api/v1/description_change_requests_controller.rb
+++ b/app/controllers/api/v1/description_change_requests_controller.rb
@@ -8,6 +8,7 @@ class Api::V1::DescriptionChangeRequestsController < Api::V1::ApplicationControl
     if @description_change_request.update(description_change_params)
       @description_change_request.update!(state: "closed")
       @planning_application.update!(description: @description_change_request.proposed_description) if @description_change_request.approved?
+
       render json: { "message": "Change request updated" }, status: :ok
     else
       render json: { "message": "Unable to update request. Please ensure rejection_reason is present if approved is false." }, status: 400
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -217,9 +217,9 @@ class PlanningApplication < ApplicationRecord
 
   def secure_change_url(application_id, secure_token)
     if ENV["DOMAIN"] == "bops-staging.services"
-      "https://#{local_authority.subdomain}.bops-applicants-staging.services/change_requests?planning_application_id=#{application_id}&change_access_id=#{secure_token}"
+      "http://#{local_authority.subdomain}.bops-applicants-staging.services/change_requests?planning_application_id=#{application_id}&change_access_id=#{secure_token}"
     elsif ENV["DOMAIN"] == "bops.services"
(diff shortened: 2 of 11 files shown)


app/controllers/api/v1/description_change_requests_controller.rb
```diff
@@ -6,8 +6,14 @@ def update
     @description_change_request = @planning_application.description_change_requests.where(id: params[:id]).first
 
     if @description_change_request.update(description_change_params)
+      previous_description = @planning_application.description
       @description_change_request.update!(state: "closed")
-      @planning_application.update!(description: @description_change_request.proposed_description) if @description_change_request.approved?
+
+      if @description_change_request.approved?
+        @description_change_request.update!(previous_description: previous_description)
```
james, line 13: Rather than filling in `previous_decision` on approval, how about we fill it in on creation of the change request. eg:

```ruby
class DescriptionChangeRequest < ApplicationRecord
  before_create :set_previous_decision
  private
  def set_previous_decision
    self.previous_description = planning_application.previous_decription
  end
```

that way the value is never null, and we can just display that on the front end instead of putting logic there to decide which one we want?

app/models/planning_application.rb
```diff
@@ -217,9 +217,9 @@ def full_address
 
   def secure_change_url(application_id, secure_token)
     if ENV["DOMAIN"] == "bops-staging.services"
```
james, line 219: Am I missing something here? Why do we have this `if` statement?

Could this method just be the last line? `"http://#{local_authority.subdomain}.#{ENV['APPLICANTS_APP_HOST']}/change_requests?planning_application_id=#{application_id}&change_access_id=#{secure_token}"`

app/views/description_change_requests/show.html.erb
```diff
@@ -19,34 +16,38 @@
...
+          <p class="govuk-body"> <strong> Suggested description:</strong><br>
+            <%= @description_change_request.proposed_description  %>
+          </p>
+        <% else %>
+          <p class="govuk-body"> <strong> Previous description:</strong><br>
+            <%= @description_change_request.previous_description  %>
+          </p>
+          <p class="govuk-body"> <strong> Suggested description:</strong><br>
+            <%= @description_change_request.proposed_description  %>
+          </p>
```
james, line 40: The "Suggested description" is repeated here, it can be removed from the if/else statement.

app/models/planning_application.rb
melzreal: I didn't add the if statement, all I did to this was change everything to http - it was introduced by @Rhian on the email PR - Rhian do you recall the reason for it?

Rhian: It was a workaround yesterday because there was an issue with environment variables on Preview that I couldn't resolve quickly

james (approved):

PR #283, opened 2021-05-13
Update application type to reflect work status

Rhian:
### Description of change

This change adds a helper method to derive the application description from application type and work status. It's currently an if statement but can be changed to a case statement when we add further certificate types.

### Story Link

https://trello.com/c/LL56t8jY/282-reflect-work-status-in-planning-application-information

### DIFF
--- a/spec/system/planning_applications/requesting_changes_spec.rb
+++ b/spec/system/planning_applications/requesting_changes_spec.rb
@@ -81,7 +81,7 @@ RSpec.describe "Requesting changes to a planning application", type: :system do
     expect(page).to have_content("Request for approval of changes to description")
     expect(page).to have_content("Application number: #{planning_application.reference}")
     expect(page).to have_content("At: #{planning_application.full_address}")
-    expect(page).to have_content("Request sent:#{description_change_request.created_at.strftime('%e %B %Y')}")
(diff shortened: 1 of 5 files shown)


spec/system/planning_applications/requesting_changes_spec.rb
```diff
@@ -81,7 +81,7 @@
     expect(page).to have_content("Request for approval of changes to description")
     expect(page).to have_content("Application number: #{planning_application.reference}")
     expect(page).to have_content("At: #{planning_application.full_address}")
-    expect(page).to have_content("Request sent: #{description_change_request.created_at.strftime('%e %B %Y')}")
+    expect(page).to have_content("Request sent: #{description_change_request.created_at.strftime('%-e %B %Y')}")
```
james, line 84: I don't understand what this commit is doing?

Rhian: It's an unrelated fix, which is why it's a separate commit. It changes %e to %-e to remove the leading padding on this date. Otherwise, if it's an application that was created in the first 10 days of the month, the spec fails because it's not expecting an extra space here. It has only just started failing because the spec was written when the relevant dates were all double-digit

melzreal (approved):
lgtm 👍

PR #285, opened 2021-05-14
Add document change requests

melzreal:
### Description of change

Add document change requests

### Story Link

https://trello.com/c/DUvW42Ly/293-request-for-change-replacing-a-document

### Decisions

- Created a change requests controller with no db entry to redirect to the right action based on params
- Implemented all new design changes whilst conferring with Ali - no document show page for the planning officer is desired; the requests index page on the planning officer side is moved to the new view after clicking on "View or Show" requests and all the text showing in the change requests index has been verified by Ali too; 
- We have chosen to use multipart upload instead of base64 for file uploads using PATCH requests as it has less overhead regarding file encryption;


app/controllers/api/v1/change_requests_controller.rb
```diff
@@ -1,7 +1,11 @@
 class Api::V1::ChangeRequestsController < Api::V1::ApplicationController
   before_action :check_token_and_set_application, only: %i[index], if: :json_request?
 
-  def index; end
+  def index
+    @documents_for_change_request = @planning_application.document_change_requests.map do |doc|
```
james, line 5: @documents_for_change_request doesn't seem to be being used anywhere?

But also another way of doing this would be:

`@planning_application.document_change_requests.map(&:document)`

app/views/document_change_requests/new.html.erb
```diff
@@ -0,0 +1,60 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h2 class="govuk-heading-l govuk-!-padding-top-6">
+      Request replacement documents
+    </h2>
+    <p class="govuk-body govuk-!-padding-bottom-3">
+      The following documents have been marked as invalid.
+    </p>
+    <%= form_with model: [@planning_application, @document_change_request], local: true, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |form| %>
+      <% if @document_change_request.errors.any? %>
```
james, line 15: We'll never have errors on this object (500 instead), so we can remove this HTML

app/helpers/planning_application_helper.rb
```diff
@@ -125,4 +125,12 @@ def cancelled_at(planning_application)
...
     end
   end
+
+  def invalid_documents(planning_application)
+    planning_application.documents.active.invalidated
+  end
+
+  def change_requests(planning_application)
+    planning_application.description_change_requests.order(created_at: :desc) + planning_application.document_change_requests.order(created_at: :desc)
+  end
```
james, line 135: These two methods make more sense on the planning application model

app/controllers/api/v1/document_change_requests_controller.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class Api::V1::DocumentChangeRequestsController < Api::V1::ApplicationController
+  skip_before_action :verify_authenticity_token, only: :update
+  before_action :check_token_and_set_application, only: :update, if: :json_request?
+  before_action :check_file_params_are_present, only: :update, if: :json_request?
```
james, line 6: This actually asks questions for the previous work we've done too, but why do we have `if: :json_request?` on these? Surely we *always* want to run these security checks?

app/controllers/api/v1/document_change_requests_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def update
+    @document_change_request = @planning_application.document_change_requests.where(id: params[:id]).first
```
james, line 9: `@planning_application.document_change_requests.find(params[:id])` is more rails idiomatic

app/controllers/api/v1/document_change_requests_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    new_document = @planning_application.documents.create!(file: params[:new_file])
+    @document_change_request.update!(state: "closed", new_document: new_document)
+
```
james, line 12: we need to archive the old document.

app/controllers/api/v1/document_change_requests_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    else
+      render json: { "message": "Unable to update request" }, status: 400
+    end
+  end
+
+private
+
+  def new_filename(name)
+    name.split("/")[-1]
+  end
```
james, line 24: This method isn't used

app/controllers/api/v1/document_change_requests_controller.rb
melzreal: Good point - I had copied it over from another controller, will take it off

app/controllers/api/v1/document_change_requests_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def check_file_params_are_present
+    if params[:new_file].blank?
+      render json: { "message": "A file must be selected to proceed." }, status: 400
+    end
+  end
+
+  def check_token_and_set_application
+    @planning_application = current_local_authority.planning_applications.where(id: params[:planning_application_id]).first
```
james, line 33: again `@planning_application = current_local_authority.planning_applications.find(params[:planning_application_id])`

app/models/planning_application.rb
```diff
@@ -225,6 +226,18 @@ def secure_change_url(application_id, secure_token)
     end
   end
 
+  def invalid_documents_without_change_request
+    documents.invalidated.reject { |x| document_change_requests.where(old_document: x).any? }
```
james, line 230: we probably want to exclude archived documents from this, otherwise we might send a change request for an archived document? We could even use the `invalid_documents` method you have below.

app/views/change_requests/index.html.erb
```diff
@@ -0,0 +1,84 @@
...
+<% add_parent_breadcrumb_link "Application", validate_documents_form_planning_application_path(@planning_application) %>
+
+<div class="govuk-grid-column-full">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column govuk-!-padding-top-6">
+      <fieldset class="govuk-fieldset">
+        <h2 class="govuk-heading-l">
+          Requests for changes
+        </h2>
+        <% if @planning_application.change_requests %>
```
james, line 12: Should be `@planning_application.change_requests.present?`, as will return truthy with an empty array otherwise.

public/api-docs/v1/_build/swagger_doc.yaml
```diff
@@ -477,7 +477,7 @@ paths:
                         rejection_reason: nil
```
james, line 477: Into this example, could we include what an example document change request would look like? Would be very helpful to someone trying to build an application against our API!

spec/models/document_change_request_spec.rb
```diff
@@ -0,0 +1,5 @@
+require "rails_helper"
+
+RSpec.describe DocumentChangeRequest, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
james, line 5: we can delete this file

spec/requests/api/document_change_request_put_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    patch "/api/v1/planning_applications/#{planning_application.id}/document_change_requests/#{document_change_request.id}?change_access_id=#{planning_application.change_access_id}",
+          params: { new_file: fixture_file_upload("../images/proposed-floorplan.png") },
+          headers: { "Authorization": "Bearer #{api_user.token}" }
+
+    expect(response).to be_successful
+
+    document_change_request.reload
+    planning_application.reload
+    expect(document_change_request.state).to eq("closed")
+    expect(document_change_request.new_document).to be_a(Document)
```
james, line 26: also add the expectation that the old document has been archived :-)

app/controllers/api/v1/document_change_requests_controller.rb
melzreal: so the reason for using where is that its faster, I remember reading about it whilst working on reducing n+1 queries. Per this stackoverflow using where is preferred from rails 3:
https://stackoverflow.com/questions/9574659/rails-where-vs-find

and this one also advocates find does a lot more:
https://stackoverflow.com/questions/6379099/performance-differences-between-find-and-where-methods


I don't think we should compromise speed over syntax?

melzreal: ok, so instead of find it seems find_by uses "where" under the hood, so we can use that to return a single record and still be efficient https://stackoverflow.com/questions/11161663/find-vs-find-by-vs-where

melzreal: (I really should just read the source code instead of stackoverflow I know, one day)

james (approved):

PR #287, opened 2021-05-17
Add expiry date and change target date

Rhian:
### Description of change

We need an expiry date of 8 weeks, and the target date is now 7 weeks. Target date continues to be displayed in the same places as before

### Story Link

https://trello.com/c/TbyyMZf6/257-change-key-application-dates

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -14,9 +14,9 @@ class PlanningApplication < ApplicationRecord
   belongs_to :user, optional: true
   belongs_to :local_authority
 
-  before_create :set_target_date
+  before_create :set_key_dates
   before_create :set_change_access_id
-  before_update :set_target_date
+  before_update :set_key_dates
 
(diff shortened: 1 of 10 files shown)


app/models/planning_application.rb
```diff
@@ -14,9 +14,9 @@ class PlanningApplication < ApplicationRecord
   belongs_to :user, optional: true
   belongs_to :local_authority
 
-  before_create :set_target_date
+  before_create :set_target_date, :set_expiry_date
   before_create :set_change_access_id
-  before_update :set_target_date
+  before_update :set_target_date, :set_expiry_date
```
melzreal, line 20: would it not be best to just do it all in one method? why have two separate method calls 🤔

melzreal (approved):
👍  although maybe a double check by @james would be good as I don't know if its good practice to have multiple before_create actions, but maybe there is nothing wrong

james (approved):

PR #288, opened 2021-05-17
Update applicant and agent helper

Rhian:
### Description of change

The agent? and applicant? methods now return true if only one field is returned

### Story Link

https://trello.com/c/Z4DqLRhJ/312-show-any-contact-details-if-any-contact-details-exist

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -124,11 +124,11 @@ class PlanningApplication < ApplicationRecord
   end
 
   def agent?
-    agent_first_name? && agent_last_name? && (agent_phone? || agent_email?)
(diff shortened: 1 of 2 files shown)


james:
Is there any point to these methods now? Are they only used in places it could be deleted, and then we can delete the methods?

Rhian:
> Is there any point to these methods now? Are they only used in places it could be deleted, and then we can delete the methods?

It's not a huge advantage, but it DRYs up the code in the view, as otherwise we would have to add an `if exists` condition to 8 different fields to stop the page breaking instead of wrapping each group of 4 fields in the helper

james (approved):

PR #292, opened 2021-05-18
Display UK timezone on audit log

Rhian:
### Description of change

Time displayed was previously UTC. This update forces the app to use London, regardless of server timezone or user's timezone

### Story Link

https://trello.com/c/q7YTMdcA/272-time-in-bops-is-all-gmt-never-got-updated-to-bst

### Decisions 

I decided to go down the route of simply displaying London timezone times in the audit log instead of UTC because this way it avoids messing with what is stored in the DB and also avoids a dependency on server time. Following the "principle of least surprise", I reckon that simply updating the display to UK current time in this way would be what a planner might expect, even if they were temporarily in another timezone.

### DIFF
--- a/app/views/audits/index.html.erb
+++ b/app/views/audits/index.html.erb
@@ -17,7 +17,7 @@
   <tbody class="govuk-table__body">
     <% @audits.each do |audit_item| %>
       <tr class="govuk-table__row">
-        <td class="govuk-table__cell"><%= audit_item.created_at.strftime("%d-%m-%Y %H:%M") %></td>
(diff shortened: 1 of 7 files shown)


james (changes requested):
I would much prefer to use `config.time_zone` and set the timezone globally, so that we don't accidentally miss times in the application.

james:
[Time in BOPS is all GMT (never got updated to BST)](https://trello.com/c/q7YTMdcA/272-time-in-bops-is-all-gmt-never-got-updated-to-bst)

melzreal:
closing in favour of #308

PR #296, opened 2021-05-25
Remove show page for description requests

melzreal:
### Description of change
Design change requests:
- Remove show page for description requests
- Link new document edit page to change requests index
image


james:
I think we need to remove the show action from the controller too?

james (approved):

PR #298, opened 2021-05-25
Expose file representation, instead of file itself

melzreal:
The front end needs a picture preview, which cannot be done with a PDF,
so we transform the PDF to an image for the frontend to use.

Co-authored-by: james <james@abscond.org>


melzreal:
skipped a test as I was going through heartache to fix a test I'm not sure can be fixed (https://stackoverflow.com/questions/52071890/how-can-i-access-an-activestorage-object-via-url-in-a-test-environment)

image

to review.

james (approved):

PR #300, opened 2021-05-28
Ensure map doesn't overlap page break in PDF

james:
Sometimes in the PDF rendering (which uses print CSS), the map would overlap a page break, causing the map to be split over 2 pages.

I've added `page-break-inside: avoid;` to the map which stops this happening! (new to me).

I have also put a wrapping container on the map and added padding to that, otherwise the map hugs the very top of the page.

Before:

Screenshot_2021-05-19_at_13 18 33

After: 

Screenshot 2021-05-28 at 16 38 25

### DIFF
--- a/app/javascript/stylesheets/decision_notice.scss
+++ b/app/javascript/stylesheets/decision_notice.scss
@@ -11,4 +11,12 @@
   dd {
     width: 60%;
   }
+
+  .map {
+    padding-top: 1em;
+  }
+}
+
+.map {
+  page-break-inside: avoid;
 }
--- a/app/views/shared/_location_map.html.erb
+++ b/app/views/shared/_location_map.html.erb
@@ -1,8 +1,10 @@
-<% if locals[:div_id] == "decision_map" %>
-  <div id="decision_map" style="width: 550px; height: 400px"></div>
-<% else %>
-  <div id="accordion_map" style="width: 100%; height: 400px"></div>
-<% end %>
+<div class="map">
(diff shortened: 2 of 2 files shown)


app/javascript/stylesheets/decision_notice.scss
```diff
@@ -11,4 +11,12 @@
...
   }
+
+  .map {
+    padding-top: 1em;
+  }
 }
+
+.map {
+  page-break-inside: avoid;
+}
```
melzreal, line 22: end of file line missing

app/views/shared/_location_map.html.erb
```diff
@@ -1,8 +1,10 @@
...
-  <div id="decision_map" style="width: 550px; height: 400px"></div>
-<% else %>
-  <div id="accordion_map" style="width: 100%; height: 400px"></div>
-<% end %>
+<div class="map">
+    <% if locals[:div_id] == "decision_map" %>
+    <div id="decision_map" style="width: 550px; height: 400px;"></div>
+    <% else %>
+    <div id="accordion_map" style="width: 100%; height: 400px;"></div>
+    <% end %>
```
melzreal, line 6: can we adjust the if else back a level

melzreal (approved):
Approved with little detail comments 🏋️

app/views/shared/_location_map.html.erb
james: good catch

PR #301, opened 2021-06-01
Flip text when clicking to view all applications

melzreal:
### Description of change
image
 
image

### DIFF
--- a/app/views/planning_applications/index.html.erb
+++ b/app/views/planning_applications/index.html.erb
@@ -1,20 +1,26 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-l govuk-!-margin-bottom-3">
-      Your planning applications
-    </h1>
-
-    <p class="govuk-body-m govuk-!-padding-top-2">
-      <strong><%= current_user.name %>,</strong> <%= role_name %>
-    </p>
-
-    <p class="govuk-body govuk-!-padding-bottom-2">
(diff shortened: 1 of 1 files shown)


james:
This works, but I think I'd prefer not to duplicate those lines of code to get out of not doing the if statement twice? What was your thinking on that?

melzreal:
> This works, but I think I'd prefer not to duplicate those lines of code to get out of not doing the if statement twice? What was your thinking on that?

I could create a helper but thought this was more readable so the helper wasn't justified just to avoid repeating two lines of code (writing more code to avoid 2 lines of code seemed like overkill?)

james:
My proposal was:

```
    <h1 class="govuk-heading-l govuk-!-margin-bottom-3">
      <% if exclude_others? %>
        Your planning applications
      <% else %>
         All planning applications
      <% end %>
    </h1>

    <p class="govuk-body-m govuk-!-padding-top-2">
      <strong><%= current_user.name %>,</strong> <%= role_name %>
    </p>

    <p class="govuk-body govuk-!-padding-bottom-2">
      <%= link_to "Add new application", new_planning_application_path %>
    </p>

    <% if exclude_others? %>
      <p class="govuk-body"><%= link_to "View all applications", planning_applications_path, class: "govuk-button" %></p>
    <% else %>
      <p class="govuk-body"><%= link_to filter_text, planning_applications_path(q: "exclude_others"), class: "govuk-button" %></p>
    <% end %>
  </div>
```

james (approved):
Anyway, this isn't a blocker if you don't want it to be

PR #303, opened 2021-06-01
Default valid date to last change request date

melzreal:
### Description of change
Change the default application "valid" date from the documents_validated_at to the last updated change request if any change requests exist on the application.

Also take care of deprecation warning about ".html.erb" partials

### Story Link
https://trello.com/c/6vC0codP/317-default-validation-date-based-on-last-change-request

### DIFF
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -82,7 +82,11 @@ class PlanningApplicationsController < AuthenticationController
   end
 
   def validate_documents_form
-    @planning_application.documents_validated_at ||= @planning_application.created_at
+    @planning_application.documents_validated_at ||= if @planning_application.closed_change_requests.present?
+                                                       @planning_application.last_change_request_date
+                                                     else
+                                                       @planning_application.created_at
+                                                     end
   end
 
   def validate_documents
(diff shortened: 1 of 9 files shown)


james (changes requested):
Everything in planning_application.rb is very ellegant, but I'm not sure it works? This needs tests!

What I'm worried about is that you've created a 'getter' on the model called 'last_change_request_at', and then set that as the value for the date field, which will accurately set the the default value, but then what happens if you change it? Where is that getting saved to?

So I don't *think* it works, but even if it does, it still needs tests to prove it.

james:
The tests are good and useful, thanks :-)

However, in the implementation, this sets the `documents_validated_at` even when the application has not yet been validated. I can imagine this being confusing in the future. What we really want to be doing is only overriding the default value in the HTML form, and leaving the DB alone.

I notice that [the form builder](https://github.com/DFE-Digital/govuk_design_system_formbuilder/blob/master/lib/govuk_design_system_formbuilder/elements/date.rb) doesn't support overriding the default value, which is annoying, but one way around this would be to set on the object on the `validate_documents_form` action. In fact, we're already doing that in planning_applications_controller line 85, so that's what needs editing.

james (approved):

PR #309, opened 2021-06-03
Disable 3D in leaflet JS

james:
This is an attempt to fix a bug in the rendering of the leaflet map in the PDF.

As I am unable to reproduce this bug locally, and it only happens on staging, I am basing this fix on https://github.com/wkhtmltopdf/wkhtmltopdf/issues/1648


Rhian (approved):
Looks good 👍

PR #311, opened 2021-06-04
Attempt 2 at fixing map in PDF

james:
7f2c7fd did not change anything, so I am trying the same thing, but explicitly before the leaflet JS is loaded, as well as adding overflow: hidden on the containing div.

### DIFF
--- a/app/views/application/_head.html.erb
+++ b/app/views/application/_head.html.erb
@@ -15,7 +15,7 @@
   <link rel="apple-touch-icon" sizes="167x167" href="<%= asset_pack_url('media/images/govuk-apple-touch-icon-167x167.png') %>">
   <link rel="apple-touch-icon" sizes="152x152" href="<%= asset_pack_url('media/images/govuk-apple-touch-icon-152x152.png') %>">
   <link rel="apple-touch-icon" href="<%= asset_pack_url('media/images/govuk-apple-touch-icon.png') %>">
-
+  <script>L_DISABLE_3D = true;</script>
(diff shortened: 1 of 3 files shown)


app/views/application/_head.html.erb
```diff
@@ -15,7 +15,7 @@
   <link rel="apple-touch-icon" sizes="167x167" href="<%= asset_pack_url('media/images/govuk-apple-touch-icon-167x167.png') %>">
   <link rel="apple-touch-icon" sizes="152x152" href="<%= asset_pack_url('media/images/govuk-apple-touch-icon-152x152.png') %>">
   <link rel="apple-touch-icon" href="<%= asset_pack_url('media/images/govuk-apple-touch-icon.png') %>">
-
+  <script>L_DISABLE_3D = true;</script>
```
melzreal, line 18: script tags should not be placed in the header of the application, but in the body - just like the one we currently have in ```app/views/layouts/application.html.erb``` which would be the right place for this?

melzreal (changes requested):

app/views/application/_head.html.erb
james: [According to the documentation](https://leafletjs.com/reference-1.7.1.html#global-switches), it is important that this flag is set before leaflet is loaded. I thought that adding it before it is included in the packed JS might work, but it didn't, so now I'm making sure it is set before *any* js is loaded, just to make sure that that isn't the problem.

melzreal (approved):
🏈

PR #313, opened 2021-06-07
switch away from lvh.me

james:
lvh.me has gone down. I've switched us over to a personal DNS record of mine.


Rhian (approved):
Looks good, thanks

PR #321, opened 2021-06-10
Amend valid date to use all change requests

melzreal:
- Previous implementation created a new collection of change requests based on weather these were in a closed state.
This meant that when the new type of change request for red line boundary was added, the method was not doing what it needed to do, because it did not account for it. Instead, now we rely on a method that will always consider existing change requests.

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -241,7 +241,7 @@ class PlanningApplication < ApplicationRecord
   end
 
   def closed_change_requests
-    description_change_requests.closed + document_change_requests.closed + document_create_requests.closed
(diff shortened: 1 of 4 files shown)


app/models/planning_application.rb
```diff
@@ -241,7 +241,7 @@ def change_requests
   end
 
   def closed_change_requests
-    description_change_requests.closed + document_change_requests.closed + document_create_requests.closed
+    change_requests.each { |cr| cr.state.eql?("closed") }
```
Rhian, line 244: 🤩

Rhian (approved):
Good improvement!

PR #323, opened 2021-06-11
Monkey patch Accordion state

james:
We don't want accordions to remember their state on page refresh. Previously we did this by breaking the aria-controls connection, which breaks accessibility requirements, as well as not being very clear on why that is happening.

This approach monkey patches gouk.Accordion methods.

### DIFF
--- a/app/javascript/packs/govuk.js
+++ b/app/javascript/packs/govuk.js
@@ -8,6 +8,9 @@ govukall.Accordion.prototype.setExpanded = function (expanded, $section) {
   }  
 }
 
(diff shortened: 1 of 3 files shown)


james:
[Add aria-controls back in for accessibility, fork and customize gov-uk npm package behaviour](https://trello.com/c/wCI8Osu6/182-add-aria-controls-back-in-for-accessibility-fork-and-customize-gov-uk-npm-package-behaviour)

Rhian (approved):
This is much better!

PR #324, opened 2021-06-11
Upgrades

james:
Upgrade gems and js packages


Rhian (approved):
All looks much better!

PR #328, opened 2021-06-15
Update secure_change_url to use env var

Rhian:
### Description of change

This simplifies the url construction to use the APPLICANTS_APP_HOST env var

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -219,11 +219,8 @@ class PlanningApplication < ApplicationRecord
   end
 
   def secure_change_url(application_id, secure_token)
-    case ENV["DOMAIN"]
-    when "bops-staging.services"
-      "http://#{local_authority.subdomain}.bops-applicants-staging.services/change_requests?planning_application_id=#{application_id}&change_access_id=#{secure_token}"
-    when "bops.services"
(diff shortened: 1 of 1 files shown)


app/models/planning_application.rb
```diff
@@ -219,11 +219,8 @@ def full_address
   end
 
   def secure_change_url(application_id, secure_token)
-    case ENV["DOMAIN"]
-    when "bops-staging.services"
-      "http://#{local_authority.subdomain}.bops-applicants-staging.services/change_requests?planning_application_id=#{application_id}&change_access_id=#{secure_token}"
-    when "bops.services"
-      "http://#{local_authority.subdomain}.bops-applicants.services/change_requests?planning_application_id=#{application_id}&change_access_id=#{secure_token}"
+    if ENV["DOMAIN"]
```
james, line 222: Why do we have this `if` here?

Rhian: To differentiate between http and https - ENV["DOMAIN"] is only set on Prod and Staging. Ideally we'd replace this with a ENV["PROTOCOL"] but that would mean another Ansible update on both Prod and Staging that we would have to coordinate.

james: What about `if Rails.env.production?`

Rhian: Much better, thanks! Will do now.

james (approved):

PR #329, opened 2021-06-15
View decision notice

Rhian:
view_rec

### Description of change

This is an extra view, for the officer only, so the decision notice can be viewed while it's in awaiting determination status.

### Story Link

https://trello.com/c/7Drv7H4R/364-case-officer-can-view-a-decision-notice-when-application-is-in-awaiting-determination

### DIFF
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -15,6 +15,7 @@ class PlanningApplicationsController < AuthenticationController
                                                     determine
                                                     validate_documents_form
                                                     validate_documents
+                                                    view_recommendation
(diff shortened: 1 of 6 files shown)


james:
You need to update the tests for your copy change, but could we also have a system test that views this new page as an officer?

Rhian:
> You need to update the tests for your copy change, but could we also have a system test that views this new page as an officer?

Ah. I've only just seen this comment - it wasn't quite ready to review before - sorry. I'd incorporated the 'visit page' test into a pre-existing spec, but if you'd prefer me to do an entirely new test, also happy to do that.

james (approved):
nope, that's good!

PR #331, opened 2021-06-17
Add receipt email acknowledgement

Rhian:
### Description of change

We now send a receipt email as soon as the application is created. We are calling the Notify integration when an application is either created via the form or via the API.

### Story Link

https://trello.com/c/PxpL2KoR/390-notify-applicant-that-their-application-has-been-received

### Known issues [OPTIONAL]

Things you know need further follow on work:

- Payment amount is TBA until the dependent story has been done
- Phone number has been removed pending product owner decision on whether we need to store these

### DIFF
--- a/app/mailers/planning_application_mailer.rb
+++ b/app/mailers/planning_application_mailer.rb
@@ -32,7 +32,7 @@ class PlanningApplicationMailer < Mail::Notify::Mailer
 
     view_mail(
       NOTIFY_TEMPLATE_ID,
-      subject: "Your planning application has been received",
+      subject: "We have received your application",
       to: @planning_application.applicant_email,
     )
   end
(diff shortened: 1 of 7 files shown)


app/mailers/planning_application_mailer.rb
```diff
@@ -26,6 +26,17 @@ def validation_notice_mail(planning_application, host)
...
   end
 
+  def receipt_notice_mail(planning_application, host)
+    @host = host
+    @planning_application = planning_application
+
+    view_mail(
+      NOTIFY_TEMPLATE_ID,
+      subject: "We have received your application",
+      to: @planning_application.applicant_email,
```
james, line 36: If it we don't have an applicant email, but do have an agent email, should we not send that to them?

Rhian: This is the current methodology for all correspondence. We'll need a new ticket that looks at both applicant and agent details and makes a decision.

james (approved):
Approved for when you have confirmation on the copy

PR #333, opened 2021-06-21
Audit change requests

Rhian:
audit

### Description of change

This PR adds entries to the audit log every time a change request is created or updated.

### Story Link

https://trello.com/c/EhLn4tmq/302-create-audit-logs-for-change-requests

### DIFF
--- a/app/helpers/audit_helper.rb
+++ b/app/helpers/audit_helper.rb
@@ -38,6 +38,8 @@ module AuditHelper
       "Application validated"
     when "withdrawn"
       "Application withdrawn"
+    when "description_change_request_sent"
+      "Sent: request for change (description##{args})"
     end
   end
 end
--- a/app/helpers/audit_helper.rb
+++ b/app/helpers/audit_helper.rb
@@ -59,6 +59,6 @@ module AuditHelper
 
   def define_api_user(audit)
     change_requests = %w[description_change_request_received red_line_boundary_change_request_received document_change_request_received document_create_request_received]
-    change_requests.include?(audit.activity_type) ? "Applicant / Agemt" : audit.api_user.name
+    change_requests.include?(audit.activity_type) ? "Applicant / Agent via #{audit.api_user.name}" : audit.api_user.name
   end
 end
--- a/app/helpers/audit_helper.rb
+++ b/app/helpers/audit_helper.rb
@@ -40,6 +40,10 @@ module AuditHelper
       "Application withdrawn"
     when "description_change_request_sent"
       "Sent: request for change (description##{args})"
+    when "document_change_request_sent"
+      "Sent: request for change (replacement document##{args})"
+    when "document_create_request_sent"
+      "Sent: request for change (new document##{args})"
     end
   end
 end
--- a/app/helpers/audit_helper.rb
+++ b/app/helpers/audit_helper.rb
@@ -46,6 +46,19 @@ module AuditHelper
       "Sent: request for change (new document##{args})"
     when "red_line_boundary_change_request_sent"
       "Sent: request for change (red line boundary##{args})"
+    when "description_change_request_received"
+      "Received: request for change (description##{args})"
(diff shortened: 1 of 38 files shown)


app/models/description_change_request.rb
```diff
@@ -21,4 +23,9 @@ def rejected_reason_is_present?
   def set_previous_application_description
     self.previous_description = planning_application.description
   end
+
+  def set_sequence
+    change_requests = PlanningApplication.find(planning_application.id).description_change_requests
+    increment_sequence(change_requests)
+  end
```
james, line 30: Could this repeated method not be pushed up into the parent ChangeRequest model?

james: oh, I see, it's a different type each time. 🤔

james: Would:
```
  def set_sequence
    self.sequence = self.where(planning_application: planning_application).count + 1
  end
```

work in ChangeRequest for all types?

app/views/audits/index.html.erb
```diff
@@ -21,13 +21,17 @@
...
           <td class="govuk-table__cell"> User deleted </td>
         <% end %>
         <td colspan="2" class="govuk-table__cell">
-          <%= activity(audit_item.activity_type, audit_item.activity_information) %>
+            <%= activity(audit_item.activity_type, audit_item.activity_information) %>
+          <% if audit_item.activity_type.include?("change") %>
+              <%= simple_format(audit_item.audit_comment, wrapper_tag: "div",) if audit_item.audit_comment.present? %>
+          <% else %>
             <%= simple_format(audit_item.audit_comment, { class: "audit_details" }, wrapper_tag: "div",) if audit_item.audit_comment.present? %>
+          <% end %>
```
james, line 34: I'm thinking this is getting quite complicated, and now might be a good time for a refactor of how we do this:

I was thinking the dumber (simpler) option is to do something like:

```
<% @audits.each do |audit_item| %>
   <%= render partial: "audit_logs/types/#{audit_item. activity_type}" %>
<% end %>
```

and then create a partial for each activity type row. What do you think about that?

app/helpers/audit_helper.rb
```diff
@@ -59,6 +59,6 @@ def activity(type_of_activity, args = nil)
 
   def define_api_user(audit)
     change_requests = %w[description_change_request_received red_line_boundary_change_request_received document_change_request_received document_create_request_received]
-    change_requests.include?(audit.activity_type) ? "Applicant / Agemt" : audit.api_user.name
+    change_requests.include?(audit.activity_type) ? "Applicant / Agent via #{audit.api_user.name}" : audit.api_user.name
```
james, line 62: I think we want to say "Applicant / Agent via #{audit.api_user.name}" for all api actions (I mean the only other one is the creation of an application, which is also the applicant via the API).

app/views/audits/index.html.erb
```diff
@@ -21,16 +21,18 @@
...
         <% elsif audit_item.api_user %>
-          <td class="govuk-table__cell"><%= define_api_user(audit_item) %></td>
+          <td class="govuk-table__cell"><%= audit_item.api_user.name %></td>
+        <% elsif audit_item.activity_type.include?("received") %>
+          <td class="govuk-table__cell"> Agent / Applicant via API </td>
         <% else %>
           <td class="govuk-table__cell"> User deleted </td>
         <% end %>
         <td colspan="2" class="govuk-table__cell">
             <%= activity(audit_item.activity_type, audit_item.activity_information) %>
```
james, line 31: Could we also move this into the partials and remove this `activity` method?

james (approved):

app/helpers/audit_helper.rb
Rhian: Unfortunately at the moment we don't submit the api_user from `bops-applicants` so it's a very temporary fix until we can deploy the permanent fix to `bops_applicants`

PR #335, opened 2021-06-29
Add payment amount field

james:
Accept payment_amount via the API, save it to a new field on planning application, and show it in application information


app/views/shared/_application_information.html.erb
```diff
@@ -53,6 +53,14 @@
               </p>
             </td>
           </tr>
+          <tr class="govuk-table__row">
+            <td class="govuk-table__cell"><strong>Payment Amount:</strong></td>
+            <td class="govuk-table__cell">
+              <p class="govuk-body">
+                <%= number_to_currency (@planning_application.payment_amount.to_i/100), unit: '£' %>
```
Rhian, line 60: Not sure if you'd gain anything by moving the calculation to a helper, or if it's just adding code unnecessarily?

Rhian (approved):
Looks good to me - one optional comment. Plus also, do we need to prompt POs to let us know any other places where the payment info should be displayed (in emails, etc)?

PR #343, opened 2021-07-12
Notification banners

Rhian:
validation

### Description of change

We are adding two notification banners to the Planning Application Show view. One is a warning banner that displays if there are overdue validation requests that have not been responded to, and the other one is an information banner that shows validation requests that have been responded to.

### Story Link

https://trello.com/c/DKaQY1c7/346-add-notification-banner-when-change-requests-are-updated

### DIFF
--- a/app/helpers/validation_request_helper.rb
+++ b/app/helpers/validation_request_helper.rb
@@ -38,8 +38,4 @@ module ValidationRequestHelper
       "green"
     end
   end
-
-  def overdue_requests(validation_requests)
-    validation_requests.select { |req| req.overdue? && req.state == "open" }
-  end
 end
--- a/app/helpers/validation_request_helper.rb
+++ b/app/helpers/validation_request_helper.rb
@@ -38,4 +38,8 @@ module ValidationRequestHelper
       "green"
     end
   end
+
+  def overdue_requests(validation_requests)
+    validation_requests.select { |req| req.overdue? && req.state == "open" }
+  end
 end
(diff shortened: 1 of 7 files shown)


app/helpers/validation_request_helper.rb
```diff
@@ -38,4 +38,12 @@ def display_request_status(validation_request)
...
     end
   end
+
+  def overdue_requests(validation_requests)
+    validation_requests.select { |req| req.overdue? && req.state == "open" }
+  end
+
+  def closed_requests(validation_requests)
+    validation_requests.select { |req| req.state == "closed" }
+  end
```
james, line 48: I think these two methods make more sense on the planning_application model than a helper.

app/views/planning_applications/_assessment_dashboard.html.erb
```diff
@@ -8,6 +8,13 @@
 
 <%= render "proposal_header" %>
 
+<div class="govuk-grid-column-two-thirds">
+  <% if @planning_application.validation_requests %>
+    <%= render "overdue_banner" unless overdue_requests(@planning_application.validation_requests).empty? %>
+    <%= render "response_banner" unless closed_requests(@planning_application.validation_requests).empty? %>
```
james, line 14: minor tweak, but I prefer `if .present?` as it's not a double negative.

james (approved):
Looks good! Two non-blocking refactor suggestions, so feel free to do them before merging, or plough on ahead :-)

app/views/planning_applications/_assessment_dashboard.html.erb
Rhian: Fixed!

PR #345, opened 2021-07-12
Switch to Road_3857 OS map layer

james:
There have been a few issues matching geojson data between RIPA and BOPS. We have also been given an API key for OS map layers, so this PR switches us over to OS Map Layer.


Rhian (approved):
Looks good!

PR #346, opened 2021-07-12
Update receipt notification email

Rhian:
### Description of change

Add human-readable application type and remove redundant sentence

### Story Link

https://trello.com/c/ys3oOxMH/422-update-lambeth-notification-email

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -260,7 +260,14 @@ class PlanningApplication < ApplicationRecord
   end
 
   def parsed_application_type
-    application_type == "lawfulness_certificate" ? "Certificate of Lawfulness" : "Full"
+    case application_type
+    when "lawfulness_certificate"
+      "Certificate of Lawfulness"
+    when "full"
+      "Full"
+    else
+      application_type.humanize
+    end
   end
 
 private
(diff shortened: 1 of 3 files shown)


app/models/planning_application.rb
```diff
@@ -259,6 +259,10 @@ def closed_requests
     validation_requests.select { |req| req.state == "closed" }
   end
 
+  def parsed_application_type
+    application_type == "lawfulness_certificate" ? "Certificate of Lawfulness" : "Full"
+  end
```
james, line 264: I'm a bit nervous about saying that if it's not "lawfulness_certificate", then it's definitely "Full"? What if in the future we add a new type and then it's that?

I think this should map "lawfulness_certificate" to "Certificate of Lawfulness", "full" to "Full", and then if it's neither, we either error or call `.humanize` on whatever it is?

Rhian: Sounds like a much better idea!

Rhian: I've updated and added a spec - note there is no spec for the case where it's something other than full or lawfulness as those are the only values we currently allow.

james (approved):

PR #347, opened 2021-07-13
Production deploy

james:
* Switch to using Ordinance Survey map tile service
* Rename change requests to validation requests
* Add "Other Validation Request"
* Add evidence tags to documents
* Add a banner to show how many overdue validation requests are on an application
* Add a banner to show how many validation request responses are on an application
* Align red line map legends better

Things to consider:
- need to deploy applicant updates to reflect the API changes
- We have one real application in the system, but has no validation requests, so I don't think we need to do a data migration, but might be wrong!


Rhian (approved):
LGTM

PR #354, opened 2021-07-19
Ripa result

Rhian:
result

### Description of change

This PR adds an element in the Planning Application accordion that displays any fields that are submitted from the non-mandatory Results json. It displays the name of the API user if one is supplied, and if no result is submitted (or there is no API user), different default text is displayed.

Note that an api_user column has been added to the planning_applications table so we can directly access the creator via API if there was one.

### Story Link

https://trello.com/c/PQegfWX4/381-show-ripa-result

### Known issues 

We'll need to rename the RIPA default user on Staging and Production to RIPA if this is not already the case.


james (approved):
Looks great! I enjoyed having the atomic commits :-)

PR #355, opened 2021-07-21
Ripa questions

Rhian:
proposal_question

### Description of change

We need to find the question(s) that have a flag that matches the flag that is present in the `result` object that RIPA sends us. These are displayed with a number that matches their number in the ordered list in the proposal details array. 

We derive this from the question's index in the proposal_details array, offset by +1.

### Story Link

https://trello.com/c/I99MPhuL/411-as-an-officer-i-want-to-show-the-questions-that-led-to-ripas-decision

### DIFF
--- a/app/helpers/planning_application_helper.rb
+++ b/app/helpers/planning_application_helper.rb
@@ -145,4 +145,8 @@ module PlanningApplicationHelper
       planning_application.returned_at
     end
   end
+
+  def display_number(proposal_details, element)
+    proposal_details.find_index(element) + 1
+  end
 end
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -216,17 +216,17 @@ class PlanningApplication < ApplicationRecord
     proposal_details.present? ? JSON.parse(proposal_details) : []
   end
 
-  def proposals_with_metadata
+  def proposal_details_with_metadata
     parsed_proposal_details.select { |proposal| proposal["metadata"].present? }
   end
 
-  def proposals_with_flags
-    proposals_with_metadata.select { |proposal| proposal["metadata"]["flags"].present? }
+  def proposal_details_with_flags
+    proposal_details_with_metadata.select { |proposal| proposal["metadata"]["flags"].present? }
   end
 
-  def flagged_proposals(flag)
-    proposals_with_flags.select do |proposal|
-      proposal["metadata"]["flags"].select { |element| element == flag }
+  def flagged_proposal_details(flag)
+    proposal_details_with_flags.select do |proposal|
+      proposal["metadata"]["flags"].include?(flag)
     end
   end
 
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -217,16 +217,16 @@ class PlanningApplication < ApplicationRecord
   end
 
   def proposals_with_metadata
-    parsed_proposal_details.select { |detail| proposal["metadata"].present? }
+    parsed_proposal_details.select { |proposal| proposal["metadata"].present? }
   end
 
   def proposals_with_flags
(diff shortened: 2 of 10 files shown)


app/models/planning_application.rb
```diff
@@ -216,6 +216,20 @@ def parsed_proposal_details
...
+    parsed_proposal_details.select { |detail| proposal["metadata"].present? }
+  end
+
+  def proposals_with_flags
+    proposals_with_metadata.select { |proposal| proposal["metadata"]["flags"].present? }
+  end
+
+  def flagged_proposal(flag)
+    proposals_with_flags.select do | proposal|
+        proposal["metadata"]["flags"].any? { |element| element == flag }
```
james: indentation

app/models/planning_application.rb
```diff
@@ -216,6 +216,20 @@ def parsed_proposal_details
     proposal_details.present? ? JSON.parse(proposal_details) : []
   end
 
+  def proposals_with_metadata
```
james, line 219: I think 'proposals' is the wrong word here really. I think `proposal_details` is probably wrong too, but I think calling each question/answer a proposal is a further step into confusing words. Struggling to think of something else though. `proposal_questions_with_metadata`? `proposal_detail_answers_with_metadata`? What do you think?

james: Or I guess just `proposal_details_with_metadata`. That's at least consistent.

app/models/planning_application.rb
james: If `proposal["metadata"]["flags"].include?(flag)` works, then I think that's a bit more readable

app/views/shared/_result_information.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="govuk-accordion__section">
+<div class="govuk-accordion__section result_information">
   <div class="govuk-accordion__section-header">
     <h3 class="govuk-accordion__section-heading">
       <button type="button" id="accordion-default-heading-2" aria-controls="accordion-default-heading-3" class="govuk-accordion__section-button" aria-expanded="false">
```
james, line 4: I've just noticed that the id and aria-control id are different here. I must have missed it in the last PR 😬

app/views/shared/_result_information.html.erb
```diff
@@ -27,6 +27,21 @@
...
           </p>
         </div>
+        <% if @planning_application.flagged_proposals(@planning_application.result_flag).present? %>
+          <p class="govuk-body">
+            <strong>Details identified <%= @planning_application.api_user.present? ? "by #{@planning_application.api_user.name}" : "" %> as relevant to the result</strong>
+          </p>
+          <% @planning_application.flagged_proposals(@planning_application.result_flag).each do |proposal_detail| %>
+          <p class="govuk-body">
+            <strong><%= proposal_detail["question"] %></strong>
+          </p>
```
james, line 39: indentation :-)

app/helpers/planning_application_helper.rb
```diff
@@ -145,4 +145,8 @@ def cancelled_at(planning_application)
       planning_application.returned_at
     end
   end
+
+  def display_number(proposal_details, element)
+    proposal_details.find_index(element) + 1
```
james, line 150: Didn't know about `find_index`!

james (approved):
I've left some tweaks, but no blockers

app/views/shared/_result_information.html.erb
Rhian: Badly copy-pasted from elsewhere and forgot to update, sorry 😬

app/models/planning_application.rb
Rhian: It worked!

PR #357, opened 2021-07-28
Officer adds boundary

james:
Comments in commit messages

### DIFF
--- a/app/views/shared/_interactive_map.html.erb
+++ b/app/views/shared/_interactive_map.html.erb
@@ -1,4 +1,4 @@
-<div class="map">
+<div class="map" style="margin-bottom: 1em">
   <div id="<%= locals[:div_id] %>" style="width: 680px; height: 400px;"></div>
   <div id="coordinates"></div>
 </div>
--- a/app/views/shared/_interactive_map.html.erb
+++ b/app/views/shared/_interactive_map.html.erb
@@ -1,55 +1,59 @@
-<% if locals[:geojson].present? %>
-  <div class="map">
-    <div id="<%= locals[:div_id] %>" style="width: 680px; height: 400px;"></div>
-    <div id="coordinates"></div>
-  </div>
+<div class="map">
+  <div id="<%= locals[:div_id] %>" style="width: 680px; height: 400px;"></div>
+  <div id="coordinates"></div>
+</div>
 
-  <script>
-    var <%= locals[:div_id] %>Map = L.map(<%= locals[:div_id] %>);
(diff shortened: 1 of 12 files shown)


james:
[Allow officer to draw a digital red line based upon a submitted red line pdf](https://trello.com/c/aIwoyv9E/423-allow-officer-to-draw-a-digital-red-line-based-upon-a-submitted-red-line-pdf)

spec/system/planning_applications/drawing_sitemap_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+      click_link "Redraw digital sitemap"
+
+      execute_script 'document.getElementById("planning_application_boundary_geojson").setAttribute("value", \'{"type":"Feature","properties":{},"geometry":{"type":"Polygon","coordinates":[[[-0.054597,51.537331],[-0.054588,51.537287],[-0.054453,51.537313]]]}}\')'
+      click_button "Save"
+
+      expect(page).to have_content("Site boundary has been updated")
+
+      click_button "Key application dates"
+      click_link "Activity log"
+      expect(page).to have_content("Red line drawing created")
```
Rhian, line 41: Should we also add an assertion that it displays the expected case officer?

app/views/shared/_interactive_map.html.erb
```diff
@@ -1,55 +1,59 @@
...
-  </div>
-
-  <script>
-    var <%= locals[:div_id] %>Map = L.map(<%= locals[:div_id] %>);
-      L.tileLayer('https://api.os.uk/maps/raster/v1/zxy/Road_3857/{z}/{x}/{y}.png?key=OmwuvG9HvHq4qKgm8QwaTtTRCMJ1psvk', {
-      attribution: 'Contains OS data &copy; Crown copyright and database rights 2021 OS (0)100019252',
-      maxZoom: 20
-    }).addTo(<%= locals[:div_id] %>Map);
-
+<div class="map" style="margin-bottom: 1em">
```
Rhian, line 1: It's not in the ACs but I'm wondering if it could be an improvement to have the map zoomed to the same resolution as the pre-drawn one? Mine is defaulting to showing a big chunk of the SouthEast - maybe we could zoom it to the postcode plus or minus however many clicks?

Rhian (approved):
I've left a couple of small comments/suggestions but no blockers. Looks great!

spec/system/planning_applications/drawing_sitemap_spec.rb
james: Maybe? I kind of left it because I didn't actually write any code to do that part, it's done elsewhere.

app/views/shared/_interactive_map.html.erb
james: Yes, I think this is a good idea too, but I didn't want to implement it (requires doing geolookups, which we don't have at the moment), without discussing with product/design to prioritise appropiately.

PR #358, opened 2021-07-29
Refactor validation flow to reflect new requirements

Rhian:
### Description of change

This PR completely changes the logic of the validation/invalidation flow and introduces a new view for displaying the validation notification.

- Individual emails are not now sent when validation requests are created. Instead they are saved and updated with a new notified_at field when the officer hits the Invalidate application button
- An email is sent on invalidation (copy needs to be agreed so it's lorem ipsum for now) also containing a link to the validation request overview page on `bops-applicants`
- When a validation request is created and the application is already invalid, an email is sent immediately
- The invalidate and validate buttons are not visible when an application is in a state where this is not possible (for example, closed or withdrawn)
- The planning application cannot now be invalidated unless at least one validation request is open (aka unresolved)
- The planning application cannot be validated if there are any open (unresolved) validation requests
- The number of resolved and unresolved validation requests are shown on the validation form
- The date of the invalidation notification is shown on the validation form if the application is invalidated
- The date of the validation notice, with a link to its content, is shown on the validation form if the application has been validated. A new route has been created for this - design to be agreed but I've based it on the decision notice view

### Story Link

https://trello.com/c/Ub5vP1dd/413-require-a-validation-request-to-mark-application-as-invalid


spec/system/planning_applications/description_change_validation_request_spec.rb
```diff
@@ -99,11 +99,4 @@
     expect(page).to have_text("approved")
     expect(page).to have_text("Applicant / Agent via Api Wizard")
   end
-
-  it "only displays a new validation request option if application is invalid" do
-    planning_application.update!(status: "in_assessment")
-    click_link "Validate application"
-
-    expect(page).not_to have_content("Start new or view existing validation requests")
-  end
```
james, line 108: Why did these get removed? Is this no longer true?

app/models/planning_application.rb
```diff
@@ -35,6 +35,7 @@ class PlanningApplication < ApplicationRecord
   validate :documents_validated_at_date
   validate :public_comment_present
   validate :decision_with_recommendations
+  # validate :can_invalidate
```
james, line 38: Is this meant to be commented out?

app/controllers/planning_applications_controller.rb
```diff
@@ -312,6 +314,13 @@ def validation_notice_mail
     ).deliver_now
   end
 
+  def invalidation_notice_mail
+    PlanningApplicationMailer.invalidation_notice_mail(
+        @planning_application,
+        request.host,
+        ).deliver_now
+  end
```
james, line 327: indentation

app/mailers/planning_application_mailer.rb
```diff
@@ -26,6 +26,17 @@ def validation_notice_mail(planning_application, host)
...
 
+  def invalidation_notice_mail(planning_application, host)
+    @host = host
+    @planning_application = planning_application
+
+    view_mail(
+        NOTIFY_TEMPLATE_ID,
+        subject: "Your planning application is invalid",
+        to: @planning_application.applicant_email,
+        )
```
james: indentation

app/models/planning_application.rb
```diff
@@ -292,6 +292,10 @@ def overdue_requests
     validation_requests.select { |req| req.overdue? && req.state == "open" }
   end
 
+  def invalidation_response_due
+    15.business_days.after(invalidated_at.to_date)
+  end
```
james, line 309: More of a product question, but we already have methods for figuring out when each validation request is due. Would this not be the latest due date of any validation requests?

spec/system/planning_applications/description_change_validation_request_spec.rb
Rhian: The new designs now allow planners to view the change requests regardless of the status, so this link *should* be there when an application is in assessment. Instead, we remove the button for a new request on the validation requests index page

app/models/planning_application.rb
Rhian: It's a legal thing - we need to give the applicant 15 days from when they get the notification. So if a planner raised one or two validation requests then didn't do anything for a couple of days because they were checking something and then  invalidated the application, the date on the email would be wrong.

james: Ah! Well in that case, then the due date for the validation requests should be the date they were sent then?

Rhian: Very good point! I'll add some extra logic for that.

Rhian: I wasn't thinking this through properly. Any validation requests that are bundled together in the invalidation notice will have the same due date as 15 days after the planning application `invalidated_at` date. So I could either add some new logic to look at all the validation requests and select the latest due date or I could simply call the planning_application method as above. Thoughts?

spec/models/planning_application_spec.rb
```diff
@@ -12,6 +12,10 @@
              numbers: "number"
     end
 
+    let!(:description_change_validation_request) do
+      create :description_change_validation_request, planning_application: planning_application, state: "open", created_at: 12.days.ago
+    end
+
```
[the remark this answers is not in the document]
Rhian: Because the `not_started` planning application needs an open validation request attached to it or else it can't invalidate it

app/models/planning_application.rb
james: Ah right, yes, we're sort of going in circles here of logic. We could probably tidy it up a bit for clarity, but this is at least correct.

james:
This is awaiting email copy

app/controllers/additional_document_validation_requests_controller.rb
```diff
@@ -8,8 +8,8 @@ def create
     @additional_document_validation_request.user = current_user
 
     if @additional_document_validation_request.save
-      flash[:notice] = "Document create request successfully sent."
-      send_validation_request_email
+      flash[:notice] = "Document create request successfully created."
```
melzreal, line 11: maybe 
```
Document create request successful
```
would probably avoid repeated the "create", just an idea

spec/system/planning_applications/replacement_document_validation_request_spec.rb
```diff
@@ -93,4 +95,27 @@
...
+
+      click_button "Invalidate application"
+
+      expect(page).to have_content("Application has been invalidated")
+
+      new_planning_application.reload
+      expect(new_planning_application.status).to eq("invalidated")
+
+      request.reload
+      expect(request.notified_at.class).to eql(Date)
```
melzreal, line 118: we could probably move these repeated tests to the validating_spec, just for the sake of having a little less code repetition since they are all testing the same thing but with different types of request?

melzreal (approved):
✨ Awesome work! Tiny comments / ideas but no blockers, works a charm. 
 :shipit:

db/schema.rb
```diff
@@ -75,9 +76,9 @@
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.bigint "api_user_id"
-    t.index ["api_user_id"], name: "index_audits_on_api_user_id"
-    t.index ["planning_application_id"], name: "index_audits_on_planning_application_id"
-    t.index ["user_id"], name: "index_audits_on_user_id"
+    t.index ["api_user_id"], name: "ix_audits_on_api_user_id"
+    t.index ["planning_application_id"], name: "ix_audits_on_planning_application_id"
+    t.index ["user_id"], name: "ix_audits_on_user_id"
```
melzreal, line 81: I guess all the "ix_..._on_user_id" are because we couldn't fit the index name, although "index_something" would be a bit clearer about what they are doing I think. Still one only needs to read the whole line to understand so it matters not 🤷

app/controllers/additional_document_validation_requests_controller.rb
Rhian: Have updated - thanks!

PR #362, opened 2021-08-10
Restore archived

Rhian:
### Description of change

This PR adds the ability for a planner to restore a document to active status if it has been archived in error. It includes a new route `unarchive` and introduces two new fields for documents - an `unarchived_at` column and an `archived` boolean.

### Story Link

https://trello.com/c/TMKc8xIp/368-allow-officers-to-restore-a-document-that-has-been-put-into-archive

### Decisions 

I thought it was better to introduce the `archived` boolean to simply toggle between archived and active because our previous way of defining `archived?` relied on the `archived_at` field being populated, which could now be true of either archived or restored documents


app/views/documents/index.html.erb
```diff
@@ -131,10 +131,15 @@
             <p class="govuk-body govuk-!-margin-bottom-1">
               <%= link_to image_tag(document.file.representation(resize: "300x212")),
                           url_for_document(document), target: :_blank %>
+            </p>
             <p class="govuk-body">
-              <%= document.name %><br/>
-              <%= link_to "View in new window", url_for_document(document), target: :_new %>
+              <%= document.name %></br>
             </p>
+            <%= link_to "View in new window", url_for_document(document), target: :_new %>
```
james, line 138: Why did this get moved outside the `p` tag?

app/views/documents/index.html.erb
```diff
@@ -131,10 +131,15 @@
...
+            <p class="govuk-body">
+              <%= form_for document, url: planning_application_document_unarchive_path(document_id: document.id), class: "restore", method: :patch do |form| %>
```
james, line 140: I don't think you need to specify `method: :patch` here, as `document` is an existing record, so Rails will do magic. Not important though.

db/migrate/20210810093230_add_unarchived_at_to_documents.rb
```diff
@@ -0,0 +1,5 @@
+class AddUnarchivedAtToDocuments < ActiveRecord::Migration[6.1]
+  def change
+    add_column :documents, :unarchived_at, :datetime
```
james, line 3: I think you meant to remove this?

app/views/documents/index.html.erb
Rhian: Thanks for the catch - it was an accident :/ Fixed now.

db/migrate/20210810093230_add_unarchived_at_to_documents.rb
Rhian: I removed the `archived` boolean but I thought this datestamp was probably still useful?

spec/system/documents/archive_spec.rb
```diff
@@ -144,6 +144,46 @@
...
+      fill_in "Why do you want to archive this document?", with: "Scale was wrong"
+      click_button "Archive"
+
+      within(".archived-documents") do
+        expect(page).to have_text("Scale was wrong")
+        expect(page).to have_text("proposed-floorplan.png")
+
+        document_tags.each do |tag|
+          expect(page).to have_css(".govuk-tag", text: tag)
+        end
```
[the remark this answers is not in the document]
Rhian: This was from a pre-existing spec that I copy-pasted to recreate the existing assertions - can remove if you think it's superfluous?

app/views/documents/index.html.erb
Rhian: Ah, didn't realise. Will remove.

db/migrate/20210810093230_add_unarchived_at_to_documents.rb
james: Ah, missed that you're still setting it. I don't know why it would be useful? I'm resistant to adding things we don't explicitly need.

PR #367, opened 2021-08-16
Reinstate lvh.me for local testing

Rhian:
### Description of change

This DNS service became unavailable but has since been restored, so we are reverting to it in order to use subdomains on localhost.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ Back Office Planning System (BOPS)
 
 ## Preflight
 
+Pre-installed dependencies include ImageMagick and Poppler for generating and previewing thumbnails for images and PDFs respectively.
+
 ### Clone the project
 
 ```sh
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -12,6 +12,8 @@ Back Office Planning System (BOPS)
 
 ## Preflight
 
+Pre-installed dependencies include ImageMagick and Poppler for generating and previewing thumbnails for images and PDFs respectively.
+
```
james, line 16: Where did this come from?

config/environments/development.rb
```diff
@@ -66,7 +66,7 @@
   # routes, locales, etc. This feature depends on the listen gem.
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
-  config.action_mailer.default_url_options = { host: (ENV["DOMAIN"] || "local.abscond.org"), port: 3000 }
+  config.action_mailer.default_url_options = { host: (ENV["DOMAIN"] || "local.lvh.me"), port: 3000 }
```
james, line 69: We don't need the `local` bit in this, that was just because I set it as the local subdomain on my domain name. `lvh.me` will work on it's own.

README.md
Rhian: Oh, I forgot to mention! When I was setting up my laptop from scratch, I discovered/remembered that you can't just pull the repo, bundle install and build, ImageMagick and Poppler already need to be on your system. So I thought as I was updating the README I'd add this for info.

PR #372, opened 2021-08-19
Update subdomain testing instructions

melzreal:
- As of 2021 both chrome and firefox support subdomain testing, treating names like *.localhost the same as localhost. This means we no longer need to use a live domain to test subdomains, so I've updated README and config.hosts to reflect this.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -66,7 +66,5 @@ Rails.application.configure do
   # routes, locales, etc. This feature depends on the listen gem.
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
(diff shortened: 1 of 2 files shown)


config/environments/development.rb
```diff
@@ -67,6 +67,4 @@
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
   config.action_mailer.default_url_options = { host: (ENV["DOMAIN"] || "local.abscond.org"), port: 3000 }
```
Rhian, line 69: We should also update this line to remove `abscond.org`

Rhian (approved):
Approved with one small change - and also we need to update the README.

PR #376, opened 2021-08-24
Add notified at and closed dates to validation index

melzreal:
### Description of change
https://trello.com/c/rL7CZA0b/363-add-created-and-received-dates-on-to-validation-request-table
image

### DIFF
--- a/app/helpers/validation_request_helper.rb
+++ b/app/helpers/validation_request_helper.rb
@@ -38,4 +38,14 @@ module ValidationRequestHelper
       "green"
     end
   end
+
+  def request_sent_at(validation_request)
+    validation_request.notified_at.strftime("%e %B %Y") if validation_request.notified_at.present?
+  end
+
+  def request_closed_at(validation_request)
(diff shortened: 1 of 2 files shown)


app/helpers/validation_request_helper.rb
```diff
@@ -38,4 +38,14 @@ def display_request_status(validation_request)
       "green"
     end
   end
+
+  def request_sent_at(validation_request)
```
Rhian, line 42: I know it's not in the designs, but wondering if we should have a "Not sent" or similar to draw it to the officer's attention if they have created requests without also invalidating?

Rhian (approved):
Approved - have made a UX suggestion but it's not a blocker for the merge

app/helpers/validation_request_helper.rb
melzreal: Its a good point - I'll merge this one and ask Tom to check during UAT / UX to see if the officers need that extra reinforcement - he might even prefer an alert box at the top or so

PR #377, opened 2021-08-24
Update notified_at when planning application is invalidated

Rhian:
### Description of change

All types of validation request currently get a `notified_at` date when the application is invalidated, while validation requests that are sent when the application is in `invalidated` state do not get this update.

This commit unifies the approach so that all validation requests have the `notified_at` field updated, regardless of the planning application status.

### Story Link

https://trello.com/c/NU4AXeJV/489-update-notifiedat-field-for-all-validation-requests

### DIFF
--- a/app/controllers/replacement_document_validation_requests_controller.rb
+++ b/app/controllers/replacement_document_validation_requests_controller.rb
@@ -1,4 +1,4 @@
-class ReplacementDocumentValidationRequestsController < ApplicationController
+class ReplacementDocumentValidationRequestsController < ValidationRequestsController
   def new
     @replacement_document_validation_request = planning_application.replacement_document_validation_requests.new
   end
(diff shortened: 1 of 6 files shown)


app/controllers/replacement_document_validation_requests_controller.rb
```diff
@@ -31,6 +31,11 @@ def send_validation_request_email
     ).deliver_now
   end
 
+  def email_and_timestamp
+    send_validation_request_email
+    @replacement_document_validation_request.update!(notified_at: Time.zone.now)
+  end
```
melzreal, line 37: seems like a lot of code repetition - could we not pop this method into ValidationRequestsController and make these controllers inherit from it?

melzreal: (same with the send_validation_request email, we could pop both into that controller theoretically and just call it with an argument or validation_request type, it seems the code is kind of exactly the same except for the request type?)

Rhian: Yes, good idea

melzreal (approved):
LGTM ⭐

PR #379, opened 2021-08-24
Add migration to update existing applications

melzreal:
### Description of change

Due to the change in implementation of when notifications for change requests get sent, we need to make sure previous change requests also have a notified_at field to inform officers of when the email notification was sent to the user

### Story Link
https://trello.com/c/rL7CZA0b/363-add-created-and-received-dates-on-to-validation-request-table

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_09_10_145259) do
+ActiveRecord::Schema.define(version: 2021_09_13_155856) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
--- a/db/migrate/20210913155856_update_notified_at.rb
+++ b/db/migrate/20210913155856_update_notified_at.rb
@@ -1,0 +1,13 @@
+class UpdateNotifiedAt < ActiveRecord::Migration[6.1]
+  def change
+    reversible do |dir|
+      dir.up do
+        PlanningApplication.find_each do |planning_application|
+          planning_application.validation_requests.each do |request|
+            request.update(notified_at: request.created_at)
+          end
+        end
+      end
+    end
+  end
+end


db/schema.rb
```diff
@@ -22,7 +22,7 @@
     t.bigint "record_id", null: false
     t.bigint "blob_id", null: false
     t.datetime "created_at", null: false
-    t.index ["blob_id"], name: "ix_active_storage_attachments_on_blob_id"
+    t.index ["blob_id"], name: "index_active_storage_attachments_on_blob_id"
```
freesteph, line 25: curious why the index names have changed?

db/migrate/20210824160127_update_notified_at_with_created_at.rb
```diff
@@ -0,0 +1,13 @@
+class UpdateNotifiedAtWithCreatedAt < ActiveRecord::Migration[6.1]
+  def change
+    reversible do |dir|
+      dir.up do
+        PlanningApplication.all.map do |planning_application|
```
freesteph, line 5: `find_each`

freesteph: and probably `include(:validation_requests)`

db/migrate/20210824160127_update_notified_at_with_created_at.rb
```diff
@@ -0,0 +1,13 @@
...
+          planning_application.validation_requests.map do |request|
+            request.update(notified_at: request.created_at.strftime("%e %B %Y"))
```
Rhian, line 7: We don't need to format the date inside the table - only when we display it

db/migrate/20210824160127_update_notified_at_with_created_at.rb
```diff
@@ -0,0 +1,13 @@
+class UpdateNotifiedAtWithCreatedAt < ActiveRecord::Migration[6.1]
+  def change
+    reversible do |dir|
+      dir.up do
+        PlanningApplication.all.map do |planning_application|
+          planning_application.validation_requests.map do |request|
```
freesteph, line 6: `each`, `map` is when you want something back (altered by whatever callback function you pass)

db/migrate/20210824160127_update_notified_at_with_created_at.rb
freesteph, line 5: also I believe it's good practice to redeclare your model class and its association if you're going to use it/them in your migration:

```ruby
class ValidationRequest < ApplicationRecord
  belongs_to :planning_application:
end

class PlanningApplication < ApplicationRecord
  has_many :validation_requests
end

end
```

db/migrate/20210824160127_update_notified_at_with_created_at.rb
melzreal: I had checked the date if the notified_at within my db and realised they were all formatted already so thought it was expected hence placing it here.. but I'll change it!

db/migrate/20210824160127_update_notified_at_with_created_at.rb
melzreal: it doesn't have many validation_requests though, so instead of that association its a scope method, within PlanningApplication line 274:

```
  def validation_requests
    (description_change_validation_requests + replacement_document_validation_requests + additional_document_validation_requests + other_change_validation_requests + red_line_boundary_change_validation_requests).sort_by(&:created_at).reverse
  end
```

melzreal: I'll declare the PlanningApplication class atop though and use it with find_each instead

db/migrate/20210824160127_update_notified_at_with_created_at.rb
```diff
@@ -0,0 +1,13 @@
+class UpdateNotifiedAtWithCreatedAt < ActiveRecord::Migration[6.1]
+  def change
+    reversible do |dir|
+      dir.up do
+        PlanningApplication.all.each do |planning_application|
```
freesteph, line 5: still `find_each`

melzreal: I tried find_each on the console and this:

```
 PlanningApplication.find_each do |planning_application| planning_application.validation_requests.each do |request| request.created_at end end
 ```
 returns: 
image

whereas this:

``` PlanningApplication.all.each do |planning_application| planning_application.validation_requests.each do |request| request.created_at end end```
returns

image

so I don't get it, find_each doesn't return the data?

freesteph: it doesn't; it just yields each record, but you don't need the returned data do you?

melzreal:
closing since we have now merged the new validation flow - meaning this field will always be present in new applications from this point forward and no request has been made to have this in historically.

PR #388, opened 2021-09-07
Email agent and applicant

melzreal:
### Description of change

Email both applicant and agent:

- decision notice
- validation notice
- receipt notice

Only validation requests go to:

- if agent, only go to agent
- if applicant, go to agent if agent exists if not got to applicant

### Story Link
https://trello.com/c/x4mQZ5j0/405-all-emails-notifications-decision-notice-to-go-to-either-agent-or-applicant

### DIFF
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -54,7 +54,7 @@ class PlanningApplicationsController < AuthenticationController
     if @planning_application.save
       audit("created", nil, current_user.name)
       flash[:notice] = "Planning application was successfully created."
-      receipt_notice_mail if @planning_application.applicant_email.present?
+      receipt_notice_mail if @planning_application.agent_email.present? || @planning_application.applicant_email.present?
       redirect_to planning_application_documents_path(@planning_application)
     else
       render :new
@@ -107,7 +107,7 @@ class PlanningApplicationsController < AuthenticationController
       @planning_application.start!
       audit("started")
       validation_notice_mail
-      flash[:notice] = "Application is ready for assessment and applicant has been notified"
+      flash[:notice] = "Application is ready for assessment and an email notification has been sent."
       render :show
     end
   end
--- a/app/mailers/planning_application_mailer.rb
+++ b/app/mailers/planning_application_mailer.rb
@@ -3,6 +3,13 @@
 class PlanningApplicationMailer < Mail::Notify::Mailer
   NOTIFY_TEMPLATE_ID = "7cb31359-e913-4590-a458-3d0cefd0d283"
 
+  def applicant_and_agent_email(planning_application)
+    @agent = planning_application.agent_email
+    @applicant = planning_application.applicant_email
+
+    [@agent, @applicant].reject(&:nil?)
+  end
+
   def decision_notice_mail(planning_application, host)
     @planning_application = planning_application
     @documents = @planning_application.documents.for_display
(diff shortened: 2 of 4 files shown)


app/mailers/planning_application_mailer.rb
```diff
@@ -3,6 +3,17 @@
...
 
+  def email_applicant_and_agent(planning_application)
+    @agent = planning_application.agent_email
+    @applicant = planning_application.applicant_email
+
+    if @agent.present? && @applicant.present?
+      [@agent, @applicant]
+    else
+      @agent.presence || @applicant
+    end
```
freesteph, line 14: You can just grab both and remove potential nils I think

```
[@agent, @applicant].reject(&:nil?)
```

app/mailers/planning_application_mailer.rb
```diff
@@ -3,6 +3,17 @@
 class PlanningApplicationMailer < Mail::Notify::Mailer
   NOTIFY_TEMPLATE_ID = "7cb31359-e913-4590-a458-3d0cefd0d283"
 
+  def email_applicant_and_agent(planning_application)
```
freesteph, line 6: applicant_and_agent_email, otherwise it sounds like it's going to actually mail them

app/mailers/planning_application_mailer.rb
```diff
@@ -44,18 +55,19 @@ def receipt_notice_mail(planning_application, host)
...
       subject: "We have received your application",
-      to: @planning_application.applicant_email,
+      to: email_applicant_and_agent(@planning_application),
     )
   end
 
   def validation_request_mail(planning_application, validation_request)
     @planning_application = planning_application
     @validation_request = validation_request
+    @application_accountable_email = @planning_application.agent_email.presence || @planning_application.applicant_email
```
freesteph, line 65: why is this not using the method you made?

app/controllers/planning_applications_controller.rb
```diff
@@ -54,7 +54,7 @@ def create
     if @planning_application.save
       audit("created", nil, current_user.name)
       flash[:notice] = "Planning application was successfully created."
-      receipt_notice_mail if @planning_application.applicant_email.present?
+      receipt_notice_mail if @planning_application.agent_email.present? || @planning_application.applicant_email.present?
```
freesteph, line 57: this could be `email_agent_and_applicant.any?`

app/mailers/planning_application_mailer.rb
melzreal: this one is the exception - in all other cases we want to email both, but for "validation requests" (where an action is required) we want to only email agent if an agent is present, or email the applicant if no agent is present

app/controllers/planning_applications_controller.rb
melzreal: but our email_agent_and_applicant method is inside  PlanningApplicationMailer - how would the controller know about this method given it doesn't belong to the planningapplication itself? (all the emails are being called by redeclaring the methods inside the controllers, like so:
```  def receipt_notice_mail
    PlanningApplicationMailer.receipt_notice_mail(
      @planning_application,
      request.host,
    ).deliver_now
  end```

how does one go around that?

app/mailers/planning_application_mailer.rb
freesteph: OK; for what it's worth one option could have been `applicant_and_agent_email.first(&:present?)` but happy with it anyway

app/controllers/planning_applications_controller.rb
freesteph: sorry I didn't realise it was in a controller; this is probably trying to tell us the logic should be on the model but I'm happy to leave it as such for the time being.

freesteph (approved):

PR #390, opened 2021-09-08
Send two emails on receipt, validation & decision

melzreal:
### Description of change
- Since notify does not support sending the same email to different to addresses, we iterate through the applicant and agent addresses and send multiple emails to fulfill the acs on this
https://trello.com/c/x4mQZ5j0/405-all-emails-notifications-decision-notice-to-go-to-either-agent-or-applicant

### DIFF
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -306,17 +306,23 @@ private
   end
 
   def decision_notice_mail
-    PlanningApplicationMailer.decision_notice_mail(
-      @planning_application,
-      request.host,
-    ).deliver_now
+    @planning_application.applicant_and_agent_email.each do |user|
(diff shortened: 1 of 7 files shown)


app/controllers/planning_applications_controller.rb
```diff
@@ -327,10 +333,20 @@ def invalidation_notice_mail
...
+    applicant_and_agent_email(@planning_application).each do |user|
+      PlanningApplicationMailer.receipt_notice_mail(
+        @planning_application,
+        request.host,
+        user,
+      ).deliver_now
+    end
+  end
+
+  def applicant_and_agent_email(planning_application)
```
freesteph, line 345: this should live on the mailer, not on the controller: the controller says "send the email", and then the mailer can handle its logic of doubling the email

freesteph (approved):

PR #395, opened 2021-09-10
Add local authority notify id to emails

melzreal:
### Description of change

- Updated the local authority db table with an email id and associated the correct id from notify to each local authority.
This allows us to add the reply-to parameter in planning application mailer, so each email can have a different reply-to address, depending on which local authority the planning application belongs to.

### Story Link

https://trello.com/c/uUqHl2wL/468-set-up-reply-to-email-from-notify

### DIFF
--- a/db/migrate/20210910145259_add_reply_to_local_authority.rb
+++ b/db/migrate/20210910145259_add_reply_to_local_authority.rb
@@ -1,0 +1,17 @@
+class AddReplyToLocalAuthority < ActiveRecord::Migration[6.1]
+  class LocalAuthority < ApplicationRecord; end
+  
+  def change
+    add_column :local_authorities, :reply_to_notify_id, :string
+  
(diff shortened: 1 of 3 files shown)


db/migrate/20210910145259_add_reply_to_local_authority.rb
```diff
@@ -0,0 +1,15 @@
+class AddReplyToLocalAuthority < ActiveRecord::Migration[6.1]
+  def change
+    add_column :local_authorities, :reply_to_notify_id, :string
+  
+    reversible do |dir|
+      dir.up do
+        LocalAuthority.find_each do |authority|
```
freesteph, line 7: can you redeclare the class in your migration?

```rb
class LocalAuthority < ApplicationRecord; end
```

freesteph (approved):

PR #396, opened 2021-09-13
Prevent validation when invalid documents exist

melzreal:
### Description of change

Prevent an application from being marked as valid if any invalid documents are present.

image

### Story Link
https://trello.com/c/y22SlsbN/379-prevent-a-case-from-being-marked-valid-if-a-live-document-is-marked-as-invalid

### DIFF
--- a/app/views/planning_applications/validate_form.html.erb
+++ b/app/views/planning_applications/validate_form.html.erb
@@ -66,8 +66,8 @@
           </h2>
           <div class="govuk-error-summary__body">
             <ul class="govuk-list govuk-error-summary__list">
-              <% @planning_application.errors.full_messages.each do |error| %>
-                <li><%= error %></li>
+              <% @planning_application.errors.each do |error| %>
+                <li><%= error.message %></li>
               <% end %>
             </ul>
           </div>
--- a/spec/system/planning_applications/validating_spec.rb
+++ b/spec/system/planning_applications/validating_spec.rb
@@ -88,7 +88,6 @@ RSpec.shared_examples "validate and invalidate" do
 
   it "allows for the user to input a validation date manually" do
     visit validate_form_planning_application_path(second_planning_application)
-
     fill_in "Day", with: "3"
     fill_in "Month", with: "6"
     fill_in "Year", with: "2022"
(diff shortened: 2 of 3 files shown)


app/views/planning_applications/validate_form.html.erb
```diff
@@ -66,8 +66,8 @@
           </h2>
           <div class="govuk-error-summary__body">
             <ul class="govuk-list govuk-error-summary__list">
-              <% @planning_application.errors.full_messages.each do |error| %>
```
freesteph, line 69: Why aren't we using the full messages anymore?

spec/system/planning_applications/validating_spec.rb
```diff
@@ -173,6 +172,21 @@
...
+             validated: false, invalidated_document_reason: "Missing a lazy Suzan"
+
+      click_link planning_application.reference
+      click_link "Validate application"
+
+      click_button "Validate application"
+      expect(page).to have_content("This application has an invalid document. You cannot validate an application with invalid documents.")
+
+      planning_application.reload
+      expect(planning_application.status).to eql("not_started")
```
freesteph, line 187: for the record, you could also do
```rb
       expect(planning_application).to be not_started
```

melzreal: ah thanks - I usually do write it like that but because this was a copy paste I didn't even look at it - will change

app/views/planning_applications/validate_form.html.erb
melzreal: Because when using the full messages, the message displayed becomes
"Planning application lalalal error"
since the errors are all being shoved into :planning_application.

This is actually what shows up in staging right now when you remove the date, but no one seems to have noticed the bug:
image

I have the idea I changed this somewhere else before but some reason we went back to the full_messages. When I upgraded the app to rails 6, the error messages changed in that version of rails and hence needing reworking as the errors are no longer a hash - now each error knows about itself: https://code.lulalala.com/2020/0531-1013.html

spec/system/planning_applications/validating_spec.rb
melzreal: actually there are about 10 tests in this spec that all use this format so I guess I'll leave it alone for consistency 🤷 unless you'd like me to change all of them?

app/views/planning_applications/validate_form.html.erb
freesteph: Gotcha

freesteph (approved):

PR #398, opened 2021-09-15
Update readme & add subdomain testing instructions

melzreal:
### Description of change

Due to bops-applicants not working with a local setup, have registered bops-care.link and added it to the development config to allow for subdomain testing with the front-end app. Also updated README instructions.


README.md
```diff
@@ -82,12 +82,23 @@ $ JS_DRIVER=selenium_chrome rspec spec/system/log_in_spec.rb
 $ rails server
 ```
 
-#### Because of the subdomain being enforced, you will need to use Chrome or Firefox to test subdomains (unless you configure your own computer etc file for this to work on any browser):
+#### Because of the subdomain being enforced, your options to test subdomains include using bops-care.link (needed to test bops-applicants[https://github.com/unboxed/bops-applicants]) or, if you're just testing bops, Chrome, Firefox and Brave currently support subdomains too. This means your app will be available on:
```
freesteph, line 85: can you capitalise all instances of BOPS?

freesteph (approved):

PR #399, opened 2021-09-15
Correct order of resolved and unresolved in helper

melzreal:
### Description of change

"Closed" should == resolved and "open" == unresolved, but the helper had those words switched on the last if statement.


freesteph (approved):

Rhian (approved):
LGTM!

PR #400, opened 2021-09-15
Yield title only when content for title is present 

melzreal:
### Description of change

Yield was generating an extra > if no title exists for a partial, but the code works the same without yielding - except it doesn't create an empty list item unnecessarily

Before:
image

After:
image

### DIFF
--- a/app/views/application/_create_breadcrumb_navigation.html.erb
+++ b/app/views/application/_create_breadcrumb_navigation.html.erb
@@ -4,6 +4,8 @@
   </li>
 <% end %>
 
-<li class="govuk-breadcrumbs__list-item" aria-current="page">
-  <%= yield :title %>
-</li>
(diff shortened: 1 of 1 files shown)


freesteph (approved):

app/views/application/_create_breadcrumb_navigation.html.erb
```diff
@@ -4,6 +4,8 @@
...
 <% end %>
 
-<li class="govuk-breadcrumbs__list-item" aria-current="page">
-  <%= yield :title %>
-</li>
+<% if content_for?(:title) %>
+  <li class="govuk-breadcrumbs__list-item" aria-current="page">
+    <%= yield :title  %>
+  </li>
+<% end %>
```
Rhian, line 11: Very tiny comment about missing newline,  but obviously no blocker :)

PR #401, opened 2021-09-17
Address email to agent or applicant names

melzreal:
### Description of change

Addressee defaulted to applicant first / last name, rules in emails changed to the below.

Receipt of notification
CURRENT: No name used
ACTION: Amend copy, use agent name (if agent) otherwise use applicant name

Application invalid
CURRENT: Applicant name
ACTION: Use agent name (if agent) otherwise use applicant name

Application valid
CURRENT: Sir / Madam
ACTION: Amend copy, use agent name (if agent) otherwise use applicant name

Validation request (email)
CURRENT: Applicant name
ACTION: Use agent name (if agent) otherwise use applicant name

Validation request (in-app)
CURRENT: No name used
ACTION: No change

Decision notice (email)
CURRENT: Applicant name
ACTION: Use agent name (if agent) otherwise use applicant name

LDC Decision Notice (current)
CURRENT: Applicant name
ACTION: No change

### Story Link
https://trello.com/c/tfXbYm1I/522-notifications-default-to-using-applicant-name-rather-than-agent-can-a-rule-be-put-on-all-notifications

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -326,6 +326,14 @@ class PlanningApplication < ApplicationRecord
     [agent_email, applicant_email].reject(&:blank?)
   end
 
+  def agent_or_applicant_name
+    if agent_first_name?
+      "#{agent_first_name} #{agent_last_name}"
+    else
+      "#{applicant_first_name} #{applicant_last_name}"
+    end
+  end
+
 private
 
   def set_key_dates
--- a/app/views/planning_application_mailer/decision_notice_mail.text.erb
+++ b/app/views/planning_application_mailer/decision_notice_mail.text.erb
@@ -1,4 +1,4 @@
-<%= @planning_application.applicant_name %>,
+<%= @planning_application.agent_or_applicant_name %>,
 
 Your Certificate of lawful development (<%= @planning_application.work_status %>) has been <%= @planning_application.decision %>.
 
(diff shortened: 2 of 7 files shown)


app/models/planning_application.rb
```diff
@@ -326,6 +326,14 @@ def applicant_and_agent_email
     [agent_email, applicant_email].reject(&:blank?)
   end
 
+  def agent_or_applicant_name
+    if agent_first_name?
```
Rhian, line 330: Just wondering if we should cater for the edgiest of edge cases where an agent/applicant has one name only? Like if it's a business name or something? Not sure whether RIPA enforce this or not, but if there's any possibility that one of these fields will be `nil` we need to prevent an error. If RIPA enforce this, then not an issue!

melzreal: we can cater for that - its actually a point of failure across different parts of the email - For example, we have a method called agent? that displays "supporting info" if any of the agent details is present - but the only thing we absolutely require is the agent or the applicant email. We also actually don't even require an address to create an application.

My thought was that if someone is hiring an agent to do something the agent should be required to insert their details properly but its easy enough to check if both fields are empty I guess

Rhian: It could maybe go in as an improvement ticket and we could clarify with product owners exactly what details we should be enforcing so it becomes part of a wider issue. Happy to approve for now.

Rhian (approved):
LGTM!

PR #402, opened 2021-09-20
Apply size limitation to document upload

Rhian:
### Description of change

Documents greater than 2MB were failing because of the nginx config default. Now files up to 30MB are permitted. Those above need to be disallowed on the controller rather than the model

### Story Link

https://trello.com/c/RjtWp2rB/97-bug-unsupported-filetype-for-a-document-submitted-through-api-breaks

### DIFF
--- a/app/controllers/api/v1/validation_requests_controller.rb
+++ b/app/controllers/api/v1/validation_requests_controller.rb
@@ -13,4 +13,10 @@ private
       @planning_application
     end
   end
+
+  def check_file_size
+    if params[:new_file].size > 30.megabytes
+      render json: { "message": "The file must be 30MB or less" }, status: :bad_request
+    end
+  end
 end
--- a/app/controllers/api/v1/additional_document_validation_requests_controller.rb
+++ b/app/controllers/api/v1/additional_document_validation_requests_controller.rb
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
-class Api::V1::AdditionalDocumentValidationRequestsController < Api::V1::ApplicationController
+class Api::V1::AdditionalDocumentValidationRequestsController < Api::V1::ValidationRequestsController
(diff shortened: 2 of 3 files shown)


melzreal:
we probably need the same check done when creating the planning application, on "upload_documents" for the PlanningApplicationsController ?

app/controllers/api/v1/validation_requests_controller.rb
```diff
@@ -13,4 +13,10 @@ def check_token_and_set_application
       @planning_application
     end
   end
+
+  def check_file_size
+    if params[:new_file].size > 30.megabytes
+      render json: { "message": "The file must be 30MB or less" }, status: :bad_request
+    end
+  end
```
melzreal, line 21: we should add a little test for this I guess - did you ever find out how to stub the size on a spec?

Rhian: Steph recommended leaving the test for now as it was blocking getting this in

Rhian:
> we probably need the same check done when creating the planning application, on "upload_documents" for the PlanningApplicationsController ?

I was in two minds about this, as the only people who will be creating using this method are planning officers, who should know this. But yeah could add to be extra-safe.

app/controllers/api/v1/validation_requests_controller.rb
Rhian: ... I may move this to a WIP PR while we think about it

freesteph: Couldn't crack a test :) Somewhere in the upload/request chain, ActionDispatch re-reads the file and erases my `size` mock (though it appears in the `Content-Length` header, but we want to check `size`, it's easier).

freesteph (approved):
Looks good :palm_tree:

PR #404, opened 2021-09-21
Updated secure_change_url method to use https on both staging and pro…

Rhian:
### Description of change

We were previously using https for `bops-applicants` on production for legacy reasons. Now we have a certificate on `bops-applicants` in the preview/staging environment, we want to use https everywhere other than our local.

### Story Link

https://trello.com/c/AKfU3h8O/481-ensure-email-links-on-prod-for-bops-applicants-are-sent-with-https-instead-of-http

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -252,7 +252,7 @@ class PlanningApplication < ApplicationRecord
   end
 
   def secure_change_url(application_id, secure_token)
-    if ENV["DOMAIN"] == "bops-services"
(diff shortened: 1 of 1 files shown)


freesteph (approved):

melzreal (changes requested):
We're still missing the security cert for staging - can do it this afternoon, if you want we can pair on it as its a quick thing - I only added it for prod and totally neglected staging from what I can remember

melzreal (approved):
🚀

PR #405, opened 2021-09-21
Feat/policy references

freesteph:
# Description

This introduce the idea of policy classes: see commit 01607bc268ec9085cb51071eaa425429ca7df860 for a brief overview of what they are and how we implement them.

Screen Shot 2021-09-21 at 18 18 18

A description of the UX flow and screenshots are available on the relevant Trello card: https://trello.com/b/KQqYJd7W/bops-private-beta-mvp-permitted-development.

# Checklist
- [x] implement the delete class action;
- [x] complete the unit tests.


config/environments/test.rb
```diff
@@ -9,7 +9,7 @@
   # Settings specified here will take precedence over those in
   # config/application.rb.
 
-  config.cache_classes = true
+  config.cache_classes = false
```
pixeltrix, line 12: Why this change? You're not reloading code during a test run

config/routes.rb
```diff
@@ -12,6 +12,16 @@
     member do
       get :assign
       patch :assign
+
+      resources :policy_assessments do
```
pixeltrix, line 16: Probably need to omit some routes here - like destroy, etc.

pixeltrix: Also nested resources don't need to live inside the member scope

Gemfile
```diff
@@ -43,11 +43,13 @@ group :development, :test do
   gem "rubocop-govuk", "4.0.0"
   gem "selenium-webdriver"
   gem "simplecov", require: false
+  gem "spring-commands-rspec"
   gem "webdrivers"
 end
 
 group :development do
   gem "listen"
+  gem "spring"
```
pixeltrix, line 52: Are you sure you want to add spring and are the other devs okay with it?

app/controllers/policy_assessments_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+                .select { |c| class_ids.include? c.id }
+                .each { |c| c.stamp_status! }
+
+    @planning_application.policy_classes += classes
+
+    if @planning_application.save
+      redirect_to @planning_application, notice: "classes successfully added"
+    else
+      render :new
+    end
```
pixeltrix, line 27: This action and the update_class one feel a little clunky but they can be iterated on.

config/routes.rb
```diff
@@ -12,6 +12,16 @@
...
+        get :part, on: :new
+
+        collection do
+          get "/:part/:policy_class", action: "show_class"
+          put "/:part/:policy_class", action: "update_class"
```
pixeltrix, line 21: These routes feel wrong - why are they on the policy assessment collection? Don't they apply to specific policy assessments?

Gemfile
freesteph: They are but I've changed my mind I think. Also Spring is going away so I've made my peace...

config/environments/test.rb
freesteph: I thought I had a Rails warning about it but I'll revert.

app/controllers/policy_assessments_controller.rb
freesteph: Agreed; I ended up writing the specs later, but once they're all green I'll start refactoring

config/routes.rb
```diff
@@ -9,6 +9,15 @@
   mount Rswag::Ui::Engine => "api-docs"
 
   resources :planning_applications, only: %i[index show new edit create update] do
+    resources :policy_assessments, only: %i[new create] do
+      get :part, on: :new
+
+      collection do
+        get "/:part/:policy_class", action: "show_class"
+        put "/:part/:policy_class", action: "update_class"
```
Rhian, line 17: We'll also need a route for `delete` as this CTA throws an error right now

app/views/policy_assessments/show_class.html.erb
```diff
@@ -0,0 +1,80 @@
...
+    <h1 class="govuk-heading-xl">Assess - <%= @klass %></h1>
+
+    <p class="govuk-body">
+      Please indicate if the application complies, does not comply, or
+      is not applicable to each of the policies. Policies are defined in the Town and Country
+      Planning (General Permitted Development) (England) Order 2015,
+      Schedule 2, <%= @klass %>.</p>
+
+    <p class="govuk-body">
+      <%= link_to "Open legislation in new window", @klass.url, class: "govuk-link" %>
```
Rhian, line 17: Needs a target, as it currently opens in the same tab

app/views/planning_applications/show.html.erb
```diff
@@ -1,108 +1,111 @@
...
+        <% end %>
+      </li>
+    </ul>
+    <h2 class="govuk-heading-l">Assess against policy</h2>
+    <ul class="app-task-list__items" id="assess-policy-section">
+      <% @planning_application.policy_classes.each do |klass| %>
+        <li class="app-task-list__item">
+          <span class="app-task-list__task-name">
+            <%= link_to klass, {
+                controller: "policy_assessments",
```
melzreal, line 26: I'm finding the use of "klass" everywhere really confusing. Is it necessary here? Would it not be easier to read if we used  "policy" or"policy_class" for the iteration? 

It feels like"klass" is everywhere and its making the code a bit tricky, not quite as straightforward as it could be with plainer english

app/controllers/policy_assessments_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+  def policy_class_params
+    params.permit(:part, policy_classes: [])
+  end
+
+  def policies_params
+    params.permit(:part, :policy_class, policies: {})
+  end
+
+  def set_policy_class
+    @klass = @planning_application.policy_classes.find { |p| p.part == params[:part] && p.id == params[:policy_class] }
```
melzreal, line 59: I feel like @policy_class would be much easier to read in the views then this.. thoughts?

config/locales/references.yml
```diff
@@ -0,0 +1,894 @@
+---
```
melzreal, line 1: so we discussed the need for officers to amend this will come very soon - how can we have yaml policies if we want these to be editable by officers? The policies change every year, are we going to create a dependency on developers having to maintain a 900 line YAML file instead of seeding these into the DB and having them be editable?

I might not understand enough though - is there a way for us to build an interface to edit things that are put in using yaml? I really, really don't want to spend hours of "development" amending text so just making sure we aren't creating that burden for future maintainers...

freesteph: YAML is best for that sort of verbose copy, at least for the time being since we don't want to build a fully-fledged system to copy the legislation. What alternatives are you thinking of?

pixeltrix: What about storing them in the database but have a nightly job to pull the latest changes from the API? That way you avoid having to build an interface.

freesteph: not keen on storing all this copy in the DB + Tom G wanted to "try out" this feature and see if it works with officers so I'd rather we stick to this plain solution now and worry about how to take it next-level once it's proven.

freesteph: I'll check with the product owners whether they're happy with the exact copy of the API (well, the XML resources). I seem to remember they wanted to use a slightly simpler copy but I'll double check.

spec/system/planning_applications/policy_assessment_spec.rb
```diff
@@ -0,0 +1,104 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Planning Application Assessment", type: :system do
+  let!(:assessor) { create :user, :assessor, local_authority: @default_local_authority }
```
Rhian, line 6: I've checked this works equivalently for the reviewer (Manager) role. Not sure if we need to add some extra specs to validate this or whether it's just going to be annoying code duplication

app/controllers/policy_assessments_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def part; end
+
+  def new
+    @part = params[:part]
+
+    redirect_to part_new_policy_assessment_path(@planning_application), alert: "Please choose one of the policy parts" if @part.blank?
+  end
+
+  def create
```
Rhian, line 13: Do we need to restrict `create` to planning applications that are `in_assessment` only?

melzreal:
typo here: 
image

melzreal:
we capitalize all the other warnings, so I'd capitalize this one too: 
image

app/views/policy_assessments/show_class.html.erb
freesteph: Thanks

app/controllers/policy_assessments_controller.rb
freesteph: yes maybe, I'll give it a try!

config/locales/references.yml
melzreal: could you help me understand why the YAML is better than adding text to the database so it can be edited later? I was trying to find some articles on YAML rendering vs retrieving the data from a db but can't find anything good, would really like to understand the logic behind that decision 🙏

freesteph: YAML is incredibly easy to read and maintain, it's also excellent for text-heavy, verbose content which is exactly what this is. I assume that's why Rails defaults to a bunch of YAML files in `config/locales/` for its translations. Those legal texts are not data, they don't need to be queried or joined (not at this point anyway) so they don't belong in the database. Building an admin interface might be as easy as dropping a gem (`activeadmin` springs to mind) but I don't want to introduce that complexity for a feature that we are only trialling at the moment. YAML just works, and works great for this usecase I think.

If the feature is approved, and the update process turns out to be a pain we can move things in the DB. This is good enough a start.

app/controllers/policy_assessments_controller.rb
freesteph: Yes that's a good shout

melzreal:
when an application has moved past "assess proposal" and no policies have been added, we're displaying the title with nothing - probably best to add a small message there like "No policy assessment added" or remove the header altogether, maybe one to ask Tom about: 
image

config/locales/references.yml
melzreal: gotcha, database entries should be for stuff we want to query and join - thanks for explaining! 
If they decide they do want this feature in the long run its likely they'll want us to display which policy classes were approved and have those display on the decision notice - but we'll cross that bridge when we come to it

app/controllers/policy_classes_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  before_action :set_planning_application
+  before_action :set_policy_class, only: %i[show update destroy]
+
+  def part; end
+
+  def new
+    @part = params[:part]
+
+    redirect_to part_new_policy_assessment_path(@planning_application), alert: "Please choose one of the policy parts" if @part.blank?
+  end
```
melzreal, line 11: was trying to test this error message but after clicking "Continue" without selecting a part I got this error -:
image

image

melzreal: I gather the error should be in create instead?

freesteph: I forgot to update the route, thanks!

app/controllers/policy_classes_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    classes = PolicyClass
+                .classes_for_part(params[:part])
+                .select { |c| class_ids.include? c.id }
+                .each(&:stamp_status!)
+
+    @planning_application.policy_classes += classes
+
+    if @planning_application.save
+      redirect_to @planning_application, notice: "Policy classes have been successfully added"
```
melzreal, line 24: we need an error here as well if the policy_classes are empty - currently when leaving all the boxes unselected and pressing "Add classes" the success message is displayed as we aren't forcing the user to choose any:

image

freesteph: Cheers, will do. Also, no need for screenshots every time :smile:

melzreal: sorry about the endless screenshots its a habit from doing tech documentation but I realise totally unnecessary for this 😆

Rhian (approved):
Looks great!

melzreal (approved):
Looks great ! :shipit: :shipit:

PR #408, opened 2021-09-22
Modify validation requests index messages

melzreal:
- This ensures we hide the validation requests table when an application has been marked valid and no validation requests were ever opened on it.
- This also makes sure we display the full message of "There are x open and x closed validation requests" anytime there are validation requests and the application is invalid, inline with the required design.

### Description of change
https://trello.com/c/8ngUjY0t/497-validation-flow-in-staging-copy-doesnt-make-sense

### DIFF
--- a/app/views/planning_applications/validate_form.html.erb
+++ b/app/views/planning_applications/validate_form.html.erb
@@ -41,21 +41,21 @@
     <p class="govuk-body">
       <%= link_to "View notification", validation_notice_planning_application_path(@planning_application), class: "govuk-link" %>
     </p>
+  <% else %>
+    <h2 class="govuk-heading-m">
+      Validation requests
+    </h2>
+    <p class="govuk-body govuk-!-margin-top-2">
+      <%= validation_request_summary(@planning_application.validation_requests, @planning_application) %>
+    </p>
+    <p class="govuk-body govuk-!-padding-bottom-6">
+      <% if @planning_application.can_invalidate? %>
+        <%= link_to "Start new or view existing validation requests", planning_application_validation_requests_path(@planning_application), class: "govuk-link" %>
+      <% else %>
+        <%= link_to "View requests", planning_application_validation_requests_path(@planning_application), class: "govuk-link" %>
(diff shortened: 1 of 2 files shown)


app/views/planning_applications/validate_form.html.erb
```diff
@@ -41,22 +41,22 @@
     <p class="govuk-body">
       <%= link_to "View notification", validation_notice_planning_application_path(@planning_application), class: "govuk-link" %>
     </p>
+  <% else %>
+    <h2 class="govuk-heading-m">
+      Validation requests
+    </h2>
+    <p class="govuk-body govuk-!-margin-top-2">
```
freesteph, line 48: why are you using a custom margin?

app/views/planning_applications/validate_form.html.erb
```diff
@@ -41,22 +41,22 @@
...
+      <%= validation_request_summary(@planning_application.validation_requests, @planning_application) %>
+    </p>
+    <p class="govuk-body govuk-!-padding-bottom-6">
```
freesteph, line 51: same here, it's best if we can stick to GOV.UK's default spacing

freesteph: Looking at the diff it was here before but I'd vote to remove the custom spacing.

freesteph (approved):
Sorry meant to approve this :+1:

app/views/planning_applications/validate_form.html.erb
melzreal: It was likely put in to match whatever design was given to us at a point in time - given the gov uk framework gives us the benefit of bootstrap instead of having to create custom classes for a little bit of padding, it makes sense to use the gov uk classes to create it as its just taking advantage of the framework

That said styling isn't consistent at all in this project so I haven't been too dear about it - you'll notice we have style tags sprinkled across some views, then use the gov uk styling classes in some others like this one, and plain old css classes in others. Bringing consistency to it.. its a bigger conversation / piece for sure

PR #412, opened 2021-09-24
Use business days instead of current date

melzreal:
### Description of change

image

### Story Link
https://trello.com/c/PvjXfcds/453-change-calculations-to-working-days-rather-than-calendar-days-on-notifications

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -12,7 +12,6 @@ AllCops:
   Exclude:
     - "db/migrate/*"
     - "lib/package_builder.rb"
-    - "config/initializers/business_time.rb"
 
 Lint/UnusedMethodArgument:
  Exclude:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -12,6 +12,7 @@ AllCops:
   Exclude:
     - "db/migrate/*"
     - "lib/package_builder.rb"
+    - "config/initializers/business_time.rb"
 
 Lint/UnusedMethodArgument:
  Exclude:
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -106,7 +106,7 @@ class PlanningApplication < ApplicationRecord
   end
 
   def days_left
-    (expiry_date - Date.current).to_i
+    Date.current.business_days_until(expiry_date)
   end
 
   def reference
@@ -341,8 +341,8 @@ class PlanningApplication < ApplicationRecord
 private
 
   def set_key_dates
-    self.expiry_date = (documents_validated_at || created_at) + 8.weeks
-    self.target_date = (documents_validated_at || created_at) + 7.weeks
+    self.expiry_date = 40.business_days.after(documents_validated_at || created_at)
+    self.target_date = 35.business_days.after(documents_validated_at || created_at)
   end
 
   def set_change_access_id
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -106,7 +106,9 @@ class PlanningApplication < ApplicationRecord
   end
 
   def days_left
-    Date.current.business_days_until(expiry_date)
+    days_left = Date.current.business_days_until(expiry_date)
+
+    days_left.positive? ? days_left : -expiry_date.business_days_until(Date.current)
   end
 
   def reference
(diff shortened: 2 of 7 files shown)


.rubocop.yml
```diff
@@ -12,6 +12,7 @@ AllCops:
   Exclude:
     - "db/migrate/*"
     - "lib/package_builder.rb"
+    - "config/initializers/business_time.rb"
```
freesteph, line 15: what are we ignoring here?

app/models/planning_application.rb
```diff
@@ -106,7 +106,9 @@ def timestamp_status_change
   end
 
   def days_left
-    (expiry_date - Date.current).to_i
+    days_left = Date.current.business_days_until(expiry_date)
+
+    days_left.positive? ? days_left : -expiry_date.business_days_until(Date.current)
```
freesteph, line 111: what's `-expiry_date.business_days_until(Date.current)`? Not familiar with reversing dates

config/initializers/business_time.rb
```diff
@@ -0,0 +1,6 @@
+BusinessTime::Config.load("#{Rails.root}/config/business_time.yml")
+
+# or you can configure it manually:  look at me!  I'm Tim Ferriss!
```
freesteph, line 3: we probably don't need these comments

freesteph (commented):
Looking good but have a couple q's

config/initializers/business_time.rb
melzreal: the comments are so people in the future now they can either amend the config using the yaml file or doing it manually, we have other config files that are nothing but comments so I thought it made sense to keep them?

app/models/planning_application.rb
melzreal: we define expiry_date as:
```
 self.expiry_date = 40.business_days.after(documents_validated_at || created_at)
 ```

so this checks how many days between the expiry date and the current date have passed, then show those days in negative because (see screenshot/ + comments for more detail) we want to inform officers of how many days past due the application is

File: .rubocop.yml (hunk 1 above)
Comment (melzreal) [reply .rubocop.yml:15]: rubocop complains that 
```
BusinessTime::Config.load("#{Rails.root}/config/business_time.yml")
```
should instead use the Rails.root.join method. But I tried that and the whole thing went crazy because the gem could no longer find the path. So I ignored it instead

Comment (melzreal) [reply .rubocop.yml:15]:
``` 
Offenses:

config/initializers/business_time.rb:1:27: C: Rails/FilePath: Prefer Rails.root.join('path/to').
BusinessTime::Config.load("#{Rails.root}/config/business_time.yml")
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Comment (melzreal) [reply .rubocop.yml:15]: when I tried changing it to:
```
BusinessTime::Config.load("#{Rails.root.join('/config/business_time.yml')}")
```
or to
```
BusinessTime::Config.load(Rails.root.join('/config/business_time.yml'))
```
I get this error when running any spec:
```
mreal $ rspec spec/models/planning_application_spec.rb

An error occurred while loading ./spec/models/planning_application_spec.rb.
Failure/Error: require File.expand_path("../config/environment", __dir__)

Errno::ENOENT:
  No such file or directory @ rb_sysopen - /config/business_time.yml
# /Users/mreal/.rvm/gems/ruby-2.6.6/gems/business_time-0.10.0/lib/business_time/config.rb:161:in `initialize'
```

Comment (melzreal) [reply .rubocop.yml:15]: ah I got it now, had an extra slash. changing...

**Review by Rhian — APPROVED** [review]
Have left one small comment/question but not a blocker. Otherwise looks good!

File: config/business_time.yml [hunk 4]
```diff
@@ -0,0 +1,30 @@
+business_time:
```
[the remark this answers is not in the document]
melzreal: there is this one - https://github.com/alphagov/calendars - but that would requiring adding another dependency to the project - bank holidays are only released a couple of years in advance so even the 2023 calendar is not yet 100% confirmed by the gov. I'll pop a ticket into the backlog to look at integrating the api as a separate ticket so we can discuss if its worth the effort / extra dependency

PR #415, opened 2021-09-28
Change validation flow UX

melzreal:
### Description of change
1. 
image

2. 
image

3. 
image

4. 
image

5. 
image

6.
image

7. 
image

8. 
image

9.
image

### Story Link
https://trello.com/c/xxEqQk1K/486-improve-ux-of-the-validation-flow


app/controllers/planning_applications_controller.rb
```diff
@@ -120,7 +123,13 @@ def invalidate
       @planning_application.invalidate!
       audit("invalidated")
       invalidation_notice_mail
-      @planning_application.unsent_validation_requests.each { |request| request.update!(notified_at: Time.zone.now) }
+      request_names = []
+      @planning_application.unsent_validation_requests.each do |request|
```
freesteph, line 127: can you make this a callback on the model's state transition instead? Something a bit like:

```diff
modified   app/models/planning_application.rb
@@ -64,7 +64,10 @@ class PlanningApplication < ApplicationRecord
     end
 
     event :invalidate do
-      transitions from: %i[not_started invalidated], to: :invalidated, guard: :validation_requests_open?
+      transitions from: %i[not_started invalidated],
+                  to: :invalidated,
+                  guard: :validation_requests_open?,
+                  after: :send_validation_requests
     end
 
     event :determine do
@@ -104,6 +107,14 @@ class PlanningApplication < ApplicationRecord
     "#{applicant_first_name} #{applicant_last_name}"
   end
 
+  def send_validation_requests
+    unsent_validation_requests.each do |request|
+      request.update!(notified_at: Time.zone.now)
+
+      audit("validation_request_sent", nil, request)
+    end
+  end
+
   def timestamp_status_change
     update("#{aasm.to_state}_at": Time.zone.now)
   end
```

app/helpers/audit_helper.rb
```diff
@@ -60,6 +68,8 @@ def activity(type_of_activity, args = nil)
       "Sent: validation request (other validation##{args})"
     when "other_change_validation_request_received"
       "Received: request for change (other validation##{args})"
+    when "validation_requests_sent"
```
freesteph, line 71: can't we use an event per request and use the already present:

```
./audit_helper.rb:43:    when "description_change_validation_request_sent"
./audit_helper.rb:45:    when "replacement_document_validation_request_sent"
./audit_helper.rb:47:    when "additional_document_validation_request_sent"
./audit_helper.rb:49:    when "red_line_boundary_change_validation_request_sent"
./audit_helper.rb:67:    when "other_change_validation_request_sent"
```

?

freesteph (commented):
Still trying to wrap my head around it but some early questions

app/views/documents/index.html.erb
```diff
@@ -50,8 +50,10 @@
       <% filter_current(@documents).each do |document| %>
         <tr class="govuk-table__row">
           <td class="govuk-table__cell govuk-!-width-one-quarter">
+            <% if document.file.representable? %>
```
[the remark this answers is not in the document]
melzreal: it is in on the index page, but not on the documents so I was still getting that error here - its out of scope for this pr though so I'll take it out, thanks for the catch

app/helpers/audit_helper.rb
melzreal: this is so we only add one audit entry for all the requests that were sent at the same time - that's why the method creates an array outside of the loop so we don't repeat the exact same audit entry, instead just ending up with:
image

melzreal: the code is a bit ugly but it does achieve this purpose - I'm sure it could be refactored somehow this way we avoid creating audit entries unnecessarily when we can (each audit is a db entry as well) - plus we capture each change at the time of controller interaction, so we'd need some extra logic to grab the past audit action and pop it in as an audit argument... can of worms really

app/controllers/planning_applications_controller.rb
melzreal: I can try for sure! Will give it a spin - Silly question but, we are ok to call mailer methods from the model? Will it know the request.host still? since we redefine on the controller this method:
``` 
def invalidation_notice_mail
    PlanningApplicationMailer.invalidation_notice_mail(
      @planning_application,
      request.host,
    ).deliver_now
  end
```

freesteph: good question; the mailing should remain in the controller I think, it's in charge of orchestrating that stuff: invalidate application + mail about it. The whole "when an application is invalidated all requests should be sent out" belongs to the model. I was actually in Show&Tell prep yesterday and didn't realise that was how the flow worked, so I think a

```ruby
    event :invalidate do
      transitions from: %i[not_started invalidated],
                  to: :invalidated,
                  guard: :validation_requests_open?,
                  after: :send_all_pending_validation_requests # or before, not sure
    end
```

will help a lot in that respect.

app/helpers/audit_helper.rb
freesteph: yes OK. I thought it'd be nicer to have each one separately, it's finer-grained which makes sense for the audit, but if the AC is to group them happy with it.

Rhian (approved):
LGTM.

PR #420, opened 2021-10-05
Feat/modern rubocop

freesteph:
* inherit the Rails/RSpec/Capybara config from govuk-rubocop;
* run an auto-fix on the codebase.


Rhian (approved):
Cool. I definitely learned a lot from reviewing this :)

PR #422, opened 2021-10-06
deploy: fix the package builder

freesteph:
This script isn't loading Rails so `Time.zone` is not a thing, as
Rubocop wrongly assumed.


Rhian (approved):
Looks good.

PR #423, opened 2021-10-06
ci: attempt to modernise CI

freesteph:
- don't reinvent gems caching;
- don't reinstall everything;
- run all tools/suites separately.


Rhian (approved):
Looks good!

PR #427, opened 2021-10-11
Leverage the validation requests state machine

freesteph:
General code cleanups, but mainly:

* default validation requests as "pending";
* try and use state events (`mark_as_sent!`);
* leverage `may_<state>?` on the planning application;
* some layout cleanups;
* some feature tests for auditing and validation requests.


features/managing_validation_requests.feature
```diff
@@ -8,14 +8,29 @@ Feature: Managing validation requests
...
     When I press "Invalidate"
     And I view the application's validations requests
-    Then there is a new document request for a "Picture of the dog" that shows "15 days"
-    Then there is a new document request for a "Picture of the cat" that shows "15 days"
+    Then there is a validation request for a "Picture of the dog" that shows "15 days"
+    Then there is a validation request for a "Picture of the cat" that shows "15 days"
+
+  Scenario: As an assessor any request past invalidation is sent immediately
+    Given the planning application is invalidated
+    Then the page contains "Assigned to: "
```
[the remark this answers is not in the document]
freesteph: Sorry yeah that's a leftover from a time where the page was not the one I wanted :smile:

app/controllers/planning_applications_controller.rb
```diff
@@ -124,23 +124,18 @@ def validate
...
       audit("invalidated")
 
       invalidation_notice_mail
 
-      # FIXME: move this into the state machine
-      request_names = @planning_application.unsent_validation_requests.map do |request|
-        request.update!(notified_at: Time.zone.now)
-        request.audit_name
-      end
+      request_names = @planning_application.pending_validation_requests.map(&:audit_name)
```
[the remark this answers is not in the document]
freesteph: which goes to https://github.com/unboxed/bops/blob/fix/pending-validation-requests/app/models/validation_request.rb#L20

app/models/validation_request.rb
```diff
@@ -9,8 +9,17 @@ module ValidationRequest
     aasm.attribute_name :state
 
     aasm do
-      state :open, initial: true
+      state :pending, initial: true
+      state :open
       state :closed
+
+      event :mark_as_sent! do
+        transitions from: :pending, to: :open
```
melzreal, line 17: so each validation request goes from a default event of pending to open... but could you help me understand where we are updating the requests to closed using the state machine? is it the helpers?

app/views/validation_requests/index.html.erb
```diff
@@ -133,23 +67,23 @@
         <p class="govuk-body">
           Send further requests for this application. Any new requests will be immediately sent.
         </p>
-      <% elsif @planning_application.can_invalidate? %>
+      <% elsif @planning_application.not_started? %>
```
melzreal, line 70: are we sure this is the one we want here? I remember the can_invalidate being necessary for some reason.. will UAT in the front end

app/models/validation_request.rb
freesteph: so yeah I guess I forgot this bit... I wasn't sure where we do close them. The only trace I can find is:

```

Comment (melzreal) [reply app/models/validation_request.rb:17]: yes indeed - validation_requests already had a state of "closed" that gets set whenever a reply from the customer is hit through the api - so we could tie the state machine transition to that I suppose

File: db/migrate/20211009130553_remove_default_value_for_validation_request_states.rb [hunk 5]
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class RemoveDefaultValueForValidationRequestStates < ActiveRecord::Migration[6.1]
+  def change
+    change_column_default(:description_change_validation_requests, :state, from: "pending", to: nil)
+    change_column_default(:additional_document_validation_requests, :state, from: "pending", to: nil)
```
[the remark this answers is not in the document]
freesteph: Ha, I spent some time removing the changing migration this morning only to cancel everything an hour later. They're both required:

1. we want to update our current database to move "open" requests to "pending";
2. once that's done, we just remove the default value on the column.

So we're actually updating real data, then removing a column default, they're not exactly the same but it is a bit confusing.

app/views/validation_requests/index.html.erb
melzreal: looks great actually :D had a full on uat, as far as I can see its good to go!

melzreal (approved):
LGTM great code cleanup ✨  :shipit:

PR #431, opened 2021-10-12
ruby: move to Ruby 2.7.4 and Bundler 2.2.29

freesteph:
Also some minor updates in the Gemfile.


.github/workflows/production.yml
```diff
@@ -12,10 +12,10 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
-    - name: Setup ruby
+    - name: Setup Ruby
```
melzreal, line 15: lol "Ruby, ladies and gents, deserves some uppercasing over here"

melzreal:
closing as we've updated these separately I believe. Otherwise we'll open another pr.

PR #438, opened 2021-10-13
Change "Send" to "Add" when adding a change request to an application that has not yet been invalidated

benbaumann95:
### Description of change

Change "Send" to "Add" when adding a change request to an application that has not yet been invalidated

- Move action buttons "Add" and "Back" into its own partial as this is being used multiple times. Easier to stay consistent with just needing a change in one place compared to all files.

### Story Link

https://trello.com/c/umrf5O7j/588-change-send-to-add-whenever-adding-a-change-request-to-an-application-that-has-not-yet-been-invalidated


app/views/shared/_validation_request_form_actions.html.erb
```diff
@@ -0,0 +1,2 @@
+<%= form.govuk_submit "Add" %>
```
freesteph, line 1: can you make this a [button group](https://design-system.service.gov.uk/components/button/#grouping-buttons)

freesteph (approved):
Looking good

PR #442, opened 2021-10-14
Change copy and format decision notice

melzreal:
### Description of change
https://trello.com/c/YbXQsATF/595-minor-text-change-to-decision-notice

### DIFF
--- a/app/views/planning_applications/_decision_notice.html.erb
+++ b/app/views/planning_applications/_decision_notice.html.erb
@@ -20,88 +20,91 @@
         "Application received" => planning_application.created_at.to_formatted_s(:day_month_year),
         "Decision date" => planning_application.determined_at&.to_formatted_s(:day_month_year) || "TBD",
         "Site address" => planning_application.full_address,
-        "Use/development" => description_including_status(planning_application),
       }.each do |key, value|
     %>
-  <div class="govuk-summary-list__row">
-        <dt class="govuk-summary-list__key">
-          <%= key %>
-        </dt>
-        <dd class="govuk-summary-list__value">
-          <%= value %>
-        </dd>
-      </div>
-    <% end %>
-  </dl>
-  <p class="govuk-body">
-      We certify that on the date of the application, the <%= @planning_application.work_status %> use or operations described in the application and supporting plans were <%= @planning_application.granted? ? "lawful" : "not lawful" %> for the purposes of <%= @planning_application.work_status == "proposed" ? "S.192" : "S.191" %> of the Town and Country Planning Act 1990.
-  </p>
-  <p class="govuk-body govuk-!-margin-bottom-1">
-    <strong>
-      The proposal <%= @planning_application.granted? ? "complies" : "does not comply" %> due to the following reason(s):
-    </strong>
-  </p>
-  <p class="govuk-body">
-      <%= planning_application.public_comment %>
-  </p>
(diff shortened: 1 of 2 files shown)


app/views/planning_applications/_decision_notice.html.erb
```diff
@@ -20,88 +20,90 @@
...
+</dl>
+<p class="govuk-body">
+  <strong>
+    Use/development
+  </strong>
+</p>
+<p class="govuk-body">
+  We certify that on the date of the application, the <%= @planning_application.work_status %> use or operations described in the application and supporting plans were <%= @planning_application.granted? ? "lawful" : "not lawful" %> for the purposes of <%= @planning_application.work_status == "proposed" ? "S.192" : "S.191" %> of the Town and Country Planning Act 1990.
+</p>
+<p class="govuk-body govuk-!-margin-bottom-1">
```
freesteph, line 43: let's try it without the custom classes and see if Tom's happy with it.

app/views/planning_applications/_decision_notice.html.erb
```diff
@@ -20,88 +20,90 @@
...
+  <strong>
+    The proposal <%= @planning_application.granted? ? "complies" : "does not comply" %> due to the following reason(s):
+  </strong>
+</p>
+<p class="govuk-body">
+  <%= planning_application.public_comment %>
+</p>
+<% if @planning_application.granted? %>
+  <h3 class="govuk-heading-s govuk-!-margin-bottom-1">
```
freesteph, line 52: same

app/views/planning_applications/_decision_notice.html.erb
```diff
@@ -20,88 +20,90 @@
...
+    This decision is based on the following approved plans:
   </h3>
-  <% if @planning_application.boundary_geojson.present? %>
-    <%= render "shared/location_map", locals: { geojson: @planning_application.boundary_geojson } %>
+  <% if planning_application.documents.for_display.present? %>
```
freesteph, line 55: I think `any?` is 0.0001% more natural but subjective preference

app/views/planning_applications/_decision_notice.html.erb
```diff
@@ -20,88 +20,90 @@
...
+    <table class="govuk-table">
```
freesteph, line 56: [tables should have captions](https://www.w3.org/WAI/WCAG21/Techniques/html/H39)

app/views/planning_applications/_decision_notice.html.erb
```diff
@@ -20,88 +20,90 @@
...
+      <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+          <th scope="col" class="govuk-table__header app-custom-class">Document reference</th>
+          <th scope="col" class="govuk-table__header app-custom-class">Description</th>
+          <th scope="col" class="govuk-table__header app-custom-class">Date received</th>
+        </tr>
+      </thead>
+      <tbody class="govuk-table__body">
+        <% @planning_application.documents.for_display.each do |document| %>
```
freesteph, line 65: can you make this an instance variable?

melzreal: have popped it into the planning application model instead. If this isn't good I feel we should just go back to what it was instead of changing 3 different controller methods and a setter just to save 2 words..

PR #443, opened 2021-10-19
Cancel validation requests

benbaumann95:
Cancel and delete validation change requests 

- Add cancel_reason and cancelled_at columns to validation request tables
- Only permit cancelling when an application has been made invalidated
- Only permit deleting when a request is still pending i.e. before an application is made invalidated
- Extract common logic across validation types into shared concern
- Send email to applicant and add to audit log when cancellation is made
- Add cancel confirmation summary page views for each of the validation request types

**Reviewers:**
- Quite a few changes here (sorry). Unconvinced for the reason we've created 5 separate models/tables for each of the request types but iterating on that is out of scope for this anyway
- Probably easiest to just pull down the branch yourself and test locally / view the different screens rather than me posting a bunch of screenshots here

Needs to be deployed with https://github.com/unboxed/bops-applicants/pull/43

https://trello.com/c/2uRDsTzD/538-cancel-validation-request

### DIFF
--- a/app/models/validation_request.rb
+++ b/app/models/validation_request.rb
@@ -3,21 +3,42 @@
 module ValidationRequest
   extend ActiveSupport::Concern
 
+  class RecordCancelError < RuntimeError; end
+
+  class NotDestroyableError < StandardError; end
+
+  class CancelledEmailError < StandardError; end
+
   included do
+    before_create :set_sequence
+
+    before_destroy :ensure_validation_request_destroyable!
+
+    validates :cancel_reason, presence: true, if: :cancelled?
+
     include AASM
 
     aasm.attribute_name :state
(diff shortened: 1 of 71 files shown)


app/models/validation_request.rb
```diff
@@ -3,15 +3,20 @@
...
+
+  class CancelledEmailError < StandardError; end
+
   included do
     include AASM
 
     aasm.attribute_name :state
 
-    aasm do
+    aasm whiny_persistence: true do
```
benbaumann95, line 15: https://www.rubydoc.info/github/aasm/aasm/AASM%2FConfiguration:whiny_persistence

Needed to raise error on bang! methods or it will just return false

PR #444, opened 2021-10-19
Remove description change

melzreal:
### Description of change

- Remove description change requests from validation flow
- Amend flow to immediately send description change request emails as soon as action is performed, regardless of planning application state
- Add validation to prevent more than one request in an open state each time
- Add cucumber specs to description change request flow
- Add cancelation route for description change requests
- Amend audit wording to match actions
- Description request cannot be made after assessment has been submitted
- Test new mailer methods for auto update
- Test notification banner for auto-close

TODO
- Setup cron job on instance to call on rake task

### Story Link
https://trello.com/c/at1G8s6E/561-remove-description-change-as-a-validation-request-and-instead-allow-proposals-for-description-change-outside-of-the-validation-r


app/controllers/description_change_validation_requests_controller.rb
```diff
@@ -1,37 +1,44 @@
...
-              @description_change_validation_request.sequence)
-      end
-      redirect_to planning_application_validation_requests_path(@planning_application)
-
+    if @description_change_request.save
+      email_and_timestamp(@description_change_request)
+      flash[:notice] = "Description change request successfully sent."
+      audit("description_change_validation_request_sent", description_audit_item(@description_change_request, @planning_application),
+            @description_change_request.sequence)
+      redirect_to planning_application_path(@planning_application)
```
freesteph, line 20:
```ruby
redirect_to path, notice: "flash notice"
```

will save you a line

app/controllers/description_change_validation_requests_controller.rb
```diff
@@ -1,37 +1,44 @@
...
+
+  def cancel
+    description_change_request.update!(state: "closed", approved: false, rejection_reason: "Request cancelled by planning officer.")
+    flash[:notice] = "Description change request successfully cancelled."
+    redirect_to planning_application_path(@planning_application)
+  end
+
   private
 
+  def description_change_request
```
freesteph, line 38: can you call this `set_description_change_request` and set the instance variable here?

app/models/planning_application.rb
```diff
@@ -279,7 +279,17 @@ def result_present?
...
+    (replacement_document_validation_requests + additional_document_validation_requests + other_change_validation_requests + red_line_boundary_change_validation_requests).sort_by(&:created_at).reverse
+  end
+
+  def open_description_change_requests
+    description_change_validation_requests.select(&:open?)
+  end
+
+  def rejected_description_change
+    description_change_validation_requests.select do |request|
+      request&.approved == false
```
freesteph, line 291: why would the request be nil?

freesteph: so here you're selecting the latest request that is not approved?

app/models/planning_application.rb
```diff
@@ -347,6 +357,12 @@ def documents_for_decision_notice
     documents.for_display
   end
 
+  def allows_only_one_open_description_change
+    if open_description_change_requests.size > 1
```
freesteph, line 361: didn't `any?` work here?

app/views/description_change_validation_requests/new.html.erb
```diff
@@ -1,50 +1,64 @@
-<%= render "validation_requests/validation_requests_breadcrumbs" %>
-
+<% add_parent_breadcrumb_link "Home", planning_applications_path %>
+<% add_parent_breadcrumb_link "Application", validate_form_planning_application_path(@planning_application) %>
 <% content_for :title, "Description change" %>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h2 class="govuk-heading-l govuk-!-padding-top-6">
```
freesteph, line 6: yasssss

app/views/description_change_validation_requests/show.html.erb
```diff
@@ -0,0 +1,40 @@
...
+    <div>
+      <p class="govuk-body"> <strong> Previous description:</strong><br>
+        <%= @planning_application.description  %>
+      </p>
+      <p class="govuk-body"> <strong> Proposed description:</strong><br>
+        <%= @description_change_request.proposed_description  %>
+      </p>
+    </div>
+    <p class="govuk-body">
+      <% if @description_change_request.state == "open" %>
```
freesteph, line 25: anything `state` you should never update or compare directly; use the state machine for that with `.open?` and `.some_event!` to update the states (I know we still do it in the API controllers but we'll get to that at some point).

app/views/planning_applications/_form.html.erb
```diff
@@ -14,13 +15,22 @@
...
-    <%= form.text_area :description, class: "govuk-textarea", rows: "5" %>
+    <h2 class="govuk-heading-m">
+      Description
+    </h2>
+    <p class="govuk-body">
+      <%= @planning_application.description %>
+    </p>
+    <p class="govuk-body">
+      <% if @planning_application.open_description_change_requests.any? %>
+        <%= link_to "View requested change", planning_application_description_change_validation_request_path(@planning_application, @planning_application.open_description_change_requests.last.id) %>
```
freesteph, line 29: please always use `govuk-link` for our links

app/views/description_change_validation_requests/show.html.erb
```diff
@@ -0,0 +1,40 @@
...
+        Sent on <%= @description_change_request.created_at.strftime("%e %B %Y") %>. Agent or applicant has not yet responded.
+      <% else %>
+        <%= @description_change_request.approved? ? 'Approved' : 'Rejected' %>
+      <% end %>
+    </p>
+    <% if @description_change_request.rejection_reason.present? %>
```
freesteph, line 31: can you tuck this in a model method?

```ruby
def rejected?
  !approved
end
```

app/views/planning_applications/new.html.erb
```diff
@@ -7,6 +7,6 @@
     <h1 class="govuk-heading-l govuk-!-margin-top-7">
       New Application: Residential Lawful Development Certificate
     </h1>
-    <%= render 'form', planning_application: @planning_application %>
+    <%= render 'new_form', planning_application: @planning_application %>
```
freesteph, line 10: why not update `_form.html.erb`?

features/adding_description_change.feature
```diff
@@ -0,0 +1,49 @@
...
+  
+  Scenario: I can add and view a new description change request after cancelling the previous one
+    Given I create a description change request from the application edit page with "Its margarita time ole!"
+    Then the page contains "Description change request successfully sent."
+    When I press "Application information"
+    And I press "View requested change"
+    Then the page contains "Previous description"
+    Then the page contains "Add a statue of Melissa striking poses"
+    Then the page contains "Proposed description"
+    Then the page contains "Its margarita time ole!"
```
freesteph, line 27: :pizza: :woman_dancing:

features/step_definitions/description_change_steps.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+require "faker"
+
+Given("a new planning application with a description of {string}") do |string|
```
freesteph, line 5: Can you try and make these override `@planning_application` instead? The idea is that during the span of an example (scenario, spec, whatever you want to call it), we refer to the same instance variable throughout our steps, it's effectively part of our state.

Something akin to:

```
Given a new planning application # step already exists
And the planning application has a description of "Foobar"
```

features/step_definitions/description_change_steps.rb
```diff
@@ -0,0 +1,75 @@
...
+    And I fill in "Please suggest a new application description" with "#{details}"
+    And I press "Add"
+  )
+end
+
+When("I visit the new description change request link") do
+  visit new_planning_application_description_change_validation_request_path(@planning_application_with_description)
+end
+
+When("I view the planning application with description") do
```
freesteph, line 64: if you reuse `@application` you don't need this step

features/step_definitions/description_change_steps.rb
```diff
@@ -0,0 +1,75 @@
...
+  visit planning_application_path(@planning_application_with_description)
+end
+
+When("I view the determined application") do
```
freesteph, line 68: same

features/step_definitions/description_change_steps.rb
```diff
@@ -0,0 +1,75 @@
...
+Given("a new planning application with a description of {string}") do |string|
+  @planning_application_with_description = FactoryBot.create(
+    :planning_application,
+    :not_started,
+    description: string,
+    local_authority: @officer.local_authority
+  )
+end
+
+Given("an existing description change request") do
```
freesteph, line 14: What's the difference with the user-simulated (hence, better) ` Given("I create a description change request from the application index with {string}") do |details|`?

app/models/planning_application.rb
melzreal: yes, I'm selecting the last request that was not approved because we want to display rejected description changes on the "new" page whenever a new request is created - so if an applicant has rejected a description change or a request was cancelled, that will be visible on the new page

app/models/planning_application.rb
melzreal: ah, this one is a leftover - we don't need it anymore since we've moved it - thanks, nice catch

features/step_definitions/description_change_steps.rb
freesteph: no, all tests start from an empty state and setup things through `Background:` (which is your equivalent of `before each`) and their own steps.

app/controllers/validation_requests_controller.rb
```diff
@@ -38,8 +36,19 @@ def send_validation_request_email(request)
...
+  def send_description_request_email(request)
+    PlanningApplicationMailer.description_change_mail(
+      @planning_application,
+      request,
+    ).deliver_now
+  end
+
   def email_and_timestamp(request)
-    send_validation_request_email(request)
+    if request.class.name == "DescriptionChangeValidationRequest"
```
freesteph, line 47: prefer `object.is_a? ClassName`

Guardfile
```diff
@@ -75,7 +75,7 @@ cucumber_options = {
   # Below are examples overriding defaults
 
   # cmd: 'bin/cucumber',
-  cmd_additional_args: "--publish-quiet --fail-fast",
+  cmd_additional_args: "--publish-quiet --fail-fast --profile wip",
```
freesteph, line 78: We don't want to check this in

app/jobs/close_description_change_job.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def perform
+    description_change_requests.each do |change_request|
+      auto_update_and_notify(change_request)
+    end
+  end
+
+  def auto_update_and_notify(description_change_request)
+    description_change_request.planning_application.update!(description: description_change_request.proposed_description)
+    description_change_request.update!(state: "closed", approved: true)
```
freesteph, line 18: We don't want to update the state manually, that's what we hired `aasm` for. Instead I suggest we reopen AASM in `DescriptionChangeValidationRequest` and do something like:

```ruby

  aasm do
    event :approve! do
      after do |description|
        update!(proposed_description: description)
      end

      transitions from: :open, to: :closed
    end

    event :reject! do
      after do
        update!(rejection_reason: I18n.t("some localised message"))
      end

      transitions from: %i[pending open], to: :closed
    end
  end
  ```

File: app/jobs/close_description_change_job.rb [hunk 18]
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class CloseDescriptionJob < ApplicationJob
+  queue_as :default
+
+  def self.description_change_requests
+    DescriptionChangeValidationRequest.open_change_created_over_5_days_ago
```
freesteph, line 7: nice, but there is already part of the scope available, see actual line

app/jobs/close_description_change_job.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+  end
+
+  def auto_update_and_notify(description_change_request)
+    description_change_request.planning_application.update!(description: description_change_request.proposed_description)
+    description_change_request.update!(state: "closed", approved: true)
+    email_description_closure_notification(description_change_request)
+  end
+
+  def email_description_closure_notification(request)
```
freesteph, line 22: don't think you need all these separate methods:

```ruby
request.each do |request|
  request.auto_approve! # whatever that might be

  PlanningApplicationMailer.do_the_mailing
```

app/mailers/planning_application_mailer.rb
```diff
@@ -57,13 +56,40 @@ def receipt_notice_mail(planning_application, host, user)
...
+    view_mail(
+      NOTIFY_TEMPLATE_ID,
+      subject: "Your planning application at: #{@planning_application.full_address}",
+      to: application_accountable_email(@planning_application),
+      reply_to_id: @planning_application.local_authority.reply_to_notify_id
+    )
+  end
+
+  def application_accountable_email(planning_application)
+    planning_application.applicant_and_agent_email.first
```
freesteph, line 93: feels like this should live on the model

app/models/description_change_validation_request.rb
```diff
@@ -12,6 +12,10 @@ class DescriptionChangeValidationRequest < ApplicationRecord
 
   validates :proposed_description, presence: true
   validate :rejected_reason_is_present?
+  validate :allows_only_one_open_description_change, on: :create
+  validate :planning_application_has_not_been_determined, on: :create
+
+  scope :open_change_created_over_5_days_ago, -> { where("state = 'open' AND created_at > ?", 5.business_days.ago) }
```
freesteph, line 18: you get a free `open` scope with AASM:

```ruby
scope :autoaccept, -> { open.where("created_at >= ? ", 5.business.days.ago) }
```

is it greater or greater or equal? always get confused about these things

app/models/planning_application.rb
```diff
@@ -279,7 +279,23 @@ def result_present?
   end
 
   def validation_requests
-    (description_change_validation_requests + replacement_document_validation_requests + additional_document_validation_requests + other_change_validation_requests + red_line_boundary_change_validation_requests).sort_by(&:created_at).reverse
+    (replacement_document_validation_requests + additional_document_validation_requests + other_change_validation_requests + red_line_boundary_change_validation_requests).sort_by(&:created_at).reverse
+  end
+
+  def open_description_change_requests
+    description_change_validation_requests.select(&:open?)
```
freesteph, line 286: don't think you need this: `description_change_validation_requests.open` does it

app/models/planning_application.rb
```diff
@@ -279,7 +279,23 @@ def result_present?
...
+  end
+
+  def responded_description_requests
+    description_change_validation_requests.select do |request|
+      request.state == "closed" && request.rejection_reason != "Request cancelled by planning officer."
```
freesteph, line 291: this should be a scope on DescriptionChangeRequest

app/models/planning_application.rb
```diff
@@ -279,7 +279,23 @@ def result_present?
...
+    end.last
+  end
+
+  def rejected_description_change
+    description_change_validation_requests.select do |request|
+      request.approved == false
```
freesteph, line 297: another  scope

features/adding_description_change.feature
```diff
@@ -0,0 +1,47 @@
...
+    And the planning application has a description of "Add a statue of Melissa striking poses"
+    And I view the planning application
+  
+  Scenario: I can add a new description change request
+    Given I create a description change request with "Add a backyard cinema"
+    Then the page contains "Description change request successfully sent."
+
+  Scenario: I can cancel a description change request
+    Given I create a description change request with "Add a backyard cinema"
+    Given I cancel the existing description change request
```
freesteph, line 14: Given should be present once

features/adding_description_change.feature
```diff
@@ -0,0 +1,47 @@
...
+    Given I create a description change request with "A yard full of bananas"
+    When I visit the new description change request link
+    And I fill in "Please suggest a new application description" with "Mambo number 2"
+    And I press "Add"
+    Then the page contains "An open description change already exists for this planning application."
+
+  Scenario: I cannot create a description change request on a determined planning application
+    Given a determined planning application
+    When I view the determined application
+    When I press "Application information"
```
freesteph, line 43: And

app/models/description_change_validation_request.rb
melzreal: went with > to give it a little more time since the business day is defined as a 9 to 5 it can be a little strict - I'm not sure I understand this autoaccept scope comment ?

app/jobs/close_description_change_job.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class CloseDescriptionChangeJob < ApplicationJob
```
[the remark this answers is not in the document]
benbaumann95: It's more about the code failing somehow rather than the job not executing but i think it's ok for now

app/mailers/planning_application_mailer.rb
```diff
@@ -32,12 +32,11 @@ def validation_notice_mail(planning_application, host, user)
   def invalidation_notice_mail(planning_application, host)
     @host = host
     @planning_application = planning_application
-    @application_accountable_email = @planning_application.applicant_and_agent_email.first
```
benbaumann95, line 35: why remove this ?

app/models/planning_application.rb
```diff
@@ -279,13 +279,31 @@ def result_present?
...
+
+  def responded_description_request
+    description_change_validation_requests.closed.select do |request|
+      request.rejection_reason != "Request cancelled by planning officer." && request.auto_closed.blank?
+    end.last
+  end
+
+  def auto_closed_description_requests
+    description_change_validation_requests.select(&:auto_closed?).last
+  end
```
benbaumann95, line 301: method name is misleading in that it expects multiple requests where here you are just pulling out the last record?
`latest_auto_closed_description_request` ?

app/models/planning_application.rb
```diff
@@ -279,13 +279,31 @@ def result_present?
...
+
+  def rejected_description_change
+    description_change_validation_requests.select(&:rejected?).last
+  end
```
benbaumann95, line 305: same here as above

app/models/planning_application.rb
```diff
@@ -279,13 +279,31 @@ def result_present?
...
     validation_requests.filter(&:cancelled?).sort_by(&:cancelled_at).reverse
   end
 
+  def open_description_change_requests
+    description_change_validation_requests.open
+  end
+
+  def responded_description_request
+    description_change_validation_requests.closed.select do |request|
+      request.rejection_reason != "Request cancelled by planning officer." && request.auto_closed.blank?
```
benbaumann95, line 295: why do you need this check `request.rejection_reason != "Request cancelled by planning officer."` ?
Can the request be cancelled another way ? If not then you can just check if `cancelled_at` is set.
And if `auto_closed` is a boolean why are we using `blank?` here ? Should it not be false

Then could we not do something like  `scope :responded, -> { closed.where(cancelled_at: nil, auto_closed: false) }` 

Then call it with `description_change_validation_requests.responded.last`

benbaumann95: Maybe for all these `last` calls we need to be sorting by order(created_at: :asc)

db/migrate/20211021152725_add_auto_update_to_description_change.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddAutoUpdateToDescriptionChange < ActiveRecord::Migration[6.1]
+  def change
+    add_column :description_change_validation_requests, :auto_closed, :boolean
```
benbaumann95, line 5: we need to set a default value of false here

benbaumann95 (commented):
Almost there!

I think we could also do with adding some mailer specs for the new email templates (unless i missed that)

app/mailers/planning_application_mailer.rb
melzreal: instead of calling this method repeatedly, I've replaced it directly on the view_mail method:
```
  to: @planning_application.applicant_and_agent_email.first,
  ```
  
  ```
    def invalidation_notice_mail(planning_application, host)
    @host = host
    @planning_application = planning_application

    view_mail(
      NOTIFY_TEMPLATE_ID,
      subject: "Your planning application is invalid",
      to: @planning_application.applicant_and_agent_email.first,
      reply_to_id: @planning_application.local_authority.reply_to_notify_id
    )
  end
  ```

app/models/planning_application.rb
melzreal: so this was before we had a "cancelled state" I was just checking the rejection reason since the state could be "closed" instead of cancelled - you're right, now that we have cancelled I can just check that instead, thanks!

app/views/planning_applications/new.html.erb
melzreal: have re-instated the old form and added a partial depending on whether the request is for "new"/"create" or "edit"

benbaumann95 (approved):

PR #446, opened 2021-10-20
planning application: simplify secure_change_url

freesteph:


app/models/planning_application.rb
```diff
@@ -258,11 +258,11 @@ def full_address
     "#{address_1}, #{town}, #{postcode}"
   end
 
-  def secure_change_url(application_id, secure_token)
+  def secure_change_url
```
benbaumann95, line 261: Good spot to remove these redundant arguments 👍

PR #459, opened 2021-10-29
Use promote to production style deploy strategy

benbaumann95:
### Description of change

Rather than having a separate branch for production, it will be easier to manage deployments through a manual approval process to promote to production.
I've configured a staging and production [environment](https://github.com/unboxed/bops/settings/environments) to our repo with some protection rules. They are only able to deploy from our master branch and deploying to production requires a review/approval from the team before the workflow can occur.


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,70 @@
+name: Deploy
+
+on:
+  push:
+    branches: [master]
```
melzreal, line 5: I'm guessing this means all jobs get queues on push to master? don't we need a push to branch prod as well? (I don't understand it well enough)

benbaumann95: The idea is that we aren't going to be using a separate branch for production. It will entirely be based off master

melzreal: ok so I guess once this is merged there will be a button somewhere to promote? well, will approve it then I trust you know :)

melzreal (approved):

benbaumann95:
Thinking I want to rework this to use the scottish petitions approach

https://github.com/ScottishParliament/e-petitions/tree/main/.github/workflows

PR #466, opened 2021-11-05
Add notes to planning applications

benbaumann95:
### Description of change

- Create notes table with belongs to relation with planning application
- Add ability to create notes and list in created_at descending order
- Display current note as the latest note

### Story Link

https://trello.com/c/p8rukRaF/465-officer-notes

### **When there are no notes**
Screenshot 2021-11-05 at 11 31 19

### **When there are multiple notes**
Screenshot 2021-11-05 at 11 31 58

### **Current note preview on planning application index**
Screenshot 2021-11-05 at 11 34 45

### DIFF
--- a/app/controllers/planning_application/notes_controller.rb
+++ b/app/controllers/planning_application/notes_controller.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+class PlanningApplication
+  class NotesController < ApplicationController
+    before_action :set_planning_application
+
+    def index
+      @notes = @planning_application.notes
+      @note = Note.new
+
+      respond_to do |format|
+        format.html
+      end
+    end
+
+    def create
+      @note = @planning_application.notes.new(note_params)
+      @note.user = current_user
+
+      respond_to do |format|
+        if @note.save
+          format.html do
+            redirect_to planning_application_path(@planning_application), notice: "Note was successfully created."
+          end
+        else
+          format.html { render :index }
+        end
+      end
+    end
+
+    private
+
+    def set_planning_application
+      @planning_application = planning_applications_scope.find(planning_application_id)
+    end
+
+    def planning_applications_scope
+      current_local_authority.planning_applications.includes(:notes)
+    end
+
+    def planning_application_id
+      Integer(params[:planning_application_id])
+    end
+
+    def note_params
+      params.require(:note).permit(:entry)
+    end
+  end
+end
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class Note < ApplicationRecord
+  belongs_to :planning_application
+  belongs_to :user
+
+  validates :entry, presence: true, length: { maximum: 500 }
+
+  scope :by_created_at_desc, -> { order(created_at: :desc) }
+end
--- a/app/views/planning_application/notes/index.html.erb
+++ b/app/views/planning_application/notes/index.html.erb
@@ -1,0 +1,38 @@
+<% add_parent_breadcrumb_link "Home", planning_applications_path %>
+<% add_parent_breadcrumb_link "Application", planning_application_path(@planning_application) %>
+
+<% content_for :title, "Notes" %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">
+      Notes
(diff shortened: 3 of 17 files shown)


app/controllers/planning_application/notes_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  class NotesController < ApplicationController
+    before_action :set_planning_application
+
+    def index
+      @notes = @planning_application.notes
+      @note = Note.new
+
+      respond_to do |format|
+        format.html
+      end
```
melzreal, line 13: we can get rid of this one given there is only one format of response

app/models/note.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Note < ApplicationRecord
+  belongs_to :planning_application
+  belongs_to :user
+
+  validates :entry, presence: true, length: { maximum: 500 }
```
melzreal, line 7: was this length agreed somewhere? Had a look at the A/C's but there's no suggestion of it - we don't have any length enforcements across the app so unless the partners specified this as a requirement we shouldn't enforce one without them giving it the 👍

app/views/planning_application/notes/index.html.erb
```diff
@@ -0,0 +1,38 @@
...
+
+    <% if @notes.empty? %>
+      <p class="govuk-body">There are no notes yet.</p>
+    <% else %>
+      <ul id="notes" class="govuk-list">
+        <% @notes.each_with_index do |note, i| %>
+          <% if i == 0 %>
+            <h2 class="govuk-heading-m">Current note</h2>
+          <% end %>
+          <% if i == 1 %>
```
melzreal, line 22: should this not be a >= ?

db/migrate/20211103111313_create_notes.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateNotes < ActiveRecord::Migration[6.1]
+  def change
+    create_table :notes do |t|
+      t.integer :planning_application_id, null: false
+      t.integer :user_id, null: false
```
freesteph, line 7: I think these two might be `t.references`

features/step_definitions/datetime_steps.rb
```diff
@@ -3,3 +3,7 @@
 Given("the date is year: {int}, month: {int}, day: {int}") do |year, month, day|
   travel_to Time.zone.local(year, month, day)
 end
+
+Given("the date is year: {int}, month: {int}, day: {int}, hour: {int}") do |year, month, day, hour|
```
freesteph, line 7: Can we try and make this slightly more natural? Something like "Given the time is 14h00 on the 4/12/2021"?

freesteph (approved):
:+1:

freesteph:
Also is there a better formatting than `[Author name in brackets]`? Should be italics, or bold, or just... plain.

db/migrate/20211103111313_create_notes.rb
benbaumann95: woops, nice spot. This is what i get for writing it manually

features/step_definitions/datetime_steps.rb
benbaumann95: > Also is there a better formatting than `[Author name in brackets]`? Should be italics, or bold, or just... plain.

What I interpreted from the designs. But i'll discuss with Tom.

app/models/note.rb
benbaumann95: More of a common sense check. Probably shouldn't let our db just take a silly amount of characters but I'll confirm this exact amount

app/controllers/planning_application/notes_controller.rb
benbaumann95: It's better to always be explicit for what formats we will accept
https://blog.arkency.com/2016/07/respond-to-format-is-useful-even-without-multiple-formats/

melzreal: rails docs say the action without respond_to implicitly responds to all formats - 
https://apidock.com/rails/ActionController/MimeResponds/respond_to - there was also an old respond_to method that is now deprecated and may have been what that blog was about.

I only suggested it because a senior suggested the same to me in the past, saying it was unnecessary - but its all the same to me if you want to keep it 🤷

melzreal (approved):

PR #469, opened 2021-11-09
Remove character limit from notes

benbaumann95:
- Update to use Latest note
- Truncate text after 200 characters for latest note preview on planning applications index


db/migrate/20211109132114_remove_character_limit_from_notes_entry.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveCharacterLimitFromNotesEntry < ActiveRecord::Migration[6.1]
+  def change
+    change_column :notes, :entry, :string, limit: nil
+    change_column :notes, :entry, :text
```
benbaumann95, line 4: Accidentally added as a string first time round oops

freesteph (approved):
Looks great :+1:

PR #470, opened 2021-11-10
Feat/rework page header

freesteph:
Rework the header according to: https://trello.com/c/mGcy8i0q/503-show-days-left-on-application-screen

Also try and refactor some bits and bobs, namely by introducing a presenter to handle some fiddly logic in the view.

# Before
Screen Shot 2021-11-10 at 15 35 47

# After
Screen Shot 2021-11-10 at 15 36 13


features/step_definitions/description_change_steps.rb
```diff
@@ -2,10 +2,6 @@
 
 require "faker"
 
-Given("a determined planning application") do
-  @planning_application.update!(status: "determined", determined_at: Time.zone.now)
```
freesteph, line 6: for the record (no pun intended), this creates a wrong determined application: because the status is updated by bypassing AASM, it's hard to notice that the resulting object is a determined application without a decision, which doesn't make any sense. 

"There should be a guard for that!" - [yes](https://github.com/unboxed/bops/blob/master/app/models/planning_application.rb#L66) but it's bypassed.

app/controllers/planning_applications_controller.rb
```diff
@@ -1,31 +1,7 @@
...
-                                                    view_recommendation
-                                                    edit_constraints_form
-                                                    edit_constraints
-                                                    cancel_confirmation
-                                                    cancel
-                                                    draw_sitemap
-                                                    update_sitemap
-                                                    decision_notice
-                                                    validation_notice]
+  before_action :set_planning_application, except: %i[new create index]
```
benbaumann95, line 4: 👍

app/models/planning_application.rb
```diff
@@ -8,6 +8,7 @@ class PlanningApplication < ApplicationRecord
   enum application_type: { lawfulness_certificate: 0, full: 1 }
 
   with_options dependent: :destroy do
+    has_many :audits
```
benbaumann95, line 11: huh how did this not exist before

app/models/planning_application.rb
```diff
@@ -48,7 +49,7 @@ class PlanningApplication < ApplicationRecord
 
   aasm.attribute_name :status
 
-  aasm do
+  aasm no_direct_assignment: true do
```
benbaumann95, line 52: 🥇

app/views/planning_applications/show.html.erb
```diff
@@ -1,118 +1,5 @@
 <%= render "assessment_dashboard" do %>
```
freesteph, line 1: I split each task into a partial

app/models/planning_application.rb
benbaumann95: should probably add this to validation requests too (but out of scope here)

app/presenters/planning_application_presenter.rb
```diff
@@ -0,0 +1,113 @@
...
+    def type_and_work_status
+      "#{type} (#{work_status.humanize})"
+    end
+  end
+
+  private
+
+  def status_tag_colour
+    if planning_application.determined?
+      planning_application.granted? ? "green" : "red"
```
benbaumann95, line 92: As you previously said, eventually would be good if this was it's own status rather than the decision string  so we can do less here

benbaumann95 (approved):
Looks really good , simplifies the view layer considerably. 🎉

PR #474, opened 2021-11-15
Ensure previous description displays

melzreal:
### Description of change
Ensure previous description is displayed in change request show page


benbaumann95 (approved):
Would a system spec be beneficial here to cover us?

melzreal:
> Would a system spec be beneficial here to cover us?

good point, I have a whole cucumber spec for this but missed out on the "show" page after a request is responded to. Will add a little spec

PR #476, opened 2021-11-16
Set scheduler to perform_later

melzreal:
Call perform_later instead of perform on scheduler


benbaumann95 (approved):

freesteph (commented):
I _think_ it's usually best to use `perform_later`, as it makes the job run as soon as the queuing system is free, rather than pushing it immediately. In our case, they will probably do the exact same thing (`perform_now/later`) since we don't have that much load, but it's good practice I think.

PR #477, opened 2021-11-16
Amend decision notice per design updates

melzreal:
### Description of change
https://trello.com/c/iYAf73W5/562-design-improvements-to-decision-notice
image

### DIFF
--- a/app/views/description_change_validation_requests/show.html.erb
+++ b/app/views/description_change_validation_requests/show.html.erb
@@ -15,7 +15,7 @@
 
     <div>
       <p class="govuk-body"> <strong> Previous description:</strong><br>
-        <%= @planning_application.description  %>
+        <%= @description_change_request.previous_description  %>
       </p>
       <p class="govuk-body"> <strong> Proposed description:</strong><br>
         <%= @description_change_request.proposed_description  %>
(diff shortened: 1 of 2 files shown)


app/views/planning_applications/_decision_notice.html.erb
```diff
@@ -12,27 +12,22 @@
...
-        "Decision date" => planning_application.determined_at&.to_formatted_s(:day_month_year) || "TBD",
-        "Site address" => planning_application.full_address,
-        "Use/development" => planning_application.description,
-      }.each do |key, value|
-    %>
-    <div class="govuk-summary-list__row">
-      <dt class="govuk-summary-list__key">
-        <%= key %>
-      </dt>
-      <dd class="govuk-summary-list__value">
```
freesteph, line 30: why did we move away from the definition list?

db/schema.rb
```diff
@@ -303,8 +301,6 @@
   add_foreign_key "audits", "planning_applications"
   add_foreign_key "description_change_validation_requests", "planning_applications"
   add_foreign_key "description_change_validation_requests", "users"
-  add_foreign_key "notes", "planning_applications"
```
freesteph, line 306: you need to rebase and re-migrate as this is undoing Ben's work, looks like :smile:

app/views/planning_applications/_decision_notice.html.erb
melzreal: design request

PR #478, opened 2021-11-17
Add design changes to decision notice

melzreal:
### Description of change

Design changes to decision notice

### Story Link
https://trello.com/c/iYAf73W5/562-design-improvements-to-decision-notice


app/views/planning_applications/_decision_notice.html.erb
```diff
@@ -12,27 +12,22 @@
...
+      "Applicant" => planning_application.applicant_name,
+      "Application number" => planning_application.reference,
+      "Application received" => planning_application.created_at.to_formatted_s(:day_month_year),
+      "Decision date" => planning_application.determined_at&.to_formatted_s(:day_month_year) || "TBD",
+      "Site address" => planning_application.full_address,
+      "Use or development" => planning_application.description,
+    }.each do |key, value|
+  %>
+  <p class="govuk-body">
+   <strong><%= key %></strong> <br>
```
freesteph, line 26: re: decision list; either make it a definition list and apply some CSS to it, or use some heading (`h3/h4`, whatever level makes sense here) to make it semantic. But `strong` + `br` isn't optimal.

db/migrate/20211116114852_update_local_authority.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class UpdateLocalAuthority < ActiveRecord::Migration[6.1]
+  class LocalAuthority < ApplicationRecord; end
+
+  def change
+    reversible do |dir|
```
freesteph, line 7: this isn't reversible, so I'd remove the whole reversible thing.

db/migrate/20211116114852_update_local_authority.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class UpdateLocalAuthority < ActiveRecord::Migration[6.1]
+  class LocalAuthority < ApplicationRecord; end
+
+  def change
```
benbaumann95, line 6: Should this be `up_only` ?

benbaumann95: probs unnecessary for this but you could also provide a down block to update the columns back to it's original state

melzreal: I'm not sure, wouldn't up be only used for a "new" column? the documentation isn't very clear: https://apidock.com/rails/v6.0.0/ActiveRecord/Migration/up_only

melzreal: ah re the down block - its irreversible as its a content change from what we have in the seeds. I should probably change the seeds too though 🤔

benbaumann95: Yeah `up_only` just refers to the up migration. Can be any modification but is just run when you do db:migrate and not db:rollback

benbaumann95: I guess what I meant with the down block is add some code to update the enquiries_paragraph to what it was before the migration would have taken place. But it's unlikely we'll have to revert back to the old copy anyway

.rubocop_todo.yml
```diff
@@ -338,3 +338,4 @@ Layout/LineLength:
     - 'spec/system/planning_applications/show_spec.rb'
     - 'spec/system/planning_applications/validating_spec.rb'
     - 'db/migrate/20211020153256_create_delayed_jobs.rb'
+    - 'db/migrate/20211116114852_update_local_authority.rb'
```
freesteph, line 341: this probably isn't necessary anymore

freesteph (approved):

benbaumann95 (approved):

db/schema.rb
```diff
@@ -21,7 +21,7 @@
     t.bigint "record_id", null: false
     t.bigint "blob_id", null: false
     t.datetime "created_at", null: false
-    t.index ["blob_id"], name: "index_active_storage_attachments_on_blob_id"
+    t.index ["blob_id"], name: "ix_active_storage_attachments_on_blob_id"
```
benbaumann95, line 24: huh how come  all this stuff is changing ?
Seems its happening to all the indexes

PR #479, opened 2021-11-18
Add worker during deployment

melzreal:


freesteph (commented):
The problem here is that the `notify_appsignal` will create a duplicate entry for both of these deployments; maybe just call it last (and so neither in the webservers or workers block)?

lib/package_builder.rb
```diff
@@ -127,9 +127,9 @@ def create_archive
   end
 
   def create_deployments!
-    create_deployment!("Webservers") do
-      notify_appsignal
-    end
+    create_deployment!("Webservers")
+    create_deployment!("Workers")
+    notify_appsignal
```
melzreal, line 132: @freesteph I guess like so?

freesteph (approved):

PR #480, opened 2021-11-23
Edit document date

melzreal:
### Description of change
- Add created_at timestamp column to documents
- Add an optional association between documents and users, setting the current_user to the api user when the api controller is accessed, so we can keep track of a document being "manually" uploaded or via an api type user
- Add audit entries for date changes

### Story Link

https://trello.com/c/TW56Gdne/196-as-a-planning-officer-i-need-to-be-able-to-edit-the-date-of-when-a-document-is-received-so-that-it-appears-on-the-decision-notic


app/models/document.rb
```diff
@@ -75,6 +86,18 @@ def published?
...
+  def created_by
+    if user.present?
+      "This document was manually uploaded by #{user.name}"
+    elsif api_user.present?
+      "This document was uploaded by #{api_user.name}"
+    end
+  end
+
+  def date_received_or_created
+    date_received.present? ? date_received.to_date.strftime("%e %B %Y") : created_at.to_date.strftime("%e %B %Y")
```
freesteph, line 98: shouldn't this be using our `day_month_year` format?

also I'd write `(date_received || created_at).to_date.to_formatted_s(:day_month_year)`

app/models/document.rb
```diff
@@ -41,6 +43,7 @@ class Document < ApplicationRecord
   validate :file_attached
   validate :numbered
   validate :invalidated_comment_present?
+  validate :created_date_is_in_the_past
```
freesteph, line 46: you can use numericality here (a date is an amount of seconds): https://stackoverflow.com/a/49063513

app/views/audits/index.html.erb
```diff
@@ -26,7 +26,7 @@
           <td class="govuk-table__cell"> User deleted </td>
         <% end %>
         <td colspan="2" class="govuk-table__cell">
-          <% if audit_item.activity_type.include?("request") %>
+          <% if audit_item.activity_type.include?("request") || audit_item.activity_type.include?("document_date_received_changed") %>
```
freesteph, line 29:
```ruby
<% if audit_item.activity_type.in? ["request", "document_date_received_changed"] %>
```

app/views/audits/types/_document_date_received_changed.html.erb
```diff
@@ -0,0 +1,11 @@
+<%= activity(locals[:item].activity_type, locals[:item].activity_information) %>
+
+<p>
+  <% if JSON.parse(locals[:item].audit_comment)["previous_received_date"].present? %>
```
freesteph, line 4: this should decide of the outside `<p>` as well

db/migrate/20211129155744_add_date_received_to_documents.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddDateReceivedToDocuments < ActiveRecord::Migration[6.1]
+  def change
+    add_column :documents, :date_received, :date
```
freesteph, line 5: this should be a timestamp, and named `received_at`

spec/system/documents/edit_documents_spec.rb
```diff
@@ -56,6 +56,30 @@
...
+    it "does not allow for a future date to be inserted in created_date" do
+      click_link "Edit"
+
+      fill_in "Day", with: "19"
+      fill_in "Month", with: "11"
+      fill_in "Year", with: "3021"
+
+      click_button("Save")
+      expect(page).to have_content("Date must be today or earlier. You cannot insert a future date.")
+    end
```
freesteph, line 81: is there any chance you can port these to Cucumber specs rather than adding to these?

app/models/document.rb
```diff
@@ -75,6 +86,18 @@ def published?
     self.class.for_publication.where(id: id).any?
   end
 
+  def created_by
```
freesteph, line 89: this is presentational logic, it shouldn't live here - `document_helper.rb` feels like the right place

freesteph (changes requested):

app/controllers/api/v1/application_controller.rb
```diff
@@ -50,6 +50,7 @@ def authenticate
 
       def current_api_user
         @current_api_user ||= authenticate
+        Current.user = @current_api_user
```
benbaumann95, line 53: Why can't we do `Current.api_user` here? And then set it as an attribute in `current.rb`.
This seems like it will override things

spec/system/documents/edit_documents_spec.rb
melzreal: I recall you shied away from doing the document specs in cucumber because of having problems loading the pdfs into it?

freesteph: OK. Would you mind if I give it a shot and open a PR to merge into your PR?

melzreal: yes, happy to pair on it as well - was just a bit afraid since I remember it was difficult for you so I figured I wouldn't even know where to start!

app/models/document.rb
melzreal: keeping in record following pairing - we're not using numericality as it requires the data value to be an integer for the time comparison so it didn't do exactly what we were after

app/models/document.rb
```diff
@@ -52,6 +55,11 @@ class Document < ApplicationRecord
 
   scope :with_tag, ->(tag) { where("tags @> ?", "\"#{tag}\"") }
 
+  before_create do
+    self.api_user ||= Current.api_user
+    self.user ||= Current.user
```
benbaumann95, line 60: Will this always be set because we do `before_action :set_current_user` in the application_controller ?

melzreal: no this is to assign the user to the document object - it won't always be set since in the api controller we won't have a user logged in necessarily - you can, but you can also log out and send the application with token and that's when api user gets set

app/views/audits/index.html.erb
```diff
@@ -26,7 +26,7 @@
           <td class="govuk-table__cell"> User deleted </td>
         <% end %>
         <td colspan="2" class="govuk-table__cell">
-          <% if audit_item.activity_type.include?("request") %>
+          <% if audit_item.activity_type.match(/validation|document_received_at_changed/) %>
```
freesteph, line 29: can we avoid a Regexp here and use `in?`

features/editing_documents.feature
```diff
@@ -0,0 +1,57 @@
...
+    Given I fill in "Day" with "19"
+    And I fill in "Month" with "11"
+    And I fill in "Year" with "2021"
+    And I press "Save"
+    And I view the document with reference "FOOBAR"
+    And I fill in "Day" with "22"
+    And I fill in "Month" with "11"
+    And I fill in "Year" with "2021"
+    When I press "Save"
+    Given I view the planning application audit
```
freesteph, line 39: There's a step for this: https://github.com/unboxed/bops/blob/master/features/step_definitions/auditing_steps.rb#L3

features/editing_documents.feature
```diff
@@ -0,0 +1,57 @@
...
+    Then the page contains "to: 22 November 2021"
+
+  Scenario: I can mark a document as not valid
+    Given I choose "No" for "Is the document valid?"
+    And I fill in "Describe in full why the document is invalid." with "BANANAS"
+    When I press "Save"
+    Then the page contains "Invalid documents: 1"
+    Then the page contains "BANANAS"
+
+  Scenario: I can see in the audit log when a document is marked as invalid
```
freesteph, line 51: Can you merge this into the previous spec ("I can mark a document as not valid"); it's okay to have a "And there is an audit entry for XXX" as an extra assertion.

spec/system/documents/upload_spec.rb
```diff
@@ -98,6 +98,23 @@
         end
       end
 
+      it "displays the date when uploaded and user who uploaded the document" do
```
freesteph, line 101: can we maybe move this to Cucumber?

freesteph (commented):
Very close :steam_locomotive:

features/editing_documents.feature
```diff
@@ -0,0 +1,51 @@
...
+    Then the page contains "Date received: 19 November 2021"
+
+  Scenario: I can mark a document as not valid
+    Given I choose "No" for "Is the document valid?"
+    And I fill in "Describe in full why the document is invalid." with "BANANAS"
+    When I press "Save"
+    Then the page contains "Invalid documents: 1"
+    Then the page contains "BANANAS"
+
+  Scenario: I can see different document edit actions in the audit log
```
melzreal, line 36: I realised of a little bug where only one action got audited if two actions that required auditing happened, this spec covered it - hence doing these two repeated here. Maybe I should abstract them into steps? @freesteph

app/views/audits/index.html.erb
melzreal: in? only takes a single string, you'd instead need an iterator to pass through the two strings and check - which is much more logic then just having the two conditionals with an or - so the match is less code

melzreal: (I did try for a while to make in? work fyi, in light of your previous suggestion - but the problem is that we are checking if a substring is contained in audit_item, we aren't checking if a whole string is contained - ["document_received_at_changed", "validation"].in? audit_item.activity_type doesn't do the partial match of the word, hence the regex. Another option was to iterate with .any but to do that much to check two strings seemed too much overhead instead of just checking two conditions

freesteph: to be fair, does this need its own audit template at all? the generic one looks fine to me:

```ruby
  def audit_date_comment(document)
    previous, updated = document.saved_change_to_received_at


    { previous_received_date: previous.to_formatted_s(:day_month_year),
      updated_received_date: updated.to_formatted_s(:day_month_year)
    }.to_json
  end
```

image

melzreal: I did try that - its what I was saying - did you pull the code and try it? because when I did it it didn't pickup on the template with that exact code
```
audit.item_activity_type.in? ["document_received_at_changed", "request"]
```

freesteph: so what's the logic? whether it contains `request` or it's equal to `document_received_at_changed`?

melzreal: the logic is that in? doesn't take a partial string, so it can't match blahblah_request

app/views/audits/index.html.erb
```diff
@@ -26,7 +26,7 @@
           <td class="govuk-table__cell"> User deleted </td>
         <% end %>
         <td colspan="2" class="govuk-table__cell">
-          <% if audit_item.activity_type.include?("request") %>
+          <% if audit_item.activity_type.include?("request") || audit_item.activity_type.include?("document_received_at_changed") %>
```
melzreal, line 29: @freesteph I've changed it back to the previous version that doesnt use regex and works fine. Your suggestion doesn't work as in? cannot account for partial string matches so it'd always need an extra iteration - with any? for ex to match the two strings, which seems silly to check just two conditions.

freesteph: in does account partial matches

```ruby
# in a rails console
"request".in? "foobar_request"
=> true
```

melzreal: steph, again, your example doesn't work - in? is unable to check against two values, because it only takes a single string for a partial match, it can't do a partial match on two strings - it looks for the exact string. 

You don't need to illustrate something I already spent a long time trying with different variations because I really wanted to make you happy and put in that example. But it doesn't work. I know that a single string can be checked - I have tried your example, I have tried it with every possible iteration and I wish that you would trust me and try it with the example we need - which is checking against two strings - and then show me that it works:

image
image

and if you swap them around, and have it check against the description_change_value, it can't check the partial match again:
image
image

freesteph: That's my bad, I'm really sorry. I got confused about what matches on which side of `in?` and `include?`. Very, very sorry.

freesteph (approved):

PR #481, opened 2021-11-23
Add validation request endpoints to the API

benbaumann95:
### Description of change

- Includes index and show resources for each of the validation requests
- Update swagger docs and improve layout
- Add request specs for new endpoints

### Story Link

https://trello.com/c/g8LtnBJG/660-add-validation-request-show-endpoints-to-bops-and-use-this-in-bops-applicant

Screenshot 2021-11-22 at 10 08 50


app/controllers/api/v1/additional_document_validation_requests_controller.rb
```diff
@@ -3,13 +3,33 @@
...
-      before_action :check_token_and_set_application,
-                    :check_file_params_are_present,
+      before_action :check_token_and_set_application
+      before_action :check_file_params_are_present,
                     :check_file_size,
                     :check_file_type, only: :update
 
+      def index
+        respond_to do |format|
+          format.json do
```
freesteph, line 15: should we default the API routes to `json` format? https://guides.rubyonrails.org/routing.html#defining-defaults

freesteph (approved):
Looks good

app/controllers/api/v1/additional_document_validation_requests_controller.rb
benbaumann95: ooh nice, haven't seen this before

benbaumann95: I'll definitely add this but do it in the next PR. Need to add a test to api_index_spec too (but planning to do something clever with Melissa on this).

PR #482, opened 2021-11-24
Fix/map into pdfs

freesteph:
Please see commit list


benbaumann95 (approved):
Looks good!
Just need to make sure anytime in the app we render pdfs it still works, will test this locally

PR #483, opened 2021-11-25
Amend target date to expiry date

melzreal:
### Description of change

Bug fix: expiry date should be communicated in validation email, not target date - due to legal target of when an application can be appealed.

### Story Link

https://trello.com/c/kGdQXdX0/672-application-valid-letter-is-showing-the-wrong-date


freesteph (approved):
You might want to rebase on master

PR #488, opened 2021-11-26
Spike ward using mapit

melzreal:
### Description of change
Spike for a possible approach on how we can use mapit to retrieve and set ward for planning applications. 
Does not have an api_key which we'll need of course, if the partners do decide they want this

### Story Link
https://trello.com/c/gYoE98cy/435-display-ward-for-a-given-application

image

### DIFF
--- a/app/views/shared/_application_information.html.erb
+++ b/app/views/shared/_application_information.html.erb
@@ -44,6 +44,14 @@
             <td class="govuk-table__cell"><%= link_to 'View site on Google Maps', map_link(@planning_application.full_address), target: '_blank', class: "govuk-link" %></td>
           </tr>
           <tr class="govuk-table__row">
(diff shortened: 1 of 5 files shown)


app/views/shared/_application_information.html.erb
```diff
@@ -43,6 +43,14 @@
             <td class="govuk-table__cell"><strong>Location:</strong></td>
             <td class="govuk-table__cell"><%= link_to 'View site on Google Maps', map_link(@planning_application.full_address), target: '_blank', class: "govuk-link" %></td>
           </tr>
+          <tr class="govuk-table__row">
```
melzreal, line 46: we'd also need a little catch statement here for all the previous planning applications that were creted without ward.

melzreal:
closing spike ticket

PR #489, opened 2021-11-26
Fix/valid from date

freesteph:
* Add `PlanningApplication#received_at` to reflect the fact that if an application is received on a Saturday, it's legally received on Monday;
* Add `PlanningApplication#valid_from` which designates the last business-day-friendly update to the application _before_ it was made valid.

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -354,6 +354,16 @@ class PlanningApplication < ApplicationRecord
     Time.next_immediate_business_day(created_at)
   end
 
+  def valid_from
+    return nil unless validated?
+
+    if closed_validation_requests.any?
+      Time.next_immediate_business_day(last_validation_request_date)
+    else
+      received_at
+    end
+  end
+
   private
 
   def set_key_dates
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -350,6 +350,10 @@ class PlanningApplication < ApplicationRecord
     documents.for_display
   end
 
+  def received_at
+    Time.next_immediate_business_day(created_at)
+  end
+
   private
 
   def set_key_dates
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -367,8 +367,8 @@ class PlanningApplication < ApplicationRecord
   private
 
   def set_key_dates
-    self.expiry_date = 40.business_days.after(documents_validated_at || created_at)
-    self.target_date = 35.business_days.after(documents_validated_at || created_at)
+    self.expiry_date = 40.business_days.after(documents_validated_at || received_at)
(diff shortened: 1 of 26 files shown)


app/models/planning_application.rb
```diff
@@ -350,11 +350,25 @@ def documents_for_decision_notice
     documents.for_display
   end
 
+  def received_at
```
melzreal, line 353: we also have a helper method with this name in planning application helpers, should we get rid of the helper and replace view calls to received_at with the calls to the model method?

freesteph: Yes good spot, thanks :eyeglasses:

config/initializers/business_time.rb
```diff
@@ -6,3 +6,17 @@
 #  BusinessTime::Config.beginning_of_workday = "10:00 am"
 #  BusinessTime::Config.end_of_workday = "11:30 am"
 #  BusinessTime::Config.holidays << Date.parse("August 4th, 2010")
+
+module Bops
+  module TimeExtensions
+    module ClassMethods
+      def next_immediate_business_day(time)
+        Time.roll_forward(time).to_date
+      end
```
melzreal, line 15: wouldn't it be easier to use the first_business_day the gem gives us, which achieves the same? would be less code to call it:
https://github.com/bokmann/business_time/blob/17ce786e4fde7ec19b1a6f2426d69aee3acb0068/lib/business_time/time_extensions.rb#L75

freesteph: I tried using it but the `Time.first_business_day` doesn't work like I expect it to:

```ruby
irb(main):016:0> Time.first_business_day(Time.zone.now.change(hour: 03))
=> Mon, 29 Nov 2021 03:00:00.000000000 GMT +00:00
irb(main):017:0> Time.first_business_day(Time.zone.now.change(hour: 12))
=> Mon, 29 Nov 2021 12:00:00.000000000 GMT +00:00
irb(main):018:0> Time.first_business_day(Time.zone.now.change(hour: 23))
=> Mon, 29 Nov 2021 23:00:00.000000000 GMT +00:00
irb(main):019:0> 
```

It's not business-hours aware.

melzreal: ahh you're right, means the business_time config isn't loading the beginning and end of workdays properly... had a look if someone had raised a bug in the gem and its in the todos - https://github.com/bokmann/business_time#todo-
nice catch, makes me wonder though if everywhere else we are using business time we need to patch this though

melzreal (approved):
LGTM

config/initializers/business_time.rb
freesteph: I think it is, just this method doesn't account for it. I was in the beginning of writing an issue but the gem doesn't seem that maintained, looks like https://github.com/zendesk/biz is a potential successor but something for a rainy day...

PR #491, opened 2021-11-29
date/time: make our :day_month_year format the default

freesteph:
Realistically this format is the one way we present all of our dates
so make it the default and save a hefty amount of typing on the rest
of the app.


spec/requests/api/additional_document_validation_request_spec.rb
```diff
@@ -30,7 +30,7 @@
             {
               "id" => additional_document_validation_request.id,
               "state" => "open",
-              "response_due" => additional_document_validation_request.response_due.to_s,
+              "response_due" => additional_document_validation_request.response_due.to_s(:db),
```
melzreal, line 33: ohh what does the :db thing to? never seen it before, how would one know how to use it?

melzreal (approved):
yeaaassss SPRING CLEAN !

spec/requests/api/additional_document_validation_request_spec.rb
freesteph: so the API doesn't format the data it sends, so it send the "raw" timestamps (as opposed to our prettification of it), which are in the `db` format: https://railsdatetimeformats.com/

PR #497, opened 2021-11-30
Re-add deleted partial

melzreal:
### Description of change
Re-add audit partial that got accidentally deleted


config/environments/development.rb
```diff
@@ -42,14 +42,11 @@
...
-
-  if ENV["NOTIFY_API_KEY"].present?
-    config.action_mailer.delivery_method = :notify
-    config.action_mailer.notify_settings = {
-      api_key: ENV.fetch("NOTIFY_API_KEY")
-    }
-  end
-
+ 
+  config.action_mailer.delivery_method = :notify
```
freesteph, line 46: how does this relate to the chance?

freesteph (approved):

config/environments/development.rb
melzreal: does not, was just in the process of removing it got pushed by accident

PR #499, opened 2021-12-01
End if statement in correct line

melzreal:
### Description of change
We were ending an if statement in the wrong line, meaning the result questions got excluded if not all the criteria was present.

Before:
image

After:
image


freesteph (approved):
Hint erb-lint

PR #500, opened 2021-12-01
Format list correctly and add small test

melzreal:
### Description of change
Format index so it displays correctly
Add a small cucumber test to check flags get displayed in index

https://trello.com/c/ZIPmN8VG/611-immune-prior-approval-question-flags-not-being-displayed-correctly-in-bops

### DIFF
--- a/features/checking_proposal_details.feature
+++ b/features/checking_proposal_details.feature
@@ -2,12 +2,9 @@ Feature: Checking proposal details are visible in the planning application
   Background:
     Given I am logged in as an assessor
     And a new application of type prior approval
-    And I view the planning application
 
-Scenario: I can view the correct flags have been assigned to the planning application
-  Then the page contains "Planning permission / Prior approval"
-  And the page contains "Details identified as relevant to the result"
-  And the page contains "I will add"
-  And the page contains "1-2 new storeys"
-  And the page contains "The height of the new roof will be higher than the old roof by"
-  And the page contains "7m or less"
+  Scenario: I can view the correct flags have been assigned to the planning application
+    When I view the planning application
+    Then there is a relevant proposal detail for "The height of the new roof will be higher than the old roof by" with a response of "7m or less"
+    And there is a relevant proposal detail for "I will add" with a response of "1-2 new storeys"
+    And there is a relevant proposal detail for "The new storeys will be" with a response of "on the principal part of the building only"
--- a/features/checking_proposal_details.feature
+++ b/features/checking_proposal_details.feature
@@ -1,0 +1,13 @@
+Feature: Checking proposal details are visible in the planning application
+  Background:
+    Given I am logged in as an assessor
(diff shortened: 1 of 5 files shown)


benbaumann95 (approved):

features/checking_proposal_details.feature
```diff
@@ -0,0 +1,13 @@
...
+    And a new application of type prior approval
+    And I view the planning application
+
+Scenario: I can view the correct flags have been assigned to the planning application
+  Then the page contains "Planning permission / Prior approval"
+  And the page contains "Details identified as relevant to the result"
+  And the page contains "I will add"
+  And the page contains "1-2 new storeys"
+  And the page contains "The height of the new roof will be higher than the old roof by"
+  And the page contains "7m or less"
```
freesteph, line 13: You can't have a scenario that's just assertions

freesteph (changes requested):

features/checking_proposal_details.feature
freesteph: I don't think this deserves a Cucumber test to be honest; it's just testing content and it's not worth the overhead. Let's write some tests when we have actual user journeys to test.

melzreal: I added a test since this was broken but we didn't have enough coverage to pick it up, so just trying to prevent for that in the future... we surely need a test since its to prevent that? Should it be a system test then?

freesteph: system tests are replaced by Cucumber; what was broken? I thought this was just fixing style?

benbaumann95: The reason we had this bug in the first place was due to a lack of tests. If we had this in place we wouldn't have had the issue with the content suddenly not showing

benbaumann95: It's quite a crucial feature of the app so i'd like some testing in place for this

freesteph: what was the bug?

benbaumann95: This was in relation to this right https://github.com/unboxed/bops/pull/499

benbaumann95: or add some erb linting at least

freesteph: I'm always happy with test, but having it just check content based on some other content found in the factory doesn't really sound like a good "Arrange-Act-Assert" pattern

benbaumann95: If only there was something like cypress screenshot tests for this kind of thing.

Just thinking that this test would have failed and picked up the accidental mistake of not closing the tags properly

benbaumann95: We have nothing in place currently to prevent that from happening again which gives me a bit less confidence

freesteph (approved):

PR #501, opened 2021-12-02
Call rake task instead of runner job

melzreal:
### Description of change
When using the whenever gem, we can either call "runner + job" or we can call "rake + rake task that calls job". When using just runner, we are seeing CRON complaining of not being able to email anyone at the end of the task. When using rake, our appsignal integration should take care of picking up on any failures - so we're changing the scheduler to call rake and the rake task, then the rake task calls the job. Phew.


config/schedule.rb
```diff
@@ -22,5 +22,5 @@
 # Learn more: http://github.com/javan/whenever
 
 every 1.day, at: "9:00 am" do
-  runner "CloseDescriptionChangeJob.perform_later"
+  rake "CloseDescriptionChange.perform_later"
```
benbaumann95, line 25: Is this how you call it?

Thought it might be something like
```rake jobs:close_description_change```

benbaumann95: have a look under `rake --tasks`

freesteph (approved):

PR #503, opened 2021-12-02
Fix/flags crash

freesteph:
Fix a bug where a certain kind of payload could crash the app; please see commit list :green_heart: 

Also some CSS cleanups which result in this:

## Before
Screen Shot 2021-12-02 at 11 12 55

## After
Screen Shot 2021-12-02 at 11 13 02

### DIFF
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -340,12 +340,4 @@ class PlanningApplicationsController < AuthenticationController
   def ensure_constraint_edits_unlocked
     render plain: "forbidden", status: :forbidden and return unless @planning_application.can_validate?
   end
-
-  def documents_validated_at_missing?
-    if params["planning_application"]["documents_validated_at(3i)"].blank? ||
-       params["planning_application"]["documents_validated_at(2i)"].blank? ||
-       params["planning_application"]["documents_validated_at(1i)"].blank?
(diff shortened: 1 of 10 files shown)


benbaumann95 (approved):

melzreal:
image
getting this when pulling?

freesteph:
my bad @melzreal I did a refactor on the fly and used the wrong method... fixed now I believe.

app/controllers/planning_applications_controller.rb
```diff
@@ -340,12 +340,4 @@ def ensure_user_is_reviewer
   def ensure_constraint_edits_unlocked
     render plain: "forbidden", status: :forbidden and return unless @planning_application.can_validate?
   end
-
-  def documents_validated_at_missing?
```
melzreal, line 344: image
this was in place to prevent the application crashing when only one data param is entered, if we are removing it then we need a replacement validation.

melzreal (changes requested):

app/controllers/planning_applications_controller.rb
melzreal: image

melzreal: op nevermind it was the previous method that we already removed that validated this step - just realised this is also broken in staging

melzreal (approved):

PR #504, opened 2021-12-06
Amend audit log and start moving tests to cucumber

melzreal:
### Description of change

Show latest audit log activity directly on main application page
Start moving tests to cucumber

### Story Link
https://trello.com/c/c6NBIWBN/664-the-activity-log-audit-link-needs-to-be-in-a-more-visible-place-so-its-more-easier-to-find-currently-in-key-application-dates-se


spec/system/planning_applications/constraints_spec.rb
```diff
@@ -32,7 +32,6 @@
       expect(page).to have_content("Listed building")
       expect(page).to have_content("Broads")
       expect(page).to have_content("Flood zone 3")
-      expect(page).to have_no_content("Conservation Area")
```
melzreal, line 35: note: this is because the audit log now has the text "Conservation Area" being displayed as one of the things removed, and the info is displayed on the same page although in a different accordion.

app/controllers/planning_applications_controller.rb
```diff
@@ -16,7 +16,9 @@ def index
                              end
   end
 
-  def show; end
+  def show
+    @audit = @planning_application.audits.order("created_at ASC").last
```
freesteph, line 20: I think it'd be nicer to have the ordering as the default scope on the association, in the model

app/views/shared/_supporting_information.html.erb
```diff
@@ -115,6 +112,38 @@
...
+          <% elsif @audit.activity_type.include?("request") || @audit.activity_type.include?("document_received_at_changed") %>
+            <%= render "audits/types/#{@audit.activity_type}", locals: { item: @audit } %>
+          <% else %>
+            <%= render "audits/types/generic_audit_entry", locals: { item: @audit } %>
+          <% end %>
+          <p class="govuk-body">
+            <%= define_audit_user(@audit) %>
+          </p>
+          <p class="govuk-body">
+            <%= @audit.created_at.to_date %> at <%= @audit.created_at.strftime("%H:%M") %>
```
freesteph, line 138: This should be the regular `day_month_date`, it's defined for Time too: https://github.com/unboxed/bops/blob/master/config/initializers/date_formatter.rb#L6

features/auditing.feature
```diff
@@ -1,25 +1,72 @@
...
     Given I create an additional document validation request with "Meme of the dog"
     And the planning application is invalidated
     When I view the application's validations requests
     And I cancel a validation request for a "Meme of the dog" with "Meme of dog is no longer needed"
     When I view the planning application audit
     Then there is an audit entry containing "Cancelled: validation request (new document#1)"
     Then there is an audit entry containing "Reason: Meme of dog is no longer needed"
+
+  Scenario: I can view in the audit log when a planning application has been withdrawn
+    Given I press "Cancel application"
```
freesteph, line 30: Given: setup
When: trigger
Then: assertions

features/auditing.feature
```diff
@@ -1,25 +1,72 @@
...
+    And I choose "Withdrawn by applicant"
+    And I fill in "Can you provide more detail?" with "Applicant is moving to Bermuda, because heck this"
```
freesteph, line 32: lol

features/auditing.feature
```diff
@@ -1,25 +1,72 @@
...
+    And I check "National Park"
+    And I check "Broads"
+    And I uncheck "Conservation Area"
+    And I press "Save"
+    Then there is an audit entry containing "Constraint added"
+    And there is an audit entry containing "National Park"
+    And there is an audit entry containing "Constraint added"
+    And there is an audit entry containing "Broads"
+    And there is an audit entry containing "Constraint removed"
+    And there is an audit entry containing "Conservation Area"
```
freesteph, line 57: Can we merge these pairs? `Constraint added:\nNational Park`? It's not _great_ but it ties it together, rather than having loose bit of texts checked through all audit entries.

features/step_definitions/auditing_steps.rb
```diff
@@ -3,15 +3,19 @@
...
+When("I click on the audit accordion") do
+  page.find("#audit-log").click
+end
+
 When("I visit the audit log") do
   steps %(
     When I view the planning application
-    Then I press "Key application dates"
-    Then I press "Activity log"
+    Then I click on the audit accordion
```
freesteph, line 18: why not keep `Then I press "Audit log"`?

features/step_definitions/planning_application_steps.rb
```diff
@@ -133,3 +133,12 @@
     expect(page).to have_selector("li", text: [question, response].join("\n"))
   end
 end
+
+Given("a created at audit entry") do
```
freesteph, line 137: This steps feels really artificial, is it used anywhere?

features/step_definitions/auditing_steps.rb
melzreal: it can't find it, that's why I added the other "page.find(#audit-log) since with just "press "Audit log" it gives:
```
    And I cancel a validation request for a "Meme of the dog" with "Meme of dog is no longer needed" # features/step_definitions/validation_requests_steps.rb:102
    When I view the planning application audit                                                       # features/step_definitions/planning_application_steps.rb:105
    Then there is an audit entry containing "Cancelled: validation request (new document#1)"         # features/step_definitions/auditing_steps.rb:3
      Unable to find link or button "Activity log" (Capybara::ElementNotFound)
      ./features/step_definitions/web_steps.rb:71:in `nil'
      ./features/step_definitions/auditing_steps.rb:16:in `nil'
      ./features/step_definitions/auditing_steps.rb:4:in `nil'
      features/auditing.feature:26:in `there is an audit entry containing "Cancelled: validation request (new document#1)"'
      ```

Comment (melzreal) [reply features/step_definitions/auditing_steps.rb:18]: because it can't find it, so I added the id finder just above that:
```
    Then the page contains "Description change request successfully cancelled"     # features/step_definitions/web_steps.rb:3
    And there is an audit entry containing "Cancelled: description change request" # features/step_definitions/auditing_steps.rb:3
      Unable to find link or button "Audit log" (Capybara::ElementNotFound)
      ./features/step_definitions/web_steps.rb:71:in `nil'
      ./features/step_definitions/auditing_steps.rb:16:in `nil'
      ./features/step_definitions/auditing_steps.rb:4:in `nil'
      features/adding_description_change.feature:16:in `there is an audit entry containing "Cancelled: description change request"'
      ```

features/auditing.feature
melzreal: I've added an extra step to check two values instead of one, does it look a bit better or does it need reworking?

app/views/shared/_supporting_information.html.erb
melzreal: but that gives me the whole thing, when I just want the date so then I can match the design to have the created_date at xtime?

so just created at:
image
vs this:
image

features/step_definitions/auditing_steps.rb
freesteph: OK I've had a look, it's because GOV.UK is inconsistent in their presentation of the accordion component. The markup on the component page says the heading text is a span, but the actual HTML example uses a button: https://design-system.service.gov.uk/components/accordion/default/index.html. Let's use a button and remove the ID match.

features/step_definitions/planning_application_steps.rb
melzreal: will just take it out for now

app/views/shared/_supporting_information.html.erb
```diff
@@ -115,6 +112,38 @@
...
+          <p class="govuk-body">
+            <%= define_audit_user(@audit) %>
+          </p>
+          <p class="govuk-body">
+            <%= @audit.created_at.to_date %> at <%= @audit.created_at.strftime("%H:%M") %> 
+          </p>
+        </p>
+      <% end %>
+      <p class='govuk-body'>
+        <%= link_to "View all", planning_application_audits_path(@planning_application), class: "govuk-link" %>
```
freesteph, line 143: This should be `View all activity` or `View all audits` or something descriptive rather than the generic `View all` (all what?)

freesteph: I know it's the design but this is the sort of thing we must challenge, it's good practice and a WCAG rule https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-refs.html

features/auditing.feature
freesteph: Can you update all of them to have the GAWT :goat: in them? Given-And(optional)-When-Then

features/auditing.feature
```diff
@@ -1,25 +1,69 @@
...
     And the planning application is invalidated
     When I view the application's validations requests
     And I cancel a validation request for a "Meme of the dog" with "Meme of dog is no longer needed"
     When I view the planning application audit
     Then there is an audit entry containing "Cancelled: validation request (new document#1)"
     Then there is an audit entry containing "Reason: Meme of dog is no longer needed"
+
+  Scenario: I can view in the audit log when a planning application has been withdrawn
+    Given I press "Cancel application"
+    And I choose "Withdrawn by applicant"
```
freesteph, line 31: This should be a step called "Given the application is withdrawn by the applicant", the logic of doing it doesn't belong in this file

freesteph (commented):
Almost there; in the long term it'd be nice to remove the `auditing.feature` file and have all of these checks performed in the relevant feature file,where the action is performed. For example:

```
# validation_requests.features
Feature: I can create a validation request etc
  Given
  When
  Then
  And there is an audit log for "XXX"
```
Otherwise we have to duplicate all of these contexts separately in our dedicated `auditing.feature` file, which is going to get quite verbose (and huge) in the long term.

Something for a rainy day

freesteph:
> Almost there; in the long term it'd be nice to remove the `auditing.feature` file and have all of these checks performed in the relevant feature file,where the action is performed. For example:
> 
> ```
> # validation_requests.features
> Feature: I can create a validation request etc
>   Given
>   When
>   Then
>   And there is an audit log for "XXX"
> ```
> 
> Otherwise we have to duplicate all of these contexts separately in our dedicated `auditing.feature` file, which is going to get quite verbose (and huge) in the long term.
> 
> Something for a rainy day

PS: well aware I started this file and put the first specs in there :smile: my mistake, but as a team we should probably move towards side-effect assertions in dedicated feature files is what I'm saying

app/views/shared/_supporting_information.html.erb
melzreal: can do, never gave it a second thought but its a good point to keep in mind thanks 👌

features/auditing.feature
melzreal: I've changed it, but now the steps below this which state "contains the text moved to bermuda" have that text being set in a completely invisible part to this test - which makes it less explicit - we could be checking for any text and thus not knowing how it was set - which makes them more obscure and less readable?

app/views/shared/_supporting_information.html.erb
freesteph: For tone consistency I suggest we update the format of the timely one to include the "at". Will look better in our emails too.

app/views/shared/_supporting_information.html.erb
```diff
@@ -115,6 +112,38 @@
...
+    <div class="govuk-accordion__section-header">
+      <h2 class="govuk-accordion__section-heading">
+        <button class="govuk-accordion__section-button" id="audit-log">
+          Audit log
+        </button>
+        <span class="govuk-accordion__icon" aria-hidden="true"></span>
+      </h2>
+    </div>
+    <div class="govuk-accordion__section-content" aria-labelledby="audit-log">
+      <% if @audit.present? %>
```
freesteph, line 125: that's a lot of logic in the view. Also can we call it `@last_audit`, on its own it's hard to figure out which special audit it is

app/views/shared/_supporting_information.html.erb
```diff
@@ -115,6 +112,38 @@
...
+        <p class="govuk-body">
+          <% if @audit.activity_type.match("/*_validation_request_cancelled") %>
+            <%= render "audits/types/validation_request_cancelled", locals: { item: @audit } %>
+          <% elsif @audit.activity_type.include?("request") || @audit.activity_type.include?("document_received_at_changed") %>
+            <%= render "audits/types/#{@audit.activity_type}", locals: { item: @audit } %>
+          <% else %>
+            <%= render "audits/types/generic_audit_entry", locals: { item: @audit } %>
```
freesteph, line 132: can you abstract this into the helper? That's too much logic for the view.

```ruby
def audit_template(audit_item)
  if # etc...
end
```

features/step_definitions/auditing_steps.rb
```diff
@@ -3,15 +3,24 @@
...
 end
 
+Then("there is an audit entry containing {string} and {string}") do |content, value|
+  step "I visit the audit log"
+
+  steps %(
+    Then the page contains "#{content}"
+    And the page contains "#{value}"
+  )
+end
```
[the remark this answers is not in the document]
melzreal: I'll try the approach of searching within a table for the step to be more precise - I can't add two different lines within a single "And there is xx" because of the \n separators between the two lines (plus it'd look hellufmessy saying "Application returned \/n\ Applicant blah"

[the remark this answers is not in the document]
freesteph: I'd rather you have the copy merged together because it ties them together, otherwise the above case (where two different entries have the separate pieces of copy - whereas it should be within one) still pases.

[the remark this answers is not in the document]
melzreal: ohh wait I've found something saying to use triple quotes: https://stackoverflow.com/questions/35639150/is-it-possible-to-write-a-gherkin-step-on-multiple-lines
gonna give it a go

app/helpers/audit_helper.rb
```diff
@@ -87,7 +87,23 @@ def activity(type_of_activity, args = nil)
...
+    end
+  end
+
+  def audit_type(audit)
+    if audit.activity_type.match("/*_validation_request_cancelled")
+      render "audits/types/validation_request_cancelled", locals: { item: audit }
+    elsif audit.activity_type.include?("request") || audit.activity_type.include?("document_received_at_changed")
+      render "audits/types/#{audit.activity_type}", locals: { item: audit }
+    else
+      render "audits/types/generic_audit_entry", locals: { item: audit }
```
freesteph, line 106: sorry I was unclear but can you make this function solely compute the template name and keep the rendering in the view? I've made you a patch, if you're happy with it :gift: 

```patch
From f367cfa86d03e9d3f3402a0bbecbbe159eb11485 Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?St=C3=A9phane=20Maniaci?=
 <stephane.maniaci@unboxedconsulting.com>
Date: Fri, 10 Dec 2021 11:06:44 +0000
Subject: [PATCH] audit: make the helper do less

---
 app/helpers/audit_helper.rb                       | 8 ++++----
 app/views/shared/_supporting_information.html.erb | 6 +++---
 2 files changed, 7 insertions(+), 7 deletions(-)

diff --git a/app/helpers/audit_helper.rb b/app/helpers/audit_helper.rb
index 35a19acb..e2031a37 100644
--- a/app/helpers/audit_helper.rb
+++ b/app/helpers/audit_helper.rb
@@ -97,13 +97,13 @@ module AuditHelper
     end
   end
 
-  def audit_type(audit)
+  def audit_entry_template(audit)
     if audit.activity_type.match("/*_validation_request_cancelled")
-      render "audits/types/validation_request_cancelled", locals: { item: audit }
+      "validation_request_cancelled"
     elsif audit.activity_type.include?("request") || audit.activity_type.include?("document_received_at_changed")
-      render "audits/types/#{audit.activity_type}", locals: { item: audit }
+      audit.activity_type
     else
-      render "audits/types/generic_audit_entry", locals: { item: audit }
+      "generic_audit_entry"
     end
   end
 end
diff --git a/app/views/shared/_supporting_information.html.erb b/app/views/shared/_supporting_information.html.erb
index ec873b31..f67f6508 100644
--- a/app/views/shared/_supporting_information.html.erb
+++ b/app/views/shared/_supporting_information.html.erb
@@ -124,13 +124,13 @@
     <div class="govuk-accordion__section-content" aria-labelledby="audit-log">
       <% if @last_audit.present? %>
         <p class="govuk-body">
-          <%= audit_type(@last_audit) %>
-  
+          <%= render "audits/types/#{audit_entry_template(@last_audit)}", locals: { item: @last_audit } %>
+
           <p class="govuk-body">
             <%= define_audit_user(@last_audit) %>
           </p>
           <p class="govuk-body">
-            <%= @last_audit.created_at.to_date %> at <%= @last_audit.created_at.strftime("%H:%M") %> 
+            <%= @last_audit.created_at.to_date %> at <%= @last_audit.created_at.strftime("%H:%M") %>
           </p>
         </p>
       <% end %>
-- 
2.33.1

```

features/step_definitions/auditing_steps.rb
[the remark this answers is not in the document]
freesteph: good shout on the triple quotes - I couldn't get them to work, but don't know if you had any more luck? Here's a solution that I actually am finding nicer, because it removes the whole `\n` awkwardness; it uses the `normalize_ws` option of Capybara that turns these pesky newlines into normal spaces: https://rubydoc.info/github/jnicklas/capybara/Capybara%2FNode%2FMatchers:assert_text


```patch
From 74d5a0aea54f36d299d1d657b67709847cc65018 Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?St=C3=A9phane=20Maniaci?=
 <stephane.maniaci@unboxedconsulting.com>
Date: Fri, 10 Dec 2021 11:19:51 +0000
Subject: [PATCH] audit: simplify the feature tests

---
 features/auditing.feature                   |  8 ++++----
 features/step_definitions/auditing_steps.rb | 11 +----------
 2 files changed, 5 insertions(+), 14 deletions(-)

diff --git a/features/auditing.feature b/features/auditing.feature
index 46fac3b2..22701a87 100644
--- a/features/auditing.feature
+++ b/features/auditing.feature
@@ -43,9 +43,9 @@ Feature: Auditing a planning application
     And I check "Broads"
     And I uncheck "Conservation Area"
     And I press "Save"
-    Then there is an audit entry containing "Constraint added" and "National Park"
-    And there is an audit entry containing "Constraint added" and "Broads"
-    And there is an audit entry containing "Constraint removed" and "Conservation Area"
+    Then there is an audit entry containing "Constraint added National Park"
+    And there is an audit entry containing "Constraint added Broads"
+    And there is an audit entry containing "Constraint removed Conservation Area"
 
   Scenario: I can view an entry in the audit log showing application updates
     Given I press "Application information"
@@ -55,7 +55,7 @@ Feature: Auditing a planning application
     Then there is an audit entry containing "Address 1 updated"
     And there is an audit entry containing "Miyazaki"
     And there is an audit entry containing "Changed to: 20 leafy gardens"
-    
+
   Scenario: I can view an entry in the audit showing the application being validated
     When I press "Validate application"
     And I press "Mark the application as valid"
diff --git a/features/step_definitions/auditing_steps.rb b/features/step_definitions/auditing_steps.rb
index 890b0b35..982d775a 100644
--- a/features/step_definitions/auditing_steps.rb
+++ b/features/step_definitions/auditing_steps.rb
@@ -4,16 +4,7 @@ Then("there is an audit entry containing {string}") do |content|
   step "I visit the audit log"
 
   within_table("Audit log") do
-    expect(page).to have_content content
-  end
-end
-
-Then("there is an audit entry containing {string} and {string}") do |content, value|
-  step "I visit the audit log"
-
-  within_table("Audit log") do
-    expect(page).to have_content content
-    expect(page).to have_content value
+    expect(page).to have_content(content, normalize_ws: true)
   end
 end
 
-- 
2.33.1

```

app/helpers/audit_helper.rb
melzreal: ahh awesome I hadn't understood that's how you were intending it, will put in that logic thanks

features/step_definitions/auditing_steps.rb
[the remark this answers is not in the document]
melzreal: they weren't working for me either so I went with the variation you suggested, cheers!

features/auditing.feature
```diff
@@ -1,25 +1,64 @@
...
+    Then there is an audit entry containing "Application returned"
+    And there is an audit entry containing "Applicant sent selfies instead of floor plans"
+
+  Scenario: Updating constraints displays all updates in the audit log
+    Given I press "Constraints"
+    And I press "Update"
+    And I check "National Park"
+    And I check "Broads"
+    And I uncheck "Conservation Area"
+    And I press "Save"
```
freesteph, line 45: Ideally this would be When. Always Given-When-Then with Ands sprinkled if needed, but no repeat and they should all be there.

freesteph (approved):
Looks good, well done :coconut:

features/auditing.feature
melzreal: we can rework them all in the refactor (I got a separate branch for it called audit-refactor, that one is an 8 point story vs this 2 pointer) and also make sure that coverage is good for all of them, as the scope of that story is meant to be a big bigger

PR #508, opened 2021-12-14
Audit refactor

melzreal:
### Description of change

Make our controllers skinnier and have our models take over the audit process

### Story Link
https://trello.com/c/aEtEryaK/673-refactor-audit-into-model


app/controllers/planning_applications_controller.rb
```diff
@@ -58,12 +49,9 @@ def update
...
   def assign
     if request.patch?
-      @planning_application.user = if params[:planning_application][:user_id] == "0"
-                                     nil
-                                   else
-                                     current_local_authority.users.find(params[:planning_application][:user_id])
-                                   end
-      audit("assigned", nil, @planning_application.user&.name)
+      user = params[:planning_application][:user_id].eql?("0") ? nil : User.find(params[:planning_application][:user_id])
+      @planning_application.assign(user)
```
benbaumann95, line 53: Do we still need to scope users to `current_local_authority.users` ?
And a bit confused with the logic here of setting user to nil if the params is 0 🤔 
I mean will there ever be a user with id of 0

app/models/planning_application.rb
```diff
@@ -103,11 +126,52 @@ class PlanningApplication < ApplicationRecord
...
+      after do
+        audit("withdrawn", cancellation_comment)
+      end
     end
 
     after_all_transitions :timestamp_status_change # FIXME: https://github.com/aasm/aasm#timestamps
   end
 
+  def audit_created
+    audit("created", nil, Current.user.name)
```
freesteph, line 138: should we use just callbacks or just in-line blocks?

app/models/planning_application.rb
```diff
@@ -103,11 +126,52 @@ class PlanningApplication < ApplicationRecord
...
+  end
+
+  # we already have an auditable module but the arguments need to change. Worth calling super or not?
```
freesteph, line 141: why do the arguments need to change?

spec/factories/planning_application.rb
```diff
@@ -4,6 +4,8 @@
 
 FactoryBot.define do
   factory :planning_application do
+    before(:create) { |pa| pa.define_singleton_method(:audit_created) { true; } }
```
freesteph, line 7: what does this do?

melzreal: factory bot fails because it doesn't run the before create hooks by default, and so all planning applications were failing to create. The explanation / case for using the singleton method to allow for the planning application factory to create successfully is here:

https://stackoverflow.com/a/35562805

app/models/planning_application.rb
melzreal: the method in auditable expects an instance variable of @planning_application to be set, in order to grab the id - but its also a controller concern not a model concern - so separate concerns right?

app/models/planning_application.rb
melzreal: it is a callback - so this gets called after_create - but I couldn't pass an argument to the hook inline method to get the current user. Have also had to modify this one to make it a little more tolerant to nil users, see updated version

app/controllers/planning_applications_controller.rb
```diff
@@ -58,12 +49,9 @@ def update
 
   def assign
     if request.patch?
-      @planning_application.user = if params[:planning_application][:user_id] == "0"
-                                     nil
-                                   else
-                                     current_local_authority.users.find(params[:planning_application][:user_id])
-                                   end
-      audit("assigned", nil, @planning_application.user&.name)
+      user = params[:planning_application][:user_id].eql?("0") ? nil : User.find(params[:planning_application][:user_id])
```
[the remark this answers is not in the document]
melzreal: the same it did before but now in a single line ;) 

it assigns an application to nil if the params are 0 or to a user if user params are present. its for the purpose of this functionality here:
image
image

app/models/planning_application.rb
melzreal: (maybe a proc on the after_create hook could achieve the same I just didn't see how)

app/controllers/planning_applications_controller.rb
melzreal: see - https://github.com/unboxed/bops/pull/508#discussion_r768660347

melzreal: in the view: app/views/planning_applications/assign.html.erb

image

spec/factories/planning_application.rb
melzreal: I think it was actually failing not due to the before_create but due to there being nil users preventing the audit creation and thus the planning application creation, have added a bit more tolerance for nil values so the error went away - removed it

spec/models/planning_application_spec.rb
```diff
@@ -251,9 +251,7 @@
 
   describe "#reference" do
     it "pads the ID correctly" do
-      planning_application.update!(id: 1000)
-
-      expect(planning_application.reference).to eq "00001000"
+      expect(planning_application.reference).to eq planning_application.id.to_s.rjust(8, "0").to_s
     end
```
melzreal, line 255: this randomly started failing due to the planning application foreign key being bound with the audit foreign key, and thus not being editable. But de facto, wasn't a good test as we don't go around changing id's anyway?

app/controllers/planning_applications_controller.rb
benbaumann95: Hm I think we need to scope it to be safe. Because it's using params there is ability for a bad actor to tamper with it

app/models/concerns/auditable_model.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module AuditableModel
```
benbaumann95, line 3: Why call it this ? It already lives under the `models/` directory
Just call it `auditable`

spec/system/planning_applications/drawing_sitemap_spec.rb
```diff
@@ -38,7 +38,6 @@
 
       click_button "Audit log"
       click_link "View all audits"
-      expect(page).to have_content("Red line drawing created")
```
benbaumann95, line 41: why remove this spec ?

app/models/planning_application.rb
```diff
@@ -103,11 +126,52 @@ class PlanningApplication < ApplicationRecord
...
 
     after_all_transitions :timestamp_status_change # FIXME: https://github.com/aasm/aasm#timestamps
   end
 
+  def audit_created
+    audit("created", nil, Current.user&.name || Current.api_user&.name)
+  end
+
+  # we already have an auditable module but the arguments need to change. Worth calling super or not?
+  def audit(activity_type, audit_comment = nil, activity_information = nil, _api_user = nil)
```
benbaumann95, line 142: guessing we can remove this method and include the concern

config/environments/development.rb
```diff
@@ -68,6 +68,5 @@
 
   config.action_mailer.default_url_options = { host: (ENV["DOMAIN"] || "bops-care.link"), port: 3000 }
 
-  config.hosts << ".bops-care.link"
-  config.hosts << "southwark.southwark.web"
+  config.hosts.clear
```
benbaumann95, line 71: mistake commit ?

app/models/planning_application.rb
```diff
@@ -103,11 +126,52 @@ class PlanningApplication < ApplicationRecord
...
+      when "boundary_geojson"
+        audit("red_line_updated", "Red line drawing updated")
+      else
+        audit("updated",
+              "Changed from: #{saved_change_to_attribute(attribute_name).first}
+                \r\n Changed to: #{saved_change_to_attribute(attribute_name).second}",
+              attribute_name.humanize)
+      end
+    end
+  end
```
benbaumann95, line 173: This seems to be doing quite a lot of stuff.
Could we make sure the test coverage is high for different scenarios here (if not already)
Will this run for every update action? So even like address / contact info change on the planning application would hit the else block? Is this what we are supposed to be doing

app/models/concerns/auditable_model.rb
melzreal: we already have auditable under controllers, its confusing to call it the same?

spec/system/planning_applications/drawing_sitemap_spec.rb
melzreal: I unified the red line drawing audit actions, so now only the update action gets registered on the audit - since both create and update are the same its doesn't make sense to have two separate wordings for the same thing

PR #509, opened 2021-12-14
Support uploading multiple documents via additional document validation requests

benbaumann95:
### Description of change

- Update association so that AdditionalDocumentValidationRequest has many documents
- Document (optionally) belongs to AdditionalDocumentValidationRequest
- Wrap documents creation process in transaction, change state from open to closed and audit the request

### Story Link

https://trello.com/c/rs1kJhqb/487-allow-multiple-file-uploads-for-a-new-document-request-on-bops-applicant

### DIFF
--- a/app/controllers/api/v1/additional_document_validation_requests_controller.rb
+++ b/app/controllers/api/v1/additional_document_validation_requests_controller.rb
@@ -6,9 +6,14 @@ module Api
       skip_before_action :verify_authenticity_token
 
       before_action :check_token_and_set_application
-      before_action :check_file_params_are_present,
-                    :check_file_size,
-                    :check_file_type, only: :update
+      before_action :check_files_params_are_present,
+                    :check_files_size,
+                    :check_files_type,
+                    :current_api_user, only: :update
+
+      rescue_from AdditionalDocumentValidationRequest::UploadFilesError do |_exception|
+        render_failed_request
+      end
 
       def index
         respond_to do |format|
@@ -32,30 +37,34 @@ module Api
 
       def update
         @additional_document_validation_request = @planning_application.additional_document_validation_requests.find_by(id: params[:id])
-        new_document = @planning_application.documents.create!(file: params[:new_file])
-        @additional_document_validation_request.update!(state: "closed", new_document: new_document)
 
-        if @additional_document_validation_request.save
-
-          audit("additional_document_validation_request_received", document_audit_item(new_document),
-                @additional_document_validation_request.sequence, current_api_user)
+        if @additional_document_validation_request.can_upload?
+          @additional_document_validation_request.upload_files!(params[:files])
 
           render json: { message: "Validation request updated" }, status: :ok
         else
-          render json: { message: "Validation request could not be updated" }, status: :bad_request
+          render_failed_request
         end
       end
 
       private
 
-      def check_file_params_are_present
-        if params[:new_file].blank?
-          render json: { message: "A file must be selected to proceed." }, status: :bad_request
+      def check_files_params_are_present
+        if params[:files].empty?
+          render json: { message: "At least one file must be selected to proceed." }, status: :bad_request
         end
       end
 
-      def document_audit_item(document)
-        document.name
+      def check_files_size
+        if params[:files].map(&:size).sum > 30.megabytes
+          render json: { message: "The total file size must be 30MB or less" }, status: :bad_request
+        end
+      end
+
+      def check_files_type
+        if params[:files].any? { |file| Document::PERMITTED_CONTENT_TYPES.exclude? file.content_type }
+          render json: { message: "The file type must be JPEG, PNG or PDF" }, status: :bad_request
+        end
       end
     end
   end
--- a/app/models/additional_document_validation_request.rb
+++ b/app/models/additional_document_validation_request.rb
@@ -1,12 +1,43 @@
 # frozen_string_literal: true
 
 class AdditionalDocumentValidationRequest < ApplicationRecord
+  class UploadFilesError < RuntimeError; end
+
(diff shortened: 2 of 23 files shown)


app/controllers/api/v1/additional_document_validation_requests_controller.rb
```diff
@@ -32,30 +37,34 @@ def show
...
+      def check_files_params_are_present
+        if params[:files].empty?
+          render json: { message: "At least one file must be selected to proceed." }, status: :bad_request
         end
       end
 
-      def document_audit_item(document)
-        document.name
+      def check_files_size
+        if params[:files].map(&:size).reduce(0) { |total, size| total + size } > 30.megabytes
```
freesteph, line 59: `map(&:size).sum`

app/models/additional_document_validation_request.rb
```diff
@@ -1,12 +1,43 @@
...
+
+  has_many :documents, dependent: :destroy
 
   validates :document_request_type, presence: { message: "Please fill in the document request type." }
   validates :document_request_reason, presence: { message: "Please fill in the reason for this document request." }
+
+  def upload_files!(files)
+    transaction do
+      files.each do |file|
+        documents.create!(file: file, planning_application: planning_application)
```
freesteph, line 19: could you go straight to `planning_application.documents.create!`?

.rubocop_todo.yml
```diff
@@ -312,6 +312,7 @@ Layout/LineLength:
     - 'app/models/red_line_boundary_change_validation_request.rb'
     - 'app/views/api/v1/validation_requests/index.json.jbuilder'
     - 'spec/factories/planning_application.rb'
+    - 'spec/helpers/validation_request_helper_spec.rb'
```
freesteph, line 315: can't this be amended instead?

app/controllers/api/v1/additional_document_validation_requests_controller.rb
benbaumann95: 👀

.rubocop_todo.yml
benbaumann95: It's a string `url = "<a href=\"/planning_applications/#{planning_application.id}/documents/#{document.id}/edit\">proposed-floorplan.png</a>"`
I could just wrap this line in a rubocop:disable ?

app/models/additional_document_validation_request.rb
benbaumann95: yes you're probably right

app/models/additional_document_validation_request.rb
```diff
@@ -1,12 +1,43 @@
...
+
+  has_many :documents, dependent: :destroy
 
   validates :document_request_type, presence: { message: "Please fill in the document request type." }
   validates :document_request_reason, presence: { message: "Please fill in the reason for this document request." }
+
+  def upload_files!(files)
+    transaction do
+      files.each do |file|
+        planning_application.documents.create!(file: file, additional_document_validation_request: self)
```
benbaumann95, line 19: probably need to amend an association so that I don't need to do `additional_document_validation_request: self` .

.rubocop_todo.yml
freesteph: or use some string formatting method https://ruby-doc.org/core-3.0.3/String.html#method-i-25, but yeah I'm always uncomfortable with shelving a rule for a whole file because of a single offense... Up to you, not critical

app/views/audits/types/_additional_document_validation_request_received.html.erb
```diff
@@ -1,5 +1,4 @@
 <%= activity(locals[:item].activity_type, locals[:item].activity_information) %>
-<p class="govuk-body">
```
melzreal, line 2: this needs the p-govuk-body tag otherwise it shows without formatting in the audit log accordion, can you please re-add it?
image

app/views/validation_requests/index.html.erb
```diff
@@ -56,8 +56,18 @@
                   <td class="govuk-table__cell limit-column-width">
                     <%= request_closed_at(validation_request)  %>
                   </td>
-                  <td class="govuk-table__cell limit-column-width">
-                    <%= applicant_response(validation_request) %>
+                   <td class="govuk-table__cell limit-column-width">
+                    <% if validation_request.is_a?(AdditionalDocumentValidationRequest) %>
+                      <ul id="additional-documents" class="govuk-list">
+                        <% validation_request.documents.each do |document| %>
+                          <%= content_tag(:li, id: dom_id(document)) do %>
```
melzreal, line 63: nice one using content tags! I just learnt about this little dom_id magic, but after reading on it it seems from rails 5 you can also just call this a "tag" element, so this content_tag can be replaced with 
```
       <%= tag.validation_request do %>
       ```
and same thing with ids shows up on the dom! 
[https://www.justinweiss.com/articles/the-lesser-known-features-in-rails-5-dot-1/]

melzreal (approved):
Great job! Made a tiny request and a small suggestion on using tags but no blockers, its looking good ⭐

app/views/audits/types/_additional_document_validation_request_received.html.erb
benbaumann95: woops total accident. Nice catch!

app/views/validation_requests/index.html.erb
benbaumann95: Nice. To replicate what there was I think I need to do `<%= tag.li id: dom_id(document) do %>`

melzreal: I gave it a go with the little snippet above and it auto generated ids so that may be all you need

PR #511, opened 2021-12-14
Remove status from awaiting_determination

melzreal:
### Description of change
https://trello.com/c/0Id6IVnc/686-remove-status-column-on-awaiting-determination-tab

### DIFF
--- a/app/views/planning_applications/_planning_application_table.html.erb
+++ b/app/views/planning_applications/_planning_application_table.html.erb
@@ -1,20 +1,10 @@
-<div class="govuk-tabs__panel" id="<%= id %>">
+<div class="govuk-tabs__panel" id="<%= planning_application_status %>">
   <h2 class="govuk-heading-l"><%= title %></h2>
   <table class="govuk-table">
     <thead class="govuk-table__head">
     <tr class="govuk-table__row">
-      <th scope="col" class="govuk-table__header">Application number</th>
-      <th scope="col" class="govuk-table__header">Site address</th>
-      <th scope="col" class="govuk-table__header">Application type</th>
-      <th scope="col" class="govuk-table__header">Expiry date</th>
-      <% if id != "closed" %>
-        <th scope="col" class="govuk-table__header">Days left to expiry</th>
-      <% end %>
-      <th scope="col" class="govuk-table__header">Status</th>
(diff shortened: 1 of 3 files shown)


app/views/planning_applications/_planning_application_table.html.erb
```diff
@@ -3,18 +3,8 @@
...
-      <th scope="col" class="govuk-table__header">Expiry date</th>
-      <% if id != "closed" %>
-        <th scope="col" class="govuk-table__header">Days left to expiry</th>
-      <% end %>
-      <th scope="col" class="govuk-table__header">Status</th>
-      <% if id == "awaiting_determination" %>
-        <th scope="col" class="govuk-table__header">Recommendation date</th>
-      <% elsif id == "closed" %>
-        <th scope="col" class="govuk-table__header">Determination date</th>
+      <% header_options(id).each do |option| %>
```
benbaumann95, line 6: I know this existed already in the code but it's a bit confusing to have "id" mean the "status" here imo. Should we update the rendering of these partials to reflect what this really is.

app/helpers/index_helper.rb
```diff
@@ -0,0 +1,26 @@
...
+module IndexHelper
+  def header_options(id)
+    options = [
+      "Application number",
+      "Site address",
+      "Application type",
+      "Expiry date",
+      "Days left to expiry",
+      "Status"
+    ]
```
benbaumann95, line 12: Could be a constant?

benbaumann95 (commented):
Overall like what you're trying to do here but just a few comments

app/helpers/index_helper.rb
melzreal: I'm modifying it though in the next few lines though, isn't it a bit weird to be modifying a constant?

app/views/planning_applications/_planning_application_table.html.erb
melzreal: you're right, id is confusing - I've changed it to "planning_application_status" so its more readable

benbaumann95 (approved):
I think the helper could be more appropriately named, now that we've updated to reflect these are statuses and not id's.
Otherwise looks good!

PR #513, opened 2021-12-15
Show last audit item in validate form

melzreal:
### Description of change
Needed to set the @last_audit instance variable in two views instead of just one in order to show the latest audit from both show page and from validate_form which is also the show page -.-

### DIFF
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -6,6 +6,8 @@ class PlanningApplicationsController < AuthenticationController
   before_action :ensure_user_is_reviewer, only: %i[review review_form]
   before_action :ensure_constraint_edits_unlocked, only: %i[edit_constraints_form edit_constraints]
 
+  before_action :set_last_audit, only: %i[show validate_form]
+
   def index
     @planning_applications = if helpers.exclude_others? && current_user.assessor?
                                current_local_authority.planning_applications.where(user_id: current_user.id).order("created_at DESC").or(
@@ -16,9 +18,7 @@ class PlanningApplicationsController < AuthenticationController
                              end
   end
 
-  def show
-    @last_audit = @planning_application.audits.last
-  end
+  def show; end
 
   def new
     @planning_application = PlanningApplication.new
@@ -341,5 +341,9 @@ class PlanningApplicationsController < AuthenticationController
 
   def ensure_constraint_edits_unlocked
     render plain: "forbidden", status: :forbidden and return unless @planning_application.can_validate?
+  end
+
+  def set_last_audit
+    @last_audit = @planning_application.audits.last if @planning_application.present?
   end
 end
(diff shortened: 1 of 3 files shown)


app/controllers/planning_applications_controller.rb
```diff
@@ -17,7 +17,7 @@ def index
   end
 
   def show
-    @last_audit = @planning_application.audits.last
+    last_audit
```
benbaumann95, line 20: can this belong in a `before_action` ?

app/controllers/planning_applications_controller.rb
```diff
@@ -68,7 +68,9 @@ def assign
     end
   end
 
-  def validate_form; end
+  def validate_form
+    last_audit
```
benbaumann95, line 72: same

app/controllers/planning_applications_controller.rb
melzreal: I did try that (mentioned in PR comment) but unexpected consequences happened - the policy references class broke. For some reason, adding a before action to planning application controller made it so the policy references also attempted to call the show action. 

From reading on it, callbacks are dangerous and where possible should be avoided - so whenever we can find an implementation that doesn't strictly need a callback (if we don't need it to be tied to an exact transaction). Its dangerous to sprinkle them all over due to the unwanted side effects - and in this case, using a callback to set an instance variable isn't quite a "strictly necessary" case - https://engineering.gusto.com/the-rails-callbacks-best-practices-used-at-gusto/

 https://thoughtbot.com/upcase/videos/callbacks

after that

benbaumann95: You've given examples of _active record_ callbacks used badly. They aren't dangerous if used in the correct way.
But this is just a standard controller action which means execute this block before we render the show or validate form method. Nothing dangerous about it. 
Because you are using `@planning_application` in the `last_audit` method you will need to add the routes to the `set_planning_application` callback too. I suspect this is what is blowing up ?

benbaumann95: Why would there be an unwanted side effect in this use case ?

melzreal: the unwanted side effect happens because of a behaviour in PolicyClasses executing the code in PlanningApplication before its own.

it calls the set_last_audit method before it calls its own set planning application method:

image


you can check this out by popping in a byebug in the set_audit method and in the set_planning_application policy class controller, then

 ```bundle exec cucumber features/adding_policy_assessment_areas.feature``` 

Whenever we get to the edit page of a policy, it can't find the planning application that was set by its own controller, because its running its own version of set_planning_application but first it runs the methods in planning application, so this happens:
image


I can go around it with an if statement just to check if the planning application is there so I'll push that, but it does mean more calls are being made than with the precious version of the code.

benbaumann95: hmm this makes me think we might have some non ideal architecture going on.
Looks like there is some inheritance from the two controllers and therefore calling this method when it's probably not appropriate.
I think what you've done is fine for now but maybe a deeper look will be good going forward. 

Although when I comment out the `set_last_audit` action, none of the tests fail. You think it's worth adding something to cover this ? I would have expected a spec to fail if this code didn't exist

melzreal: ah, yes I didn't add coverage for this bit of text showing, I'll pop in a quick cucumber

benbaumann95 (approved):

PR #514, opened 2021-12-20
Add original geojson for red line boundary change validation requests

benbaumann95:
### Description of change

- Add original geojson for red line boundary change validation requests
- Add necessary migrations, including non null constrains for existing columns
- Keep track of original/new geojson data so we can display this state change rather than overriding the data

### Story Link

https://trello.com/c/K4yH4Ijr/532-keep-track-record-of-previous-red-line-in-bops-applicants


spec/models/red_line_boundary_change_validation_request_spec.rb
```diff
@@ -3,5 +3,77 @@
 require "rails_helper"
 
 RSpec.describe RedLineBoundaryChangeValidationRequest, type: :model do
+  # rubocop:disable Layout/LineLength
```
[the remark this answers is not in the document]
benbaumann95: Can do for now. This cop is very annoying 😠 , as long as we aren't silly with line lengths i'm not sure we need this

app/views/red_line_boundary_change_validation_requests/show.html.erb
```diff
@@ -21,14 +21,14 @@
     <p class="govuk-body">
       <%= @red_line_boundary_change_validation_request.reason  %>
     </p>
-    <%= render "shared/location_map", locals: { geojson: @planning_application.boundary_geojson } %>
+    <%= render "shared/location_map", locals: { geojson: @red_line_boundary_change_validation_request.original_geojson } %>
```
melzreal, line 24: I'm getting this error for the shared/location_map view:

image

app/views/red_line_boundary_change_validation_requests/show.html.erb
```diff
@@ -21,14 +21,14 @@
...
 
     <h2 class="govuk-heading-m govuk-!-padding-top-6">
      Proposed red line boundary
     </h2>
     <%= render "shared/location_map", locals: { geojson: @red_line_boundary_change_validation_request.new_geojson } %>
```
melzreal, line 29: we need to prevent some errors on this bit, some ideas:
1. Update the db for old red_line boundaries to have both old and new geojson (the old geojson can be set to the current one for ex (like this migration update ex: db/migrate/20210910145259_add_reply_to_local_authority.rb)

2. Add a little if statement to make sure this doesn't blow up if no "old" and "new" fields exist

3. Wrap this in a helper method that renders a default string for "No previous geojson" available if the view doesn't exist else render partial

app/views/red_line_boundary_change_validation_requests/show.html.erb
benbaumann95: hm would like to see your db for this

melzreal: seems to have resolved itself

melzreal (approved):
:shipit:

PR #515, opened 2021-12-20
Split out CI build into smaller jobs

benbaumann95:
### Description of change

- Break up jobs into smaller steps for much quicker parallel builds

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,12 +4,9 @@ name: CI
 on: [push, pull_request]
 
 jobs:
-  test:
+  bundle-audit:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -22,6 +205,99 @@ jobs:
...
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.6.6
+          bundler-cache: true
+
+      - name: Install Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+          cache: 'yarn'
```
melzreal, line 226: should we use double quotes everywhere for consistency?

PR #516, opened 2021-12-30
Edit validation requests

melzreal:
### Description of change

Edit red line, other validations and new document requests.

Doesn't edit "replacement document validation requests" since those are always to replace an existing invalid document and don't have any amendable features, so they need to be cancelled and the invalid document changed if amendments were needed.

### Story Link

https://trello.com/c/bYbWqYxC/684-edit-function-when-writing-validation-request

### DIFF
--- a/app/helpers/validation_request_helper.rb
+++ b/app/helpers/validation_request_helper.rb
@@ -58,6 +58,12 @@ module ValidationRequestHelper
          validation_request)
   end
 
+  def edit_validation_request_url(validation_request)
+    link_to "Edit",
+            send("edit_planning_application_#{request_type(validation_request)}_path", validation_request.planning_application,
+                 validation_request)
+  end
+
   def document_url(document)
     link_to(document.name.to_s,
             edit_planning_application_document_path(document.planning_application, document.id))
--- a/app/helpers/validation_request_helper.rb
+++ b/app/helpers/validation_request_helper.rb
@@ -58,10 +58,12 @@ module ValidationRequestHelper
          validation_request)
   end
 
-  def edit_validation_request_url(validation_request)
-    link_to "Edit",
-            send("edit_planning_application_#{request_type(validation_request)}_path", validation_request.planning_application,
-                 validation_request)
+  def edit_validation_request_url(planning_application, validation_request)
+    type = request_type(validation_request)
+
+    unless type.eql?("replacement_document_validation_request")
+      link_to "Edit", [:edit, planning_application, validation_request]
+    end
   end
 
   def document_url(document)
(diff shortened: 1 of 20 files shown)


app/helpers/validation_request_helper.rb
```diff
@@ -58,6 +58,14 @@ def cancel_request_url(planning_application, validation_request)
...
   end
 
+  def edit_validation_request_url(validation_request)
+    link_to "Edit",
+            send("edit_planning_application_#{request_type(validation_request)}_path", validation_request.planning_application,
+                 validation_request)
+  rescue NoMethodError
+    nil
+  end
+
```
melzreal, line 68: note: this is so document replacements just have an empty string under the column instead of "Edit": 
image
had tried leaving the rescue empty but rubocop complained so we're returning nil instead.

config/environments/development.rb
```diff
@@ -68,6 +68,5 @@
 
   config.action_mailer.default_url_options = { host: (ENV["DOMAIN"] || "bops-care.link"), port: 3000 }
 
-  config.hosts << ".bops-care.link"
-  config.hosts << "southwark.southwark.web"
+  config.hosts.clear
```
melzreal, line 71: I keep having to do this for the local environment to work, should I separate it into its own commit and pop it outta here?

app/helpers/validation_request_helper.rb
```diff
@@ -58,6 +58,14 @@ def cancel_request_url(planning_application, validation_request)
          validation_request)
   end
 
+  def edit_validation_request_url(validation_request)
+    link_to "Edit",
+            send("edit_planning_application_#{request_type(validation_request)}_path", validation_request.planning_application,
+                 validation_request)
+  rescue NoMethodError
```
pixeltrix, line 65: Using exceptions like this isn't a good pattern - better to check the value returned by `request_type` is valid, e.g.

``` ruby
def edit_validation_request_url(validation_request)
  type = request_type(validation_request)
  planning_application = validation_request.planning_application

  unless type == :replacement_document
    link_to "Edit", [:edit, planning_application, validation_request]
  end
end
```

The array in `link_to` triggers polymorphic routing:
https://api.rubyonrails.org/classes/ActionDispatch/Routing/PolymorphicRoutes.html

config/environments/development.rb
pixeltrix: I think this was fixed in the latest 6.1 release? I remember seeing something about `config.hosts` starting with `.`

pixeltrix (changes requested):
Everything else seems fine but that helper method needs changing.

config/environments/development.rb
melzreal: it only started happening after the 6.1 release, now no matter what I put in config.hosts it asks me to add it to config.hosts - unless I call config.hosts.clear it doesn't work locally 🤷

app/helpers/validation_request_helper.rb
melzreal: thanks, change made

pixeltrix (approved):

benbaumann95 (approved):

PR #518, opened 2022-01-04
Travel to specific time in spec to keep consistent runs

benbaumann95:
- This spec would fail depending on what day you run the test due to the dynamic nature in which it was previously running in


melzreal (approved):
:shipit:

PR #527, opened 2022-01-14
Draw site outline is zoomed in on latitude/longitude when no boundary set

benbaumann95:
### Description of change

- If no boundary is set/drawn then planning applications created via RIPA with latitude/longitude fields within the site object is used as a starting point for the site outline

### Story Link

https://trello.com/c/hV8lG0L0/700-draw-site-outline-should-start-on-site-address

### DIFF
--- a/app/views/planning_applications/draw_sitemap.html.erb
+++ b/app/views/planning_applications/draw_sitemap.html.erb
@@ -18,6 +18,8 @@
 <%= form_with model: @planning_application, url: update_sitemap_planning_application_path(@planning_application), local: true, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |form| %>
   <%= form.hidden_field :boundary_geojson %>
   <%= render "shared/location_map", locals: { geojson: @planning_application.boundary_geojson, geojson_field: 'planning_application_boundary_geojson' } %>
-  <%= form.govuk_submit "Save" %>
-  <%= link_to "Back", planning_application_path(@planning_application), class: "govuk-button govuk-button--secondary" %>
+  <div class="govuk-button-group govuk-!-margin-top-3">
+    <%= form.govuk_submit "Save" %>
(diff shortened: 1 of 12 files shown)


app/views/planning_applications/draw_sitemap.html.erb
```diff
@@ -18,6 +18,8 @@
 <%= form_with model: @planning_application, url: update_sitemap_planning_application_path(@planning_application), local: true, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |form| %>
   <%= form.hidden_field :boundary_geojson %>
   <%= render "shared/location_map", locals: { geojson: @planning_application.boundary_geojson, geojson_field: 'planning_application_boundary_geojson' } %>
-  <%= form.govuk_submit "Save" %>
-  <%= link_to "Back", planning_application_path(@planning_application), class: "govuk-button govuk-button--secondary" %>
+  <div class="govuk-button-group govuk-!-margin-top-3">
```
EGiataganas, line 21: 👍

db/migrate/20220112155627_add_latitude_and_longitude_to_planning_applications.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddLatitudeAndLongitudeToPlanningApplications < ActiveRecord::Migration[6.1]
+  # rubocop:disable Rails/BulkChangeTable
+  def change
+    add_column :planning_applications, :latitude, :string
+    add_column :planning_applications, :longitude, :string
+  end
+  # rubocop:enable Rails/BulkChangeTable
+end
```
EGiataganas, line 10: How about including the `bulk` option 

```ruby
def change
  change_table :planning_applications, bulk: true do |t|
    t.string :latitude
    t.string :longitude
  end
end
```

benbaumann95: Ah, good to know. Will update

PR #531, opened 2022-01-18
Extract planning application AASM status into concern

benbaumann95:
### Description of change

- This logic can live in a separate concern module and be included in the model PlanningApplication. This should help with easier navigation and declutter our model a bit
- This PR also moves over the state specific specs to its own file


EGiataganas (approved):
Great job, LGTM :shipit:

PR #532, opened 2022-01-18
Update to ruby v2.7.5

EGiataganas:
### Description of change

Update to ruby v2.7.5 & bundler v2.3.5


benbaumann95:
Do we need to coordinate this with our terraform?
I guess we can update https://github.com/unboxed/bops-applicants/pulls too

benbaumann95 (approved):

pixeltrix:
This needs to be deployable on the current 2.6 instances so when instance refresh is triggered it can deploy the last successful deploy - if the last deploy doesn't work on 2.7 the new instance will get stuck in a failure loop

PR #538, opened 2022-01-21
Update master to main as default branch

benbaumann95:


lib/package_builder.rb
```diff
@@ -158,20 +158,16 @@ def current_revision
     end
   end
 
-  def master_ref?
-    ENV.fetch("GITHUB_REF", "") == "refs/heads/master"
-  end
-
-  def production_ref?
```
benbaumann95, line 165: we don't deploy to prod like this anymore. It is no a review process and manual promote

EGiataganas (approved):

PR #540, opened 2022-01-26
Withdraw submitted application

benbaumann95:
### Description of change

Submit and withdraw recommendation for planning applications

### Story Link

https://trello.com/c/TiHzjcNz/710-officer-should-be-able-to-edit-their-assessment-post-submission-currently-locked

### DIFF
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -6,7 +6,9 @@ class PlanningApplicationsController < AuthenticationController
   before_action :ensure_user_is_reviewer, only: %i[review review_form]
   before_action :ensure_constraint_edits_unlocked, only: %i[edit_constraints_form edit_constraints]
 
-  before_action :set_last_audit, only: %i[show validate_form]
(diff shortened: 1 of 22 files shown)


app/controllers/planning_applications_controller.rb
```diff
@@ -135,35 +143,65 @@ def recommend
...
 
-  def submit_recommendation; end
+  def submit_recommendation
+    respond_to do |format|
+      if @planning_application.can_submit_recommendation?
+        format.html { render :submit_recommendation }
+      else
+        format.html { render plain: "Not Found", status: :not_found }
+      end
+    end
```
EGiataganas, line 160: is there any particular reason of using `respond_to` here since you're returning always html?

benbaumann95: https://blog.arkency.com/2016/07/respond-to-format-is-useful-even-without-multiple-formats/

benbaumann95: I'm just quite used to this style of rendering layouts too.

EGiataganas (approved):

PR #546, opened 2022-02-01
Audit refactor part 1

EGiataganas:
### Description of change

Audit refactor

### Story Link

https://trello.com/c/aEtEryaK

### DIFF
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -68,12 +68,9 @@ class PlanningApplicationsController < AuthenticationController
 
   def assign
     if request.patch?
-      @planning_application.user = if params[:planning_application][:user_id] == "0"
-                                     nil
-                                   else
-                                     current_local_authority.users.find(params[:planning_application][:user_id])
-                                   end
-      audit("assigned", nil, @planning_application.user&.name)
+      user = params[:planning_application][:user_id].eql?("0") ? nil : current_local_authority.users.find(params[:planning_application][:user_id])
+      @planning_application.assign(user)
+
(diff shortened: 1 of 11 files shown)


app/controllers/planning_applications_controller.rb
```diff
@@ -70,12 +68,9 @@ def update
 
   def assign
     if request.patch?
-      @planning_application.user = if params[:planning_application][:user_id] == "0"
-                                     nil
-                                   else
-                                     current_local_authority.users.find(params[:planning_application][:user_id])
-                                   end
-      audit("assigned", nil, @planning_application.user&.name)
+      user = params[:planning_application][:user_id].eql?("0") ? nil : current_local_authority.users.find(params[:planning_application][:user_id])
```
benbaumann95, line 71: I know this was from before but this just looks a bit wrong getting the user like this haha. I think this whole controller action needs a rethink

app/models/planning_application.rb
```diff
@@ -376,6 +379,12 @@ def withdraw_last_recommendation!
     raise WithdrawRecommendationError, e.message
   end
 
+  def assign(user)
+    self.user = user
+
+    audit_created!("assigned", self.user&.name)
+  end
```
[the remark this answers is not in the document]
EGiataganas: Happy to rename it to `assigned_to`, and further refactor can be on a later PR once we move all the audit actions from controllers to models

app/controllers/planning_applications_controller.rb
EGiataganas: Agreed can be on a later PR

benbaumann95 (approved):

PR #553, opened 2022-02-07
Add user_role enum to planning applications

benbaumann95:
### Description of change

- Add applicant_role to planning applications
- Update this field on legacy planning applications based on if an agent contact detail has been provided
- Improve error handling on planning application post requests

### Story Link

https://trello.com/c/yBWBLTkn/716-applicant-role-types


EGiataganas:
This is a good PR, although I am wondering whether is best to make a new table `Applicant` which will have the following:
- role
- first_name
- last_name
- phone
- email
And then the `planning_application` will have one `applicant`

benbaumann95:
> This is a good PR, although I am wondering whether is best to make a new table `Applicant` which will have the following:
> 
> * role
> * first_name
> * last_name
> * phone
> * email
>   And then the `planning_application` will have one `applicant`

Yeah I originally was thinking about doing this but when role is proxy it becomes a bit problematic filling in the rest of those fields with the correct values. (I'll discuss this with you offline)

EGiataganas (approved):

PR #554, opened 2022-02-07
Audit refactor p5

EGiataganas:


app/models/planning_application.rb
```diff
@@ -482,6 +478,36 @@ def create_audit!
...
+        next if saved_change_to_attribute(attribute_name).all?(&:blank?)
+
+        attribute_to_audit(attribute_name)
+      end
+    end
+  end
+
+  def attribute_to_audit(attribute_name)
+    if attribute_name.eql?("constraints")
+      audit_constraits!(saved_changes)
```
benbaumann95, line 493: I have a feeling we should be creating a class to help with this. Probably out of scope for now but there's probably enough complexity to warrant this

app/models/concerns/auditable_model.rb
```diff
@@ -7,13 +7,23 @@ module AuditableModel
...
+        )
+      else
+        planning_application.audits.create!(
+          user: Current.user,
+          activity_type: activity_type,
+          activity_information: activity_information,
+          audit_comment: audit_comment,
+          api_user: Current.api_user
+        )
+      end
```
benbaumann95, line 26: How about an `audits` method in `document.rb` and `validation_request.rb`?

```ruby
def audits
  planning_application.audits
end
  ```
  
Then this can just be:

```ruby
def audit_created!(activity_type:, activity_information: nil, audit_comment: nil)
   audits.create!(
      user: Current.user,
      activity_type: activity_type,
      audit_comment: audit_comment,
      api_user: Current.api_user
   )
end
  ```

So no conditional would be needed here

File: app/models/document.rb [hunk 3]
```diff
@@ -78,11 +82,16 @@ def referenced_in_decision_notice?
 
   def archive(archive_reason)
     unless archived?
-      update!(archive_reason: archive_reason,
-              archived_at: Time.zone.now)
+      update!(archive_reason: archive_reason, archived_at: Time.zone.now)
+      audit_created!(activity_type: "archived", activity_information: file.filename, audit_comment: archive_reason)
```
benbaumann95, line 86: should this be wrapped in a transaction ? And do we need any error handling

app/models/document.rb
```diff
@@ -78,11 +82,16 @@ def referenced_in_decision_notice?
...
     end
   end
 
+  def unarchive!
+    update!(archived_at: nil)
+    audit_created!(activity_type: "unarchived", activity_information: file.filename)
```
benbaumann95, line 92: Same comment as above here

app/models/planning_application.rb
```diff
@@ -482,6 +478,36 @@ def create_audit!
     audit_created!(activity_type: "created", activity_information: Current.api_user&.name || Current.user&.name)
   end
 
+  def audit_updated!
+    if saved_changes?
+      saved_changes.keys.intersection(PLANNING_APPLICATION_PERMITTED_KEYS).map do |attribute_name|
+        next if saved_change_to_attribute(attribute_name).all?(&:blank?)
```
benbaumann95, line 484: I think we should add a spec for this. Maybe just a simple system spec will do . There wasn't any coverage before so would be good to add something now

app/models/concerns/planning_application_status.rb
```diff
@@ -80,6 +80,8 @@ module PlanningApplicationStatus
 
       event :request_correction do
         transitions from: :awaiting_determination, to: :awaiting_correction
+
+        after { audit_created!(activity_type: "challenged", audit_comment: recommendations&.last&.reviewer_comment) }
```
benbaumann95, line 84: Wondering if there is a way to actually pass in the relevant db record because I've just realised that maybe grabbing the last record like this may be subject to a race condition.

Have a look at https://github.com/aasm/aasm/issues/506#issuecomment-532434609

I realise I may not have done this for some previous work but we should probably use this Proc style approach to ensure we are using the right record.

what are your thoughts?

app/models/concerns/auditable_model.rb
benbaumann95: Or even better i think you can just use a delegator
i.e. 
```ruby
delegate :audits, to: :planning_application
```

benbaumann95 (approved):

PR #555, opened 2022-02-08
Add osVectorTilesApiKey to  AWS param store

benbaumann95:
### Description of change

Not sure why we were exposing this API key publicly but this moves the logic to AWS parameter store

### Story Link

https://trello.com/c/vbc7MVod/741-move-api-keys-to-aws-parameter-store

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,6 +47,7 @@ services:
       API_USER: api_user
       PORT: 3001
       PROTOCOL: http
+      OS_VECTOR_TILES_API_KEY: ${OS_VECTOR_TILES_API_KEY}
     ports:
       - "3001:3001"
     volumes:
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -36,7 +37,7 @@ services:
         target: /app
   applicants:
     build: ./bops-applicants
-    entrypoint: ./bops-applicants/docker-entrypoint.sh
```
EGiataganas, line 39: This should stay as by changing it to `./docker-entrypoint.sh` it will use the `./docker-entrypoint.sh` from bops, not from `bops-applicants`

benbaumann95: Does it work for you doing this though?

I'm getting an error

benbaumann95: I pasted the error message here: https://github.com/unboxed/bops/pull/555/commits/0d5f9d770593097839682290981e3effbdc1246f

EGiataganas (approved):

PR #557, opened 2022-02-10
Display warning text when address / boundary geojson has been manually updated

benbaumann95:
### Description of change

- Display warning text when address / boundary geojson has been manually updated

### Story Link

https://trello.com/c/OpJtIGV9/765-include-a-prompt-when-address-or-polygons-are-updated-to-check-constraints-and-proposal-details

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -37,6 +37,7 @@ class PlanningApplication < ApplicationRecord
   after_create :create_audit!
   before_update :set_key_dates
   after_update :audit_updated!
+  after_update :address_or_boundary_geojson_updated?
 
   WORK_STATUSES = %w[proposed existing].freeze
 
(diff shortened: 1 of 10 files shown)


db/schema.rb
```diff
@@ -237,6 +237,8 @@
     t.datetime "closed_at"
     t.jsonb "planx_data"
     t.datetime "determination_date"
+    t.integer "applicant_role"
```
EGiataganas, line 240: This shouldn't be in this schema 😄

app/models/planning_application.rb
```diff
@@ -488,6 +497,14 @@ def audit_updated!
     end
   end
 
+  def address_or_boundary_geojson_updated?
+    return if updated_address_or_boundary_geojson
+
+    if saved_changes.keys.intersection(ADDRESS_AND_BOUNDARY_GEOJSON_FIELDS).any?
+      update!(updated_address_or_boundary_geojson: true)
+    end
```
EGiataganas, line 505: How about add an audit for this?

benbaumann95: The updates to any of those fields are already captured in the audit log so hopefully that should cover this

db/schema.rb
benbaumann95: woops !

EGiataganas (approved):

PR #567, opened 2022-02-23
Extract the upload_document functionality to job

EGiataganas:
### Description of change

Extract the upload_document functionality to job


benbaumann95 (approved):
Is there anything with AWS we need to do here?
Probably worth testing with swagger straight after we deploy this to see if the documents are being uploaded properly

PR #579, opened 2022-03-10
Validation wizard/constraints

EGiataganas:
### Description of change

Add checked for constraints on validation wizard

### Story Link

https://trello.com/c/bfwwyA1D

### DIFF
--- a/app/controllers/constraints_controller.rb
+++ b/app/controllers/constraints_controller.rb
@@ -4,6 +4,7 @@ class ConstraintsController < AuthenticationController
   before_action :set_planning_application
   before_action :ensure_constraint_edits_unlocked, only: %i[edit update]
 
+  def show; end
 
   def edit; end
 
@@ -12,6 +13,24 @@ class ConstraintsController < AuthenticationController
       redirect_to @planning_application, notice: "Constraints have been updated"
     else
       render :edit
+    end
+  end
+
+  def check
+    @planning_application.constraints_checked!
+
+    respond_to do |format|
+      if @planning_application.constraints_checked?
+        format.html do
+          redirect_to planning_application_validation_tasks_path(@planning_application),
+                      notice: "Constraints was successfully checked."
+        end
+      else
+        format.html do
+          redirect_to planning_application_validation_tasks_path(@planning_application),
+                      alert: "Couldn't check constraints - please contact support."
+        end
+      end
     end
   end
 
--- a/app/controllers/constraints_controller.rb
+++ b/app/controllers/constraints_controller.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+class ConstraintsController < AuthenticationController
+  before_action :set_planning_application
+  before_action :ensure_constraint_edits_unlocked, only: %i[edit update]
+
+
+  def edit; end
+
+  def update
+    if @planning_application.update(constraints: constraints_params[:constraints].compact_blank)
+      redirect_to @planning_application, notice: "Constraints have been updated"
+    else
+      render :edit
+    end
+  end
+
+  private
+
+  def constraints_params
+    params.require(:planning_application).permit(constraints: [])
+  end
+
+  def ensure_constraint_edits_unlocked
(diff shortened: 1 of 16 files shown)


app/controllers/constraints_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  before_action :set_planning_application
+  before_action :ensure_constraint_edits_unlocked, only: %i[edit update]
+
+  def show; end
+
+  def edit; end
+
+  def update
+    @planning_application.constraints = params[:planning_application][:constraints].reject(&:blank?)
+    if @planning_application.save!
```
benbaumann95, line 13: If we use a `bang!` here we need to rescue the error I think otherwise it will just blow up and won't reach the `else` block

app/controllers/constraints_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+class ConstraintsController < AuthenticationController
+  before_action :set_planning_application
+  before_action :ensure_constraint_edits_unlocked, only: %i[edit update]
+
+  def show; end
+
+  def edit; end
+
+  def update
+    @planning_application.constraints = params[:planning_application][:constraints].reject(&:blank?)
```
benbaumann95, line 12: Can we not update this by using the conventional Rails whitelisting params ?

benbaumann95, line 12: Can we not update this by using the conventional Rails whitelisting params ?

app/controllers/constraints_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    if @planning_application.save!
+      redirect_to @planning_application, notice: "Constraints have been updated"
+    else
+      render :edit
+    end
+  end
+
+  def check
+    @planning_application.constraints_checked!
```
benbaumann95, line 21: This will blow up if we don't rescue the error in the method unless that is the intention

benbaumann95, line 21: This will blow up if we don't rescue the error in the method unless that is the intention

app/controllers/constraints_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+        end
+      end
+    end
+  end
+
+  private
+
+  def ensure_constraint_edits_unlocked
+    render plain: "forbidden", status: :forbidden and return unless @planning_application.can_validate?
+  end
```
benbaumann95, line 42: 👍

benbaumann95, line 42: 👍

app/views/constraints/show.html.erb
```diff
@@ -0,0 +1,13 @@
+<h1 class="govuk-heading-l">Check the constraints</h1>
+
+<h2 class="govuk-heading-m">Constraints identified by <%= nil %></h2>
```
benbaumann95, line 3: Why is there `nil` here ?

benbaumann95, line 3: Why is there `nil` here ?

config/routes.rb
```diff
@@ -43,6 +41,10 @@
       get :validation_decision
     end
 
+    resource :constraints, only: %i[show edit update] do
+      post :check
```
benbaumann95, line 45: Is it post or patch here?
I have a feeling it could be patch as we are amending a record rather than creating a new resource

benbaumann95, line 45: Is it post or patch here?
I have a feeling it could be patch as we are amending a record rather than creating a new resource

benbaumann95 (commented):
Please could we also add a system spec of someone actually checking the constraints and then checking the audit log for the entry.

benbaumann95 (approved):
Looks good. Can we just add a tiny check in your spec `constraints_spec.rb` to check the audit log entry so we can cover https://github.com/unboxed/bops/pull/579/files#diff-9a62a01a9fac72dcc6b007a04caffa572946ff4a2b9b2ff57f2c3ea94867100cR446

PR #583, opened 2022-03-23
Validation wizard - missing documents

benbaumann95:
### Description of change

Validation wizard - add missing documents to the tasklist 
(see commits for more detail)

### Story Link

https://trello.com/c/A5kaoYRs/724-validation-wizard-6-add-missing-documents-to-the-tasklist


db/migrate/20220321164352_add_documents_missing_to_planning_applications.rb
```diff
@@ -0,0 +1,25 @@
...
+class AddDocumentsMissingToPlanningApplications < ActiveRecord::Migration[6.1]
+  class PlanningApplication < ApplicationRecord
+    has_many :additional_document_validation_requests
+  end
+
+  class AdditionalDocumentValidationRequest < ApplicationRecord
+    belongs_to :planning_application
+
+    scope :open_or_pending, -> { where(state: %i[open pending]) }
+  end
```
EGiataganas, line 12: This is not necessary here

app/controllers/validation_requests_controller.rb
```diff
@@ -10,16 +10,6 @@ def index
 
   def new; end
```
EGiataganas: Do we want this, also shall we remove these actions from [routes](https://github.com/unboxed/bops/blob/main/config/routes.rb#L62)?

benbaumann95: I did remove - look at the last commit https://github.com/unboxed/bops/pull/583/commits/4ad264dfde7cef634b780c65644dc5f20b441081

app/controllers/validation_requests_controller.rb
```diff
@@ -10,16 +10,6 @@ def index
...
-      redirect_to new_planning_application_additional_document_validation_request_path
-    else
-      flash.now[:error] = "You must select a validation request type to proceed."
-      render "new"
-    end
-  end
-
   private
 
   def send_validation_request_email(request)
```
EGiataganas, line 13: I am also wondering whether we can move these methods on [validation_requests](https://github.com/unboxed/bops/blob/main/app/controllers/concerns/validation_requests.rb)

app/controllers/validation_requests_controller.rb
EGiataganas: I was going commit by commit so I missed it! Good job 😄

EGiataganas (approved):

PR #591, opened 2022-04-12
Add 2FA to login process

benbaumann95:
### Description of change

Read commits for details

### Story Link

https://trello.com/c/6SdLXh6c/885-add-2fa-for-council-staff


app/controllers/application_controller.rb
```diff
@@ -1,20 +1,29 @@
...
+  def after_sign_in_path_for(resource)
+    if session[:mobile_number] && !resource.mobile_number?
+      resource.assign_mobile_number!(session[:mobile_number])
+      session.delete(:mobile_number)
+    end
+
+    respond_to do |format|
+      format.html do
+        flash[:notice] = "Signed in successfully."
+        root_path
```
EGiataganas, line 22: redirect_to root_path, notice: "Signed in successfully."

EGiataganas:
I would recommend to add a condition to use 2fa on dev/locally only when an environment variable is set. For example `2FA_ENABLED`

benbaumann95:
> I would recommend to add a condition to use 2fa on dev/locally only when an environment variable is set. For example `2FA_ENABLED`

You can just update user to have `otp_required_for_login` set to false which will bypass this 2fa process. I'm not sure what benefit having this env variable on top would do?

app/controllers/application_controller.rb
benbaumann95: Yeah I wanted to do this but I get
```
AbstractController::DoubleRenderError:
   Render and/or redirect were called multiple times in this action.
```

benbaumann95:
> 2FA_ENABLED

Or maybe I can add to the `otp_two_factor_enabled?` method:

```ruby
  def otp_two_factor_enabled?
    if Rails.env.development?
      ENV["2FA_ENABLED"] == true && find_user.try(:otp_required_for_login)
    else
      find_user.try(:otp_required_for_login)
    end
  end
```

This could be ok perhaps

benbaumann95:
> > 2FA_ENABLED
> 
> Or maybe I can add to the `otp_two_factor_enabled?` method:
> 
> ```ruby
>   def otp_two_factor_enabled?
>     if Rails.env.development?
>       ENV["2FA_ENABLED"] == true && find_user.try(:otp_required_for_login)
>     else
>       find_user.try(:otp_required_for_login)
>     end
>   end
> ```
> 
> This could be ok perhaps

I added the commit here https://github.com/unboxed/bops/pull/591/commits/f67cdc4e4b276b2b8d678e4a4329b8d1e07797c1 @EGiataganas

app/controllers/concerns/authenticate_with_otp_two_factor.rb
```diff
@@ -56,7 +56,11 @@ def user_params
   end
 
   def otp_two_factor_enabled?
-    find_user.try(:otp_required_for_login)
+    if Rails.env.development?
+      ENV["2FA_ENABLED"] == "true" && find_user.try(:otp_required_for_login)
+    else
+      find_user.try(:otp_required_for_login)
+    end
```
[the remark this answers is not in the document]
benbaumann95: I'm confused. The if and else statement returns the exact same thing?

PR #623, opened 2022-05-19
Update case numbering for planning applications

benbaumann95:
### Description of change

**Update case numbering for planning applications**
- Add migration to add council codes for local authorities
- Add migration to add application numbers starting from 100
- Add reference and reference_in_full instance methods to construct a string using a combination of created at year, council codes, application number and application type
- Only display full reference on email / notifications
- Council code not shown in BOPS pages as a user will be scoped to this local authority anyway

### Story Link

https://trello.com/c/zSxcxMr8/910-case-numbering-mvp

Screenshot 2022-05-19 at 15 24 52

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -39,6 +39,7 @@ class PlanningApplication < ApplicationRecord
 
   before_create :set_key_dates
   before_create :set_change_access_id
+  before_create :set_application_number
 
   after_create :set_ward_information
   after_create :create_audit!
@@ -87,7 +88,7 @@ class PlanningApplication < ApplicationRecord
   validates :work_status,
             inclusion: { in: WORK_STATUSES,
                          message: "Work Status should be proposed or existing" }
-  validates :application_type, presence: true
+  validates :application_type, :application_number, presence: true
   validates :payment_amount, :invalid_payment_amount, numericality: { greater_than_or_equal_to: 0 }, allow_nil: true
 
   validate :applicant_or_agent_email
@@ -112,7 +113,15 @@ class PlanningApplication < ApplicationRecord
   end
 
   def reference
-    @reference ||= id.to_s.rjust(8, "0")
+    @reference ||= "#{created_at_year}-#{application_number}-#{application_type_code}"
+  end
+
+  def reference_in_full
+    @reference_in_full ||= "#{local_authority.council_code}-#{reference}"
+  end
+
+  def application_number
+    self[:application_number].to_s.rjust(5, "0")
   end
 
   def correction_provided?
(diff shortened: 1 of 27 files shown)


app/models/planning_application.rb
```diff
@@ -112,7 +113,15 @@ def days_overdue
...
-    @reference ||= id.to_s.rjust(8, "0")
+    @reference ||= "#{created_at_year}-#{application_number}-#{application_type_code}"
+  end
+
+  def reference_in_full
+    @reference_in_full ||= "#{local_authority.council_code}-#{reference}"
+  end
+
+  def application_number
+    super.to_s.rjust(5, "0")
```
allbecauseyoutoldmeso, line 124: I wonder if `self[:application_number]` would be clearer than `super`? I think that way it's more obvious that this is overwriting a getter method. https://github.com/rubocop/rails-style-guide#read-attribute

app/models/planning_application.rb
```diff
@@ -560,4 +569,16 @@ def determination_date_is_not_in_the_future
       errors.add(:determination_date, "Determination date must be today or in the past")
     end
   end
+
+  def set_application_number
+    self.application_number = local_authority.planning_applications.count + 100
```
allbecauseyoutoldmeso, line 574: Would it be possible for 2 records initialised at the same time to end up with the same `application_number`? If so it's probably worth at least adding a validation to check it's unique.

I think there's a way to use PSQL to create an auto-incrementing column (https://www.geeksforgeeks.org/postgresql-create-auto-increment-column-using-serial/) but that's probably overkill!

db/migrate/20220510131942_add_application_number_to_planning_applications.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class AddApplicationNumberToPlanningApplications < ActiveRecord::Migration[6.1]
+  def up
+    add_column :planning_applications, :application_number, :bigint
+
+    add_index :planning_applications, %i[application_number local_authority_id], unique: true
```
allbecauseyoutoldmeso, line 7: 👍

app/models/planning_application.rb
allbecauseyoutoldmeso: Ah, just seen the migration 😁

allbecauseyoutoldmeso (approved):
Looks great!

app/models/planning_application.rb
benbaumann95: Ah okay, thanks for sharing. I have no strong preference for this so if it helps makes thing clearer then will update

app/models/planning_application.rb
```diff
@@ -560,4 +569,16 @@ def determination_date_is_not_in_the_future
...
+  end
+
+  def created_at_year
+    created_at.year % 100
```
EGiataganas, line 578: How about `created_at.strftime("%y")` which is a [ruby method](https://apidock.com/ruby/DateTime/strftime) and does that

app/models/planning_application.rb
benbaumann95: Ah okay, thanks for sharing. I have no strong preference for this so if it helps makes thing clearer then will update

PR #627, opened 2022-05-20
Update how application number is set

benbaumann95:
- Previous way might be a bit problematic if planning applications are deleted since an existing application number may clash when calculating against a count

### DIFF
--- a/spec/models/planning_application_spec.rb
+++ b/spec/models/planning_application_spec.rb
@@ -128,6 +128,24 @@ RSpec.describe PlanningApplication, type: :model do
           expect(planning_application3.application_number).to eq("00100")
         end
       end
+
+      context "when a planning application is deleted" do
+        let(:local_authority) { create :local_authority }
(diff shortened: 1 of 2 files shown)


spec/models/planning_application_spec.rb
```diff
@@ -128,6 +128,24 @@
...
+        it "updates the application number incrementing after the existing maximum application number" do
+          expect(planning_application1.application_number).to eq("00100")
+          expect(planning_application2.application_number).to eq("00101")
+          expect(planning_application3.application_number).to eq("00102")
+
+          planning_application2.destroy
+
+          expect(planning_application4.application_number).to eq("00103")
+        end
+      end
```
benbaumann95, line 148: This spec fails with the previous implementation as it will try count all the planning applications which would be 2 after the planning application is destroyed and therefore try to set the application number as 102 which already exists
```
ActiveRecord::RecordNotUnique: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "ix_planning_applications_on_application_number__local_authority"
```

PR #635, opened 2022-05-23
Include signatory name in decision notice

allbecauseyoutoldmeso:
### Description of change

Include signatory name in postal address rendered in decision notice.

### Story Link

https://trello.com/c/H0B3Uvwa/946-directors-name-missing-from-decision-notice

### Screenshot

Screenshot 2022-05-23 at 14 35 26

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -5,4 +5,8 @@ class LocalAuthority < ApplicationRecord
   has_many :planning_applications, dependent: :destroy
 
   validates :council_code, presence: true
+
+  def signatory
+    "#{signatory_name}, #{signatory_job_title}"
+  end
 end
--- a/spec/system/planning_applications/determine_application_spec.rb
+++ b/spec/system/planning_applications/determine_application_spec.rb
@@ -3,7 +3,15 @@
 require "rails_helper"
 
 RSpec.describe "Planning Application Assessment", type: :system do
-  let(:default_local_authority) { create(:local_authority, :default) }
+  let(:default_local_authority) do
+    create(
+      :local_authority,
+      :default,
+      signatory_name: "Jane Smith",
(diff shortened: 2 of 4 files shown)


app/models/local_authority.rb
```diff
@@ -5,4 +5,8 @@ class LocalAuthority < ApplicationRecord
   has_many :planning_applications, dependent: :destroy
 
   validates :council_code, presence: true
+
+  def signatory
+    "#{signatory_name}, #{signatory_job_title}"
+  end
```
allbecauseyoutoldmeso, line 11: Not 100% sure about 
1. Whether a view helper would be more appropriate than a method in the model.
2. Whether this should be called `#signatory_name_and_job_title` instead!

benbaumann95: We could probably update the email text to use this method now. I think they all currently sign off with `@planning_application.local_authority.signatory_name %>, <%= @planning_application.local_authority.signatory_job_title`

allbecauseyoutoldmeso: I've actually just updated them all to use `<%= @planning_application.local_authority.name %>` instead 😞  😆

spec/system/planning_applications/determine_application_spec.rb
```diff
@@ -67,6 +67,14 @@
           href: decision_notice_planning_application_path(planning_application)
         )
 
+        click_link("View decision notice")
+
+        expect(page).to have_content(
+          "Certificate of Lawful Use or Development Granted"
+        )
+
+        expect(page).to have_content(default_local_authority.signatory)
```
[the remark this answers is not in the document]
allbecauseyoutoldmeso: Good shout!

benbaumann95 (approved):

PR #641, opened 2022-05-26
Add user management interface

allbecauseyoutoldmeso:
### Description of change

- Add simple interface for adding, updating and viewing users.
- Add `admin` column to users, and only allow users with `admin` true to access interface.

### Story Link

https://trello.com/c/bCi8PiVJ/882-user-management-interface

### Decisions [OPTIONAL]

~~We could add a new user `role` of admin to the existing ones. But this would give us two choices:~~

~~1. Existing real users have two accounts, an admin one and a reviewer/assessor one. Given that user accounts have to have a unique email per local authority I think this could be unsatisfactory!~~
~~2. Modifying the permissions in some way so that assessors had, eg, the same permissions as assessors PLUS access to the user management interface. And I think this might get a bit messy..?~~

Following discussion with the partners, we're creating a new `administrator` role.

Screenshot 2022-06-01 at 09 31 13

Screenshot 2022-06-01 at 09 30 48


app/controllers/users_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+  end
+
+  private
+
+  def user_params
+    params
+      .require(:user)
+      .permit(:name, :email, :password, :mobile_number, :role)
+      .tap { |p| p.delete(:password) if p[:password].blank? }
```
allbecauseyoutoldmeso, line 43: In the edit form, the password is not rendered. This means that it passes `password: ""` in the params, although obviously the admin isn't trying to set the password to ""!

config/brakeman.ignore
```diff
@@ -1,83 +1,23 @@
...
-      "warning_type": "Redirect",
-      "warning_code": 18,
-      "fingerprint": "3c8044aad0dfc6077d787528ad6b41c1869e156860948fd78d088ae46fd11e6f",
-      "check_name": "Redirect",
-      "message": "Possible unprotected redirect",
-      "file": "app/controllers/planning_applications_controller.rb",
-      "line": 156,
-      "link": "https://brakemanscanner.org/docs/warning_types/redirect/",
-      "code": "redirect_to(current_local_authority.planning_applications.find(params[:id]))",
+      "warning_type": "Mass Assignment",
```
allbecauseyoutoldmeso, line 4: This is raised by `params.require(:user).permit(:role)`. Although it's not breaking the basic rule, since we're explicitly whitelisting the params, apparently 'Brakeman will warn on potentially dangerous attributes that are whitelisted.'

See https://brakemanscanner.org/docs/warning_types/mass_assignment/

benbaumann95 (commented):
This looks neat, loving the locale stuff! Well done for doing it so speedy too 🚀 
I think it's worth sharing a screenshot of the table to the partners as they may want a say in the content for the headers etc 

Also take a look at error messages / summary for govuk projects https://design-system.service.gov.uk/components/error-summary/

I think we need to follow this pattern

allbecauseyoutoldmeso:
> Also take a look at error messages / summary for govuk projects https://design-system.service.gov.uk/components/error-summary/
> 
> I think we need to follow this pattern

Ah, thank you, I hadn't seen this!

allbecauseyoutoldmeso:
> This looks neat, loving the locale stuff! Well done for doing it so speedy too 🚀 I think it's worth sharing a screenshot of the table to the partners as they may want a say in the content for the headers etc

...and this was a great call 😁

benbaumann95:
> > This looks neat, loving the locale stuff! Well done for doing it so speedy too 🚀 I think it's worth sharing a screenshot of the table to the partners as they may want a say in the content for the headers etc
> 
> ...and this was a great call 😁

Yeah usually there is a design so good to  sense check but people seem to love it! Great stuff 👏

app/models/user.rb
```diff
@@ -15,6 +15,8 @@ class User < ApplicationRecord
 
   before_create :generate_otp_secret
 
+  validates_format_of :mobile_number, with: /\A\d*\z/
```
allbecauseyoutoldmeso, line 18: Thought it was worth some basic validation for this.

I'm not checking the length as I don't know if it will be UK numbers only!

app/controllers/planning_applications_controller.rb
```diff
@@ -289,6 +289,11 @@ def validate_documents
 
   private
 
+  def validation_date_fields_invalid?
+    validation_date_fields.any?(&:blank?) ||
+      validation_date_fields.any? { |field| !field.match(/\A[0-9]*\z/) }
```
allbecauseyoutoldmeso, line 294: Prior to updating the `govuk_design_system_formbuilder` gem, an HTML5 validation handled this without submitting the form. Now the form overrides this, allowing us to use back-end validation and render a proper gov-uk styled error message ☺️

spec/system/planning_applications/validating_spec.rb
```diff
@@ -189,8 +189,8 @@
       click_button "Mark the application as valid"
 
       new_planning_application.reload
-      # This is stopped by HTML 5 validations, which are hard to test the UI for.
```
allbecauseyoutoldmeso, line 192: Fixed by updating the gem.

Gemfile.lock
```diff
@@ -221,11 +220,11 @@ GEM
     formatador (0.3.0)
     globalid (1.0.0)
       activesupport (>= 5.0)
-    govuk_design_system_formbuilder (2.7.2)
-      actionview (>= 6.0)
-      activemodel (>= 6.0)
-      activesupport (>= 6.0)
-      deep_merge (~> 1.2.1)
+    govuk_design_system_formbuilder (3.0.3)
```
allbecauseyoutoldmeso, line 223: Updating this because version 3 has an error summary helper!

app/controllers/planning_applications_controller.rb
benbaumann95: Probably out of scope but we probably don't need this method https://github.com/unboxed/bops/blob/61908f34fc9013d9df22131ea02f2653b0f89311/app/controllers/planning_applications_controller.rb#L338 as date params should be handled by rails. I think I had to do similar for determination date and didn't need to do any hacky things. Feels like we are doing a bit much here to validate this date field but possibly out of scope for this pr anyway

app/presenters/error_presenter.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class ErrorPresenter
```
benbaumann95, line 3: ooh nice idea extracting this out

benbaumann95 (commented):
Other than what we just discussed regarding logging out / adding navigation buttons to new and edit pages this is super stuff. 🎉  Nice job

benbaumann95 (approved):

PR #643, opened 2022-05-26
Add not null constraint on name / subdomain for local authorities

benbaumann95:
- These fields are mandatory and we don't have any UI to create local authority records so we must at least enforce integrity at the db level

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -4,7 +4,7 @@ class LocalAuthority < ApplicationRecord
   has_many :users, dependent: :destroy
   has_many :planning_applications, dependent: :destroy
 
-  validates :council_code, presence: true
+  validates :council_code, :name, :subdomain, presence: true
 
   def signatory
     "#{signatory_name}, #{signatory_job_title}"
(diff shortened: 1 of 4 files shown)


allbecauseyoutoldmeso (approved):
I think perhaps we should also have a `validates_presence_of` in the model?

And I'm sure you have, but double-check that no existing records are missing these attributes!

benbaumann95:
> I think perhaps we should also have a `validates_presence_of` in the model?
> 
> And I'm sure you have, but double-check that no existing records are missing these attributes!

- Yeah I checked staging and prod db so think we are ok!
- Is validates_presence_of necessary when we don't have any UI for adding local authorities. Is it to future protect when we want to add some errors to the page ?

allbecauseyoutoldmeso:
> * Is validates_presence_of necessary when we don't have any UI for adding local authorities. Is it to future protect when we want to add some errors to the page ?

Yes, I think good to future protect, and also apparently it means an error is raised without a query to the db!

https://stackoverflow.com/questions/2367281/ruby-on-rails-is-it-better-to-validate-in-the-model-or-the-database#answer-2367302

benbaumann95:
> > * Is validates_presence_of necessary when we don't have any UI for adding local authorities. Is it to future protect when we want to add some errors to the page ?
> 
> Yes, I think good to future protect, and also apparently it means an error is raised without a query to the db!
> 
> https://stackoverflow.com/questions/2367281/ruby-on-rails-is-it-better-to-validate-in-the-model-or-the-database#answer-2367302

Ah that's good to know about the lack of query, will add. Thanks!

PR #652, opened 2022-06-08
Handle Active Storage missing image errors

allbecauseyoutoldmeso:
### Description of change

- In API response for `ReplacementDocumentValidationRequest`, rescue `ActiveStorage::PreviewError` thrown when image is not present in S3, and return url as `nil`.

### Story Link

https://trello.com/c/Jp0SevPC/986-500-error-in-bops-applicants

### Decisions [OPTIONAL]

The error in the ticket is because someone has managed to create a `document` without successfully uploading an image. I don't think they should be able to do this, and we're still trying to work out how it happened.

However if it _has_ happened, it should be possible for the planner to ask the applicant to replace the image. Therefore we want the API request from bops-applicants to be successful, and get the data of the image-less document.

### DIFF
--- a/spec/requests/api/replacement_document_validation_request_spec.rb
+++ b/spec/requests/api/replacement_document_validation_request_spec.rb
@@ -6,72 +6,123 @@ RSpec.describe "Replacement document validation requests API", type: :request, s
   let!(:api_user) { create(:api_user) }
   let!(:default_local_authority) { create(:local_authority, :default) }
   let!(:planning_application) { create(:planning_application, :invalidated, local_authority: default_local_authority) }
+  let(:document) { create(:document) }
+
   let!(:replacement_document_validation_request) do
-    create(:replacement_document_validation_request, planning_application: planning_application)
+    create(
+      :replacement_document_validation_request,
+      planning_application: planning_application,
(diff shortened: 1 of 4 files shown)


spec/requests/api/replacement_document_validation_request_spec.rb
```diff
@@ -6,72 +6,123 @@
...
-              "id" => replacement_document_validation_request3.id,
-              "state" => "cancelled",
-              "response_due" => replacement_document_validation_request3.response_due.to_s(:db),
-              "days_until_response_due" => 15,
-              "cancel_reason" => "Made by mistake!",
-              "cancelled_at" => json_time_format(replacement_document_validation_request3.cancelled_at),
-              "old_document" => {
-                "name" => "proposed-floorplan.png",
-                "invalid_document_reason" => nil,
-                "url" => json["data"][2]["old_document"]["url"]
```
allbecauseyoutoldmeso, line 71: I've refactored these tests a bit because this is kind of saying `expect(json).to eq({ foo: json[:foo] }`, which will pass whatever the value of `foo` is!

EGiataganas (approved):

PR #654, opened 2022-06-10
Spec fixes

allbecauseyoutoldmeso:
### Description of change

Fix specs now that `subdomain` of `local_authorities` is enum.

### Story Link

https://link-to-issue


spec/system/local_authorities_spec.rb
```diff
@@ -3,8 +3,15 @@
 require "rails_helper"
 
 RSpec.describe "Accessing correct local authority", type: :system do
-  let!(:lambeth) { create(:local_authority, :lambeth) }
-  let!(:southwark) { create(:local_authority, :southwark) }
+  let!(:lambeth) do
+    LocalAuthority.find_by(subdomain: "lambeth") ||
+      create(:local_authority, :lambeth)
+  end
```
allbecauseyoutoldmeso, line 9: These aren't necessary if you run the test alone, but seem to be if you run the whole suite. I don't like this - seems like we're not cleaning up some data somewhere 😬

allbecauseyoutoldmeso:
Closing so github actions doesn't keep running the specs twice!

PR #655, opened 2022-06-13
Fix specs

allbecauseyoutoldmeso:
### Description of change

Fix the tests!


.github/workflows/build.yml
```diff
@@ -225,7 +225,6 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: "14"
-          cache: "yarn"
```
allbecauseyoutoldmeso, line 228: I'm getting the error 'Error: getCacheEntry failed: Cache service responded with 503' for 3 out of 4 runs. Seems like it's an ongoing bug - see https://github.com/bahmutov/npm-install/issues/80. To date the only solution is to remove cacheing 😞

.github/actions/setup/action.yml
```diff
@@ -19,4 +19,5 @@ runs:
         RAILS_ENV: test
         OTP_SECRET_ENCRYPTION_KEY: 83ccc50b-3ba9-45d9-8047-9518a935f065
       run: |
-        bundle exec rake db:setup
+        bundle exec rake db:create
+        bundle exec rake db:schema:load
```
allbecauseyoutoldmeso, line 23: Don't add seed data for rspec tests.

features/support/e2e.rb
```diff
@@ -10,3 +10,7 @@
 After("@e2e") do
   Capybara.server_port = nil
 end
+
+Before do
+  Rails.application.load_seed
+end
```
allbecauseyoutoldmeso, line 16: The cucumber tests are set up to work with seed data. However we don't want to seed the db, since it's also used for rspec tests. So instead this sets the data up on a test-by-test basis.

benbaumann95 (approved):

EGiataganas (approved):

PR #660, opened 2022-06-15
Show proposal refs on fee page

allbecauseyoutoldmeso:
### Description of change

Show proposal refs on fee page.

Also fix bug where proposal refs weren't rendering properly on planning application show page.

### Story Link

https://trello.com/c/91wqsf1s/852-show-officers-all-information-related-to-the-fee-calculation

### Screenshots

Screenshot 2022-06-15 at 16 25 26

before:
Screenshot 2022-06-15 at 16 40 23

after:
Screenshot 2022-06-15 at 16 39 36

### DIFF
--- a/app/helpers/proposal_details_helper.rb
+++ b/app/helpers/proposal_details_helper.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+module ProposalDetailsHelper
+  def proposal_details_group_id(group)
+    proposal_details_group_name(group).downcase.gsub(/[^0-9a-z]/i, "")
+  end
+
+  def proposal_details_group_title(group)
+    proposal_details_group_name(group).downcase.underscore.humanize
+  end
+
+  def proposal_details_group_name(group)
+    group || t("proposal_details.other")
+  end
+
+  def proposal_detail_item(proposal_detail)
+    tag.p(class: "govuk-body") do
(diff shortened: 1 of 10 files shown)


app/helpers/proposal_details_helper.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  def proposal_details_group_title(group)
+    proposal_details_group_name(group).downcase.underscore.humanize
+  end
+
+  def proposal_details_group_name(group)
+    group || t("proposal_details.other")
+  end
+
+  def proposal_detail_item(proposal_detail)
```
allbecauseyoutoldmeso, line 16: I've moved all these methods out of `ProposalDetailsPresenter` as that was essentially defining them on `planning_application` which doesn't really make sense. This change means that instead of 

```
planning_application.proposal_details.each { |proposal_detail| planning_application.my_method(proposal_detail) ```
we can write

```
planning_application.proposal_details.each { |proposal_detail| my_method(proposal_detail) }
```

allbecauseyoutoldmeso: An alternative is to have an _actual_ `ProposalDetailsPresenter` that wraps a `proposal_detail` and define them there. Would that be nicer? Please tell me what you think!

benbaumann95 (approved):

EGiataganas (approved):

PR #661, opened 2022-06-16
Refactor local authorities

EGiataganas:
### Description of change

Refactor local authorities:
- Add an enum for subdomains
- Drop council_code and use locales
- Drop name and use locales
- Add not null to local authorities fields

### Story Link

https://trello.com/c/ynTH3Zbq

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -4,7 +4,7 @@ class LocalAuthority < ApplicationRecord
   has_many :users, dependent: :destroy
   has_many :planning_applications, dependent: :destroy
 
-  validates :name, :subdomain, presence: true
+  validates :subdomain, presence: true
 
   enum subdomain: {
     lambeth: "lambeth",
@@ -19,5 +19,9 @@ class LocalAuthority < ApplicationRecord
 
   def council_code
     I18n.t("council_code.#{subdomain}")
+  end
+
+  def council_name
+    ripa? ? subdomain.capitalize : "#{subdomain.capitalize} Council Authority"
   end
 end
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -4,7 +4,7 @@ class LocalAuthority < ApplicationRecord
   has_many :users, dependent: :destroy
   has_many :planning_applications, dependent: :destroy
 
-  validates :council_code, :name, :subdomain, presence: true
+  validates :name, :subdomain, presence: true
 
   enum subdomain: {
     lambeth: "lambeth",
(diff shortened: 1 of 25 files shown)


allbecauseyoutoldmeso (approved):
👍

db/migrate/20220615151407_add_not_null_to_local_authorities.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddNotNullToLocalAuthorities < ActiveRecord::Migration[6.1]
+  def change
+    change_column_null :local_authorities, :signatory_name, false
+    change_column_null :local_authorities, :signatory_job_title, false
+    change_column_null :local_authorities, :enquiries_paragraph, false
+    change_column_null :local_authorities, :email_address, false
+    change_column_null :local_authorities, :feedback_email, false
```
benbaumann95, line 9: I think these are all set in our staging/prod db but maybe worth double checking otherwise this migration will fail

spec/mailer/planning_application_mailer_spec.rb
```diff
@@ -7,12 +7,10 @@
     create(
       :local_authority,
       :default,
-      name: "Cookie authority",
       signatory_name: "Mr. Biscuit",
       signatory_job_title: "Lord of BiscuitTown",
       enquiries_paragraph: "reach us on postcode SW50",
-      email_address: "biscuit@somuchbiscuit.com",
-      council_code: "ABC"
+      email_address: "biscuit@somuchbiscuit.com"
```
benbaumann95, line 13: lol

benbaumann95 (approved):

app/models/local_authority.rb
```diff
@@ -4,9 +4,25 @@ class LocalAuthority < ApplicationRecord
...
   def signatory
     "#{signatory_name}, #{signatory_job_title}"
   end
+
+  def council_code
+    I18n.t("council_code.#{subdomain}")
+  end
+
+  def council_name
+    ripa? ? subdomain.capitalize : "#{subdomain.capitalize} Council Authority"
```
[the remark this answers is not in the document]
EGiataganas: I also thought of this but since all of these would just have `Council Authority` for now at the end I think is ok. Maybe in the future when we find that some other councils have different name then we can change slighting a bit our locales files maybe to this:
```
councils:
  lambeth:
    name: Lambeth Council Authority
    code: LBH
  southwark:
    name: Southwark Council Authority
    code: SWK
  buckinghamshire:
    name: Buckinghamshire Council Authority
    code: BUC
  ripa:
    name: Ripa
    code: RIPA
```

PR #663, opened 2022-06-16
Apply UAT feedback to grouped proposal details

allbecauseyoutoldmeso:
### Description of change

- Wrap grouping of proposal details in environment variable. If this isn't present they're rendered as before, as a single list.
- Change the list numbering so that every question has a unique number.

### Story Link

https://trello.com/c/PEm3MXs5/992-put-the-proposal-details-from-ripa-into-groups-using-the-portal-name-supplied-in-json-blob

### Screenshot

Screenshot 2022-06-16 at 10 55 01

(there's a fix for the bug with link rendering in https://github.com/unboxed/bops/pull/660)


app/presenters/concerns/proposal_details_presenter.rb
```diff
@@ -18,6 +18,13 @@ def fee_related_proposal_details
       end
     end
 
+    def grouped_proposal_details_with_start_numbers
+      grouped_proposal_details.each_with_index.map do |group, index|
+        question_count = grouped_proposal_details.first(index).map(&:last).flatten.count
+        group.dup.push(question_count + 1)
+      end
```
allbecauseyoutoldmeso, line 25: Please tell me if there's a nicer way to do this 😅

PR #664, opened 2022-06-16
Apply UAT feedback to grouped proposal details

allbecauseyoutoldmeso:
### Description of change

- Wrap grouping of proposal details in environment variable. If this isn't present they're rendered as before, as a single list.
- Change the list numbering so that every question has a unique number.

### Story Link

https://trello.com/c/PEm3MXs5/992-put-the-proposal-details-from-ripa-into-groups-using-the-portal-name-supplied-in-json-blob

### Screenshot

Screenshot 2022-06-16 at 10 55 01

(there's a fix for the bug with link rendering in https://github.com/unboxed/bops/pull/660)
### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


app/presenters/concerns/proposal_details_presenter.rb
```diff
@@ -18,6 +18,13 @@ def fee_related_proposal_details
       end
     end
 
+    def grouped_proposal_details_with_start_numbers
+      grouped_proposal_details.each_with_index.map do |group, index|
+        question_count = grouped_proposal_details.first(index).map(&:last).flatten.count
+        group.dup.push(question_count + 1)
+      end
```
benbaumann95, line 25: Ooh this seems a nice puzzle to solve. Code seems fine and hopefully won't use too much memory if the proposal details we get is very long. Just wanted to ask what the context behind this was?

benbaumann95 (approved):

spec/presenters/planning_application_presenter_spec.rb
```diff
@@ -322,4 +322,88 @@
       )
     end
   end
+
+  describe "#grouped_proposal_details_with_start_numbers" do
```
benbaumann95, line 326: nice spec 👍

app/presenters/concerns/proposal_details_presenter.rb
allbecauseyoutoldmeso: It's so that even though the proposal details are split into separate groups, each with a separate list, the numbering can continue sequentially. This is to make it easier for planners to reference a specific question. They can just say 'question 10', instead of 'group 3, question 4'.

PR #667, opened 2022-06-17
Return to task list if document not marked as invalid

allbecauseyoutoldmeso:
### Description of change

- Return user to task list if document is not marked as valid or invalid.

### Story Link

https://trello.com/c/0rwD2nSQ/988-check-document-page-takes-user-to-invalid-page-if-is-the-document-valid-yes-or-no-is-not-selected

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -28,7 +28,7 @@ class DocumentsController < AuthenticationController
     respond_to do |format|
       format.html do
         if @document.update(document_params)
-          if !@document.validated? && @validate_document
+          if validate_document? && @document.validated == false
             redirect_to new_planning_application_replacement_document_validation_request_path(document: @document)
           else
             redirect_to redirect_url, notice: "Document has been updated"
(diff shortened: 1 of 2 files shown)


app/controllers/documents_controller.rb
```diff
@@ -28,7 +28,7 @@ def update
     respond_to do |format|
       format.html do
         if @document.update(document_params)
-          if !@document.validated? && @validate_document
+          if validate_document? && @document.validated == false
```
EGiataganas, line 31: I would leave it as
`!@document.validated?`

app/controllers/documents_controller.rb
```diff
@@ -112,7 +112,7 @@ def ensure_document_edits_unlocked
   end
 
   def validate_document?
-    @validate_document = params[:validate] == "yes"
+    @validate_document ||= params[:validate] == "yes"
```
EGiataganas, line 115: 👍

EGiataganas (approved):

PR #668, opened 2022-06-21
Make mobile number required for users

allbecauseyoutoldmeso:
### Description of change

- Validate users on presence of `mobile_number` if `otp_required_for_login` is true.
- Set `otp_required_for_login` to true as a `before_validate` callback that only runs on create.
- This means that admins can't create users without mobile numbers, but existing users on preview with no mobile number and `otp_required_for_login` set to false are still valid (and that we can manually create more like this if required).

### Story Link

https://trello.com/c/bCi8PiVJ/882-user-management-interface#comment-62b1b2c6fd65fa5adf959191 (see recent comments)

### Screenshot

Screenshot 2022-06-21 at 14 24 12

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -42,6 +42,8 @@ local_authorities.each do |authority|
       last_name = Faker::Name.last_name
       user.name = "#{first_name} #{last_name}"
       user.local_authority = authority
+      user.mobile_number = Faker::Base.numerify('07#########')
+
       if Rails.env.development?
         user.password = user.password_confirmation = "password"
       else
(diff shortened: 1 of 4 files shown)


db/seeds.rb
```diff
@@ -42,6 +42,8 @@
       last_name = Faker::Name.last_name
       user.name = "#{first_name} #{last_name}"
       user.local_authority = authority
+      user.mobile_number = Faker::Base.numerify('07#########')
```
EGiataganas, line 45: Nipitip `double-quoted`

PR #672, opened 2022-06-24
Make response time for description change requests 5 days

allbecauseyoutoldmeso:
### Description of change

Overwrite default `#response_due` method that `DescriptionChangeValidationRequest` inherits from `ValidationRequest`, which returns date 15 days from the created at date, with one that uses 5 days instead. 

Previously 15 days was correct for description change requests, but this has been changed and they are now auto-closed after 5 days.

Linked to https://github.com/unboxed/bops-applicants/pull/82

### Story Link

https://trello.com/c/O1qA5sVv/682-support-both-15-day-validation-and-5-day-post-validation-time-limited-request-from-applicants-3

### DIFF
--- a/app/models/description_change_validation_request.rb
+++ b/app/models/description_change_validation_request.rb
@@ -3,6 +3,8 @@
 class DescriptionChangeValidationRequest < ApplicationRecord
   include ValidationRequest
 
+  RESPONSE_TIME_IN_DAYS = 5
+
   before_create :set_previous_application_description
 
   belongs_to :planning_application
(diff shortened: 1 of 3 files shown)


app/models/description_change_validation_request.rb
```diff
@@ -59,6 +61,10 @@ def audit_cancel!
     )
   end
 
+  def response_due
+    RESPONSE_TIME_IN_DAYS.business_days.after(created_at).to_date
```
allbecauseyoutoldmeso, line 65: I tried overwriting the constant rather than the whole method, but it seems that a method defined in a concern will pick up the constant from that concern, not from the class it's included in 🤔

benbaumann95 (approved):

PR #673, opened 2022-06-29
Added redirect to public_url

EGiataganas:
### Description of change

Added redirect to council's public_url

### BEFORE DEPLOYING
- we need to add the new environment variable on AWS
- we need to change `APPLICANTS_APP_HOST` environment variable on AWS

### Story Link

https://trello.com/c/K3mnhPvO

### Tasks

- Change the environment variable `APPLICANTS_APP_HOST` to map to the new one
- Add the environment variable `PUBLIC_URL`


app/models/planning_application.rb
```diff
@@ -261,7 +261,15 @@ def flagged_proposal_details(flag)
   def secure_change_url
     protocol = Rails.env.production? ? "https" : "http"
 
-    "#{protocol}://#{local_authority.subdomain}.#{ENV['APPLICANTS_APP_HOST']}/validation_requests?planning_application_id=#{id}&change_access_id=#{change_access_id}"
+    ENV.fetch("PUBLIC_URL", "false") == "true" ? "#{protocol}://#{public_url}" : "#{protocol}://#{internal_url}"
+  end
+
+  def public_url
```
benbaumann95, line 267: Can this be a private method?

app/models/planning_application.rb
```diff
@@ -261,7 +261,15 @@ def flagged_proposal_details(flag)
...
+    "#{ENV['APPLICANTS_APP_HOST']}.#{local_authority.subdomain}.gov.uk/validation_requests?planning_application_id=#{id}&change_access_id=#{change_access_id}"
+  end
+
+  def internal_url
```
benbaumann95, line 271: private method?

benbaumann95 (commented):
If that ENV is set globally then would Lambeth (who don't have the public url set up) get provided with the wrong url?

EGiataganas:
I was thinking that @benbaumann95 but technically we shouldn't have any council that is not ready. So lambeth shouldn't use our service until they are full onboard

benbaumann95 (commented):
Is it worth adding a spec for this too ?

EGiataganas:
### BEFORE DEPLOYING
- we need to add the new environment variable on AWS
- we need to change `APPLICANTS_APP_HOST` environment variable on AWS

benbaumann95 (approved):

allbecauseyoutoldmeso (approved):

PR #680, opened 2022-07-06
Retreive parish name from mapit API

benbaumann95:
### Description of change

Retrieve parish name from mapit API

- Annoyingly there is no shortcut id to access the relevant parish object so we have to iterate and find the object based on the parish type code. Example json: https://mapit.mysociety.org/postcode/HP9%202HA
- We then access the "name" key in this object to reference the parish name
- From docs, there seem to be 3 types of parish types: "Non-civil parished area (NCP), Civil parish/community (CPC), Civil parish/community, Civil parish/community ward (CPW)"
- Save this to db after planning application create callback

### Story Link

https://trello.com/c/BUe9EUhm/1033-show-parish-information

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -474,10 +474,11 @@ class PlanningApplication < ApplicationRecord
   def set_ward_information
     return if postcode.blank?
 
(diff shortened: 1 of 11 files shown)


allbecauseyoutoldmeso (approved):
🙌

PR #682, opened 2022-07-08
Show alert dialogue for unsaved changes

allbecauseyoutoldmeso:
### Description of change

- Add [stimulus](https://stimulus.hotwired.dev/) to the project using the [stimulus-rails](https://github.com/hotwired/stimulus-rails) gem.
- Intercept `beforeunload` eventsand show alert dialogue if policy class forms have changes that have not been submitted.
- Fix small bug where part is deselected if you click 'Back' in multipart policy form.

### Story Link

https://trello.com/c/rllvuyz8/1035-when-assessing-against-legislation-provide-warning-that-data-may-be-lost-when-navigating-away

### Screenshot

Screenshot 2022-07-14 at 14 43 17

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,6 +30,7 @@ gem "puma", "~> 4"
 gem "rails", "~> 6.1"
 gem "rswag-ui"
 gem "sassc", "~> 2.4.0"
+gem "stimulus-rails"
 gem "webpacker", "~> 5.4"
 gem "whenever"
 
--- a/app/controllers/policy_classes_controller.rb
+++ b/app/controllers/policy_classes_controller.rb
@@ -4,7 +4,9 @@ class PolicyClassesController < PlanningApplicationsController
   before_action :set_planning_application
   before_action :set_policy_class, only: %i[show update destroy]
 
-  def part; end
+  def part
+    @part_number = params[:part]&.to_i
+  end
 
   def new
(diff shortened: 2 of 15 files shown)


Gemfile
```diff
@@ -30,6 +30,7 @@ gem "puma", "~> 4"
 gem "rails", "~> 6.1"
 gem "rswag-ui"
 gem "sassc", "~> 2.4.0"
+gem "stimulus-rails"
```
allbecauseyoutoldmeso, line 33: There are a few ways to add stimulus to a Rails 6 project, but Rails 7 used the gem by default so I think it's probably the way to go.

app/controllers/policy_classes_controller.rb
```diff
@@ -4,7 +4,9 @@ class PolicyClassesController < PlanningApplicationsController
   before_action :set_planning_application
   before_action :set_policy_class, only: %i[show update destroy]
 
-  def part; end
+  def part
+    @part_number = params[:part]&.to_i
+  end
```
allbecauseyoutoldmeso, line 9: This is just a fix for a small bug I found as I was clicking around. Currently if you click 'Add assessment area', select a part, click 'Continue' and then click 'Back', the part you selected is no longer selected.

benbaumann95 (approved):
🎉

PR #683, opened 2022-07-11
Show description change requests in post-validation validation requests table

allbecauseyoutoldmeso:
### Description of change

- Update `PlanningApplication#validation_requests` to take an optional `include_description_change_validation_requests` option.
- In `ValidationRequestsController#post_validation_requests` use the `include_description_change_validation_requests` option when setting `@validation_requests`. 
- Update partial so aligns with existing table headers.

### Story Link

https://trello.com/c/l0E2kVlx/713-add-description-changes-to-post-validation-requests-table-2

### DIFF
--- a/app/models/concerns/planning_application/validation_request.rb
+++ b/app/models/concerns/planning_application/validation_request.rb
@@ -31,9 +31,16 @@ class PlanningApplication
       end
     end
 
-    def validation_requests(post_validation: false)
-      (replacement_document_validation_requests + additional_document_validation_requests + other_change_validation_requests + red_line_boundary_change_validation_requests)
-        .send(enumerable_method(post_validation), &:post_validation?).sort_by(&:created_at).reverse
+    def validation_requests(post_validation: false, include_description_change_validation_requests: false)
+      request_types = validation_request_types(
+        include_description_change: include_description_change_validation_requests
+      )
+
+      requests = request_types.map do |request_type|
+        send("#{request_type}_validation_requests").where(post_validation: post_validation)
+      end
+
+      requests.flatten.sort_by(&:created_at).reverse
     end
 
     def active_validation_requests(post_validation: false)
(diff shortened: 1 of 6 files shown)


app/models/concerns/planning_application/validation_request.rb
```diff
@@ -31,9 +31,16 @@ module ValidationRequest
       end
     end
 
-    def validation_requests(post_validation: false)
-      (replacement_document_validation_requests + additional_document_validation_requests + other_change_validation_requests + red_line_boundary_change_validation_requests)
-        .send(enumerable_method(post_validation), &:post_validation?).sort_by(&:created_at).reverse
+    def validation_requests(post_validation: false, include_description_change_validation_requests: false)
```
allbecauseyoutoldmeso, line 34: This method's doing a lot now, and I've done my best but I'm sure it could be prettier! Suggestions very welcome 😅

spec/models/concerns/planning_application/validation_request_spec.rb
```diff
@@ -34,24 +34,52 @@
...
+        :description_change_validation_request,
+        planning_application: planning_application,
+        created_at: 5.days.ago
+      )
     end
 
     context "when no argument is supplied" do
       it "returns validation requests where post_validation is false" do
-        expect(planning_application.validation_requests).to match_array(
+        expect(planning_application.validation_requests).to eq(
```
allbecauseyoutoldmeso, line 82: I noticed we're ordering by `created_at`, so seemed worth checking the order as well as the contents of the array.

app/models/concerns/planning_application/validation_request.rb
benbaumann95: Not sure if it would be worth turning this stuff into a class perhaps. 

We can initialize it with what we need and then run methods in there? Perhaps if this needs any more logic down the line we do this. Or whatever the best OOP design is for this type of thing

benbaumann95 (approved):

app/models/concerns/planning_application/validation_request.rb
allbecauseyoutoldmeso: Yes, I like the sound of that! Let's say if it needs any further changes (and I feel like it may, given it's masquerading as an active record relation!) we will extract it out into its own thing

PR #684, opened 2022-07-11
Allow post validation additional document requests

allbecauseyoutoldmeso:
### Description of change

- Remove validation that prevents creation of `additional_document_validation_request` after application has been validated.
- Render 'Request a new document' button on documents page if application is post validation.
- List post validation `additional_document_validation_requests` on documents page.
- Prevent post validation `additional_document_validation_requests` from being rendered on validation documents page.

### Story Link

https://trello.com/c/LQG1atOK/735-enable-amendment-requests-for-new-documents-that-can-be-actioned-after-validation-2

### Screenshot

Screenshot 2022-07-12 at 12 26 49


spec/system/planning_applications/post_validation_requests_spec.rb
```diff
@@ -17,13 +16,110 @@
       create :planning_application, :in_assessment, local_authority: default_local_authority
     end
 
-    it "displays the post validation request table" do
```
allbecauseyoutoldmeso, line 20: Because the file name didn't have `_spec` at the end this file wasn't being run, so this test was failing silently because of a css typo 😆 . However I think we've now got so many tests covering this table that we probably don't need this one, which just checks that the user can follow a link to the page.

app/controllers/additional_document_validation_requests_controller.rb
```diff
@@ -58,4 +61,12 @@ def additional_document_validation_request_params
   def set_additional_document_validation_request
     @additional_document_validation_request = @planning_application.additional_document_validation_requests.find(params[:id])
   end
+
+  def cancel_redirect_url
```
benbaumann95, line 65: Do we have a method for this already or is this redirect behaviour supposed to be different for this type of request?

https://github.com/unboxed/bops/blob/main/app/controllers/concerns/validation_requests.rb#L105

app/helpers/validation_request_helper.rb
```diff
@@ -73,17 +73,26 @@ def document_url(document)
...
+    url = show_validation_request_url(application, request)
+    link_to(text, url)
+  end
 
-    text = planning_application.validated? ? "View" : "View and update"
+  def show_validation_request_url(application, request)
+    if request.is_a?(AdditionalDocumentValidationRequest)
+      show_additional_document_validation_request_url(application, request)
+    else
+      polymorphic_path([application, request])
```
benbaumann95, line 86: 🎉

app/models/additional_document_validation_request.rb
```diff
@@ -15,7 +15,6 @@ class UploadFilesError < RuntimeError; end
   validates :document_request_type, presence: { message: "Please fill in the document request type." }
   validates :document_request_reason, presence: { message: "Please fill in the reason for this document request." }
 
-  before_create :ensure_planning_application_not_validated!
```
benbaumann95, line 18: 🙌

app/controllers/additional_document_validation_requests_controller.rb
allbecauseyoutoldmeso: I'm deliberately over-riding that method here because it takes the user to the post-validation requests table. Since they navigate to cancelling the document via the documents page, I think it's maybe bit disorientating to take them to the table page - they won't necessarily have every been there before 😅 .

app/views/additional_document_validation_requests/_additional_document_validation_requests_table.html.erb
```diff
@@ -21,7 +21,7 @@
         </td>
         <td class="govuk-table__cell govuk-!-width-one-sixth">
           <% if validation_request.open_or_pending? %>
-            <% if @planning_application.invalidated? %>
+            <% if @planning_application.invalidated? || validation_request.post_validation? %>
```
benbaumann95, line 24: Huh looks like I added a method `can_cancel?` but it's not being used anymore. Wonder if it's worth adding this logic in there and reviving it!

I think I may have overlooked this too when adding this same if statement for other validation request types so i'm happy to just create a separate PR to update all of these if you prefer that ?

https://github.com/unboxed/bops/blob/main/app/models/concerns/validation_request.rb#L118

benbaumann95 (approved):

app/controllers/additional_document_validation_requests_controller.rb
benbaumann95: Ah okay sounds good. We may wanna speak with Tom about some consistency here going forward . I reckon after we finish these PR's for post validation requests we'll do a cleanup of some bits

benbaumann95 (commented):
I think this wasn't on the ticket but there is some text here on this prototype at the top we may need to add
https://beta-bops-prototypes.herokuapp.com/v19/non-validation-request

I didn't include it when doing red line validation requests because obviously we didn't have this functionality yet of new document requests

```For new document requests, requests will remain open until responded, cancelled or the application is decided.```

app/views/additional_document_validation_requests/_additional_document_validation_requests_table.html.erb
allbecauseyoutoldmeso: Oh nice!

allbecauseyoutoldmeso:
> I think this wasn't on the ticket but there is some text here on this prototype at the top we may need to add https://beta-bops-prototypes.herokuapp.com/v19/non-validation-request
> 
> I didn't include it when doing red line validation requests because obviously we didn't have this functionality yet of new document requests
> 
> `For new document requests, requests will remain open until responded, cancelled or the application is decided.`

Eep, good spot, thank you!

PR #685, opened 2022-07-12
Auto close red line boundary change requests after 5 business days

benbaumann95:
### Description of change

Auto close red line boundary change requests after 5 business days

- Add migration for auto_closed and auto_closed_at column on red line boundary change requests
- Add missing migration for auto_closed_at column on description changes
- Wrap event and subsequent updates in transaction as we want to update everything successfully or nothing at all. If there is an error raised we send the error to AppSignal 

### Story Link

https://trello.com/c/yA3qHp8h/1030-red-line-requests-auto-accept-after-5-days-according-to-rules-in-use-for-description-changes

### DIFF
--- a/app/helpers/validation_request_helper.rb
+++ b/app/helpers/validation_request_helper.rb
@@ -29,10 +29,6 @@ module ValidationRequestHelper
     end
   end
 
-  def request_closed_at(validation_request)
-    validation_request.updated_at if validation_request.closed?
-  end
-
   def edit_request_url(planning_application, validation_request, classname: nil)
     link_to "Edit request",
             send("edit_planning_application_#{request_type(validation_request)}_path", planning_application,
--- a/db/migrate/20220711085009_add_auto_closed_at_to_description_change_validation_requests.rb
+++ b/db/migrate/20220711085009_add_auto_closed_at_to_description_change_validation_requests.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class AddAutoClosedAtToDescriptionChangeValidationRequests < ActiveRecord::Migration[6.1]
+  class DescriptionChangeValidationRequest < ApplicationRecord
+    scope :auto_closed, -> { where(auto_closed: true) }
+  end
+
+  def change
+    add_column :description_change_validation_requests, :auto_closed_at, :datetime
+
+    DescriptionChangeValidationRequest.auto_closed.find_each do |request|
+      request.update(auto_closed_at: request.updated_at)
+    end
+  end
+end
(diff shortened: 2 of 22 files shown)


db/migrate/20220711085009_add_auto_closed_at_to_description_change_validation_requests.rb
```diff
@@ -0,0 +1,15 @@
...
+  class DescriptionChangeValidationRequest < ApplicationRecord
+    scope :auto_closed, -> { where(auto_closed: true) }
+  end
+
+  def change
+    add_column :description_change_validation_requests, :auto_closed_at, :datetime
+
+    DescriptionChangeValidationRequest.auto_closed.find_each do |request|
+      request.update(auto_closed_at: request.updated_at)
+    end
```
benbaumann95, line 13: When the closed status was added it seems as if the person missed out the `closed_at` timestamp so we actually have no way of knowing the actual value of this.

Not sure whether we just leave as nil for historical records or assume the updated_at column is correct as usually a closed request does not have further action taken on it

app/helpers/validation_request_helper.rb
```diff
@@ -30,6 +30,8 @@ def display_request_status(validation_request)
   end
 
   def request_closed_at(validation_request)
```
allbecauseyoutoldmeso, line 32: Where is this being called?

spec/mailer/previews/planning_application_mailer_preview.rb
```diff
@@ -57,6 +57,10 @@ def validation_request_mail
     PlanningApplicationMailer.validation_request_mail(planning_application)
   end
 
+  def validation_request_closure_mail
+    PlanningApplicationMailer.validation_request_closure_mail(planning_application)
```
allbecauseyoutoldmeso, line 61: 🎉

spec/models/validation_request_spec.rb
```diff
@@ -276,5 +276,57 @@
...
+
+    describe "#auto_close_request!" do
+      let(:request) { create(:red_line_boundary_change_validation_request, :open) }
+
+      describe "when successful" do
+        it "auto closes the request and creates an audit record" do
+          expect { request.auto_close_request! }
+            .to change(request, :auto_closed_at).from(nil).to(Time.current)
+                                                .and change(request, :state).from("open").to("closed")
+                                                                            .and change(request, :auto_closed).from(false).to(true)
```
allbecauseyoutoldmeso, line 288: This alignment looks a bit weird to me...was it a rubocop autocorrect?

allbecauseyoutoldmeso (approved):
Really nice!

app/helpers/validation_request_helper.rb
benbaumann95: good catch, nowhere it seems

spec/models/validation_request_spec.rb
benbaumann95: lol yes, rubocop on a mad one

EGiataganas (approved):
Well done, this look really good 🎉

PR #689, opened 2022-07-14
Show feedback

benbaumann95:
### Description of change

Add feedback column to planning applications

### Story Link

Part 1 for https://trello.com/c/mbS4PsmF/393-display-the-free-text-collected-ripa-related-to-a-any-disagreements-about-the-applicants-planning-constraints-or-b-the-ripa-deci

- Still need to add to the frontend when designs are completed

### DIFF
--- a/app/controllers/api/v1/planning_applications_controller.rb
+++ b/app/controllers/api/v1/planning_applications_controller.rb
@@ -59,7 +59,8 @@ module Api
             end
           end
         end
-      rescue Errors::WrongFileTypeError, Errors::GetFileError, ActiveRecord::RecordInvalid, ArgumentError => e
+      rescue Errors::WrongFileTypeError, Errors::GetFileError, ActiveRecord::RecordInvalid, ArgumentError,
+             NoMethodError => e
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -37,7 +37,6 @@ def create
             constraints: constraints_array_from_param(params[:constraints]),
             planx_data: (params[:planx_debug_data].to_json if params[:planx_debug_data].present?),
             api_user: current_api_user,
-            audit_log: params.to_json,
```
allbecauseyoutoldmeso, line 40: Should we double check we don't want this?

allbecauseyoutoldmeso (approved):
👍

app/controllers/api/v1/planning_applications_controller.rb
benbaumann95: I'll add it back and we can remove this as another PR if not needed.
Looks like it is just storing all the params from the request so not sure why it's called audit_log anyway

PR #693, opened 2022-07-19
Allow admins to update reviewer_group_email

allbecauseyoutoldmeso:
### Description of change

- Allow admins to add and update the reviewer group email.

### Story Link

https://trello.com/c/3fHsvIUm/776-feature-request-email-notification-to-officer-of-changes-in-bops

### Screenshots

Screenshot 2022-07-19 at 16 41 13

Screenshot 2022-07-19 at 16 41 00


spec/system/managing_council_information_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+require "rails_helper"
+
+RSpec.describe "managing council information", type: :system do
+  let(:local_authority) { create(:local_authority, :default) }
+  let(:user) { create(:user, :administrator, local_authority: local_authority) }
+
+  it "allows the administrator to manage the reviewer group email" do
+    sign_in(user)
+    visit(administrator_dashboard_path)
+    row = find_all("tr").find { |tr| tr.has_content?("Manager group email") }
```
benbaumann95, line 12: Not in scope for this PR but I'm thinking we could move these specific finders to do with tables or any other relevant bits into an Rspec::Matcher class https://rubydoc.info/github/rspec/rspec-expectations/RSpec/Matchers to improve readability and for reuse

For another project, we did something like:

```ruby
RSpec::Matchers.define :have_table_row do |cells, row_index|
  match do |actual|
    page.within(:tr, row_index) do
      children = page.find_all(:xpath, "./*")
      expect(children.size).to eq(cells.size)

      cells.each_with_index do |cell, index|
        expect(page).to have_table_cell(cell, index + 1)
      end
    end
  end
end
```

Then in the spec you would just do ` expect(page).to have_table_row ....`

allbecauseyoutoldmeso: That's an excellent idea! I will do.

config/routes.rb
```diff
@@ -128,4 +128,12 @@
...
 
   get :healthcheck, to: proc { [200, {}, %w[OK]] }
+
+  resources :local_authorities, only: %i[edit update]
+
+  get(
+    :administrator_dashboard,
+    action: :show,
+    controller: :administrator_dashboard
+  )
```
benbaumann95, line 138: why not use resources here ?

benbaumann95 (approved):
This looks really great!
Could learn a thing or two from all your locale work 👍

config/routes.rb
allbecauseyoutoldmeso: To avoid having to use an id! https://guides.rubyonrails.org/routing.html#singular-resources

There's probably more than one way to do this though 🤔

EGiataganas (approved):

config/routes.rb
benbaumann95: Ah I think I ran into this issue before and I ended up doing
https://github.com/unboxed/bops/blob/main/config/routes.rb#L107-L110

But if the docs say to do as you did then no issues with me

EGiataganas: you can use the word `resource` singular

so it can be:
```
resource : administrator_dashboard, only: :show
```

EGiataganas: The singular `resource`, will do the trick without referencing an ID and that's why it doesn't have an `index` action

allbecauseyoutoldmeso: aha...nice, thanks both!

PR #698, opened 2022-07-22
Add search applications by reference feature

allbecauseyoutoldmeso:
### Description of change

- Add `reference` column to planning applications, replacing existing method.
- Add a `Search` form object which searches a set of planning applications by full or partial reference.
- Add an 'All applications' tab with a search form to the planning applications dashboard.
- Use some javascript to submit the search form and update the tab with the returned data.

### Story Link

https://trello.com/c/A3O5ieno/1013-dashboard-search-feature-for-planning-application-dashboard-application-number-only

### Decisions

I wound up using javascript to submit the form and refresh the view because trying to refresh the whole page with query parameters and also the correct tab pre-selected was not pretty.

### Screenshots

Screenshot 2022-07-22 at 12 13 16

Screenshot 2022-07-22 at 12 16 29


app/javascript/controllers/search_controller.js
```diff
@@ -0,0 +1,18 @@
+import { Controller } from "@hotwired/stimulus"
+import Rails from "@rails/ujs"
+
+export default class extends Controller {
+  handleSubmit(event) {
```
allbecauseyoutoldmeso, line 5: Submits the form to `SearchesController#create` and replaces the existing partial with the one returned.

app/models/planning_application.rb
```diff
@@ -423,6 +420,14 @@ def send_update_notification_to_reviewers
 
   private
 
+  def set_reference
```
allbecauseyoutoldmeso, line 423: Possibly my obsession with not going over the wrap guide line in Atom is out of control 😅

app/javascript/controllers/search_controller.js
allbecauseyoutoldmeso, line 5: Submits the form to `SearchesController#create` and replaces the existing partial with the one returned.

app/models/search.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class Search
+  include ActiveModel::Model
+
+  attr_accessor :query, :planning_application_ids
```
allbecauseyoutoldmeso, line 6: Passing the planning application ids to the search because they've already been filtered by things like local authority, user type and view type.

app/javascript/controllers/search_controller.js
allbecauseyoutoldmeso, line 5: Submits the search form to `SearchesController.create` and replaces the existing partial with the updated one returned.

app/models/planning_application.rb
```diff
@@ -423,6 +419,14 @@ def send_update_notification_to_reviewers
...
```
allbecauseyoutoldmeso, line 422: Possibly my obsession with not going over the guide line in Atom is out of control 😅 .

app/models/user.rb
```diff
@@ -8,8 +8,7 @@ class User < ApplicationRecord
 
   devise :database_authenticatable if Rails.env.development? && ENV["2FA_ENABLED"] != "true"
 
-  has_many :decisions, dependent: :restrict_with_exception
-  has_many :planning_applications, through: :decisions
```
allbecauseyoutoldmeso, line 12: Just spotted this while I was trying out different ways of searching. The `decisions` table no longer exists, and `planning_applications` now has `user_id`.

spec/system/planning_applications/recommending_spec.rb
```diff
@@ -30,7 +30,11 @@
     context "with no previous recommendations" do
       it "can create a new recommendation, edit it, and submit it" do
         click_link "In assessment"
-        click_link planning_application.reference
+
+        within("#under_assessment") do
```
allbecauseyoutoldmeso, line 34: Because Capybara can 'see' the content of all the tabs, it doesn't know whether to click the planning application reference in the 'In assessment' tab or the 'All applications' tab.

db/migrate/20220722101202_add_reference_to_planning_applications.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class AddReferenceToPlanningApplications < ActiveRecord::Migration[6.1]
+  def up
+    add_column :planning_applications, :reference, :string
+    add_index :planning_applications, :reference
+
+    PlanningApplication.all.find_each do |planning_application|
+      planning_application.send(:set_reference)
```
benbaumann95, line 9: Not sure this can find a private method.
My migration failed locally for this

db/migrate/20220722101202_add_reference_to_planning_applications.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class AddReferenceToPlanningApplications < ActiveRecord::Migration[6.1]
+  def up
+    add_column :planning_applications, :reference, :string
+    add_index :planning_applications, :reference
```
benbaumann95, line 6: I guess we have a unique index on the application number / council code but wonder if we want to add it here too 🤔

app/models/search.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  validates :query, presence: true
+
+  def results
+    if valid?
+      planning_applications.where("reference LIKE ?", "%#{query}%")
```
benbaumann95, line 12: 🎉

app/models/search.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class Search
+  include ActiveModel::Model
+
+  attr_accessor :query, :planning_application_ids
+
+  validates :query, presence: true
```
benbaumann95, line 8: 👍

app/models/search.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class Search
```
benbaumann95, line 3: Should be fine as we are just searching on planning applications for now but wonder if we could make this more agnostic and less coupled with planning applications as we add different search criteria in . Or we specify that this search class is just for planning applications

benbaumann95 (commented):
Nice job!

One to pick up / discuss on monday but I think we'll want to go with query / filter / sorting params being in the url. We'll be adding pagination too which needs to be in the url along with this other search critera.

By doing it with JS like this we aren't able to share links unfortunately which will be useful as we add more criteria.

I think there's some other example applications where I can share with you the different strategies of how this is done so possibly worth comparing

db/migrate/20220722101202_add_reference_to_planning_applications.rb
benbaumann95: I guess it would never flag up because of the other unique index though

db/migrate/20220722101202_add_reference_to_planning_applications.rb
allbecauseyoutoldmeso: That's weird...I thought using `send` had fixed that 🤔

db/migrate/20220722101202_add_reference_to_planning_applications.rb
allbecauseyoutoldmeso: Might be nice just for completeness!

app/models/search.rb
allbecauseyoutoldmeso: I think I like the second idea...making this `PlanningApplicationSearch`. That should push us to create separate models if we need to search other tables!

allbecauseyoutoldmeso:
Yes, let's discuss next week!

I ran into some hairy issues trying to pass queries via the url because of the way we're using anchor links to handle tabs (`/planning_applications?query=my_query#all` messes up the query, and `/planning_applications#all?query=my_query#all` messes up the tabs). If you fancy a bit of pairing, a fresh pair of eyes might help!

benbaumann95:
> Yes, let's discuss next week!
> 
> I ran into some hairy issues trying to pass queries via the url because of the way we're using anchor links to handle tabs (`/planning_applications?query=my_query#all` messes up the query, and `/planning_applications#all?query=my_query` messes up the tabs). If you fancy a bit of pairing, a fresh pair of eyes might help!

yeah let's do some pairing !

app/models/search.rb
benbaumann95: I think we'll actually want to add a LOWER function here too so we can be case insensitive

benbaumann95: which means we may need to add that to the index on reference i.e. `add_index :planning_applications, "lower(reference)"`

allbecauseyoutoldmeso: oo...good thinking!

app/models/search.rb
EGiataganas: I think this should be a service called `PlanningApplicationSearch`, which takes some arguments and search for a planning application and returns it

allbecauseyoutoldmeso: Yes, that's what I'm thinking too 😄

allbecauseyoutoldmeso: Updates coming up!

spec/system/searching_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      end
+
+      fill_in("Find an application", with: "00100")
+      click_button("Search")
+
+      within("#all") do
+        expect(page).to have_content(planning_application1.reference)
+        expect(page).not_to have_content(planning_application2.reference)
+        expect(page).not_to have_content(planning_application3.reference)
+      end
```
benbaumann95, line 58: with the update we mentioned this morning would now be great to add a spec to check current_url to to see we have the params and correct tab (if you weren't gonna add already!)

Then also i think is worth visiting this url directly and seeing the correct scope for the search so we know the search link is sharable

spec/system/searching_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+  end
+
+  before { sign_in(user) }
+
+  context "when users views her own planning applications" do
+    it "allows user to search planning applications by reference" do
+      visit(planning_applications_path(q: "exclude_others"))
+      click_link("All your applications")
+
+      selected_tab = find("div[class='govuk-tabs__panel']")
```
allbecauseyoutoldmeso, line 36: `find("govuk-tabs__panel")` will return all the tabs with this class, whereas this will only return the tab that doesn't have the "govuk-tabs__panel___hidden" class, ie it gets the tab that is currently selected!

spec/system/searching_spec.rb
```diff
@@ -0,0 +1,142 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "managing users", type: :system do
```
benbaumann95, line 5: copy and paste 😉

db/migrate/20220722101202_add_reference_to_planning_applications.rb
```diff
@@ -0,0 +1,49 @@
...
+      :planning_applications,
+      %i[reference local_authority_id],
+      unique: true
+    )
+
+    PlanningApplication.all.find_each do |planning_application|
+      planning_application.set_reference
+      planning_application.save!
+    end
+  end
```
benbaumann95, line 44: do we need to add a not null constraint to reference too?

app/services/planning_application_search.rb
```diff
@@ -0,0 +1,19 @@
...
+  include ActiveModel::Model
+
+  attr_accessor :query, :planning_applications
+
+  validates :query, presence: true
+
+  def results
+    if valid?
+      planning_applications.where(
+        "LOWER(reference) LIKE ?", "%#{query.downcase}%"
```
benbaumann95, line 13: 😍

db/migrate/20220722101202_add_reference_to_planning_applications.rb
benbaumann95: But then we'd need to update `after_create :set_reference` to `before_create :set_reference`

spec/system/searching_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+      expect(selected_tab).to have_content("All your applications")
+
+      within(selected_tab) do
+        expect(page).not_to have_content("Query can't be blank")
+        expect(page).to have_content(planning_application1.reference)
+        expect(page).not_to have_content(planning_application2.reference)
+        expect(page).not_to have_content(planning_application3.reference)
+      end
+
+      search_url = current_url
```
benbaumann95, line 71: 🙌

benbaumann95 (approved):
Just left some minor comments but this looks super duper, nice work!

db/migrate/20220722101202_add_reference_to_planning_applications.rb
allbecauseyoutoldmeso: So I tried this and then realised the (I think) fatal flaw...we use `created_at` in the `reference`!

spec/system/searching_spec.rb
allbecauseyoutoldmeso: busted!

db/migrate/20220722101202_add_reference_to_planning_applications.rb
benbaumann95: ahhh haha ok

benbaumann95: I guess we'd use `Time.current` ?

allbecauseyoutoldmeso: Actually yes, I don't see why not! It's only the year, after all, and the changes of someone creating a planning application at the spot of midnight on new years eve and the created at falling those crucial microseconds later is...probably non existent right? 🤣

benbaumann95: haha what a scenario 😂

benbaumann95: in that case (in narnia) we could just do a query to see if created at year is different to any planning application ref year and update manually as there is no unique index on the year itself 😄

allbecauseyoutoldmeso: yes, I'm sure if the day ever comes we can handle it 😁

PR #700, opened 2022-07-25
Set 30mb limit per file size when uploading

benbaumann95:
### Description of change

- Set 30mb limit per file size when uploading
- Now that we've moved the file uploading to a background job, we should be able to lift the overall 30mb limit and do it per file instead to be consistent with the other planning services

### Story Link

https://trello.com/c/zA0j9epO/1065-increase-file-limits-to-30mb-per-item-not-30mb-for-all-items-on-one-application-aligning-with-ripa

### DIFF
--- a/app/controllers/api/v1/replacement_document_validation_requests_controller.rb
+++ b/app/controllers/api/v1/replacement_document_validation_requests_controller.rb
@@ -47,6 +47,18 @@ module Api
 
       private
 
+      def check_file_type
+        unless Document::PERMITTED_CONTENT_TYPES.include? params[:new_file].content_type
+          render json: { message: "The file type must be JPEG, PNG or PDF" }, status: :bad_request
+        end
+      end
+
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/replacement_document_validation_requests_controller.rb
```diff
@@ -47,6 +47,18 @@ def update
 
       private
 
+      def check_file_type
+        unless Document::PERMITTED_CONTENT_TYPES.include? params[:new_file].content_type
+          render json: { message: "The file type must be JPEG, PNG or PDF" }, status: :bad_request
```
EGiataganas, line 52: It will be great if we can also update the AWS policy to allow only these certain file types to be uploaded to my Amazon S3 bucket using our terraform

EGiataganas (approved):
Good job! The comment is not affecting it as is on a separate repo

PR #706, opened 2022-07-29
Block validation if application has no boundary geojson

allbecauseyoutoldmeso:
### Description of change

- Add and render error if validation form is submitted for planning application that has no `boundary_geojson`.

### Story Link

https://trello.com/c/May15sJH/711-prevent-applications-from-being-marked-as-valid-if-they-do-not-have-a-digital-sitemap

### Screenshot

Screenshot 2022-07-29 at 13 02 30

### DIFF
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -101,14 +101,21 @@ class PlanningApplicationsController < AuthenticationController
   def validate
     if validation_date_fields_invalid?
       @planning_application.errors.add(:planning_application, "Please enter a valid date")
-      render "confirm_validation"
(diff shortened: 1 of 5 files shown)


app/controllers/planning_applications_controller.rb
```diff
@@ -101,14 +101,21 @@ def confirm_validation
   def validate
```
allbecauseyoutoldmeso, line 101: I would really love to do a refactor which gets all this logic out of the controller and into the model or some kind of PORO (maybe a form object 😁). But when I started so many different things started to unravel that I thought I'd better leave it for another day!

benbaumann95 (approved):

PR #707, opened 2022-07-29
Add option to send one time password via email

allbecauseyoutoldmeso:
### Description of change

- Add `UserMailer#otp_mail`.
- Add enum `otp_delivery_method` to `users`, with `sms` and `email` options, defaulting to `sms`.
- Extract logic for sending otp into `User` model and update to send using `#otp_mail` if `otp_delivery_method` is `email`.

### Story Link

https://trello.com/c/JE7aV8rr/905-discuss-adding-email-2fa-support

### Screenshot

Screenshot 2022-07-29 at 17 29 43

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -3,6 +3,8 @@
 class User < ApplicationRecord
   enum role: { assessor: 0, reviewer: 1, administrator: 2 }
 
+  enum otp_delivery_method: { sms: 0, email: 1 }
+
   devise :recoverable, :two_factor_authenticatable, :recoverable, :timeoutable,
          :validatable, otp_secret_encryption_key: ENV["OTP_SECRET_ENCRYPTION_KEY"], request_keys: [:subdomains]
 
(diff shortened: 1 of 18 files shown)


app/models/user.rb
```diff
@@ -34,6 +36,19 @@ def valid_otp_attempt?(otp_attempt)
...
+    if send_otp_by_sms?
+      number = mobile_number.presence || session_mobile_number
+      TwoFactor::SmsNotification.new(number, current_otp).deliver!
+    else
+      UserMailer.otp_mail(self).deliver_now
+    end
+  end
+
+  def send_otp_by_sms?
+    otp_delivery_method == "sms"
```
benbaumann95, line 49: does doing just `otp_delivery_method.sms?` work here ?

benbaumann95 (approved):
Lookin good 👍 
Am I right that on the initial signup process, when a new user is created without a mobile we will be prompting for this number (as default enum is sms) .
And the way we set up email OTP is via the admin UI or just updating the db?

allbecauseyoutoldmeso:
> Lookin good 👍 Am I right that on the initial signup process, when a new user is created without a mobile we will be prompting for this number (as default enum is sms) . And the way we set up email OTP is via the admin UI or just updating the db?

Yes and yes.

Following these changes
1. In the admin UI a new user can be created with with either sms (default) or email as the OTP delivery method.
2. This can also be updated for users in the admin UI.
3. If a user is created without a mobile number and with `sms` as the delivery method they will be prompted to add a mobile number on login.
4. If a user is created without a mobile number and with `email` as the delivery method they WON'T be prompted.
5. Finally, for the user in 4, if at a later date an admin updates their delivery method to `sms`, they will be prompted to add a mobile number when they next try to log in 😅

benbaumann95:
> 

Awesome!

This makes sense. Thanks for the detailed explanation 🙏

app/models/user.rb
allbecauseyoutoldmeso: I'd thought that...but `otp_delivery_method` just returns a string. As far as I can see a lot of the enum magic only happens if the column is `status` 😞

benbaumann95: Ah is that so! Good to know

PR #711, opened 2022-08-02
Display map section on following page on decision notice

benbaumann95:
### Description of change

- Display map section on following page on decision notice
- Previous to this the map would render both on page 1 and 2 in a broken way. This change introduces a forced page break so that we can see the map in isolation
- This also removes the button controls from the map in pdf version 

### Story Link

https://trello.com/c/LmUCf7iv/1088-move-digital-sitemap-to-page-2-on-pdf

**BEFORE**

[decision_notice_(1).pdf](https://github.com/unboxed/bops/files/9240890/decision_notice_.1.pdf)

**AFTER**

[decision_notice 2.pdf](https://github.com/unboxed/bops/files/9240894/decision_notice.2.pdf)

### DIFF
--- a/app/controllers/planning_applications_controller.rb
+++ b/app/controllers/planning_applications_controller.rb
@@ -264,6 +264,12 @@ class PlanningApplicationsController < AuthenticationController
     end
   end
 
+  def decision_notice
+    respond_to do |format|
+      format.html
+    end
+  end
+
   def cancel
     case params[:planning_application][:status]
     when "withdrawn"
--- a/app/javascript/stylesheets/decision_notice.scss
+++ b/app/javascript/stylesheets/decision_notice.scss
@@ -21,3 +21,7 @@
   page-break-inside: avoid;
   overflow: hidden;
 }
+
+.decision-notice-main-section {
+  break-after: always;
+}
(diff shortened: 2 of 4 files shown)


app/controllers/planning_applications_controller.rb
```diff
@@ -264,6 +264,12 @@ def determine
     end
   end
 
+  def decision_notice
```
benbaumann95, line 267: We never actually had this action defined here which worked fine because Rails automatically render views with names that correspond to valid routes. Buuut that's probably bad practice and inconsistent

app/javascript/stylesheets/decision_notice.scss
```diff
@@ -21,3 +21,7 @@
   page-break-inside: avoid;
   overflow: hidden;
 }
+
+.decision-notice-main-section {
+  break-after: always;
```
benbaumann95, line 26: https://developer.mozilla.org/en-US/docs/Web/CSS/break-after

app/views/shared/_location_map.html.erb
```diff
@@ -20,6 +20,10 @@
   <% if locals[:geojson_field].present?  %>
     drawMode="true"
   <% end %>
+  <% if @blank_layout %>
+    hideResetControl="true"
+    staticMode="true"
+  <% end %>
```
benbaumann95, line 26: We don't want to show those interactive controls on the pdf version so these are the required props to switch that off

allbecauseyoutoldmeso (approved):
looks good!

PR #712, opened 2022-08-02
Prevent tests around reference breaking in 2023

allbecauseyoutoldmeso:
### Description of change

The `reference` for planning applications is now set using `Date.current` rather than `created_at`. This means that a number of tests will fail on 1/1/23 as they assume that the reference will include the characters `22`, derived from the year.

These changes use `travel_to` to make sure that `Date.current` will continue to return a date in 2022.

### Story Link

NA


benbaumann95 (approved):
⏲️

PR #715, opened 2022-08-03
Fix bug where all files load in the same tab

allbecauseyoutoldmeso:
### Description of change

- Change `target: :_new` to `target: :_blank` for 'View in new window' link. 
- See https://stackoverflow.com/questions/4964130/target-blank-vs-target-new#answer-4964223 - TLDR 'there is no such keyword as `_new`' in HTML5, HTML4 or XHTML 😆 

### Story Link

https://trello.com/c/avgtmnbw/1081-always-open-documents-in-new-tab-and-not-in-a-tab-already-showing-an-open-document

### Screen recording

https://user-images.githubusercontent.com/25392162/182603307-dc41eea5-0919-46d2-85ce-ac38c4793a18.mov


spec/system/documents/index_spec.rb
```diff
@@ -54,6 +54,49 @@
 
       expect(current_url).to include("/rails/active_storage/")
     end
+
+    context "when there is more than one document" do
+      let(:file2) do
+        Rack::Test::UploadedFile.new(
+          Rails.root.join("spec/fixtures/images/proposed-roofplan.png"),
```
allbecauseyoutoldmeso, line 61: I've added another png file because all the other test files in `/spec/fixtures/images` are PDFs. When I tried to write a test that opened any of these PDFs in a new window it kept timing out on github actions - I'm guessing due to the file size?

benbaumann95 (approved):

PR #716, opened 2022-08-03
Add option to search applications by description

allbecauseyoutoldmeso:
### Description of change

- Update `PlanningApplicationsSearch` so that `#results` returns applications with either references OR descriptions that match the query.
- Update query field hint.
- Update the 'All planning applications' tab as per the [dashboard redesign.](https://beta-bops-prototypes.herokuapp.com/v20/application-list3#all-cases)

### Story Link

https://trello.com/c/qWkAsPnX/1026-dashboard-add-description-to-search-feature-on-planning-application-dashboard

### Screenshot

Screenshot 2022-08-03 at 16 00 02

### DIFF
--- a/app/services/planning_application_search.rb
+++ b/app/services/planning_application_search.rb
@@ -8,12 +8,46 @@ class PlanningApplicationSearch
   validates :query, presence: true
 
   def results
-    if valid?
-      planning_applications.where(
-        "LOWER(reference) LIKE ?", "%#{query.downcase}%"
-      )
-    else
-      planning_applications
-    end
+    valid? ? records_matching_query : planning_applications
+  end
+
+  private
+
+  def records_matching_query
+    records_matching_reference.presence || records_matching_description
+  end
+
+  def records_matching_reference
+    planning_applications.where(
+      "LOWER(reference) LIKE ?",
+      "%#{query.downcase}%"
+    )
+  end
+
+  def records_matching_description
+    planning_applications
+      .select(sanitized_select_sql)
+      .where(where_sql, query_terms)
+      .order(rank: :desc)
+  end
+
+  def sanitized_select_sql
+    ActiveRecord::Base.sanitize_sql_array([select_sql, query_terms])
+  end
+
+  def select_sql
+    "*,
+    ts_rank(
+      to_tsvector('english', description),
+      to_tsquery('english', ?)
+    ) AS rank"
+  end
+
+  def where_sql
+    "to_tsvector('english', description) @@ to_tsquery('english', ?)"
+  end
+
+  def query_terms
+    @query_terms ||= query.split.join(" | ")
   end
 end
--- a/app/views/planning_applications/_all_planning_applications_table.html.erb
+++ b/app/views/planning_applications/_all_planning_applications_table.html.erb
@@ -1,0 +1,57 @@
+<div class="govuk-tabs__panel" id="all">
+  <h2 class="govuk-heading-l"><%= all_applications_tab_title %></h2>
+  <%= render(
+    partial: "search/form",
+    locals: {
+      search: search,
+      planning_applications: planning_applications,
+      tab_id: "all"
+    }
+  ) %>
+  <table class="govuk-table">
+    <thead class="govuk-table__head">
+      <tr class="govuk-table__row">
+        <th scope="col" class="govuk-table__header">
+          <%= t(".expiry_date") %>
+        </th>
+        <th scope="col" class="govuk-table__header">
+          <%= t(".application_number") %>
+        </th>
+        <th scope="col" class="govuk-table__header">
+          <%= t(".status") %>
+        </th>
+        <th scope="col" class="govuk-table__header">
+          <%= t(".site_address") %>
+        </th>
+        <th scope="col" class="govuk-table__header">
+          <%= t(".description") %>
+        </th>
+      </tr>
+    </thead>
+    <tbody class="govuk-table__body">
+      <% planning_application_presenters(self, planning_applications).each do |planning_application| %>
+        <tr class="govuk-table__row" id=<%= dom_id(planning_application) %>>
+          <td class="govuk-table__cell">
+            <%= planning_application.expiry_date.strftime("%e %b") %>
+          </td>
+          <td class="govuk-table__cell">
+            <%= link_to(
+              planning_application.reference,
+              planning_application,
+              class: "govuk-link"
+            ) %>
+          </td>
+          <td class="govuk-table__cell">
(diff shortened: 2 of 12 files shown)


app/views/planning_applications/_all_planning_applications_table.html.erb
```diff
@@ -0,0 +1,57 @@
+<div class="govuk-tabs__panel" id="all">
```
allbecauseyoutoldmeso, line 1: Adding this as a separate view for now, but I'm about to dry up all the tabs in https://github.com/unboxed/bops/pull/717!

benbaumann95:
Looks great.

I think we need to add an index to the description though as we are searching on it

benbaumann95:
> Looks great.
> 
> I think we need to add an index to the description though as we are searching on it

I think we go with a GIN index https://www.ramblingcode.dev/posts/adding_index_to_text_columns_in_rails/

`add_index :planning_applications, :description, using: :gin`

And I think we'd want to deploy this when no users are going to be using the system in case it locks up the db for a bit - won't be an issue (but otherwise we'd need to use `algorithm: :concurrently`)

Also I wonder if there is a way we can regex the input or differentiate it at least so we aren't hitting the db unnecessarily 🤔 . But I guess due to the nature of using `LIKE` we are being flexible with this input

allbecauseyoutoldmeso:
Oh doh...we talked about adding the index and I think I even put it in the original trello card, and then I totally forgot 🤦 . Coming up!

Will have a bit more of a think about a regex to recognise the input type...but I think you might be right that we'd be losing flexibility 🤔

benbaumann95:
> Oh doh...we talked about adding the index and I think I even put it in the original trello card, and then I totally forgot 🤦 . Coming up!
> 
> Will have a bit more of a think about a regex to recognise the input type...but I think you might be right that we'd be losing flexibility 🤔

Will you update this to a full text search ? https://pganalyze.com/blog/full-text-search-ruby-rails-postgres
Seems not too bad to add

config/brakeman.ignore
```diff
@@ -1,23 +1,23 @@
...
-      "code": "params.require(:user).permit(:name, :email, :password, :mobile_number, :role)",
+      "warning_type": "SQL Injection",
+      "warning_code": 0,
+      "fingerprint": "024ae0d4e7586e89ac6b9039fccc4a99055fc20b620d35c2efcb76b883abf998",
+      "check_name": "SQL",
+      "message": "Possible SQL injection",
+      "file": "app/services/planning_application_search.rb",
+      "line": 30,
+      "link": "https://brakemanscanner.org/docs/warning_types/sql_injection/",
+      "code": "planning_applications.select(\"*, ts_rank(#{tsvector}, #{tsquery}) AS rank\").where(\"#{tsvector} @@ #{tsquery}\")",
```
allbecauseyoutoldmeso, line 12: I'm using `sanitize_sql_array` on the search query. I think the issue might just be because it's buried away in a private helper method and Brakeman hasn't managed to chase it down 🤔

allbecauseyoutoldmeso:
> > Oh doh...we talked about adding the index and I think I even put it in the original trello card, and then I totally forgot 🤦 . Coming up!
> > Will have a bit more of a think about a regex to recognise the input type...but I think you might be right that we'd be losing flexibility 🤔
> 
> Will you update this to a full text search ? https://pganalyze.com/blog/full-text-search-ruby-rails-postgres Seems not too bad to add

...and added!  @benbaumann95 @EGiataganas this is ready for re-review when you have the time 🙂

app/services/planning_application_search.rb
```diff
@@ -8,12 +8,40 @@ class PlanningApplicationSearch
...
+  def tsquery
+    ActiveRecord::Base.sanitize_sql_array(
+      ["to_tsquery('english', ?)", query_terms]
+    )
+  end
+
+  def query_terms
+    query.split.join(" | ")
   end
 end
```
EGiataganas, line 47: How about moving `attr_accessor :query, :planning_applications` bellow private?

Also I would be tempted to try and add an initialize method to fetch query param.

How about renaming `query`, to `q`?

allbecauseyoutoldmeso: Ah, so it's set up the way it is because it's (thoughbot's favourite thing 😆) a form object! See https://thoughtbot.com/blog/activemodel-form-objects. 

By having `include ActiveModel::Model` and having `query` and `planning_applications` as public attributes we can treat `PlanningApplicationSearch` in a lot of ways like an active model record - eg [passing it to `form_with`](https://github.com/unboxed/bops/blob/main/app/views/search/_form.html.erb#L2), [intializing it with attributes](https://github.com/unboxed/bops/blob/main/app/controllers/planning_applications_controller.rb#L30), and also running [validations](https://github.com/unboxed/bops/blob/main/app/services/planning_application_search.rb#L8).

allbecauseyoutoldmeso: I marginally prefer `query` to `q` just because stylistically I prefer using the whole word, but happy either way ☺️

config/brakeman.ignore
benbaumann95: hmm this seems like a thing we shouldn't ignore so i wonder why it doesn't see `sanitize_sql_array`.

Are we confident that we aren't risk to sql injection, wonder if someone's pulled up an issue on brakeman regarding this

app/services/planning_application_search.rb
benbaumann95: I prefer query too

config/brakeman.ignore
EGiataganas: Have you tried to make the helper method public? Does Brakeman still raise this issue?

allbecauseyoutoldmeso: It still raises it if I just do

```
def records_matching_description
    planning_applications
      .select("*, ts_rank(to_tsvector('english', description), to_tsquery('english', 'banana | orange')) AS rank")
      .where("to_tsvector('english', #{kate}) @@ to_tsquery('english', 'banana | orange')")
      .order(rank: :desc)
  end

  def kate
    "description"
  end
```

🤯

allbecauseyoutoldmeso: I've refactored it so that brakeman is happy!

Still not sure if the problem was 1. brakeman needlessly freaks out at any use of `#{}` in a sql statement or 2. brakeman was rightfully freaking out because I was passing `sanitize_sql_array` isn't as effective if you only hand it an sql fragment instead of the whole statement.

Anyway, I will sleep better at night knowing brakeman is happy 🤣

EGiataganas (approved):

PR #717, opened 2022-08-04
Refactor planning application tabs

allbecauseyoutoldmeso:
### Description of change

- Add [View Components](https://viewcomponent.org/) to the project.
- Use to refactor planning application tabs.

### Story Link

NA

I'd love to know what people think of this! 

I'd hoped it would result in a reduction of code overall, but actually I've ended up with about the same number of lines. The good news is that there are ~100 fewer lines of HTML 😁


benbaumann95 (approved):
looks amazing! nice refactor 👏 

The only thing is that in the docs I see you can add some nice unit testing but is the assumption that all this has been covered in other system specs anyway?

PR #718, opened 2022-08-04
Add validation requests table with polymoprhic association to our existing validation request models

benbaumann95:
### Description of change

- This is a first step into refactoring the validation requests architecture as a whole
- Create validation request records adding foreign key and type of existing validation requests data

Eventually we'll probably just want the one table to hold validation requests but need to do this one step at a time

### Story Link

https://trello.com/c/MfFA7gFH/1105-add-polymorphic-validation-requests-table


allbecauseyoutoldmeso (approved):
Looking good! And like that this is moving in small steps 😅

PR #719, opened 2022-08-10
Improve auditing of auto-closed requests

allbecauseyoutoldmeso:
### Description of change

- Replace `auto_closed` audit activity type with `red_line_boundary_change_validation_request_auto_closed` and `description_change_validation_request_auto_closed` types.
- Update `#auto_close_request!` to set activity type correctly, and to set `activity_information` to `sequence` of validation request.
- Add migration to update existing audits.
- Add link to validation requests in audit log.

### Story Link

https://trello.com/c/wRntF5L6/1071-recording-the-validation-request-against-the-entry-in-the-audit-log

### Screenshot

Screenshot 2022-08-10 at 13 26 30

(I'm going to have a look at the 'User deleted' issue, but in a separate branch and PR since it's not strictly related to the above ticket and also I think might be more complicated that I'd hoped!)

### DIFF
--- a/app/helpers/audit_helper.rb
+++ b/app/helpers/audit_helper.rb
@@ -52,8 +52,6 @@ module AuditHelper
       "Application withdrawn"
     when "closed"
       "Application closed"
-    when "auto_closed"
-      "Request was auto-closed and approved after being open for 5 business days."
     when "description_change_validation_request_sent"
       "Sent: description change request (description##{args})"
     when "description_change_request_cancelled"
--- a/app/models/audit.rb
+++ b/app/models/audit.rb
@@ -30,7 +30,8 @@ class Audit < ApplicationRecord
     started: "started",
     withdrawn: "withdrawn",
     closed: "closed",
-    auto_closed: "auto_closed",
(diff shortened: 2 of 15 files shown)


app/models/audit.rb
```diff
@@ -63,4 +64,16 @@ class Audit < ApplicationRecord
   }
 
   validates :activity_type, presence: true
+
+  def validation_request
```
allbecauseyoutoldmeso, line 68: For the new audit types I've I've copied the logic we're already using for existing audit types, eg `description_change_validation_request_sent`, where `activity_information` is set to the `sequence` of the request. This means that later on we can derive the specific validation request from `activity_type` and `activity_information`. 

I don't love doing it this way - I think it might be worth having a bigger rethink about audits at some point!

app/helpers/audit_helper.rb
```diff
@@ -52,8 +52,6 @@ def activity(type_of_activity, args = nil)
       "Application withdrawn"
     when "closed"
       "Application closed"
```
allbecauseyoutoldmeso, line 54: This `#activity` helper accepts a key and returns a string...ie exactly what `I18n.t` does 😆. Future refactor to move all this to locale files 😁

benbaumann95 (approved):

PR #721, opened 2022-08-11
Fix isses with audit user

allbecauseyoutoldmeso:
### Description of change

- Fix bug where we're not setting the api user for audits associated with API activity.
- Add `automated_activity` column to `audits`
-  Set `automated_activity` to true on audit creation if there is no current user or API user.
- Only show 'User deleted' in audit log if there is no user and `automated_activity` is false.
- Update existing audits.

### Story Link

NA

### DIFF
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -48,14 +48,15 @@ module Api
       private
 
       def authenticate
-        authenticate_or_request_with_http_token do |token, _options|
+        api_user = authenticate_or_request_with_http_token do |token, _options|
           ApiUser.find_by(token: token)
         end
+
+        Current.api_user = api_user
       end
 
       def current_api_user
         @current_api_user ||= authenticate
-        Current.api_user = @current_api_user
       end
 
       protected
--- a/app/models/concerns/auditable.rb
+++ b/app/models/concerns/auditable.rb
@@ -8,12 +8,25 @@ module Auditable
 
     def audit!(activity_type:, activity_information: nil, audit_comment: nil)
       audits.create!(
-        user: Current.user,
+        user: current_user,
         activity_type: activity_type,
         activity_information: activity_information,
         audit_comment: audit_comment,
-        api_user: Current.api_user
+        api_user: current_api_user,
+        automated_activity: no_current_user?
       )
+    end
+
+    def no_current_user?
+      current_user.blank? && current_api_user.blank?
+    end
+
+    def current_user
+      @current_user ||= Current.user
+    end
+
+    def current_api_user
+      @current_api_user ||= Current.api_user
     end
   end
 end
--- a/app/helpers/audit_helper.rb
+++ b/app/helpers/audit_helper.rb
@@ -109,14 +109,34 @@ module AuditHelper
     end
   end
 
-  def define_audit_user(audit)
+  def audit_user_name(audit)
     if audit.api_user.present?
-      audit.activity_type.include?("received") ? "Applicant / Agent via #{audit.api_user.name}" : audit.api_user.name
+      audit_api_user_name(audit)
     elsif audit.user.present?
       audit.user.name
+    elsif audit.automated_activity?
+      t("audit_user_name.system")
     else
-      "User deleted"
+      t("audit_user_name.deleted")
     end
+  end
+
+  def audit_api_user_name(audit)
+    if applicant_activity_types.include?(audit.activity_type)
+      t("audit_user_name.applicant", api_user_name: audit.api_user.name)
+    else
+      audit.api_user.name
+    end
+  end
+
+  def applicant_activity_types
+    %w[
+      description_change_validation_request_received
+      replacement_document_validation_request_received
+      additional_document_validation_request_received
+      red_line_boundary_change_validation_request_received
+      other_change_validation_request_received
+    ]
   end
 
   def audit_entry_template(audit)
--- a/app/views/audits/index.html.erb
+++ b/app/views/audits/index.html.erb
@@ -25,7 +25,7 @@
       <tr class="govuk-table__row">
         <%= content_tag(:tr, id: dom_id(audit_item)) do %>
           <td class="govuk-table__cell"><%= audit_item.created_at.strftime("%d-%m-%Y %H:%M") %></td>
-          <td class="govuk-table__cell"><%= define_audit_user(audit_item) %></td>
+          <td class="govuk-table__cell"><%= audit_user_name(audit_item) %></td>
           <td colspan="2" class="govuk-table__cell">
             <% if audit_item.activity_type.include?("request") || audit_item.activity_type.include?("document_received_at_changed") %>
               <% if audit_item.activity_type.match("/*_validation_request_cancelled") %>
--- a/app/views/shared/_supporting_information.html.erb
+++ b/app/views/shared/_supporting_information.html.erb
@@ -135,7 +135,7 @@
           <%= render "audits/types/#{audit_entry_template(@last_audit)}", locals: { item: @last_audit } %>
 
           <p class="govuk-body">
-            <%= define_audit_user(@last_audit) %>
+            <%= audit_user_name(@last_audit) %>
           </p>
           <p class="govuk-body">
             <%= @last_audit.created_at.to_date %> at <%= @last_audit.created_at.strftime("%H:%M") %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,4 +1,8 @@
 en:
+  audit_user_name:
+    applicant: Applicant / Agent via %{api_user_name}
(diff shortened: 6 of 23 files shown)


app/controllers/api/v1/application_controller.rb
```diff
@@ -48,14 +48,15 @@ def check_token_and_set_application
...
       def authenticate
-        authenticate_or_request_with_http_token do |token, _options|
+        api_user = authenticate_or_request_with_http_token do |token, _options|
           ApiUser.find_by(token: token)
         end
+
+        Current.api_user = api_user
       end
 
       def current_api_user
```
allbecauseyoutoldmeso, line 58: This method is not called in all actions. This means that in many cases `Current.api_user` is not set, and therefore when an audit is created `api_user` is `nil`.

allbecauseyoutoldmeso, line 58: This method is not called in all actions. This means that in many cases `Current.api_user` is not set, and therefore when an audit is created `api_user` is `nil`.

allbecauseyoutoldmeso, line 58: This method is not called in all actions. This means that in many cases `Current.api_user` is not set, and therefore when an audit is created `api_user` is `nil`.

app/models/concerns/auditable.rb
```diff
@@ -8,12 +8,21 @@ module Auditable
...
     def audit!(activity_type:, activity_information: nil, audit_comment: nil)
       audits.create!(
-        user: Current.user,
+        user: user,
         activity_type: activity_type,
         activity_information: activity_information,
         audit_comment: audit_comment,
-        api_user: Current.api_user
+        api_user: api_user,
+        automated_activity: user.blank? && api_user.blank?
```
allbecauseyoutoldmeso, line 16: At this point if there's no current user or api user we know it's a an automated action.

db/migrate/20220810131220_add_automated_activity_to_audits.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class AddAutomatedActivityToAudits < ActiveRecord::Migration[6.1]
+  # rubocop:disable Metrics/MethodLength
```
allbecauseyoutoldmeso, line 4: SQLocop would have to be more forgiving 😆

db/migrate/20220810131220_add_automated_activity_to_audits.rb
```diff
@@ -0,0 +1,44 @@
...
+  def up
+    add_column(
+      :audits,
+      :automated_activity,
+      :boolean,
+      default: false,
+      null: false
```
allbecauseyoutoldmeso, line 11: Does it make sense to have `null: false` if the column has a default? I can't decide!

app/controllers/api/v1/application_controller.rb
allbecauseyoutoldmeso, line 58: This method isn't called in all actions, which means `Current.api_user` doesn't get set, which in turn means that `api_user` isn't set for audits.

app/models/concerns/auditable.rb
```diff
@@ -8,12 +8,21 @@ module Auditable
...
     def audit!(activity_type:, activity_information: nil, audit_comment: nil)
       audits.create!(
-        user: Current.user,
+        user: current_user,
         activity_type: activity_type,
         activity_information: activity_information,
         audit_comment: audit_comment,
-        api_user: Current.api_user
+        api_user: current_api_user,
+        automated_activity: current_user.blank? && current_api_user.blank?
```
allbecauseyoutoldmeso, line 16: At this stage if there's no current user or api user we know it's an automated task!

db/migrate/20220810131220_add_automated_activity_to_audits.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class AddAutomatedActivityToAudits < ActiveRecord::Migration[6.1]
+  # rubocop:disable Metrics/MethodLength
```
allbecauseyoutoldmeso, line 4: SQLocop would need to be more forgiving 😆

db/migrate/20220810131220_add_automated_activity_to_audits.rb
```diff
@@ -0,0 +1,41 @@
...
+  def up
+    add_column(
+      :audits,
+      :automated_activity,
+      :boolean,
+      default: false,
+      null: false
```
allbecauseyoutoldmeso, line 11: Does it make sense to have `null: false` when there's a default? I can't decide!

db/migrate/20220810131220_add_automated_activity_to_audits.rb
```diff
@@ -0,0 +1,41 @@
...
+    execute(
+      "UPDATE audits
+      SET automated_activity = TRUE
+      WHERE activity_type IN (
+        'description_change_validation_request_auto_closed',
+        'red_line_boundary_change_validation_request_auto_closed'
+      );"
+    )
+
+    execute(
```
allbecauseyoutoldmeso, line 23: Updates the 'description updated' audit associated with the auto-closure of the description change request.

db/migrate/20220811093302_update_audits_generated_by_api_user.rb
```diff
@@ -0,0 +1,35 @@
...
+      AND activity_type IN (
+        'description_change_validation_request_received',
+        'replacement_document_validation_request_received',
+        'additional_document_validation_request_received',
+        'red_line_boundary_change_validation_request_received',
+        'other_change_validation_request_received'
+      );"
+    )
+
+    execute(
```
allbecauseyoutoldmeso, line 21: Updates the 'document uploaded' audit associated with the additional document 'additional_document_validation_request_received' event.

db/migrate/20220811093302_update_audits_generated_by_api_user.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+# rubocop:disable Rails/ReversibleMigration
+class UpdateAuditsGeneratedByApiUser < ActiveRecord::Migration[6.1]
+  # rubocop:disable Metrics/MethodLength
+  def change
+    execute(
+      "UPDATE audits
+      SET api_user_id = 1
```
allbecauseyoutoldmeso, line 9: Luckily there's only one API user record on preview, with id 1, and on production only 2 - id 1 is 'api_user', and id 2 is 'RIPA.

app/controllers/api/v1/application_controller.rb
allbecauseyoutoldmeso, line 58: This method isn't called in all actions, meaning that `Current.api_user` doesn't get set, and therefore the `api_user` isn't set for audits.

app/models/concerns/auditable.rb
EGiataganas: How about moving it into a method `#has_current_api_and_user?` or any other naming you prefer?

db/migrate/20220810131220_add_automated_activity_to_audits.rb
EGiataganas: There's this [article](1) which states why is best to have `null: false, default: false`

[1]:https://thoughtbot.com/blog/avoid-the-threestate-boolean-problem

db/migrate/20220811093302_update_audits_generated_by_api_user.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class UpdateAuditsGeneratedByApiUser < ActiveRecord::Migration[6.1]
+  # rubocop:disable Rails/ReversibleMigration
+  # rubocop:disable Metrics/MethodLength
+  def change
```
EGiataganas, line 6: does this needs to be `def up`?

app/models/concerns/auditable.rb
allbecauseyoutoldmeso: Good shout!

db/migrate/20220810131220_add_automated_activity_to_audits.rb
allbecauseyoutoldmeso: Can you reshare? I think you've linked the wrong thing 😆

EGiataganas: That’s weird tried to edit the commit and the url is correct, here it is https://thoughtbot.com/blog/avoid-the-threestate-boolean-problem

allbecauseyoutoldmeso: Awesome, thank you!

benbaumann95 (approved):
🙌

EGiataganas (approved):
Great work!! 😄

PR #724, opened 2022-08-15
Remove ability to update planning application created_at date

benbaumann95:
### Description of change

Remove ability to update planning application created_at date
- Add received_at column and allow users to update this . We should never permit a created_at timestamp to be updated

### Story Link

https://trello.com/c/agAgh2He/1079-high-priority-remove-ability-to-update-createdat-timestamp-on-planning-applications


app/models/planning_application.rb
```diff
@@ -303,7 +303,7 @@ def documents_for_decision_notice
   end
 
   def received_at
-    Time.next_immediate_business_day(created_at)
+    super || (Time.next_immediate_business_day(created_at) if created_at)
```
allbecauseyoutoldmeso, line 306: I think `created_at` should still always be present?

allbecauseyoutoldmeso (approved):
Looks good!

Just wondering if there's any value to retrospectively updating `received_at` for existing records?

app/models/planning_application.rb
benbaumann95: Yeah left it out but then on planning application create page it was calling this code `Time.next_immediate_business_day(created_at)` and nil was blowing up. Maybe I can edit that [method](https://github.com/unboxed/bops/blob/main/config/initializers/business_time.rb#L13-L15) directly though just to return if the argument is nil 

What do you reckon?

benbaumann95:
> Looks good!
> 
> Just wondering if there's any value to retrospectively updating `received_at` for existing records?

My understanding was that it's just an override if the next business day code isn't quite correct.
I'm not sure what an accurate update would look like 🤔

app/models/planning_application.rb
allbecauseyoutoldmeso: Ah, I hadn't thought of new records! I think it's probably best as you've already done it, with the check in the model.

PR #726, opened 2022-08-17
Apply design changes to proposal details

allbecauseyoutoldmeso:
### Description of change

- Change styling of anchor links
- Update copy
- Allow filtering by 'auto answered'
- Add 'back to top' links

### Story Link

https://trello.com/c/3T7wafsB/1110-support-users-to-better-understand-proposal-detail-groups

### TO DO

- [ ] Fix specs
- [ ] Back to top links


app/helpers/proposal_details_helper.rb
```diff
@@ -37,7 +37,7 @@ def proposal_detail_responses(responses)
   end
 
   def proposal_detail_metadata(metadata)
-    return if metadata.nil?
+    return if metadata.blank?
```
allbecauseyoutoldmeso, line 40: Return if metdata is `{}` as well as if metadata is `nil`.

app/views/shared/_back_to_top_link.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= link_to(
+  target_id,
+  class: "govuk-link govuk-link--no-visited-state back-to-top-link"
+) do %>
+  <svg role="presentation" focusable="false" class="back-to-top-icon" xmlns="http://www.w3.org/2000/svg" width="13" height="17" viewBox="0 0 13 17">
+    <path fill="currentColor" d="M6.5 0L0 6.5 1.4 8l4-4v12.7h2V4l4.3 4L13 6.4z"></path>
+  </svg>
+  <p><%= t(".back_to_top") %></p>
+<% end %>
```
allbecauseyoutoldmeso, line 9: There's currently no 'Back to top' link in the gov UK design system, although there's one [used on their site](https://design-system.service.gov.uk/get-started/), and [an open issue to discuss it](https://github.com/alphagov/govuk-design-system-backlog/issues/5). 

I've put this in a shared partial so we can use it elsewhere as needed until, hopefully, an official one is added!

.github/workflows/build.yml
```diff
@@ -279,7 +279,7 @@ jobs:
         env:
           DATABASE_URL: postgres://postgres:postgres@localhost:5432/test
           RAILS_ENV: test
-          SPEC_OPTS: "-f doc -P spec/{helpers,presenters,mailer,jobs,services}/**/*_spec.rb"
+          SPEC_OPTS: "-f doc -P spec/{helpers,presenters,mailer,jobs,services,translations}/**/*_spec.rb"
```
allbecauseyoutoldmeso, line 282: Just spotted that we're not actually running the specs for i18n configuration in github actions 😬

benbaumann95 (commented):
Hey this looks really great, sorry it's taken long to look at 🎉 . 

I still think I would benefit if we went through this quickly together though at some point just to help my understanding of certain decisions if that's ok?

allbecauseyoutoldmeso:
> Hey this looks really great, sorry it's taken long to look at 🎉 .
> 
> I still think I would benefit if we went through this quickly together though at some point just to help my understanding of certain decisions if that's ok?

Yes of course...this was a painful piece of work so grateful for other eyes on it! Any time today works for me I think.

notapatch (approved):

PR #732, opened 2022-08-23
Refactor asess recommendation flow to be more RESTful

allbecauseyoutoldmeso:
### Description of change

- `PlanningApplicationsController#recommendation_form` -> `RecommedationsController#new` and  `PlanningApplicationsController#recommend` -> `RecommedationsController#create`
- Add a `RecommendationForm` form object to handle creating the recommendation and updating the planning application.

### Story Link

NA


benbaumann95 (approved):

app/services/recommendation_form.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
```
benbaumann95, line 1: Should this be inside `models/`?
Not sure it's a service as such

allbecauseyoutoldmeso: yes, I think you're right!

PR #737, opened 2022-08-30
FIX: PlanningApplicationMailer unexpected urls

notapatch:
- If you run spec/models/planning_application_spec everything
  worked.
- PlanningApplicationMailer url was unexpected only when you
  ran the full test suite.

Tracked down by
1. Noticing the difference in expectation had something to do
   with the URL and *eventually* that it was giving the
   public_url and not the expected internal_url
2. URL output relies on ENV settings
3. ENV is a global througout the test suite (see link below)
4. Commenting out other ENV variables and running the full test
   suite. All the mailer tests now passed.
5. ENV["PUBLIC_URL_ENABLED"] affects if public_url or internal_url
   is used.
6. FIX

- Fixes: Choose 3

Version 1) Shortest but keeps the ENV global
- spec/models/planning_application_spec.rb
- Line 800 ish
+      after do
+        ENV["PUBLIC_URL_ENABLED"] = nil
+      end

Version 2) around to stop the change in values between each test
https://stackoverflow.com/a/71208726/1511504

Version 3) Which I went was to replace the global with stubs.
I also wanted to access ENV consistently and switched
planning_application ENV calls from [] to fetch.

### Description of change

### Story Link

https://link-to-issue

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


spec/mailer/planning_application_mailer_spec.rb
```diff
@@ -59,7 +59,8 @@
   end
 
   before do
-    ENV["APPLICANTS_APP_HOST"] = "example.com"
+    allow(ENV).to receive(:[])
```
benbaumann95, line 62: Do we need this line as well, is this an extra step we need on top of the line below?

I would have thought just `allow(ENV).to receive(:[]).with("APPLICANTS_APP_HOST").and_return("example.com")` would suffice?

benbaumann95: Ah nvm i tested locally, that doesn't work..

I guess there aren't many cases we stub the env here but might be nice moving to a module under `spec/support` i.e.

```ruby
module EnvHelper
  def set_env(name:, value:)
    allow(ENV).to receive(:[])
    allow(ENV).to receive(:[]).with(name).and_return(value)
  end
end

if RSpec.respond_to?(:configure)
  RSpec.configure do |config|
    config.include(EnvHelper)
  end
end
```

then we can just do something like `set_env(name: "APPLICANTS_APP_HOST", value: "example.com")`

But not super fussed, will leave up to you

notapatch: The comment for the stackoverflow was:

`allow(ENV).to receive(:[])  # stub a default value first if message might be received with other args as well.`

When I tested the code without it and it blows up as other messages are indeed calling it:

- For the mailer it calls ENV with: "FARADAY_NO_AUTOLOAD"
- For the planning_application model spec:  "https_proxy"

I think your helper code looks cool; not seen `RSpec.resond_to?(:configure)`! I wanted to just get the fix it. There's an issue about ENV vs ENV.fetch. Some of the code uses one and some the other. 

Thoughtbot had this to say: 
`Use ENV.fetch for environment variables instead of ENV[]so that unset environment variables are detected on deploy.`

[Rubocop has introduced a cop for it.](https://msp-greg.github.io/rubocop/RuboCop/Cop/Style/FetchEnvVar.html)

It might make sense in a later commit to change ENV => ENV.fetch throughout the code and introduce a helper like you suggested.. fix it all in 1 sweep?

benbaumann95: Cool thanks for the explanation.
Yeah it's best to be consistent I suppose, might be worth adding that rubocop rule 👍 

Yep, will leave it up to you in terms of doing it as another commit / PR.

benbaumann95 (approved):

PR #738, opened 2022-08-30
Fix: Locator for selector must be an instance of String or Symbol

notapatch:
- this is a warning that will become an error in a later version
  of Capybara

- found the test causing the warning by switchinging rspec format:
   --format documenation

- object instance was a `ActiveStorage::Filename` the
  object returns a santizied string if you call to_s on it and it
  still passed the test.


allbecauseyoutoldmeso (approved):
👍

PR #739, opened 2022-08-30
Refactor use constant instead of magic number

notapatch:
- the 56 day constant should be used in other tests which are based
  on expiry date: days_left and days_overdue. But keeping the
  change as simple as possible

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -15,6 +15,8 @@ class PlanningApplication < ApplicationRecord
 
   include PlanningApplication::Notification
 
+  DAYS_TO_EXPIRE = 56
+
   enum application_type: { lawfulness_certificate: 0, full: 1 }
 
   enum user_role: { applicant: 0, agent: 1, proxy: 2 }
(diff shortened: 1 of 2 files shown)


app/models/planning_application.rb
```diff
@@ -444,7 +446,7 @@ def set_reference
   end
 
   def set_key_dates
-    self.expiry_date = 56.days.after(validated_at || received_at)
+    self.expiry_date = DAYS_TO_EXPIRE.days.after(validated_at || received_at)
     self.target_date = 35.days.after(validated_at || received_at)
```
benbaumann95, line 450: Hey any reason why we wouldn't update this to a constant too then ?

notapatch: The 35? It should be changed.

But I didn't know enough about it to give it a name and then I would have to use it in the tests. My idea would be to do the change separately later when I would change it and change the tests it is involved with.

benbaumann95: Yeah exactly, I agree with moving it to a constant and making it consistent so change separately sounds good 🙌 .

Perhaps just `DAYS_TO_TARGET`  ? But you're right I'm not even sure we use this date really or it doesn't have much significance anyway in the business context.

I think at some point we were going to do an audit of the code to pull out these dates and create some documentation like we've done with notifications so we have a better oversight and can manage it better going forward

benbaumann95 (approved):

PR #740, opened 2022-08-30
Improve review recommendation user flow

benbaumann95:
### Description of change

- Move review recommendation logic out of planning application controller to be more restful / conventional

There was far too much going on in the planning applications controller for the reviewing recommendation logic so this PR aims to resolve some of that. Also there were multiple updates done separately and not wrapped in a transaction so if a later update failed it would potentially lead to bad/inconsistent data down the line..

### Story Link

https://trello.com/c/q9ldIBw3/1120-improve-review-the-recommendation-page


app/controllers/recommendations_controller.rb
```diff
@@ -22,8 +30,57 @@ def create
     end
   end
 
+  def review_form
```
allbecauseyoutoldmeso, line 33: Could we arguably change `review_form` to `edit`, and `review` to `update`? 

If not I'm almost inclined to say we should have `reviews_controller` instead, with `new` and `create`. It would be really nice to only stick 100% to restful actions 🤔

app/controllers/recommendations_controller.rb
```diff
@@ -1,7 +1,15 @@
 # frozen_string_literal: true
 
 class RecommendationsController < AuthenticationController
-  before_action :set_planning_application
+  before_action :set_planning_application, only: %i[new create]
```
allbecauseyoutoldmeso, line 4: Oo...I think it wouldn't break anything, and would actually be more efficient, to use `set_planning_application_with_recommendations` instead of `set_planning_application` for new and create!

app/controllers/recommendations_controller.rb
```diff
@@ -22,8 +30,57 @@ def create
...
+  def set_recommendations
+    @recommendations = @planning_application.recommendations
+  end
+
+  def planning_applications_scope
+    current_local_authority.planning_applications.includes(:recommendations)
+  end
+
+  def planning_application_id
+    Integer(params[:planning_application_id])
```
allbecauseyoutoldmeso, line 69: Isn't Rails clever enough to not need this? `PlanningApplication.find("1")` should return the same as `PlanningApplication.find(1)`

app/controllers/recommendations_controller.rb
```diff
@@ -22,8 +30,57 @@ def create
...
+  end
+
+  def set_recommendation
+    @recommendation = @planning_application.recommendations.find(recommendation_id)
+  end
+
+  def recommendation_id
```
allbecauseyoutoldmeso, line 76: And again..?

app/models/recommendation.rb
```diff
@@ -11,7 +15,7 @@ class Recommendation < ApplicationRecord
 
   validate :reviewer_comment_is_present?
 
-  attr_accessor :agree
+  delegate :audits, to: :planning_application
```
allbecauseyoutoldmeso, line 18: Love it!

app/views/recommendations/_events.html.erb
```diff
@@ -1,7 +1,7 @@
 <% if recommendations.any? %>
-  <h3 class="govuk-heading-m">
+  <h2 class="govuk-heading-m">
```
allbecauseyoutoldmeso, line 2: 🙇

app/views/recommendations/review_form.html.erb
```diff
@@ -29,12 +29,14 @@
...
+  <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+
   <%= render(
     "recommendations/events",
-    { recommendations: @planning_application.recommendations }
+    { recommendations: @recommendations }
   ) %>
 
-  <%= form_for @recommendation, url: review_planning_application_path(@planning_application) do |form| %>
+  <%= form_for @recommendation, url: review_planning_application_recommendation_path(@planning_application, @recommendation) do |form| %>
```
allbecauseyoutoldmeso, line 39: While you're here it might be worth updating this form to use [the gov uk form builder gem](https://govuk-form-builder.netlify.app/introduction/get-started/)? But it's a faff, so happy for use to leave it for another day!

allbecauseyoutoldmeso (approved):
Nicely done!

Couple of comments...I do think it would be nice to only have actions like `edit`, `update` etc if we can. But appreciate it's a work in progress!

app/views/recommendations/_event.html.erb
```diff
@@ -6,7 +6,7 @@
       ".by_name_at",
       name: name,
       date: datetime.strftime("%-d  %b %Y"),
-      time: datetime.strftime("%I:%M%P")
+      time: datetime.strftime("%H:%M%P")
```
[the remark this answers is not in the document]
allbecauseyoutoldmeso: Oh wait, if we change this I think we should remove the `P`? Otherwise we get things like `13:00PM`, where the `PM` is redundant 🤔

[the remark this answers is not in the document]
benbaumann95: The wireframes had it like `12:31 PM` but I guess common sense might mean you are right to drop this?

app/controllers/recommendations_controller.rb
benbaumann95: Yeah I think you are right.. not sure why i didn't frame it like this

app/controllers/recommendations_controller.rb
benbaumann95: Didn't think any recommendation records were loaded in this action. Let me double check

app/views/recommendations/_event.html.erb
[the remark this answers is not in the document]
EGiataganas: How about moving these values in the `yml`?

For example:
```
en:
  time:
    formats:
      default: 'H:%M%P'
```

So you might not need to do `datetime.strftime("%H:%M%P")`

app/controllers/recommendations_controller.rb
benbaumann95: habit and i think Andrew told me once to do this before.. let me ask why again 😄

app/views/recommendations/_event.html.erb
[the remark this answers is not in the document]
allbecauseyoutoldmeso: oh, I like!

app/controllers/recommendations_controller.rb
```diff
@@ -22,8 +30,49 @@ def create
     end
   end
 
+  def edit
```
allbecauseyoutoldmeso, line 33: 😍

allbecauseyoutoldmeso (approved):
Doubly approved!

app/controllers/recommendations_controller.rb
allbecauseyoutoldmeso: thanks...good to know!

config/routes.rb
```diff
@@ -24,7 +24,8 @@
       get :part, on: :new
     end
 
-    resources :recommendations, only: %i[new create]
+    resources :recommendations, only: %i[new create update]
```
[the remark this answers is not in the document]
benbaumann95: No because `resources` requires an :id for edit whereas `resource` does not and since we don't have render the page for a single resource.

I want the url like `/recommendations/edit` rather than ``/recommendations/edit/:id` otherwise will have to load up a query for recommendations on the planning application show page to find this for providing the url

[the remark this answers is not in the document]
allbecauseyoutoldmeso: Ohh...yes, I get you 😄

app/views/recommendations/_event.html.erb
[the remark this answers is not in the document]
benbaumann95: There's a `date_formatter.rb` file with a default already.
I think in another PR we can do a clean up of dates and use this file to reuse different formats

benbaumann95:
> Doubly approved!

Thanks for the review, your suggestions much appreciated ! 🙏

PR #741, opened 2022-08-30
Move 'assessment_dashboard' partial to shared folder

allbecauseyoutoldmeso:
### Description of change

[This refactor](https://github.com/unboxed/bops/pull/732) means that the `assessment_dashboard` partial can be rendered in views that are not in the 'app/views/planning_applications' folder. 

Which is why we got the below error:

Screenshot 2022-09-01 at 16 54 50

So move that partial and those associated with it to the 'shared' folder, and call it correctly from all views.


notapatch (approved):
LGTM

PR #743, opened 2022-09-02
Let users add comments to policies

allbecauseyoutoldmeso:
### Description of change

- Add a text areas to the policy class form which allow users to add a comment for each policy.

### Story Link

https://trello.com/c/90BX4FSz/1140-comment-option-for-each-paragraph-of-legislation-and-check-what-is-permitted-overall-by-class (covers first two Acceptance Criteria only)

### Screenshots

No comment added:
Screenshot 2022-09-02 at 16 53 41

Comment added:
Screenshot 2022-09-02 at 16 52 50

Comment updated:
Screenshot 2022-09-02 at 16 53 14

### Decisions

We've talked about extracting the information in the `policy_classes` column in `planning_applications` into separated `policy_classes` and `policies` tables with their own models. I've left it as is for now because I'd like to have worked with this aspect of the application a bit more and understand a bit more about it before making any big architectural decisions!

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+class Comment < ApplicationRecord
+  belongs_to :user, optional: true
+  belongs_to :policy
+
+  validates :text, presence: true
+
+  before_save :set_user
+
+  def edited?
+    created_at != updated_at
+  end
+
+  private
+
+  def set_user
+    self.user = Current.user
+  end
+end
--- a/app/models/policy.rb
+++ b/app/models/policy.rb
@@ -2,6 +2,13 @@
 
 class Policy < ApplicationRecord
   belongs_to :policy_class
+  has_one :comment, dependent: :destroy
+
+  accepts_nested_attributes_for(
+    :comment,
+    update_only: true,
+    reject_if: proc { |attributes| attributes[:text].blank? }
+  )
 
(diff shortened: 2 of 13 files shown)


benbaumann95 (commented):
Hey this is looking good, these policy classes seem a bit complicated don't they! 🤯 

Think I would benefit again if we went through this with a chat ?

allbecauseyoutoldmeso:
@EGiataganas @benbaumann95 @notapatch this is now rebased, rationalised, and ready for re-review!

app/models/comment.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class Comment < ApplicationRecord
+  belongs_to :user, optional: true
+  belongs_to :policy
```
allbecauseyoutoldmeso, line 5: I can imagine a future where we switch this for a polymorphic association, given that lots of things could need comments.

app/models/policy.rb
```diff
@@ -2,6 +2,13 @@
 
 class Policy < ApplicationRecord
   belongs_to :policy_class
+  has_one :comment, dependent: :destroy
+
+  accepts_nested_attributes_for(
+    :comment,
+    update_only: true,
```
allbecauseyoutoldmeso, line 9: Update existing record, don't destroy and replace existing record.

notapatch (approved):

PR #744, opened 2022-09-06
Create policy_class and policies tables

allbecauseyoutoldmeso:
### Description of change

- Create `policy_class` and `policy` tables to store data currently held in the `policy_classes` jsonb column in `planning_applications`.
- Update `PolicyClass` to be an Active Record model.
- Add `Policy` Active Record model.
- Update code to work with these models rather that the jsonb object.

### Story Link

NA


app/controllers/policy_classes_controller.rb
```diff
@@ -19,7 +19,7 @@ def new
   end
 
   def create
```
allbecauseyoutoldmeso, line 21: I think the form and this action are due a refactor, but one for another day!

app/controllers/policy_classes_controller.rb
```diff
@@ -46,17 +45,11 @@ def create
...
-    new_policies = policies_params[:policies]
-
-    @policy_class.policies.each do |policy|
-      value = new_policies[policy["id"].to_s]
-
-      policy["status"] = value if value.present?
+    if @policy_class.update(policy_class_params)
+      redirect_to @planning_application, notice: "Successfully updated policy class"
+    else
+      render :show
```
allbecauseyoutoldmeso, line 51: Not sure how the form ended up in `show` and not `edit`...but we're about to overhaul a lot of things in upcoming work and not sure how it will look, so leaving it that way for now!

app/models/planning_application.rb
```diff
@@ -303,10 +301,6 @@ def agent_or_applicant_name
     end
   end
 
-  def policy_classes_editable
```
allbecauseyoutoldmeso, line 306: `policy_classes_changed?` obviously breaks now it's an association rather than a column, and there's a check in `PolicyClassesController` that makes this redundant so I think it's safe to remove.

notapatch (approved):
Well the first one is more internal … you’re moving from one way of doing something to another? And it’s easier to reason …. LGTM

EGiataganas (approved):

PR #745, opened 2022-09-08
Allow user to mark assess against legislation form as complete

allbecauseyoutoldmeso:
### Description of change

- Add `status` column to `policy_classes`, which defaults to `:in_assessment`.
- Add 'Save and mark as complete' and 'Save and come back later' buttons. 'Save and mark as complete' updates the status to `:complete.`
- Replace 'show' action with 'show' and 'edit' actions. When the user clicks on the policy class link they are taken to show if the status is `:complete` and edit if it's `:in_assessment`.

### Story Link

https://trello.com/c/90BX4FSz/1140-allow-users-to-save-and-mark-as-complete-or-save-and-come-back-later-on-assess-legislation-page

### Screenshots

New buttons:

Screenshot 2022-09-08 at 15 58 51

Error message if click 'Save and mark as complete' when some policies are still 'To be determined':

Screenshot 2022-09-08 at 15 59 07

Show view for 'Complete' policy classes:

Screenshot 2022-09-08 at 16 00 03

### DIFF
--- a/app/models/policy.rb
+++ b/app/models/policy.rb
@@ -4,6 +4,8 @@ class Policy < ApplicationRecord
   belongs_to :policy_class
   has_one :comment, dependent: :destroy
 
+  default_scope { order(:section) }
+
   accepts_nested_attributes_for(
     :comment,
     update_only: true,
--- a/app/views/policy_classes/_form.html.erb
+++ b/app/views/policy_classes/_form.html.erb
@@ -1,0 +1,127 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <%= form_with(
+      model: [planning_application, policy_class],
+      html: { data: unsaved_changes_data },
+      builder: GOVUKDesignSystemFormBuilder::FormBuilder
+    ) do |form| %>
+      <%= form.govuk_error_summary %>
+      <table class="govuk-table">
+        <caption class="govuk-table__caption govuk-table__caption--l">
+          <%= t(
+            ".table_caption",
+            part: policy_class.part,
+            class: policy_class.section,
+            name: policy_class.name
+          ) %>
+        </caption>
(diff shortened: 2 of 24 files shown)


app/models/policy.rb
```diff
@@ -4,6 +4,8 @@ class Policy < ApplicationRecord
   belongs_to :policy_class
   has_one :comment, dependent: :destroy
 
+  default_scope { order(:section) }
```
allbecauseyoutoldmeso, line 7: Ensure that the policies are always displayed '1a, 1b, 2a' etc. 

I tried putting the ordering in `PolicyClass` (`has_many :policies, -> { order(:section) }`) but this seems to cause errors when used in conjunction with `accepts_nested_attributes_for`. I tried looking for known bugs without much success 😬

app/views/policy_classes/_form.html.erb
```diff
@@ -0,0 +1,127 @@
+<div class="govuk-grid-row">
```
allbecauseyoutoldmeso, line 1: Most of this has just been moved from the previous 'show' page.

db/migrate/20220907161526_add_status_to_policy_classes.rb
```diff
@@ -0,0 +1,27 @@
...
+      FROM policies
+      WHERE policies.policy_class_id = policy_classes.id
+      AND policies.status = 2;"
+    )
+
+    execute(
+      "UPDATE policy_classes
+      SET status = 1
+      WHERE status IS NULL;"
+    )
```
allbecauseyoutoldmeso, line 19: If a policy class has any policies that are 'to be determined' set the status as 'in assessment'. Otherwise set the status as 'complete'.

notapatch (approved):
LGTM

PR #746, opened 2022-09-08
Fix bug where users can add policy class twice

allbecauseyoutoldmeso:
### Description of change

- Make sure previously added policy classes are checked and disabled on the add policy classes form, so that they can't be added twice!

### Story Link

NA. The issue is due to [this refactor](https://github.com/unboxed/bops/pull/744).

### Screenshot

Screenshot 2022-09-08 at 16 50 45


notapatch (approved):
LGTM

PR #748, opened 2022-09-13
Add button to delete policy comments

allbecauseyoutoldmeso:
### Description of change

- Add 'Delete comment' button to each comment in form.
- Button triggers javascript that makes delete comment request and refreshes the comment field partial.

### Story Link

https://trello.com/c/IFJqyJTg/1192-delete-comment-button-for-each-comment-on-assess-legislation-page

### Screenshots

Screenshot 2022-09-13 at 10 55 28

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+class CommentsController < AuthenticationController
+  def destroy
+    policy.comment.destroy
+    render(json: { partial: comment_form })
+  end
+
+  private
+
+  def comment_form
+    render_to_string(
+      partial: "policy_classes/comment_form",
+      locals: {
+        comment: policy.build_comment,
+        policy_index: params[:policy_index]
+      }
+    )
+  end
+
+  def policy
+    @policy ||= policy_class.policies.find(params[:policy_id])
+  end
+
+  def policy_class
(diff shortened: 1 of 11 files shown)


app/views/policy_classes/_comment_form.html.erb
```diff
@@ -0,0 +1,34 @@
+<%= content_tag(
```
allbecauseyoutoldmeso, line 1: Not using `fields_for` or govuk form builder helpers, since this partial needs to be refreshed without the context of the whole form.

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class CommentsController < AuthenticationController
+  def destroy
+    policy.comment.destroy
+    render(json: { partial: comment_form })
```
notapatch, line 6: Json because you're using Javascript?

notapatch (approved):
LGTM

EGiataganas (approved):

app/controllers/comments_controller.rb
allbecauseyoutoldmeso: Yes...although I have to admit I copied this from a pattern I'd used on another project. I think it probably makes sense though?

notapatch: Makes sense. destroy, assuming it works, can only go wrong with the destroy cascade. Otherwise it's easy street.

PR #749, opened 2022-09-15
Add consistency checklist to assessment tasks

allbecauseyoutoldmeso:
### Description of change

- Add `concistency_checklists` table and `ConistsencyChecklist` model. A planning application has one consistency checklist.
- Add consistency checklist controller and views, and functionality so that users can create and edit consistency checklist, and view it once marked as commplete.

### Story Link

https://trello.com/c/s9bqs7dY/1154-descriptions-need-to-match-documents-drawings-proposals-from-ripa

### Screenshots

Screenshot 2022-09-16 at 15 02 31

#### With errors (when user clicks 'Save and mark as complete'):

Screenshot 2022-09-16 at 15 02 53

#### 'No' conditionals:

Screenshot 2022-09-16 at 15 04 00

#### Description change requested ('Request a change to the description' link hidden because only one request can be open at a time):

Screenshot 2022-09-16 at 15 07 45

#### Additional document requested ('Request a new document' link still visible because you can open multiple requests):

Screenshot 2022-09-16 at 15 08 40

#### Marked as complete:

Screenshot 2022-09-16 at 15 11 06

### DIFF
--- a/app/controllers/concerns/commit_matchable.rb
+++ b/app/controllers/concerns/commit_matchable.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module CommitMatchable
+  extend ActiveSupport::Concern
+
+  def commit_matches?(regex)
+    params[:commit]&.downcase&.match(regex).present?
+  end
+end
--- a/app/models/consistency_checklist.rb
+++ b/app/models/consistency_checklist.rb
@@ -1,0 +1,88 @@
+# frozen_string_literal: true
+
+class ConsistencyChecklist < ApplicationRecord
+  belongs_to :planning_application
+
+  with_options if: :complete? do
+    validate :description_matches_documents_determined
+    validate :proposal_details_match_documents_determined
+    validate :documents_consistent_determined
+    validate :description_change_requests_closed
+    validate :additional_document_requests_closed
+  end
+
+  enum status: { in_assessment: 0, complete: 1 }, _default: :in_assessment
+
+  enum(
+    description_matches_documents: {
+      to_be_determined: 0,
+      yes: 1,
+      no: 2
+    },
+    _default: :to_be_determined,
+    _prefix: :description_matches_documents
+  )
+
+  enum(
+    documents_consistent: {
+      to_be_determined: 0,
+      yes: 1,
+      no: 2
+    },
+    _default: :to_be_determined,
+    _prefix: :documents_consistent
+  )
+
+  enum(
+    proposal_details_match_documents: {
+      to_be_determined: 0,
+      yes: 1,
+      no: 2
+    },
+    _default: :to_be_determined,
+    _prefix: :proposal_details_match_documents
+  )
+
+  private
+
+  def description_matches_documents_determined
+    return unless description_matches_documents_to_be_determined?
+
+    errors.add(:description_matches_documents, :not_determined)
+  end
+
+  def proposal_details_match_documents_determined
+    return unless proposal_details_match_documents_to_be_determined?
+
+    errors.add(:proposal_details_match_documents, :not_determined)
+  end
+
+  def documents_consistent_determined
+    return unless documents_consistent_to_be_determined?
+
+    errors.add(:documents_consistent, :not_determined)
+  end
+
+  def description_change_requests_closed
+    return unless open_description_change_requests?
+
+    errors.add(
+      :description_matches_documents,
+      :open_description_change_requests
+    )
+  end
+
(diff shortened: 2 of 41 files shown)


app/models/consistency_checklist.rb
```diff
@@ -0,0 +1,88 @@
...
+  with_options if: :complete? do
+    validate :description_matches_documents_determined
+    validate :proposal_details_match_documents_determined
+    validate :documents_consistent_determined
+    validate :description_change_requests_closed
+    validate :additional_document_requests_closed
+  end
+
+  enum status: { in_assessment: 0, complete: 1 }, _default: :in_assessment
+
```
allbecauseyoutoldmeso, line 15: I've made `description_matches_documents` , `documents_consistent` and `proposal_details_match_documents` fields enums rather than booleans because they need three states - not answered, answered 'Yes', and answered 'No'. 

Elsewhere we've used booleans for similar scenarios and relied on states of `nil`, `true`, and `false`, but it seems to expose us to ambiguity and bugs because ruby often treats `nil` and `false` as equivalent. Eg if I did that here `consistency_checklist.description_matches_documents?` would return `false` whether the column returned `false` whether then column contained `nil` or `false`.

I'd be interested to know if this is an approach people like or not!

allbecauseyoutoldmeso, line 15: I've made the `description_matches_documents` , `documents_consistent` and `proposal_details_match_documents` fields enums rather than booleans because they need three states - not answered, answered 'Yes', and answered 'No'. 

Elsewhere we've used booleans for similar scenarios and relied on states of `nil`, `true`, and `false`, but it seems to expose us to ambiguity and bugs because ruby often treats `nil` and `false` as equivalent. Eg if I did that here `consistency_checklist.description_matches_documents?` would return `false` when the column contained either `nil` or `false`.

I'd be interested to know if this is an approach people like or not!

app/controllers/concerns/commit_matchable.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module CommitMatchable
+  extend ActiveSupport::Concern
+
+  def commit_matches?(regex)
+    params[:commit]&.downcase&.match(regex).present?
```
allbecauseyoutoldmeso, line 7: I'm still not sure I like doing this...but I still can't find another way to change what happens on form submission based on which submit button the user presses.

It means we're just relying on the button text, which could be changed for multiple reasons (and we'd assign it 1 story point in planning 😆). Also I know it's unlikely for this project, but it would break if we started supporting multiple languages.

app/models/consistency_checklist.rb
allbecauseyoutoldmeso, line 15: I've made the `description_matches_documents` , `documents_consistent` and `proposal_details_match_documents` fields enums rather than booleans because they need three states - not answered, answered 'Yes', and answered 'No'. 

Elsewhere we've used booleans for similar scenarios and relied on states of `nil`, `true`, and `false`, but it seems to expose us to ambiguity and bugs because ruby often treats `nil` and `false` as equivalent. Eg if I did that here `consistency_checklist.description_matches_documents?` would return `false` when the column contained either `nil` or `false`.

I'd be interested to know if people like this approach or not..?

app/controllers/concerns/commit_matchable.rb
notapatch: Not sure I fully understand the problem. I think you want buttons having different results depending on which button you press??

I assume you've tried things like changing the [name](https://stackoverflow.com/a/42377507/1511504) - [this is similar](https://stackoverflow.com/a/3086833/1511504). 
Saw something where they had [different formactions](https://stackoverflow.com/a/32536941/1511504) 

Any good?

notapatch (approved):
I've read through the code but I haven't run the code. I clicked around half heartedly but not sure what this is. I've looked at the [trello card](https://trello.com/c/s9bqs7dY/1154-descriptions-need-to-match-documents-drawings-proposals-from-ripa) but this didn't help me know what was going on.

app/controllers/concerns/commit_matchable.rb
allbecauseyoutoldmeso: Ugh, I'm torn. `params[:commit]` not ideal because of tightly coupling actions and views. `formaction` potentially not ideal because it pushes us towards having eg two different `create` actions for the same model (the 'save' and the 'save and mark as complete' actions). I'm going to merge this as is for now, and then agonise some more about it 🙈 😰

PR #750, opened 2022-09-16
Hand rolled parallel tests

notapatch:
Splitting tests across CI nodes

    - RSpec doesn't offer parallel tests natively there are a number
      of options to achieve parallel tests. This PR is for making
      the Github Actions CI system faster by splitting the tests
      across multiple processes.

      Parallel Tests

      Parallel testing can be done through threading using parallel
      test gem (grosser/parallel_tests). Or through multiple
      proceses. This PR only involves multiple processes.

      Test splitting

      parallel_tests and Knapsack both split tests based on keeping
      a list of the times tests took on previous runs. Instead, This
      PR simply spreads the tests based on a regex of the starting
      letter(s) of the file the test is in.

      In the example below the models specs in files beginning a to p
      are tested with one CI runner and the remaining letters are
      placed in another CI runner.

```
specs-models-1:
  uses: ./.github/workflows/testing.yml
  with:
    name: "1"
    include: "spec/models/**/[a-p]*_spec.rb"
    secrets: inherit

specs-models-2:
  uses: ./.github/workflows/testing.yml
  with:
    name: "2"
    include: "spec/models/**/[q-z]*_spec.rb"
    secrets: inherit
```

     - there are a large number of runners available check
       Gihtub Actions usage-limits. If you are paying for CI then
       at the moment they allow 40 concurrent jobs. In the above
       example we are using 2. We could use 3 and each CI process
       have fewer letters. However, there's diminishing returns
       as a lot of the time is spent setting up the instance
       ready to run the tests.

    - tests that don't need Node and Javscript/yarn could have
      testing.yml optomized better by skipping out their installation

    Removing push

    - push is in the event of you forgetting to rebase your PR
    - not sure how common this is but I'm guessing you'd have merge
      issues as well??
    - what it does do is take the number of CI instances over 20
      which is going to slow every PR build

    Parrellel tests with turbo_tests

    I did try a 'plug and play' gem that uses parallel_test called
    turbo_tests (https://github.com/serpapi/turbo_tests) that has come
    out of the work of discourse - note: two gems called turbo_tests!

    - I don't know what I was doing! However, there were a few database
      lockups each time I ran it and it ran in 70% of the time of the
      full test suite. I didn't look further as I was more interested
      in improving the CI.


.github/workflows/build.yml
```diff
@@ -1,399 +1,104 @@
 ---
 name: CI
 
-on: [push, pull_request]
```
EGiataganas, line 4: I wouldn't delete this as we would like to know before making a PR whether our code is passing CI or not

notapatch: "before making a PR"


You're going to have to educate me.  You mean that "push" tells you if your code in main would pass?? pull_request if your PR passes?

EGiataganas: That's not a problem at all! Yes, the `push` key means that this workflow will run on the state of `push` & `pull_request` of the branch.

EGiataganas (approved):

benbaumann95 (approved):

PR #757, opened 2022-09-29
Planning history

EGiataganas:
### Description of change

- Integrate PlanningApplication API
- Match the right heading
- Planning History screen

### Story Link

https://trello.com/c/crdTSJN5


notapatch (approved):
With the understanding that I've never written an API  😄

PR #758, opened 2022-09-29
Add workaround for Active Storage misidentifying pdf file blob content type

benbaumann95:
### Description of change

- There is a bug with the signature for WMV2
- Magic priority for WMV is 60 but PDF is only 50 https://github.com/rails/marcel/blob/8e285636063d3617df6f73bc73de6574d83a53d5/data/tika.xml#L673-L693
- This workaround will identify a PDF and check to see if the bug is still present before removing the rule
https://github.com/rails/marcel/issues/77

### DIFF
--- a/config/initializers/marcel.rb
+++ b/config/initializers/marcel.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+# config/initializers/marcel.rb
+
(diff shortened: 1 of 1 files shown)


notapatch (approved):
Have you noticed that the text for the heading ends up on the next line... "pe as pdf".

I think that means you need to shorten the heading line... snappier.

PR #761, opened 2022-09-30
Add feature flag for planning history

EGiataganas:
### Description of change

Add feature flag for planning history

### Story Link

https://trello.com/c/m9CeNhsl

### DIFF
--- a/app/controllers/planning_histories_controller.rb
+++ b/app/controllers/planning_histories_controller.rb
@@ -2,6 +2,7 @@
 
 class PlanningHistoriesController < AuthenticationController
   before_action :set_planning_application
+  before_action :ensure_planning_history_is_enabled
 
   def show
     @planning_history = Apis::Paapi::Query.new.fetch(@planning_application.uprn)
@@ -9,5 +10,13 @@ class PlanningHistoriesController < AuthenticationController
     respond_to do |format|
       format.html
     end
+  end
+
+  private
+
+  def ensure_planning_history_is_enabled
+    return if @planning_application.planning_history_enabled?
+
+    render plain: "forbidden", status: :forbidden
   end
 end
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -440,6 +440,10 @@ class PlanningApplication < ApplicationRecord
     policy_classes.pluck(:section).include?(section)
   end
 
+  def planning_history_enabled?
+    ENV.fetch("PLANNING_HISTORY_ENABLED", "false") == "true"
+  end
+
   private
 
   def set_reference
--- a/app/views/planning_application/assessment_tasks/_check_consistency.html.erb
+++ b/app/views/planning_application/assessment_tasks/_check_consistency.html.erb
@@ -8,10 +8,12 @@
         Description, documents and proposal details
       </span>
     </li>
-    <li class="app-task-list__item">
-      <span class="app-task-list__task-name">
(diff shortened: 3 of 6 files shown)


app/views/planning_application/assessment_tasks/_check_consistency.html.erb
```diff
@@ -8,10 +8,12 @@
         Description, documents and proposal details
       </span>
     </li>
-    <li class="app-task-list__item">
-      <span class="app-task-list__task-name">
-        <%= link_to "History", planning_application_planning_history_path(@planning_application), class: "govuk-link" %>
-      </span>
-    </li>
+    <% if @planning_application.planning_history_enabled? %>
```
notapatch, line 11: You've made a change in the logic.

Have you added a test for this?

app/models/planning_application.rb
```diff
@@ -440,6 +440,10 @@ def has_policy_class?(section)
     policy_classes.pluck(:section).include?(section)
   end
 
+  def planning_history_enabled?
```
allbecauseyoutoldmeso, line 443: I'm not sure it makes sense to define this in the planning application model? If its dependent on an env var it won't be planning application specific.

spec/system/planning_applications/assessment_tasks_index_spec.rb
```diff
@@ -8,6 +8,8 @@
 
   before do
     sign_in assessor
+
+    allow(ENV).to receive(:fetch).with("PLANNING_HISTORY_ENABLED", "false").and_return("true")
```
allbecauseyoutoldmeso, line 12: In these feature specs it might be worth also testing for the case where the env var is blank or "false".

app/views/planning_application/assessment_tasks/_check_consistency.html.erb
EGiataganas: I didn't add a spec as there was already a spec, what I did was passing true on the env [cc][1]

[1]: https://github.com/unboxed/bops/pull/761/files#diff-364c0ea3353cade165138baacbd0c41e0e4cea4ab8990073313e6eff7762ed2b

allbecauseyoutoldmeso (approved):
Happy with this as it is, although I'm wondering if there's somewhere else to put the `#planning_history_enabled` other than the planning application model. Maybe there's a case for a PORO that handles feature flags?

app/models/planning_application.rb
EGiataganas: @allbecauseyoutoldmeso yeah I was thinking of moving it somewhere else, but as as I couldn't find a good name and this is the first case of a feature flag, I thought of adding it on the planning_application model 😂 

Happy to change it on a separate PR or when we will have more feature flags 😄

app/controllers/planning_histories_controller.rb
```diff
@@ -10,4 +11,12 @@ def show
       format.html
     end
   end
+
+  private
+
+  def ensure_planning_history_is_enabled
+    return if @planning_application.planning_history_enabled?
+
+    render plain: "forbidden", status: :forbidden
```
benbaumann95, line 20: Is a 404 perhaps a more appropriate error response here?
As technically this page doesn't exist in production rather than it being forbidden

EGiataganas: I don't think so, cause the page is there, but is forbidden as we are disabling them. By having it 403, we can start doing AB testing!

For example once we have all the data for a specific local_authority we can enable this feature to that local_authority but not to others and so on

PR #763, opened 2022-10-03
Add assessment details to planning applications

benbaumann95:
### Description of change

- Rework "summary of work" into this new model as we add more similar assessment detail functionality. These sections can be split by enum rather than creating a separate table for all which would result in more code to manage and less optimal querying performance
- Using the new "assessment detail" model, introduce the ability for officers to add additional evidence as part of the assessment process

### Story Link

https://trello.com/c/3Lo9mT1p/1131-entering-additional-information-for-a-record-of-impacts-additional-evidence

Screenshot 2022-10-03 at 16 09 17
Screenshot 2022-10-03 at 16 10 10


app/controllers/planning_application/assessment_details_controller.rb
```diff
@@ -0,0 +1,104 @@
...
+    def assessment_detail_id
+      Integer(params[:id])
+    end
+
+    def assessment_details_params
+      params.require(:assessment_detail).permit(:entry, :category)
+    end
+
+    def save_progress?
+      params[:commit] == I18n.t("form_actions.save_and_come_back_later")
```
allbecauseyoutoldmeso, line 91: Oh, smart thinking! I've been worrying about using `params[:commit]` in this way, but I think using translations makes it more robust - that way we can change the copy/language without so much breaking 😌

app/models/assessment_detail.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class AssessmentDetail < ApplicationRecord
+  belongs_to :planning_application
+  belongs_to :user
+
+  enum status: {
```
allbecauseyoutoldmeso, line 7: Is it worth having a default status?

app/presenters/concerns/assessment_tasks/assessment_detail_presenter.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+module AssessmentTasks
+  extend ActiveSupport::Concern
+
+  class AssessmentDetailPresenter < PlanningApplicationPresenter
```
allbecauseyoutoldmeso, line 6: Does this need to inherit everything from `PlanningApplicationPresenter`? Instead you could pull anything you need in that class into a concern and include it here? Otherwise this class has access to all the methods in `StatusPresenter`, `ProposalDetailsPresenter`, `ValidationTasksPresenter` and `AssessmentTasksPresenter`, since they're all included in `PlanningApplicationPresenter`...which seems a bit mad!

allbecauseyoutoldmeso: That said, maybe this just means we're due to a bit of a presenter tidy up generally, which could wait for another day 😁

app/views/planning_application/assessment_details/_form.html.erb
```diff
@@ -0,0 +1,16 @@
...
+
+  <%= form.govuk_text_area :entry, label: nil, rows: 10 %>
+
+  <%= form.hidden_field :category, value: category %>
+
+  <div class="govuk-button-group">
+    <%= form.submit "Save and mark as complete", class: "govuk-button", data: { module: "govuk-button" } %>
+    <%= form.submit "Save and come back later", class: "govuk-button govuk-button--secondary", data: { module: "govuk-button" } %>
+
+    <%= link_to "Back", :back, class: "govuk-button govuk-button--secondary" %>
```
allbecauseyoutoldmeso, line 14: Now you can use the [`back_link`](https://github.com/unboxed/bops/blob/main/app/helpers/application_helper.rb#L6) helper 😁

app/views/planning_application/assessment_details/edit.html.erb
```diff
@@ -0,0 +1,21 @@
+<% content_for :page_title do %>
+  <%= t("assessment_details.edit_#{@assessment_detail.category}") %> - <%= t('page_title') %>
+<% end %>
+
+<% add_parent_breadcrumb_link "Home", planning_applications_path %>
+<% add_parent_breadcrumb_link "Application", planning_application_path(@planning_application) %>
+<% add_parent_breadcrumb_link "Assess application", planning_application_assessment_tasks_path(@planning_application) %>
```
allbecauseyoutoldmeso, line 7: There's a [shared partial for this](https://github.com/unboxed/bops/blob/main/app/views/shared/_assessment_task_breadcrumbs.html.erb) I made earlier!

app/views/planning_application/assessment_details/new.html.erb
```diff
@@ -0,0 +1,21 @@
+<% content_for :page_title do %>
+  <%= t("assessment_details.new_#{@category}") %> - <%= t('page_title') %>
+<% end %>
+
+<% add_parent_breadcrumb_link "Home", planning_applications_path %>
+<% add_parent_breadcrumb_link "Application", planning_application_path(@planning_application) %>
+<% add_parent_breadcrumb_link "Assess application", planning_application_assessment_tasks_path(@planning_application) %>
```
allbecauseyoutoldmeso, line 7: and again

app/views/planning_application/assessment_details/show.html.erb
```diff
@@ -0,0 +1,32 @@
+<% content_for :page_title do %>
+  <%= t("assessment_details.show_#{@assessment_detail.category}") %> - <%= t('page_title') %>
+<% end %>
+
+<% add_parent_breadcrumb_link "Home", planning_applications_path %>
+<% add_parent_breadcrumb_link "Application", planning_application_path(@planning_application) %>
+<% add_parent_breadcrumb_link "Assess application", planning_application_assessment_tasks_path(@planning_application) %>
```
allbecauseyoutoldmeso, line 7: x3

app/views/planning_application/assessment_details/show.html.erb
```diff
@@ -0,0 +1,32 @@
...
+    <%= render "shared/planning_application_address_and_reference" %>
+
+    <div class="govuk-body">
+      <%= render "planning_application/assessment_details/#{@assessment_detail.category}/information", category: @assessment_detail.category %>
+
+      <p><%= @assessment_detail.entry %></p>
+    </div>
+
+    <div class="govuk-button-group">
+      <%= link_to "Back", :back, class: "govuk-button govuk-button--secondary" %>
```
allbecauseyoutoldmeso, line 26: `#back_link`

allbecauseyoutoldmeso (approved):
This looks great!

Really like having the flexible `assessment_details` table...seems like a really good approach.

app/presenters/concerns/assessment_tasks/assessment_detail_presenter.rb
benbaumann95: Good point, we certainly don't want that! I'll see what I can do

benbaumann95: On second thoughts i'm gonna make a card for this as it might bloat this PR

allbecauseyoutoldmeso: Yes, I think that's a good idea! More of a nice to have refactor than a requirement for this piece of work ☺️

app/models/assessment_detail.rb
benbaumann95: Hmm not sure, I don't think so
What would be the use case for this ?

allbecauseyoutoldmeso: I guess I was assuming that it's 'in progress' from the moment it exists...but then perhaps it's better to set that explicitly... 🤔

benbaumann95: It could be `complete` too as the record persists if the officer chooses to select "save and mark as complete"

PR #765, opened 2022-10-04
Extract shared submit buttons

allbecauseyoutoldmeso:
### Description of change

In preparation for [this story](https://trello.com/c/AOywBVuR/1121-update-button-groups-on-all-assessment-and-review-pages-to-be-consistent-with-pattern) extract shared logic for commit submit button pattern.

### Story Link

https://trello.com/c/AOywBVuR/1121-update-button-groups-on-all-assessment-and-review-pages-to-be-consistent-with-pattern

### DIFF
--- a/app/views/shared/_submit_buttons.html.erb
+++ b/app/views/shared/_submit_buttons.html.erb
@@ -1,0 +1,15 @@
+<div class="govuk-button-group">
+  <%= form.submit(
+    t("form_actions.save_and_mark_as_complete"),
+    class: "govuk-button",
+    data: { module: "govuk-button" },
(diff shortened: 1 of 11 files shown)


app/views/shared/_submit_buttons.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="govuk-button-group">
+  <%= form.submit(
```
allbecauseyoutoldmeso, line 2: For some reason, although I can't see anything about the gov uk form builder docs, `form.govuk_submit` renders a button where the text is a before pseudo-element, and therefore `params[:commit]` is nil 🤯

EGiataganas (approved):

PR #766, opened 2022-10-05
Add status to recommendations

allbecauseyoutoldmeso:
### Description of change

- Add `status` column to `recommendations`.
- Use recommendation status to display correct tags in task list for 'Assess the recommendation' and 'Review the recommendation'.
- Add `Save and mark as complete` and 'Save and come back later' buttons to 'Review the recommendation' form, which set status to `review_in_progress` and `review_complete`.

### Story Link

https://trello.com/c/gXXs6W5z/1249-update-button-groups-on-review-the-recommendation-page

### Screenshot

Screenshot 2022-10-06 at 10 09 58

### DIFF
--- a/app/controllers/recommendations_controller.rb
+++ b/app/controllers/recommendations_controller.rb
@@ -43,7 +43,12 @@ class RecommendationsController < AuthenticationController
   def update
     respond_to do |format|
       @recommendation.assign_attributes(recommendation_params)
-      @recommendation.review!
+
+      if @recommendation.review_complete?
+        @recommendation.review!
+      else
+        @recommendation.save!
+      end
 
       format.html do
         redirect_to @planning_application, notice: "Recommendation was successfully reviewed."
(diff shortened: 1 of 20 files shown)


app/models/recommendation.rb
```diff
@@ -13,10 +13,20 @@ class ReviewRecommendationError < StandardError; end
...
 
+  enum(
+    status: {
+      assessment_in_progress: 0,
+      assessment_complete: 1,
+      review_in_progress: 2,
+      review_complete: 3
+    },
+    _default: :assessment_in_progress
+  )
```
EGiataganas, line 28: How about moving the default in the migration?

allbecauseyoutoldmeso: Yes, good thought!

I wonder if there's any benefit to having both 🤔

EGiataganas: I think is better usually to have the default value on the creation on the database level and here you are extending the list of possible status 😃

allbecauseyoutoldmeso: Yes, I think you're right. Amended!

EGiataganas (approved):

app/controllers/recommendations_controller.rb
```diff
@@ -43,7 +43,12 @@ def edit
   def update
     respond_to do |format|
       @recommendation.assign_attributes(recommendation_params)
-      @recommendation.review!
+
+      if @recommendation.review_complete?
+        @recommendation.review!
+      else
+        @recommendation.save
```
benbaumann95, line 50: What if the recommendation does not save?

It will still execute the below success statement. I think we need to use ! or have some logic to say `if @recommendation.save` etc

benbaumann95 (approved):
Great stuff 🎉 
Loving all the specs too

spec/system/planning_applications/recommending_spec.rb
```diff
@@ -538,4 +538,111 @@
...
+    find("li", text: "Assess recommendation")
+  end
+
+  def review_list_item
+    find("li", text: "Review assessment")
+  end
+
+  def view_list_item
+    find("li", text: "View recommendation")
+  end
```
benbaumann95, line 647: Ah yeah guess this was a good use case of the previous suggestion to reuse this

app/controllers/recommendations_controller.rb
allbecauseyoutoldmeso: Oops, you are right. Good shout!

PR #767, opened 2022-10-06
Make status tags more consistent

allbecauseyoutoldmeso:
### Description of change

- Replace 'Completed' with 'Complete' for all tags.
- Update consistency checklist tag so that initial status is 'Not started', not 'In assessment'.

### Story Link

NA

### DIFF
--- a/app/presenters/concerns/assessment_tasks/assessment_detail_presenter.rb
+++ b/app/presenters/concerns/assessment_tasks/assessment_detail_presenter.rb
@@ -8,7 +8,7 @@ module AssessmentTasks
       super(template, planning_application)
 
       @assessment_detail = planning_application.assessment_details.send(category).first
-      @status = assessment_details_status
+      @status = assessment_detail&.status || "not_started"
       @category = category
     end
 
@@ -46,14 +46,8 @@ module AssessmentTasks
       classes = ["#{govuk_tag_class} app-task-list__task-tag"]
 
       tag.strong class: classes do
-        status
+        I18n.t("assessment_details.#{status}")
       end
-    end
-
-    def assessment_details_status
-      return "Not started" unless assessment_detail
-
-      assessment_detail.status.humanize
     end
 
     def govuk_tag_class
--- a/spec/system/planning_applications/checking_consistency_spec.rb
+++ b/spec/system/planning_applications/checking_consistency_spec.rb
@@ -28,6 +28,7 @@ RSpec.describe "checking consistency", type: :system do
   end
 
   it "lets user save draft or mark as complete" do
+    expect(task_list_item).to have_content("Not started")
     click_link("Description, documents and proposal details")
     click_button("Save and mark as complete")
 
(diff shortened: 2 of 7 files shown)


app/presenters/concerns/assessment_tasks/assessment_detail_presenter.rb
```diff
@@ -29,31 +29,31 @@ def assessment_details_link
...
+        assessment_details_status
       end
     end
 
     def assessment_details_status
-      return "Not started" unless assessment_detail
+      return I18n.t("assessment_details.not_started") unless assessment_detail
 
-      assessment_detail.status.humanize
+      I18n.t("assessment_details.#{status}")
```
allbecauseyoutoldmeso, line 56: Alternatively we could change the status enum in the `AssessmentDetail` model to be "complete" instead of "completed". But I don't wand to get into a habit of making logic or data changes every time there's a change in the front end copy! Using translations seems simpler, because if tomorrow we decide that the tag should read "Done", "Finished" or "Banana" we only have to change it in one place!

app/presenters/concerns/assessment_tasks/assessment_detail_presenter.rb
```diff
@@ -46,16 +46,10 @@ def assessment_details_status_tag
       classes = ["#{govuk_tag_class} app-task-list__task-tag"]
 
       tag.strong class: classes do
-        status
+        I18n.t("assessment_details.#{status}")
```
allbecauseyoutoldmeso, line 49: Alternatively we could change the status enum in the `AssessmentDetail` model to be "complete" instead of "completed". But I don't wand to get into a habit of making logic or data changes every time there's a change in the front end copy. Using translations seems simpler, because if tomorrow we decide that the tag should read "Done", "Finished" or "Banana" we only have to change it in one place!

benbaumann95: Yep agreed, nice approach 👍

spec/system/planning_applications/checking_consistency_spec.rb
```diff
@@ -295,7 +288,8 @@ def form_group_with_legend(legend)
     find("legend", text: legend).find(:xpath, "..")
   end
 
-  def task_list_item_with_name(name)
-    find("span", text: name).find(:xpath, "..")
+  def task_list_item
+    text = "Description, documents and proposal details"
+    find("span", text: text).find(:xpath, "..")
```
benbaumann95, line 293: Not sure if this is worth extracting out into the system spec helpers ?

And again making it dynamic if perhaps we need to use it for another spec. But not fussed if you think it's not necessary

benbaumann95 (approved):
Loving this refactor!

app/presenters/concerns/assessment_tasks/assessment_detail_presenter.rb
EGiataganas: Agreed! Good choice 😃

EGiataganas (approved):

PR #768, opened 2022-10-06
Add status tag to 'Check and assess' task

allbecauseyoutoldmeso:
### Description of change

Add a status tag to the 'Check and assess' task.

### Story Link

https://trello.com/c/PaYsVcQH/1130-create-page-structure-for-assessment-of-pd-stories

### Screenshot

Screenshot 2022-10-06 at 15 57 52

### DIFF
--- a/spec/support/system_spec_helpers.rb
+++ b/spec/support/system_spec_helpers.rb
@@ -8,4 +8,8 @@ module SystemSpecHelpers
   def selected_govuk_tab
     find("div[class='govuk-tabs__panel']")
   end
+
+  def list_item(text)
+    find("li", text: text)
+  end
 end
--- a/spec/system/planning_applications/summary_of_works_spec.rb
+++ b/spec/system/planning_applications/summary_of_works_spec.rb
@@ -16,11 +16,9 @@ RSpec.describe "Summary of works", type: :system do
   end
 
   context "when planning application is in assessment" do
-    before do
-      click_link "Check and assess"
-    end
-
     it "I can view the information on the summary of works page" do
+      click_link "Check and assess"
+
       within("#assessment-information-tasks") do
         expect(page).to have_content("Not started")
         click_link "Summary of works"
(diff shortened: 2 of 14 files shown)


spec/system/planning_applications/summary_of_works_spec.rb
```diff
@@ -84,9 +85,14 @@
       within("#assessment-information-tasks") do
         expect(page).to have_content("In progress")
       end
```
benbaumann95, line 87: I guess we may not need this specs if you are doing `expect(check_and_assess_list_item).to have_content("In progress")` anyway?

allbecauseyoutoldmeso: Surely you can't check too many tags 🤣

benbaumann95: There's a few examples of these scattered around. I guess this one is checking extra that it's within the id`"#assessment-information-tasks"`

spec/support/system_spec_helpers.rb
```diff
@@ -8,4 +8,8 @@ def row_with_content(content, element = page)
   def selected_govuk_tab
     find("div[class='govuk-tabs__panel']")
   end
+
+  def check_and_assess_list_item
```
benbaumann95, line 12: Is it worth making this dynamic / reusable ?

i.e. passing in the text as an arg so we can apply this to other list items

benbaumann95:
```ruby
def list_item(text)
  find("li", text: text)
end
```

benbaumann95 (approved):

spec/system/planning_applications/summary_of_works_spec.rb
benbaumann95: Ahh my bad ok lol

spec/support/system_spec_helpers.rb
allbecauseyoutoldmeso: Oh, excellent thought, yes! Especially because I've previously used the same thing elsewhere. DRY 🎉

EGiataganas (approved):

PR #770, opened 2022-10-07
Re-render edit form with errors if recommendation fails to update

allbecauseyoutoldmeso:
### Description of change

- When saving recommendation with status of 'review_in_progress', render edit form if recommendation fails to save.

### Story Link

NA

### DIFF
--- a/spec/controllers/recommendations_controller_spec.rb
+++ b/spec/controllers/recommendations_controller_spec.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe RecommendationsController, type: :controller do
+  describe "#update" do
+    let(:local_authority) { create(:local_authority, :default) }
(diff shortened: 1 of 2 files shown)


spec/controllers/recommendations_controller_spec.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe RecommendationsController, type: :controller do
```
allbecauseyoutoldmeso, line 5: We need to test that if a user presses 'Save and come back later' on the review form and the recommendation doesn't save the form is re-rendered with errors.

However currently there are no real errors that can be triggered in this scenario!

So adding a couple of controller specs, including one where I stub the recommendation to fail to save.

Hope this makes sense!

benbaumann95 (approved):

PR #774, opened 2022-10-10
Allow officers to add previous application references

allbecauseyoutoldmeso:
### Description of change

- Add `additional_information` to `assessment_details`.
- Add `past_application_references` category to `assessment_details`.
-  Use existing `assessment_details` structure to add editing and updating 'History'.

### Story Link

https://trello.com/c/bk4LXxFY/1183-add-field-application-reference-numbers-to-the-history-page-in-assessment-tasklist

### Screenshots

Screenshot 2022-10-11 at 12 18 58

Validation error:

Screenshot 2022-10-11 at 12 19 31

### Decisions [OPTIONAL]

Adding an `additional_information` field to `assessment_details` instead of creating a new table. But I think this should be the last field we add! We want to keep this table as generic as possible.

### DIFF
--- a/app/views/planning_application/assessment_tasks/_check_consistency.html.erb
+++ b/app/views/planning_application/assessment_tasks/_check_consistency.html.erb
@@ -17,12 +17,20 @@
         )
       ) %>
     </li>
-    <% if @planning_application.planning_history_enabled? %>
-      <li class="app-task-list__item">
+    <li class="app-task-list__item">
+      <% if @planning_application.planning_history_enabled? %>
         <span class="app-task-list__task-name">
-          <%= link_to "History", planning_application_planning_history_path(@planning_application), class: "govuk-link" %>
+          <%= link_to(
+            t(".history"),
+            planning_application_planning_history_path(@planning_application),
+            class: "govuk-link"
+          ) %>
         </span>
-      </li>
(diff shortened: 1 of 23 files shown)


app/views/planning_application/assessment_tasks/_check_consistency.html.erb
```diff
@@ -17,12 +17,20 @@
...
+            class: "govuk-link"
+          ) %>
         </span>
-      </li>
-    <% end %>
+      <% else %>
+        <%= @planning_application.assessment_detail_tasklist(
+          :past_applications
+        ) %>
+      <% end %>
```
benbaumann95, line 33: Does this mean we won't be showing this on production ?

I had the understanding that this would be a feature on top of what the API pulls in

benbaumann95 (approved):

app/views/planning_application/assessment_tasks/_check_consistency.html.erb
allbecauseyoutoldmeso: I was going on on [this conversion](https://ubxd.slack.com/archives/C010XKV8QLR/p1665417583721359), but seems worth double checking 🤔. I will talk to the design team again tomorrow!

PR #779, opened 2022-10-19
Add assessment report to submit recommendation page

allbecauseyoutoldmeso:
### Description of change

- Remove planning application accordion from submit recommendation page.
- Replace with single-section accordion which contains assessment report.
- Add route to download assessment report as PDF, with link on submit recommendation page.

### Story Link

https://trello.com/c/MnuYnFaS/1198-include-full-assessment-summary-on-submit-recommendation-page

### Screenshot

Screenshot 2022-10-19 at 17 01 15

### DIFF
--- a/app/components/assessment_report_component.html.erb
+++ b/app/components/assessment_report_component.html.erb
@@ -1,0 +1,57 @@
+<% if constraints.any? %>
+  <h3 class="govuk-heading-m"><%= t(".constraints_including_article") %></h3>
+  <ul class="govuk-list govuk-list--bullet">
+    <% constraints.each do |constraint| %>
+      <li><%= constraint.humanize %></li>
+    <% end %>
+  </ul>
+<% end %>
+<% if past_applications.present? %>
+  <h3 class="govuk-heading-m"><%= t(".planning_history") %></h3>
+  <h4 class="govuk-heading-s"><%= past_applications.entry %></h4>
+  <%= simple_format(
+    past_applications.additional_information,
+    class: "govuk-body"
+  ) %>
+<% end %>
+<% if summary_of_work.present? %>
+  <h3 class="govuk-heading-m"><%= t(".summary_of_works") %></h3>
+  <%= simple_format(summary_of_work.entry, class: "govuk-body") %>
+<% end %>
+<% if site_description.present? %>
+  <h3 class="govuk-heading-m"><%= t(".location_description") %></h3>
+  <%= simple_format(site_description.entry, class: "govuk-body") %>
+<% end %>
+<% if consultation_summary.present? %>
+  <h3 class="govuk-heading-m"><%= t(".consultation_documents") %></h3>
+  <ul class="govuk-list govuk-list--bullet">
+    <% consultees.each do |consultee| %>
+      <li><%= "#{consultee.name} (#{consultee.origin})" %></li>
+    <% end %>
+  </ul>
+  <%= simple_format(consultation_summary.entry, class: "govuk-body") %>
+<% end %>
+<% if policy_classes.any? %>
+  <h3 class="govuk-heading-m"><%= t(".legislation") %></h3>
+  <% policy_classes.each do |policy_class| %>
+    <%= render(PolicyClassComponent.new(policy_class: policy_class)) %>
+  <% end %>
+<% end %>
+<h3 class="govuk-heading-m"><%= t(".documents_included_in") %></h3>
+<% if documents.any? %>
+  <table class="govuk-table">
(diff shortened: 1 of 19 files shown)


app/javascript/stylesheets/application.scss
```diff
@@ -300,3 +300,9 @@ html * {
     margin-bottom: 0px;
   }
 }
+
+.single-section-accordion {
+  .govuk-accordion__controls {
+    display: none;
```
allbecauseyoutoldmeso, line 306: Hide the 'Open all' button at the top of the govuk accordion, since there's only one section to open.

app/javascript/stylesheets/application.scss
```diff
@@ -300,3 +300,9 @@ html * {
...
+  }
+}
```
allbecauseyoutoldmeso, line 308: Hide the 'Open all' button at the top of the govuk accordion, since there's only one section to open.

spec/models/document_spec.rb
```diff
@@ -246,4 +246,20 @@
...
   end
+
+  describe ".referenced_in_decision_notice" do
+    let(:document) do
+      create(:document, referenced_in_decision_notice: true, numbers: "REF1")
+    end
+
+    before { create(:document, referenced_in_decision_notice: false) }
+
+    it "returns documents where referenced_in_decision_notice is true" do
```
notapatch, line 257: Interesting. I test scopes explicitly which [David Chelimsky suggested (that's my posting, but I just copied him)](https://stackoverflow.com/a/60058513/1511504). However, you've implicitly tested both cases. Don't see why not 🤷‍♂️

app/models/planning_application.rb
```diff
@@ -453,6 +453,12 @@ def planning_history_enabled?
     ENV.fetch("PLANNING_HISTORY_ENABLED", "false") == "true"
   end
 
+  AssessmentDetail.categories.each_key do |category|
+    define_method(category) do
```
notapatch, line 457: Not used this before. You're dynamically creating a method for each category??

notapatch (approved):

spec/models/document_spec.rb
allbecauseyoutoldmeso: Hello, @notapatch, nice to have you back on the team!

I think I like the explicit approach...will amend!

app/models/planning_application.rb
allbecauseyoutoldmeso: We introduced the `assessment_details` table because a lot of new features ('Summary of works', 'Consultation summary', 'Previous applications' etc) all called for a table with `planning_application_id` and just one text field. 

So instead of planning applications having one `summary_of_work`, one `consultation_summary` etc, it has many `application_details` distinguished by category.

But it's still useful to be able to call `planning_application.summary_of_work`, instead of always calling `planning_application.application_details.find_by(category: 'summary_of_work')`.

Hope that makes sense!

spec/models/planning_application_spec.rb
```diff
@@ -1068,4 +1068,24 @@
...
+
+      let!(:assessment_detail) do
+        create(
+          :assessment_detail,
+          category,
+          planning_application: planning_application
+        )
+      end
+
+      it "returns the correct assessment detail" do
```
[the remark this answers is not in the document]
notapatch: It's more that I'm highlighting things I don't know about so that others, you or whoever, can be encouraged to look and correct me (or whatever).

I think my strategy would be to keep it is. If you have an error in that part of the code then you go with explicit tests along with an error case.

spec/models/document_spec.rb
notapatch: Thank you.

app/components/assessment_report_component.html.erb
```diff
@@ -0,0 +1,57 @@
+<% if constraints.any? %>
+  <h3 class="govuk-heading-m"><%= t(".constraints_including_article") %></h3>
+  <ul class="govuk-list govuk-list--bullet">
+    <% constraints.each do |constraint| %>
+      <li><%= constraint.humanize %></li>
+    <% end %>
+  </ul>
+<% end %>
```
benbaumann95, line 8: I've moved this into a partial for my [PR](https://github.com/unboxed/bops/pull/782/files#diff-746b024295017db10b384f027a60eac862b6fae97a2837a496b7840ff9ba4556R1)

We'll be reusing this and the "planning history" sections for a few views now so probs best to use a shared source. Maybe it can live under a different directory though

benbaumann95 (approved):

EGiataganas (approved):

PR #786, opened 2022-10-21
Github actions workflow to include components

notapatch:
Github actions build.yml works by running tests on specific
directories. If the directory is missing the test isn't run!
This change adds "components" directory to the test runs


allbecauseyoutoldmeso (approved):
🎉

PR #790, opened 2022-10-27
Pr reviewer is unable to edit constraints on the legislation page

notapatch:
## Description of change

Reviewer wants to understand the reasoning of the officer who has assessed the class related to constraints. But not change the constraints. 

Assessor want to relate the constraints to the legislation and change them.

## Story Link

https://trello.com/c/RrilPSF6/1296-reviewer-is-unable-to-edit-constraints-on-the-legislation-page

### Legislation page with assessor
- current user is an assessor 
Screenshot 2022-10-27 at 22 52 04

- current user is a reviewer
Screenshot 2022-10-27 at 22 51 25

## Commits
- PR shows two commits the first is just to add a assessor context block around original tests so that the diff for the actual work is clearer
- Commits to be squashed before merge

### DIFF
--- a/app/views/shared/_constraints.html.erb
+++ b/app/views/shared/_constraints.html.erb
@@ -10,4 +10,7 @@
   </ul>
 <% end %>
 
-<%= link_to "Edit constraints", edit_planning_application_constraints_path(@planning_application), class: "govuk-link govuk-body" %>
+<%= link_to_if(current_user.assessor?,
+               "Edit constraints",
+               edit_planning_application_constraints_path(@planning_application),
+               class: "govuk-link govuk-body") {} %>
--- a/spec/system/planning_applications/assessment_against_legislation_spec.rb
+++ b/spec/system/planning_applications/assessment_against_legislation_spec.rb
@@ -124,7 +124,7 @@ RSpec.describe "assessment against legislation", type: :system do
       end
     end
 
-    it "displays the constraints" do
+    it "displays the constraints with edit" do
       click_link("Check and assess")
       add_policy_classes(["Class D - porches"])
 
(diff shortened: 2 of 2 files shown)


benbaumann95 (approved):

EGiataganas (approved):

benbaumann95 (commented):
Looks fine if we just want to hide the link but not sure if this ticket requires validation on the model layer?

notapatch:
I'm confused by your question, I'll take a shot at it as I see it in my confused state....

So the model represents the database object - in this case the  constraints associated with the planning application object. These constraints won't change depending on the user who sees it... So I wouldn't expect to validate the model layer??

Does that answer the question??

benbaumann95:
> I'm confused by your question, I'll take a shot at it as I see it in my confused state....
> 
> So the model represents the database object - in this case the constraints associated with the planning application object. These constraints won't change depending on the user who sees it... So I wouldn't expect to validate the model layer??
> 
> Does that answer the question??



> I'm confused by your question, I'll take a shot at it as I see it in my confused state....
> 
> So the model represents the database object - in this case the constraints associated with the planning application object. These constraints won't change depending on the user who sees it... So I wouldn't expect to validate the model layer??
> 
> Does that answer the question??

So i'm referring to adding a callback on the model to check whether an update action on the constraints was made by a reviewer. If so, prevent this update on the database and raise an error.

Or perhaps on the controller we could render a forbidden page if a reviewer tries to directly access the edit route. Because technically I could just visit the url `edit_planning_application_constraints_path(@planning_application)` and still perform an edit action as a reviewer?

Or maybe this ticket is purely about just not showing an edit link then ignore me..

notapatch:
With conversation Ben has left it up to me if I put in the security into the application (a reviewer can obv get around this if they wanted to by accessing the API directly). Interesting as the "Reviewers" are really the managers who in most applications are the power users but in this application we are limiting them from changing the work of an assessor.

PR #792, opened 2022-10-31
Refactor steps/review

notapatch:
- can_review_asessment == awaiting_determination && Current.user.reviewer?
 - going to simplify by have them both use awaiting_determination

```ruby
  def submit_recommendation_complete?
    awaiting_determination? || determined?
  end
```

### DIFF
--- a/app/views/planning_applications/steps/_review.html.erb
+++ b/app/views/planning_applications/steps/_review.html.erb
@@ -3,7 +3,7 @@
   <ul class="app-task-list__items">
     <li class="app-task-list__item">
       <span class="app-task-list__task-name">
-        <% if @planning_application.can_review_assessment? && current_user.reviewer? %>
+        <% if @planning_application.awaiting_determination? && current_user.reviewer? %>
           <%= link_to "Review assessment", edit_planning_application_recommendations_path(@planning_application), aria: { describedby: "review_assessment-completed" }, class: "govuk-link" %>
         <% elsif @planning_application.awaiting_determination? && !current_user.reviewer? %>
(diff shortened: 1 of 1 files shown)


app/views/planning_applications/steps/_review.html.erb
```diff
@@ -3,7 +3,7 @@
   <ul class="app-task-list__items">
     <li class="app-task-list__item">
       <span class="app-task-list__task-name">
-        <% if @planning_application.can_review_assessment? && current_user.reviewer? %>
+        <% if @planning_application.awaiting_determination? && current_user.reviewer? %>
```
benbaumann95, line 6: Does it make sense to just drop the `current_user.reviewer?` part from this instead?

as we have

```
  def can_review_assessment?
    awaiting_determination? && Current.user.reviewer?
  end
  ```
  
  But yeah no need for both (`@planning_application.can_review_assessment? && current_user.reviewer?`) conditional checks

benbaumann95 (approved):

app/views/planning_applications/steps/_review.html.erb
notapatch: I could go either way. I think to me it wasn't obvious  that` can_review_asessment?` checked `current_user.reviewer?` and `awaiting_determination?`

By making them the same I think it makes it clearer what's going on and might allow further refactor in another PR.

notapatch: I am going to introduce a fourth branch. So I want to make it as easy as possible.

PR #794, opened 2022-11-02
Pr fix remove made from legislation url

notapatch:
`config/locals/reference.yml `has a list of URL that are used to link the application to legislation. 

When discussing it with the designers they explained it should point to the latest version of the legislation and not the original version. However, some of the URLs are pointing to the original versions of the legislation.

For example reference.yml points to the original '/made' version the URL will be changed by removing '/made'

   ` url: https://www.legislation.gov.uk/uksi/2015/596/schedule/2/part/1/crossheading/class-g-chimneys-flues-etc-on-a-dwellinghouse/made`

### DIFF
--- a/config/locales/references.yml
+++ b/config/locales/references.yml
@@ -430,7 +430,7 @@ en:
               schedule: "Schedule 1"
               part: 1
               section: B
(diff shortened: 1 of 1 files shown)


EGiataganas (approved):

notapatch:
Heads up I've updated the changes to include CLASS B which also is wrong.

PR #795, opened 2022-11-04
Fix legislation-url in database with migration

notapatch:
- Update previous policy_class records with the wrong urls
- Updated the order of rubocops to put Layout/LineLength in
      alphabetical order

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,6 +16,11 @@ AllCops:
     - 'bops-applicants/**/*'
     - 'lib/package_builder.rb'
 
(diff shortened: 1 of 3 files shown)


benbaumann95 (approved):

notapatch:
This PR never got put through because I never had shell access to the server so couldn't test if it worked or not. It means that new planning applications have different legislation URL to old planning applications; It's not *that* bad as the URLs are almost the same.

I'm leaving it open but happy to have it put through or closed as long as you're aware of the situation.

PR #798, opened 2022-11-07
Allow multiple updates of documents

benbaumann95:
### Description of change

Allow multiple updates of documents

- Whether an active document is to be show on the decision notice / made public can be edited on this new review documents for recommendation page. Multiple updates to several documents can occur here via checking / unchecking checkboxes
- Show a link to "Add document reference" if a document does not have an associated reference number
- Wrap the update action in a transaction so if one update fails none of the updates will persist
- Add status tag to indicate the progress of reviewing these documents

### Story Link

https://trello.com/c/6FZ2tU7d/1299-improve-the-adding-of-documents-to-the-decision-notice

Screenshot 2022-11-07 at 18 26 44
Screenshot 2022-11-07 at 18 27 10

### DIFF
--- a/app/controllers/planning_application/review_documents_controller.rb
+++ b/app/controllers/planning_application/review_documents_controller.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+class PlanningApplication
+  class ReviewDocumentsController < AuthenticationController
+    include CommitMatchable
+    include PlanningApplicationAssessable
+
+    before_action :set_planning_application
+    before_action :set_documents
+    before_action :ensure_planning_application_is_validated
+
+    def index
+      respond_to do |format|
+        format.html
+      end
+    end
+
+    def update
+      ActiveRecord::Base.transaction do
+        @planning_application.update!(review_documents_for_recommendation_status: status)
+
+        @documents.each do |document|
+          document.update!(
+            referenced_in_decision_notice: referenced_in_decision_notice?(document.id.to_s),
+            publishable: publishable?(document.id.to_s)
+          )
+        end
+      end
+
+      redirect_to planning_application_assessment_tasks_path(@planning_application),
+                  notice: "Documents were successfully updated."
+    rescue ActiveRecord::ActiveRecordError => e
+      redirect_to planning_application_review_documents_path(@planning_application),
+                  alert: "Couldn't update documents with error: #{e.message}. Please contact support."
+    end
+
+    private
+
+    def set_planning_application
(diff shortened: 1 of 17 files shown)


app/controllers/planning_application/review_documents_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+    def update_all
+      ActiveRecord::Base.transaction do
+        @planning_application.update!(review_documents_for_recommendation_status: status)
+
+        @documents.each do |document|
+          document.update!(
+            referenced_in_decision_notice: referenced_in_decision_notice?(document.id.to_s),
+            publishable: publishable?(document.id.to_s)
+          )
+        end
```
benbaumann95, line 27: Ideally would like one query here to do the multiple updates but can't use something like `update_all` as the value of the attributes (referenced_in_decision_notice / publishable) can be different for each document

app/presenters/concerns/assessment_tasks/review_documents_presenter.rb
```diff
@@ -0,0 +1,62 @@
...
+        planning_application_review_documents_path(planning_application), class: "govuk-link"
+      )
+    end
+
+    def review_documents_status_tag
+      classes = ["#{govuk_tag_class} app-task-list__task-tag"]
+
+      tag.strong class: classes do
+        I18n.t("status_tag_component.#{status}")
+      end
```
notapatch, line 39: I assume the above block creates a "strong" tag, inserts class of "classes" into it and nests the localised string inside? 🤔

notapatch (approved):

app/controllers/planning_application/review_documents_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+    before_action :set_documents
+    before_action :ensure_planning_application_is_validated
+
+    def index
+      respond_to do |format|
+        format.html
+      end
+    end
+
+    def update_all
```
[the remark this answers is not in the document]
benbaumann95: I guess `update` usually refers to a single entity whereas this is doing multiple updates.
However, I guess I could tweak the route to enable the update action to do this 👍

app/presenters/concerns/assessment_tasks/review_documents_presenter.rb
benbaumann95: Yeah so the html formed would be something like
```
<strong class="govuk-tag govuk-tag--blue app-task-list__task-tag">Complete</strong>
```

PR #804, opened 2022-11-10
FIX: Sign-off link converts to text for rejected application

notapatch:
FIX: Sign-off link converts to text for rejected application

  - Trello for: Create review tasklist for managers signing off cases
  - Accepted applications can be rejected if the reviewer changes
      their mind
  - Rejected applications have to wait for the assessor to fix the
      issue
  - To show this visually we change the link to text on rejection

### Story Link

https://trello.com/c/cuDX5hsQ/1263-create-review-tasklist-for-managers-signing-off-cases

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -71,6 +71,10 @@ class ApplicationController < ActionController::Base
     redirect_to administrator_dashboard_path if current_user&.administrator?
   end
 
+  def ensure_user_is_reviewer
+    render plain: "forbidden", status: :forbidden and return unless Current.user.reviewer?
+  end
+
   def ensure_user_is_reviewer_checking_assessment
     render plain: "forbidden", status: :forbidden and return unless @planning_application.can_review_assessment?
   end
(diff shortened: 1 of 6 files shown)


benbaumann95 (commented):
Is this a page we want to prevent them from accessing entirely ?
I can still visit the url directly

app/controllers/application_controller.rb
```diff
@@ -71,7 +71,7 @@ def enforce_user_permissions
     redirect_to administrator_dashboard_path if current_user&.administrator?
   end
 
-  def ensure_user_is_reviewer
```
notapatch, line 74: Names should tell you what they do. In this case the name says one thing but the code says another. I need the name "ensure_user_is_reviewer" so I'm changing this name. I'm *very* open for a better name change!

app/controllers/application_controller.rb
```diff
@@ -71,7 +71,7 @@ def enforce_user_permissions
...
+  def ensure_user_is_reviewer_checking_assessment
     render plain: "forbidden", status: :forbidden and return unless @planning_application.can_review_assessment?
```
notapatch, line 79: can_review_assessment? is actually...

```
def can_review_assessment?
  awaiting_determination? && Current.user.reviewer?
end
```

benbaumann95 (approved):

EGiataganas (approved):

PR #812, opened 2022-11-17
Make comments polymorphic

allbecauseyoutoldmeso:
### Description of change

Make the `comments` table polymorphic, so that any other table can have one or many comments as a `commentable`.

Currently comments only belong to `policies`, but a lot of current and upcoming tickets look to be needing comments associated with other things!

### Story Link

NA


allbecauseyoutoldmeso:
Mostly down to the magic of factorybot, I think 😆

PR #813, opened 2022-11-18
Review permitted development rights

benbaumann95:
### Description of change

See commits

### Story Link

https://trello.com/c/V0byESvg/1203-reviewer-can-sign-off-permitted-development-rights-check

Screenshot 2022-11-18 at 10 31 09
Screenshot 2022-11-18 at 10 33 12

### DIFF
--- a/app/controllers/planning_application/review_permitted_development_rights_controller.rb
+++ b/app/controllers/planning_application/review_permitted_development_rights_controller.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+class PlanningApplication
+  class ReviewPermittedDevelopmentRightsController < AuthenticationController
+    include CommitMatchable
+    include PlanningApplicationAssessable
+    include PermittedDevelopmentRights
+
+    before_action :set_planning_application
+    before_action :ensure_planning_application_is_validated
+    before_action :set_permitted_development_right, only: %i[show edit update]
+    before_action :set_permitted_development_rights, only: %i[show edit]
+
+    def edit
+      respond_to do |format|
+        format.html
+      end
+    end
+
+    def show
+      respond_to do |format|
+        format.html
+      end
+    end
+
+    def update
+      @permitted_development_right.assign_attributes(
+        review_status: status, reviewer: current_user, reviewed_at: Time.current
+      )
+
+      respond_to do |format|
+        if @permitted_development_right.update(permitted_development_right_params)
+          format.html do
+            redirect_to planning_application_review_tasks_path(@planning_application),
+                        notice: I18n.t("permitted_development_rights.successfully_updated")
+          end
+        else
+          set_permitted_development_rights
+          format.html { render :edit }
+        end
+      end
+    end
+
+    private
+
+    def permitted_development_right_params
+      params.require(:permitted_development_right).permit(:removed_reason, :accepted, :reviewer_comment)
+    end
+
+    def status
+      save_progress? ? "review_in_progress" : "review_complete"
+    end
+  end
+end
--- a/app/controllers/planning_application/review_tasks_controller.rb
+++ b/app/controllers/planning_application/review_tasks_controller.rb
@@ -10,5 +10,21 @@ class PlanningApplication
         format.html
       end
     end
+
+    private
+
+    def set_planning_application
+      planning_application = planning_applications_scope.find(planning_application_id)
+
+      @planning_application = PlanningApplicationPresenter.new(view_context, planning_application)
+    end
+
+    def planning_applications_scope
+      current_local_authority.planning_applications
+    end
+
+    def planning_application_id
+      Integer(params[:planning_application_id])
+    end
   end
 end
--- a/app/controllers/planning_application/review_tasks_controller.rb
+++ b/app/controllers/planning_application/review_tasks_controller.rb
@@ -10,21 +10,5 @@ class PlanningApplication
         format.html
       end
     end
-
-    private
-
-    def set_planning_application
-      planning_application = planning_applications_scope.find(planning_application_id)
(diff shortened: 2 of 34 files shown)


app/presenters/concerns/review_tasks/permitted_development_right_presenter.rb
```diff
@@ -0,0 +1,73 @@
...
+        edit_planning_application_review_permitted_development_right_path(planning_application,
+                                                                          permitted_development_right)
+      when "Review in progress", "Review complete"
+        planning_application_review_permitted_development_right_path(planning_application, permitted_development_right)
+      else
+        raise ArgumentError, "The status provided: '#{status}' is not valid"
+      end
+    end
+
+    def permitted_development_rights_status_tag
```
notapatch, line 45: I think we're stretching the role of the presenter too far when there's a better alternative. It's complicated to try to understand the HTML structure. The status tag component is offering an easier solution you can use the template and then the focus of the test can be the thing that changes ... in this case` govuk-tag--blue` and not the entire html structure that you end up having to test (It's right to test the constructed HTML but comparing the whole html line means you're not sure what is the thing that varies and is most likely to break in this test). 

This same, relatively, complicated code is present in [all the presenters with status tags](https://github.com/unboxed/bops/blob/main/app/presenters/concerns/assessment_tasks_presenter.rb). We haven't tried to abstract the status tag code and we should want to when it's used so often.

Have you tried the status tag component? What did you think of the abstraction?

There haven't been tests for the status tag as the rspec setup had been incorrectly configured. However, that's now fixed and this is an example test which will be in my PR:

```ruby
# frozen_string_literal: true

require "rails_helper"

RSpec.describe StatusTags::ReviewPolicyClassComponent, type: :component do
  it "displays start status" do
    review_policy_class = create(:review_policy_class, status: :not_checked_yet)

    render_inline(described_class.new(review_policy_class: review_policy_class))

    expect(page).to have_css ".govuk-tag--grey", text: "Not checked yet"
  end

  it "displays complete status" do
    review_policy_class = create(:review_policy_class, status: :complete)

    render_inline(described_class.new(review_policy_class: review_policy_class))

    expect(page).to have_css ".govuk-tag--blue", text: "Complete"
  end
end
```
This is **unreviewed** code that goes with the test class for refrence


```ruby
module StatusTags
  class ReviewPolicyClassComponent < StatusTags::BaseComponent
    def initialize(review_policy_class:)
      @review_policy_class = review_policy_class
    end

    private

    def status
      @review_policy_class&.status&.to_sym || :not_checked_yet
    end
  end
end
```

spec/presenters/review_tasks/permitted_development_right_presenter_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+        expect(html).to include(
+          link_to(
+            "Permitted development rights",
+            edit_planning_application_review_permitted_development_right_path(planning_application, permitted_development_right),
+            class: "govuk-link"
+          )
+        )
+
+        expect(html).to include(
+          "<strong class=\"govuk-tag govuk-tag--grey app-task-list__task-tag\">Not started</strong>"
```
notapatch, line 30: See this test as I say you're right to test that strong is there, the classes are there as it's easy to muck up the html you're building but it would have been much easier to use an erb template already given to you in the status tags.

notapatch (approved):

app/models/planning_application.rb
```diff
@@ -41,6 +41,7 @@ class WithdrawRecommendationError < RuntimeError; end
     has_many :notes, -> { by_created_at_desc }, inverse_of: :planning_application
     has_many :requests, class_name: "ValidationRequest"
     has_many :assessment_details, -> { by_created_at_desc }, inverse_of: :planning_application
+    has_many :permitted_development_rights
```
allbecauseyoutoldmeso, line 44: Might be worth adding an order to this, just to make sure `#permitted_development_right` (singuler) always returns the right thing..?

allbecauseyoutoldmeso (approved):
Looks good!

EGiataganas (approved):

app/presenters/concerns/review_tasks/permitted_development_right_presenter.rb
benbaumann95: Thanks for the suggestion, I was being consistent with previous work achieving this sort of thing but I forgot about this module. 
I think we could do some refactor work moving all the status html to this format if the team agree but perhaps as a separate PR.

app/models/planning_application.rb
benbaumann95: I believe it will return ordered by primary id which would be the created_at but no harm to add !

app/controllers/planning_application/review_tasks_controller.rb
```diff
@@ -10,5 +10,21 @@ def index
         format.html
       end
     end
+
+    private
+
+    def set_planning_application
```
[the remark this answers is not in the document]
benbaumann95: Yep good spot, just habit!

app/controllers/planning_application/review_permitted_development_rights_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+      @planning_application = PlanningApplicationPresenter.new(view_context, planning_application)
+    end
+
+    def set_permitted_development_right
+      @permitted_development_right = @planning_application.permitted_development_right
+    end
+
+    def set_permitted_development_rights
+      @permitted_development_rights = @planning_application.permitted_development_rights.returned
+    end
```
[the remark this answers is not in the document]
benbaumann95: Yep !

PR #819, opened 2022-11-23
Refactor replace button-group with shared partial

notapatch:
Refactor button groups with a shared partial


allbecauseyoutoldmeso (approved):
🌟

PR #820, opened 2022-11-24
Display error when editing permitted development right if recommendation has been accepted

benbaumann95:


allbecauseyoutoldmeso (approved):

app/models/recommendation.rb
```diff
@@ -60,6 +60,10 @@ def unchallenged?
     !challenged
   end
 
+  def accepted?
+    review_complete? && challenged == false
+  end
```
allbecauseyoutoldmeso, line 65: Should probably add a test for this (I was about to write one, honest 😆)

notapatch: Rebase with main and you get free tests from ben

PR #828, opened 2022-11-29
Missing expectation in spec/jobs/upload_document_job_spec.rb

notapatch:
Test code is missing an expectation

### DIFF
--- a/spec/jobs/upload_documents_job_spec.rb
+++ b/spec/jobs/upload_documents_job_spec.rb
@@ -13,7 +13,7 @@ RSpec.describe UploadDocumentsJob, type: :job do
     let!(:service) { double }
 
     it "calls UploadDocumentsService" do
-      allow(service).to receive(:call).and_return(service)
+      expect(service).to receive(:call)
 
       allow(UploadDocumentsService).to receive(:new)
         .with(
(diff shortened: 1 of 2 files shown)


spec/jobs/upload_documents_job_spec.rb
```diff
@@ -13,17 +13,17 @@
     let!(:service) { double }
 
     it "calls UploadDocumentsService" do
-      allow(service).to receive(:call).and_return(service)
+      expect(service).to receive(:call)
 
       allow(UploadDocumentsService).to receive(:new)
```
pixeltrix, line 18: I'd recommend changing this to an `expect` as well

allbecauseyoutoldmeso (approved):
🙌

spec/jobs/upload_documents_job_spec.rb
notapatch: @allbecauseyoutoldmeso also had expect there before Rubocop's suggestion. 

When I make the change Rubocop tells me:
[RSpec/StubbedMock: Prefer allow over expect when configuring a response.
](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/StubbedMock)

Do you still prefer the change?

allbecauseyoutoldmeso: `service` only receives `#call` if `UploadDocumentsService` receives `.new`, so I guess the first expectation covers both 🤔

pixeltrix: I think what the rule is saying is that you do a `allow` for the `and_return` and then an `expect` to set an expectation. This seems arbitrarily verbose and I'd recommended disabling `RSpec/StubbedMock` in the rubocop config.

PR #834, opened 2022-11-30
Improve phone number authentication

allbecauseyoutoldmeso:
### Description of change

- Alter validation of `mobile_number` in `users` table so that it

1. Can include white spaces, `-`, `(`, `)` or `+`.
2. Excluding the above characters contains only digits and is between 8 and 15 characters long.

- Apply these validations when the user submits the mobile number during the login process.

NB I've copied the accepted characters and length criteria from [Notify](https://github.com/alphagov/notifications-utils/blob/8c8c8fb591e27f11fab5cf17c64aead0870d40d2/notifications_utils/recipients.py#L567). It's safe to be pretty generous because if Notify rejects the number the user will see an error at the next step. The important thing at our end is to apply the same validations when the user submits the mobile number as we do when we save the number to the database (we weren't doing this, which is what was causing the bug).

### Story Link

https://trello.com/c/igCiPkfe/1257-some-valid-mobile-number-formats-arent-accepted

### Screenshot

Screenshot 2022-11-30 at 18 13 32

### Decisions

I've opted not to use a gem for now, since in a sense the phone number is being validated by the log in process (if the request to notify succeeds the user gets a code etc). But we explore the gem option? [phonelib](https://github.com/daddyz/phonelib) looks like a good candidate. It uses google's [libphonenumber](https://github.com/google/libphonenumber) which is recommended by GOV.UK Design System.

### DIFF
--- a/app/validators/phone_number_validator.rb
+++ b/app/validators/phone_number_validator.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class PhoneNumberValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return if value.blank? || value.delete("-+ ()").match(/\A\d{8,15}\z/)
+
+    record.errors.add(attribute, :invalid)
+  end
+end
--- a/app/controllers/concerns/authenticate_with_otp_two_factor.rb
+++ b/app/controllers/concerns/authenticate_with_otp_two_factor.rb
@@ -4,6 +4,8 @@ module AuthenticateWithOtpTwoFactor
   extend ActiveSupport::Concern
 
   def authenticate_with_otp_two_factor
+    return unless valid_mobile_number?
+
     user = self.resource = find_user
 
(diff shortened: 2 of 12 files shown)


app/validators/phone_number_validator.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class PhoneNumberValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return if value.blank? || value.delete("-+ ()").match(/\A\d{8,15}\z/)
```
notapatch, line 5: [I assume you're calling this??](https://ruby-doc.org/core-3.1.2/String.html#method-i-delete)
value.delete("-+ ()")

notapatch (approved):

notapatch:
Worth re-testing with the changes merged from rubocop? 👮‍♂️

allbecauseyoutoldmeso:
> Worth re-testing with the changes merged from rubocop? 👮‍♂️

💯 %!

PR #841, opened 2022-12-05
Some status tag fixing and refactoring

allbecauseyoutoldmeso:
### Description of change

### Story Link

https://link-to-issue

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


app/components/status_tags/check_and_assess_component.rb
```diff
@@ -12,22 +10,18 @@ def initialize(planning_application:)
 
     attr_reader :planning_application
 
-    delegate(:recommendation, to: :planning_application, allow_nil: true)
```
notapatch, line 15: :allow_nil - If set to true, prevents a Module::DelegationError from being raised

I assume that means if planning_application is null it will now error?

notapatch (approved):

app/components/status_tags/check_and_assess_component.rb
allbecauseyoutoldmeso: It will, but `planning_application` is never null. I think when I initially wrote this I got confused, and thought it was protecting against an error if `recommendation` is null, which it can be.

PR #843, opened 2022-12-06
Rails/I18nLocaleTexts

notapatch:
Enforces use of I18n and locale files instead of locale specific strings.

- In general added a key for the verb and then a ".success" for a success string rather than the full name as it isn't very useful when debugging (you just want to know if you're on the happy path) and it increases the chance of reuse / refactoring at a later date.

- Something that broke this pattern was concerns
- Concerns

  Shared translations
  - choose to have one share the rest

  - concerns.validation_requests.cancel.success
    - additional_document_validation_requests.cancel.success
    - red_line_boundary_change_validation_requests.cancel.success - other_change_validation_requests.cancel.success - replacement_document_validation_requests.cancel.success

  - concerns.validation_requests.destroy.success - "Validation request was successfully deleted." - replacement_document_validation_requests.destroy.success - additional_document_validation_requests.destroy.success - other_change_validation_requests.destroy.success

- I followed the best practice which *seemed* to be what was suggested in Rubocop docs

- Unfixed errors there are still a number of errors unfixed and the reason I didn't was that they were without tests. In many cases it didn't even break the test when I raised an error instead of returning

**Rubcop Examples**

Models

Bad
```
class User < ApplicationRecord
  validates :email, presence: { message: "must be present" }
end
```

Good

```
 config/locales/en.yml
 en:
   activerecord:
     errors:
       models:
         user:
           blank: "must be present"
```

```
class User < ApplicationRecord
  validates :email, presence: true end
```

----------------------------------------------------------

Controllers

Bad

```
class PostsController < ApplicationController
  def create
    # ...
    redirect_to root_path, notice: "Post created!"
  end
end
```

Good
```
config/locales/en.yml
en:
  posts:
    create:
      success: "Post created!"
```

```
class PostsController < ApplicationController
  def create
    # ...
    redirect_to root_path, notice: t(".success")
  end
end
```


allbecauseyoutoldmeso (approved):
I cannot tell you how happy this makes me 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #844, opened 2022-12-06
Pr np task list item component

notapatch:
Idea to do the list item component

This is what I think AssessmentDetailsReview would look like

```
     <%= render TaskListItems::ReviewComponent.new do |c| %>
       <% c.with_link do %>
         Link here
       <% end %>

       <% c.with_tag do %>
         <%= render(
             StatusTags::AssessmentDetailsReviewComponent.new(
               planning_application: @planning_application
             )
         )%>
       <% end %>
     <% end %>
```

### DIFF
--- a/app/components/task_list_items/review_component.html.erb
+++ b/app/components/task_list_items/review_component.html.erb
@@ -1,0 +1,6 @@
+<li class="app-task-list__item">
+  <span class="app-task-list__task-name">
+    <%= link %>
+  </span>
+  <%= tag %>
+</li>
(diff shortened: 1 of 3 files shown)


allbecauseyoutoldmeso (approved):
I like!

Should we refactor all the `TaskListItem` components to work like this?

allbecauseyoutoldmeso:
I'm about to move a few more task list items into components. Maybe I'll do that first, and then we can refactor them all together 🤔

notapatch:
All at once is normally the best policy ...

PR #864, opened 2022-12-12
Staging banner

notapatch:
- Main issue is that system tests are flakey as there's only the "test" environment. I can test it doesn't appear on the test environment but it's obviously a weak test.

- The component, unit, tests are better and you can change the environment

## Can you spot it?

Screenshot 2022-12-12 at 18 02 16

### DIFF
--- a/app/components/environment_banner_component.html.erb
+++ b/app/components/environment_banner_component.html.erb
@@ -1,0 +1,5 @@
+<header class="environment-banner">
+  <div class="govuk-width-container">
+    This is staging. Only process test cases on this version of BoPS
+  </div>
+</header>
--- a/app/components/environment_banner_component.rb
+++ b/app/components/environment_banner_component.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class EnvironmentBannerComponent < ViewComponent::Base
+  def initialize(display:)
+    @display = display
+  end
+
+  def render?
+    @display
+  end
+end
--- a/app/javascript/stylesheets/application.scss
+++ b/app/javascript/stylesheets/application.scss
@@ -59,6 +59,11 @@ html * {
   padding: govuk-spacing(2);
 }
 
+.environment-banner {
+  background-color: govuk-colour("yellow");
+  padding: govuk-spacing(1);
+}
+
 .alert__icon {
   fill: govuk-colour("red");
   float: left;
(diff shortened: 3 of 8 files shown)


app/components/environment_banner_component.html.erb
```diff
@@ -0,0 +1,5 @@
+<header class="environment-banner">
+  <div class="govuk-width-container">
+    This is <%= Rails.env %>. Only process test cases on this version of BoPS
```
notapatch, line 3: I added this as I thought it might be used on other environments at some point. However, maybe I should just hardocde it for simplicity??

app/components/environment_banner_component.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class EnvironmentBannerComponent < ViewComponent::Base
+  def render?
+    Rails.env.staging?
```
benbaumann95, line 5: Does `ENV["RAILS_ENV"] == "production"` on the staging environment?

https://github.com/unboxed/bops-terraform/blob/main/ansible/templates/app/etc/default/application.j2#L5

This isn't differentiated on the `ansible_env`

benbaumann95: We might need to set our own custom environment variable if we can't distinguish production and staging with an existing differentiated env value

notapatch: Interesting, I've always added staging environment... for staging. You have production for both?

EGiataganas: Yes we have production for both. If we want to have staging we will need to have an additional `staging` file [here][1]
I think we could add another environment variable maybe `ENV_NAME`. Set it to default production, on staging set it to staging and add a check if it's staging show the banner as you did! 

[1]: https://github.com/unboxed/bops/tree/main/config/environments

EGiataganas (approved):

app/components/environment_banner_component.html.erb
```diff
@@ -0,0 +1,5 @@
+<header class="environment-banner">
+  <div class="govuk-width-container">
+    This is staging. Only process test cases on this version of BoPS
```
allbecauseyoutoldmeso, line 3: Lower case 'o' 🙌 😄

allbecauseyoutoldmeso (approved):
Nice!

app/components/environment_banner_component.html.erb
notapatch: I hope they appreciate the quality that has gone into this 🤔

PR #868, opened 2022-12-13
Remove buttons masquerading as links from consistency checklist page

allbecauseyoutoldmeso:
### Description of change

As discussed with the design team, in order to avoid having 'links' that are really buttons:

- Change links on 'Description, documents and proposal details' page to actually BE links. Currently they look like links but submit the form before the user is taken to the expected path.
- Update the logic of the 'Description, documents and proposal details' page so that 'No' is automatically selected for a check if there are open requests associated with it.

As part of the work for [this ticket](https://trello.com/c/c2HVZjBV/1382-where-there-are-links-to-actions-from-the-accordion-users-should-return-to-the-same-place-they-left-from):

- Update the logic for redirecting back to the 'Description, documents and proposal details' page after a request has been made to use a value stored in the form, not the session. This means that if the user behaves unexpectedly we don't have confusing information hanging around in the session!
- Make updates so that the user is redirected back to the previous page by default, not to a hard-coded page.

### DIFF
--- a/app/controllers/red_line_boundary_change_validation_requests_controller.rb
+++ b/app/controllers/red_line_boundary_change_validation_requests_controller.rb
@@ -5,7 +5,6 @@ class RedLineBoundaryChangeValidationRequestsController < ValidationRequestsCont
 
   before_action :ensure_no_open_or_pending_red_line_boundary_validation_request, only: %i[new]
   before_action :ensure_planning_application_is_not_closed_or_cancelled, only: %i[new create]
-  before_action :set_return_to, only: %i[new]
 
   def show
     @red_line_boundary_change_validation_request = @planning_application.red_line_boundary_change_validation_requests.find(params[:id])
@@ -22,12 +21,27 @@ class RedLineBoundaryChangeValidationRequestsController < ValidationRequestsCont
 
     if @red_line_boundary_change_validation_request.save
       redirect_to(
-        (session.delete(:return_to) || create_request_redirect_url),
+        create_request_redirect_url,
         notice: t(".validation_request_for")
       )
     else
       render :new
     end
+  end
+
+  def create_request_redirect_url
+    case return_to
+    when nil
+      super
+    when planning_application_sitemap_url(@planning_application)
+      planning_application_validation_tasks_path(@planning_application)
+    else
+      return_to
+    end
+  end
+
+  def return_to
+    params.dig(:red_line_boundary_change_validation_request, :return_to)
   end
 
   private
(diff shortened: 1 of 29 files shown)


app/controllers/red_line_boundary_change_validation_requests_controller.rb
```diff
@@ -22,14 +21,29 @@ def create
...
     else
       render :new
     end
   end
 
+  def create_request_redirect_url
+    case return_to
+    when nil
+      super
+    when planning_application_sitemap_url(@planning_application)
```
allbecauseyoutoldmeso, line 36: The default is to return to the previous page. However there is one specific user journey, where the user is directed to the request form from another form, where we want need to overwrite this behavious.

app/models/consistency_checklist.rb
```diff
@@ -23,6 +31,16 @@ class ConsistencyChecklist < ApplicationRecord
     enum(check => { to_be_determined: 0, yes: 1, no: 2 }, _prefix: check)
   end
 
+  REQUEST_TYPES.each do |check, request_type|
```
allbecauseyoutoldmeso, line 34: Is all this meta-coding unreadable?!

allbecauseyoutoldmeso:
This was not fun to write, and probably won't be fun to review! Happy to talk through it.

app/models/consistency_checklist.rb
notapatch: In truth it's readable within the file... it's not great but you can work it out - to be fair the array/hash of items creating methods, REQUEST_TYPES in this case, it's quite common.

What I didn't work out is how these records interacts with the rest of the application. I was about to put in "raise 'error'" and thought, "I shouldn't have to do that to know what's happening. I rely on "grepping" the code but it's more difficult when you don't have a unique word to search for. If I am not alone in the confusion maybe a comment??

notapatch (approved):

app/models/consistency_checklist.rb
allbecauseyoutoldmeso: Yes, that sounds like a good idea! Maybe just comments listing the public methods that are defined by each loop?

eg above line 35 `# defined #defaut_description_matches_documents_to_no?, #defaut_documents_consistent_to_no?, #defaut_site_map_correct_to_no?`

That way anyone searching the application to find where the method is defined will end up in the right place!

notapatch: It helps having the full name as I can see where it's used but I can't tell if I'm being thick or not.

EGiataganas (approved):

PR #871, opened 2022-12-14
Present helpful error message when using expired password reset link

benbaumann95:
### Story Link

https://trello.com/c/M8pegXZo/1356-what-does-reset-password-token-is-invalid-mean

An error "Reset password token is invalid" is presented when a user has clicked on an expired reset password link (perhaps a more recent reset password request was made). This PR updates that message to something a bit more helpful.

### DIFF
--- a/app/views/devise/passwords/edit.html.erb
+++ b/app/views/devise/passwords/edit.html.erb
@@ -4,8 +4,14 @@
 
 <h1 class="govuk-heading-l">Change your password</h1>
 
-<%= form_for(resource, as: resource_name, url: password_path(resource_name), html: { method: :put }) do |f| %>
-  <%= render "devise/shared/error_messages", resource: resource %>
+<%= form_with(
+  model: resource,
+  as: resource_name,
+  url: password_path(resource_name),
+  method: :put,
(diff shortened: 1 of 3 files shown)


app/views/devise/passwords/edit.html.erb
```diff
@@ -4,8 +4,14 @@
 
 <h1 class="govuk-heading-l">Change your password</h1>
 
-<%= form_for(resource, as: resource_name, url: password_path(resource_name), html: { method: :put }) do |f| %>
-  <%= render "devise/shared/error_messages", resource: resource %>
+<%= form_for(
```
allbecauseyoutoldmeso, line 7: Might be worth updating to `form_with`?

app/views/devise/passwords/edit.html.erb
```diff
@@ -4,8 +4,14 @@
...
+  resource,
+  as: resource_name,
+  url: password_path(resource_name),
+  html: { method: :put },
```
allbecauseyoutoldmeso, line 11: I know you didn't write this, but isn't it just `method: :put` rather than `html: { method: :put }` 🤔

allbecauseyoutoldmeso (approved):
Really nice!

PR #877, opened 2022-12-19
Fix RSpec/NoExpectionExample errors

allbecauseyoutoldmeso:
### Description of change

- Check that (stubbed) requests are successful, rather than stubbing Faraday, in client specs for Mapit and PAAPI APIs.

### Decisions

Alternatively we could change the `allow`s to `expects`, but I don't think it hurts to check Faraday actually makes the request!

### DIFF
--- a/spec/services/apis/mapit/client_spec.rb
+++ b/spec/services/apis/mapit/client_spec.rb
@@ -6,13 +6,8 @@ RSpec.describe Apis::Mapit::Client do
   let(:client) { described_class.new }
 
   describe "#call" do
-    let(:faraday_connection) { spy }
-    let(:url) { "https://mapit.mysociety.org" }
-
-    it "makes a Faraday connection" do
-      allow(Faraday).to receive(:new).with(url: url).and_yield(faraday_connection)
-
-      client.call("SE220HW")
+    it "is successfull" do
+      expect(client.call("SE220HW").status).to eq(200)
     end
 
     it "removes whitespace from the postcode" do
--- a/spec/services/apis/mapit/client_spec.rb
+++ b/spec/services/apis/mapit/client_spec.rb
@@ -6,7 +6,7 @@ RSpec.describe Apis::Mapit::Client do
   let(:client) { described_class.new }
 
   describe "#call" do
-    it "is successfull" do
+    it "is successful" do
       expect(client.call("SE220HW").status).to eq(200)
     end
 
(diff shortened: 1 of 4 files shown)


spec/services/apis/mapit/client_spec.rb
```diff
@@ -6,13 +6,8 @@
...
 
   describe "#call" do
-    let(:faraday_connection) { spy }
-    let(:url) { "https://mapit.mysociety.org" }
-
-    it "makes a Faraday connection" do
-      allow(Faraday).to receive(:new).with(url: url).and_yield(faraday_connection)
-
-      client.call("SE220HW")
+    it "is successfull" do
```
notapatch, line 9: typo ... successful

spec/services/apis/paapi/client_spec.rb
```diff
@@ -6,13 +6,8 @@
...
   describe "#call" do
-    let(:faraday_connection) { spy }
-    let(:url) { "https://staging.paapi.services/api/v1" }
-
-    it "makes a Faraday connection" do
-      allow(Faraday).to receive(:new).with(url: url).and_yield(faraday_connection)
-
-      client.call("100081043511")
+    it "is successful" do
+      expect(client.call("100081043511").status).to eq(200)
```
notapatch, line 10: So if I remove network access to the system the test fails: In other words a test run generates a network request each time it runs. Which is a no no.

If it doesn't make sense to bring in webmock or some such (I have no idea what's involved in testing it). I am also OK with the tests being ripped out. At least it isn't pretending to test something that it isn't.

I am no network testing expert!!

[I haven't read it through and it might not be relevant but Faraday had this??
](https://lostisland.github.io/faraday/adapters/testing)
[Which had an RSpec link to this](https://github.com/lostisland/faraday/blob/main/examples/client_spec.rb
)

notapatch (commented):
Apologies for nothing more useful.

spec/services/apis/paapi/client_spec.rb
allbecauseyoutoldmeso: Huh...so we already were stubbing the request with Webmock [here](https://github.com/unboxed/bops/blob/main/spec/support/api/paapi_helpers.rb). The test fails without network access because we had `net_http_connect_on_start` set to true. See https://github.com/bblimke/webmock#connecting-on-nethttpstart. Who knew!

I've set it to false globally now, but it looks like you can disable it for a specific request as well. Not sure which is better.

notapatch: Well, I looked at that helper I know a bit more but not sure how everything happens ... It seems to self load itself? 
Anyway, that's not your problem. I'll approve it and hope nobody asks me to change it 🙏

notapatch (approved):

PR #878, opened 2022-12-19
Complete -> Completed

allbecauseyoutoldmeso:
### Description of change

- Move task list item for reviewing documents to components.
- Move validation status tag to components.
- Change 'complete' to 'completed' for status tag components.

### Story Link

https://trello.com/c/irWJtop9/1363-replace-any-instances-of-the-complete-tag-with-the-completed-tag


allbecauseyoutoldmeso:
I guess this is kind of a POC. It moves the last of the status tags (excluding validation task-list ones) into components, so that anything we want to change about status tags can be changed in just one place. That means if we want to change the text of a status tag (in this case 'Complete' -> 'Completed'), it involves changing just one line in the `en.yml` file.

I know there's an argument for going through all the 'status' enum columns in the various models and making them consistent with front end copy. But this would be a fairly massive task and I'd argue that it's not one we want to do when the copy is still in a state of flux! 

Plus is wouldn't achieve our requirements here since not all status tag 'statuses' are derived from a single column.

EGiataganas (approved):
I am happy with this change. Yes, I would argue that it would be great to update enums to match the name for better future changes. Although, we can do that in another pr only this bit and renaming the locale key. 😄

allbecauseyoutoldmeso:
Thanks @EGiataganas! This is the quick pre-Christmas fix...my plan is to be on a beach somewhere when the rest of the team is updating all the enums 🤣 🌴

PR #880, opened 2022-12-19
Change 'Not checked yet' to 'Not started'

allbecauseyoutoldmeso:
### Description of change

- Move task list items for validation tasks from `PlanningApplicationPresenter` to view components.
- Replace `not_changed_yet` with `not_started` everywhere!

### Story Link

https://trello.com/c/mpyoV8mD/1361-replace-not-checked-tag-with-not-started-across-service


app/components/status_tags/additional_document_component.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    private
+
+    attr_reader :planning_application
+
+    def status
+      if planning_application.additional_document_validation_requests.open_or_pending.any?
+        :invalid
+      elsif planning_application.documents_missing == false
```
allbecauseyoutoldmeso, line 16: Future improvement to add 'not null' to all the boolean columns 😬

app/components/task_list_items/base_component.rb
```diff
@@ -7,5 +7,9 @@ class BaseComponent < ViewComponent::Base
     def link_active?
       true
     end
+
+    def status_tag_component
+      StatusTags::BaseComponent.new(status: status)
+    end
```
allbecauseyoutoldmeso, line 13: Make this the default so that if a task list item has a `status` method that returns a status the basic status tag component will be rendered automatically.

app/components/task_list_items/check_red_line_boundary_component.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+module TaskListItems
+  class CheckRedLineBoundaryComponent < TaskListItems::BaseComponent
```
allbecauseyoutoldmeso, line 4: I've tried to copy the logic in this component faithfully from the presenter - it's a bit hairy!

allbecauseyoutoldmeso:
I kind of wish I'd never started this, but here it is 😬

app/components/status_tags/additional_document_component.rb
notapatch: Twenty Twenty Three problems

app/components/task_list_items/base_component.rb
notapatch: My review_component is sitting in that directory but not using the base. Not sure if it should be moved out or changed to fit in.

app/models/review_policy_class.rb
```diff
@@ -4,7 +4,12 @@ class ReviewPolicyClass < ApplicationRecord
   belongs_to :policy_class, optional: true
 
   enum mark: { not_marked: 0, accept: 1, return_to_officer_with_comment: 2 }
-  enum status: { not_checked_yet: 0, complete: 1 }, _default: :not_checked_yet, _prefix: true
+
+  enum(
```
notapatch, line 8: Consistency change?

notapatch (approved):
There's quite a lot of system wide changes, however, I can confirm that "Not checked yet" was indeed changed to "Not started"

PR #883, opened 2022-12-20
Add rails_autolink to ensure links are clickable

benbaumann95:
### Description of change

- Extract simple_format and auto_link helper methods to add this formatting to all the relevant content
- The `auto_link` method was removed from Rails in version Rails 3.1 and extracted into this [gem](https://github.com/tenderlove/rails_autolink)
- Allow user to enter href / target attributes and do not sanitize this
- We sanitize the string with our own strategy and ignore the default santize behaviour with simple_format and auto_link as this removes href / target tags

(Will need to add this code to bops-applicant separately as we cannot pass this formatted content via the API and neither would we want to)

### Story Link

https://trello.com/c/P7zc4HvZ/1349-make-links-clickable-on-bops-bops-applicant


app/helpers/format_content_helper.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module FormatContentHelper
+  def auto_link_and_simple_format_content(content:, classname: nil)
+    allowed_attributes = Rails::Html::Sanitizer.white_list_sanitizer.allowed_attributes + %w[href target]
```
notapatch, line 5: > We sanitize the string with our own strategy 

Is this "our own strategy" ?

app/helpers/format_content_helper.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module FormatContentHelper
```
allbecauseyoutoldmeso, line 3: I wonder if this is a good use case for a view component?

`FormattedContentCompent.new(content: content)` ...

In theory any time we find ourselves reaching for view helpers we should be using view components instead 🤔

app/helpers/format_content_helper.rb
```diff
@@ -0,0 +1,12 @@
...
+  def auto_link_and_simple_format_content(content:, classname: nil)
+    allowed_attributes = Rails::Html::Sanitizer.white_list_sanitizer.allowed_attributes + %w[href target]
+
+    auto_link_content = auto_link(content.to_s, html: { target: "_blank" }, sanitize: false)
```
allbecauseyoutoldmeso, line 7: What happens if you have `sanitize: true` here?

spec/helpers/format_content_helper_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe FormatContentHelper do
```
allbecauseyoutoldmeso, line 5: Also if you made it a component you could have a nice component spec with `render_inline` and `expect(page)...` etc 🙂

allbecauseyoutoldmeso (approved):
This seems like a really nice solution!

I think it should use a view component, but I guess it's personal preference more than anything 😁

app/helpers/format_content_helper.rb
benbaumann95: It will remove the href and target attributes that we want to keep

allbecauseyoutoldmeso: That seems like a self-defeating option for an `#auto_link` method 🤣

benbaumann95: It's so we can enable users to enter some html instead of just a string with a link

app/helpers/format_content_helper.rb
benbaumann95: Want to keep consistent with bops-applicant so will have to add the view_component gem there too.

Reckon that's not overkill?
I guess we may want to move to that view approach on that app too

allbecauseyoutoldmeso: I am in favour of view components in All Of The Places...but possibly I'll get over it 😆

app/helpers/format_content_helper.rb
allbecauseyoutoldmeso: ohhh, yes, got it!

benbaumann95: I think if you take it out, this [spec](https://github.com/unboxed/bops/pull/883/files#diff-73cf68088b043c75735045653397a26e21544d48d508d0da92abc26b31dcb77fR21-R27) would fail

PR #885, opened 2022-12-21
Pr np check the application page

notapatch:
### Description of change
check-all-others

pr-check-check-list

- AC1 - Change "Manage document" from link to secondary CTA
- AC2 - On document accordion: If a reference has been added by the officer, use the "Document reference" instead the "Document name"
- AC3 - On Check supplied document list: If a reference has been added by the officer, use the "Document reference" instead the "Document name"
- AC4 - Document accordion: interline indented to the content

### Story Link

https://trello.com/c/VzSVWVoY/1270-check-the-application-page-check-list-changes

### DIFF
--- a/app/helpers/document_helper.rb
+++ b/app/helpers/document_helper.rb
@@ -24,4 +24,12 @@ module DocumentHelper
   def document_name_and_reference(document)
     "#{document.name}#{document.numbers? ? " - #{document.numbers}" : ''}"
   end
+
+  def titled_reference_or_file_name(document)
+    document.numbers? ? "Reference: #{document.numbers}" : "File name: #{document.name}"
+  end
+
+  def reference_or_file_name(document)
+    document.numbers.presence || document.name
+  end
 end
--- a/app/presenters/concerns/validation_tasks/documents_presenter.rb
+++ b/app/presenters/concerns/validation_tasks/documents_presenter.rb
@@ -4,6 +4,7 @@ module ValidationTasks
   extend ActiveSupport::Concern
 
   class DocumentsPresenter < PlanningApplicationPresenter
+    include DocumentHelper
     attr_reader :document
 
     def initialize(template, planning_application, document)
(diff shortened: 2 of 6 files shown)


app/presenters/concerns/validation_tasks/documents_presenter.rb
```diff
@@ -4,6 +4,7 @@ module ValidationTasks
   extend ActiveSupport::Concern
 
   class DocumentsPresenter < PlanningApplicationPresenter
```
allbecauseyoutoldmeso, line 6: I've removed this class in [this change](https://github.com/unboxed/bops/pull/880/files#diff-342eeb61b06bda5c910d2bd51dcd8cf85fec5ef02a340f3a2a3404029892c9d9)!

I think easiest if you merge this first, and then I'll do a careful rebase ☺️

app/helpers/document_helper.rb
```diff
@@ -24,4 +24,12 @@ def created_by(document)
...
     "#{document.name}#{document.numbers? ? " - #{document.numbers}" : ''}"
   end
+
+  def titled_reference_or_file_name(document)
+    document.numbers? ? "Reference: #{document.numbers}" : "File name: #{document.name}"
+  end
+
+  def reference_or_file_name(document)
+    document.numbers.presence || document.name
+  end
```
allbecauseyoutoldmeso, line 34: I guess we currently have to use helper methods given that the logic is needed in both the views and the dreaded `PlanningApplicationPresenter`.

allbecauseyoutoldmeso (approved):
Very nice!

app/helpers/document_helper.rb
notapatch: I did a Component treatment which you've seen - I think comment was overkill.
I didn't want to just add it without tests or rather I would have had to have tested both cases with a system test which is slow and confusing.
I'm definitely open to ideas or it just becomes 2023 issue.

PR #886, opened 2022-12-21
Retain information when replacing documents

allbecauseyoutoldmeso:
### Description of change

- Update API so that when applicant submits a replacement document the `tags` and `numbers` attributes are copied over from the old document.
- Update logic when assessor submits edit documents form so that if they have uploaded a new file the existing document record is archived with the old file, and a new document record is created with the new file 😅 
- Tidy up the edit document form a bit.

### Story Link

https://trello.com/c/Nek3QKHF/1287-archive-document-when-user-upload-a-replacement-file

### DIFF
--- a/app/models/replacement_document_validation_request.rb
+++ b/app/models/replacement_document_validation_request.rb
@@ -30,6 +30,19 @@ class ReplacementDocumentValidationRequest < ApplicationRecord
     raise ResetDocumentInvalidationError, e.message
   end
 
+  def replace_document!(file:, reason:)
+    transaction do
+      self.new_document = planning_application.documents.create!(
+        file: file,
+        tags: old_document.tags,
+        numbers: old_document.numbers
+      )
+
+      close!
+      old_document.update!(archive_reason: reason, archived_at: Time.zone.now)
+    end
+  end
+
   private
 
   def audit_api_comment
--- a/spec/models/document_spec.rb
+++ b/spec/models/document_spec.rb
@@ -274,4 +274,119 @@ RSpec.describe Document do
       end
     end
   end
+
+  describe "#update_or_replace" do
+    let(:planning_application) { create(:planning_application, :not_started) }
+
+    let(:file_path1) do
+      Rails.root.join("spec/fixtures/images/proposed-floorplan.png")
+    end
+
+    let(:file_path2) do
+      Rails.root.join("spec/fixtures/images/proposed-roofplan.png")
+    end
+
+    let(:file1) { Rack::Test::UploadedFile.new(file_path1, "image/png") }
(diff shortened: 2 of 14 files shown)


spec/models/document_spec.rb
```diff
@@ -274,4 +274,119 @@
...
+    let(:document) do
+      create(
+        :document,
+        file: file1,
+        planning_application: planning_application,
+        numbers: "DOC123"
+      )
+    end
+
+    context "when there is no 'file' attribute" do
```
notapatch, line 301: This is just a general question about contexts and nothing to do with this PR.

In general, contexts cover the set of data so they are often paired... so in this case you have "when there is **no** 'file' attribute" and when there is a 'file' attribute??

It's just that I've been thinking about some of the contexts I've seen in other PRs.

app/models/replacement_document_validation_request.rb
```diff
@@ -30,6 +30,19 @@ def reset_document_invalidation
...
 
+  def replace_document!(file:, reason:)
+    transaction do
+      self.new_document = planning_application.documents.create!(
+        file: file,
+        tags: old_document.tags,
+        numbers: old_document.numbers
+      )
+
+      close!
```
notapatch, line 41: What are we closing? Is it the new_document so we can change the old_document .... ?? confused

notapatch (approved):
Merry Xmas, Santa got you an approval. Virtual presents are all the rage.

app/models/replacement_document_validation_request.rb
allbecauseyoutoldmeso: The request!

spec/models/document_spec.rb
allbecauseyoutoldmeso: Hmm. I've been vaguely wondering whether Unboxed should have a 'handbook' with style guides etc. This is definitely the sort of thing that should be considered in it!

I think the way I set up contexts is probably a bit haphazard. More `context: "I needed a slightly different data set up"` than anything else 😅

notapatch: I know my test style is different from the rest of the team so I am trying to make sense of the team style.

I don't think there's anything wrong with what you've done here. But, when I tried to imitate the style it seems to be the tail wagging the dog, I need variation in the test setup so I need another context but I don't have any good reasoning for it. The only way around this is to increase the nesting which isn't idea either.

Anyway thanks.

app/models/replacement_document_validation_request.rb
notapatch: Just as well I asked. So the request is for the new document ... you've replace the old document, you close the request and then you update the new document with info. 👍

PR #889, opened 2022-12-28
Replace OpenStruct with ProposalDetail class

allbecauseyoutoldmeso:
### Description of change

- Use custom `ProposalDetail` class instead of `OpenStruct` to parse `planning_application.proposal_details`.

### Story Link

https://trello.com/c/GjWdNTTK/1369-openstruct-refactoring


app/components/concerns/proposal_detailable.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module ProposalDetailable
+  extend ActiveSupport::Concern
+
+  included do
+    def formatted_policy_refs
```
allbecauseyoutoldmeso, line 7: I've just moved this method from somewhere else, but would love suggestions for nice ways to refactor it!

spec/components/proposal_details/list_component_spec.rb
```diff
@@ -3,92 +3,60 @@
...
+  end
+
+  it "renders first group with correct numbered questions" do
+    group_a = page.find("h4", text: "Group a").find(:xpath, "../..")
+    expect(group_a).to have_content("1. \n  \n    Test question 1")
+    expect(group_a).to have_content("2. \n  \n    Test question 3")
+  end
 
-    it "returns numbered proposal details grouped by portal name" do
-      expect(list_component.send(:groups)).to eq(
```
allbecauseyoutoldmeso, line 43: I think when I initially wrote these files I'd forgotten (I think I knew once!) about how to test components nicely. So I wrote a whole bunch of tests for private methods 😬.

app/models/proposal_detail.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class ProposalDetail
```
allbecauseyoutoldmeso, line 3: There's probably more logic that can be pulled into this class over time 🙂

notapatch (approved):
The end

allbecauseyoutoldmeso:
> The end

😭

PR #892, opened 2023-01-03
Fix date (year) specific test expectations

benbaumann95:
### Description of change

Fix date (year) specific test expectations

- Before a planning application is created, we have a `before_create` callback that invokes `set_reference`
```ruby
  def set_reference
    self.reference = [
      Date.current.strftime("%y"),
      application_number,
      application_type_code
    ].join("-")
  end
```

The date used here is dynamic so the test needs to be set up in a way so that at the exact point we create the planning application it has the year we expect.

Previously we would create the planning application with it being the current date (i.e. year 2023) and then in the test itself we travel to a specified date

### Story Link

https://trello.com/c/5LZZ9vkT/1407-fix-all-date-sensitive-tests-that-relied-on-the-year-being-2022

### DIFF
--- a/spec/services/planning_application_search_spec.rb
+++ b/spec/services/planning_application_search_spec.rb
@@ -5,7 +5,7 @@ require "rails_helper"
 RSpec.describe PlanningApplicationSearch do
   describe "#results" do
     let!(:planning_application1) do
-      travel_to(Date.new(2022)) do
+      travel_to("2022-01-01") do
         create(
           :planning_application,
           work_status: "proposed",
--- a/spec/services/planning_application_search_spec.rb
+++ b/spec/services/planning_application_search_spec.rb
@@ -5,12 +5,13 @@ require "rails_helper"
 RSpec.describe PlanningApplicationSearch do
   describe "#results" do
     let!(:planning_application1) do
-      create(
-        :planning_application,
-        work_status: "proposed",
-        created_at: DateTime.new(2022, 1, 1),
-        description: "Add a chimney stack."
-      )
+      travel_to(Date.new(2022)) do
+        create(
+          :planning_application,
+          work_status: "proposed",
+          description: "Add a chimney stack."
+        )
+      end
     end
 
     let!(:planning_application2) do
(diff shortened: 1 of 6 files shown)


spec/services/planning_application_search_spec.rb
```diff
@@ -5,12 +5,14 @@
...
-        description: "Add a chimney stack."
-      )
+      travel_to(Date.new(2022)) do
+        create(
+          :planning_application,
+          work_status: "proposed",
+          created_at: DateTime.new(2022, 1, 1),
+          description: "Add a chimney stack."
+        )
+      end
```
pixeltrix, line 15: Why are you using `DateTime` and `travel_to`? You can just use `Time.current.at_beginning_of_year`.

benbaumann95: `Time.current.at_beginning_of_year` will return 2023. I think the test wants to search a reference with 2022 i.e. `"22-00100-LDCP"`

That created_at override can be taken out though

benbaumann95: Or did you mean just change that created_to to `Time.current.at_beginning_of_year` ?

benbaumann95: But anyway I can just do 
```ruby
let!(:planning_application1) do
      travel_to(Date.new(2022)) do
        create(
          :planning_application,
          work_status: "proposed",
          description: "Add a chimney stack."
        )
     end
   end
```

and the tests pass

benbaumann95: Ignore my comment about returning 2023 🤦 , within the context of that travel_to it returns 2022 obviously

pixeltrix: Sorry, thought the `created_at` was necessary - just using `travel_to` is fine.

benbaumann95: Yep I thought it was necessary too - but checked and it isn't so have dropped it 👍

pixeltrix: FYI, `travel_to` can take a string so you can do `travel_to "2022-01-01"` if the constructor is too ugly 🙈

benbaumann95: ooh nice I prefer the readability of that

pixeltrix (approved):

PR #895, opened 2023-01-05
Create local authority through a script

EGiataganas:
### Description of change

Create local authority through a script

### Story Link

https://trello.com/c/uFtUHvHl

### DIFF
--- a/lib/tasks/local_authority_creation.rake
+++ b/lib/tasks/local_authority_creation.rake
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+require "optparse"
+
+namespace :local_authority do |args|
+  desc "Creates local authority with given name: rake local_authority:create"
+  # environment is required to have access to Rails models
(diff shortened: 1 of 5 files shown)


lib/tasks/local_authority_creation.rake
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require "optparse"
+
+namespace :local_authority do |args|
```
benbaumann95, line 5: Do we want to raise any argument errors if a required parameter is not provided?

Also, perhaps overkill, but a spec might be good here where we test running the rake task and then just see the local authority output at the end

benbaumann95 (approved):

lib/tasks/local_authority_creation.rake
EGiataganas: That's not required as we're relying on database validation errors 😄

PR #897, opened 2023-01-09
Remember exact scroll position when user leaves a page

benbaumann95:
### Description of change

- When navigating back to a previous page, a user should be taken to the exact scroll position of that page
- Use sessionStorage to store this information and access it upon the dom content loading

### Story Link

https://trello.com/c/qE1chdQb/1262-save-or-back-cta-will-take-the-user-back-to-the-exact-point-on-the-previous-page


app/javascript/controllers/index.js
```diff
@@ -14,3 +14,16 @@ application.register("submit-form", SubmitFormController)
 
 import UnsavedChangesController from "./unsaved_changes_controller.js"
 application.register("unsaved-changes", UnsavedChangesController)
+
+window.addEventListener('DOMContentLoaded', () => {
+  let pathname = window.location.pathname;
+  let scrollPositionKey = `${pathname}-scrollPosition`
```
benjamineskola, line 20:
```suggestion
  let scrollPositionKey = `${pathname}-scrollPosition`;
```

nitpick :)

benjamineskola (approved):

EGiataganas (approved):

PR #899, opened 2023-01-09
Enforce secure passwords

benjamineskola:
### Description of change

Currently the only password strength requirement is that it be at least 6 characters, which is extremely low. This PR:

- Increases minimum password length to 8 characters
- Ensures that passwords contain both upper and lowercase letters, and special characters
- Uses `strong_password` gem to ensure entropy and do dictionary checking

### Story Link

https://trello.com/c/CZ7wOksm/1195-password-minimum-8-characters-long-capital-letter-symbol

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


benbaumann95 (approved):

spec/system/managing_users_spec.rb
```diff
@@ -129,3 +133,7 @@
     end
   end
 end
+
+def secure_password
+  SecureRandom.random_bytes(256).chars.select { |b| b.ord > 32 && b.ord < 0x7f }.join[0, 128]
+end
```
benbaumann95, line 139: Was going to suggest to perhaps move this into a helper under `spec/support` - maybe if we have to repeat this once more though

benjamineskola: I'm not even sure if this is the best way to generate a password, but the problem I had was that if the random passwords are even slightly borderline, then given the number of test runs it'll fail an unacceptable number of times.

EGiataganas (approved):

PR #905, opened 2023-01-13
Don't cancel all matrix tests when one fails

benjamineskola:
The default matrix behaviour is not desirable for the way we use it.


benbaumann95 (approved):
This would be good to merge in if the tests now pass

PR #912, opened 2023-01-18
Remove 2FA_ENABLED env variable and rely on user flag

benjamineskola:
### Description of change

Currently in development 2fa is only used if `2FA_ENABLED` is set to true; however, even if this is not the case `generate_otp_secret` will be run as a `before_create` hook, which may then fail because no key has been set — i.e. `2FA_ENABLED` still requires that 2fa is correctly configured but simply skips it, or partially skips it, and will break if it is not correctly configured even though it will never be used.

Trying to fix that lead to additional complexity in the test environment. What seems like the better solution therefore is just to rely on the `otp_required_for_login` flag on a user.

### Story Link

https://link-to-issue

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


app/models/user.rb
```diff
@@ -15,7 +15,7 @@ class User < ApplicationRecord
   has_many :comments, dependent: :nullify
   belongs_to :local_authority, optional: false
 
-  before_create :generate_otp_secret
+  before_create :generate_otp_secret unless !Rails.env.development? || ENV["2FA_ENABLED"] == "true"
```
EGiataganas, line 18: That's really handy for our dev environment, can you please add it on a private method? There's the revert kind logic on [L11](https://github.com/unboxed/bops/blob/main/app/models/user.rb#L11)

benbaumann95: It only uses `database_authenticatable` in development and if 2FA has been disabled.
I think this is because the login mechanism expects `two_factor_authenticatable` instead.

You can try taking the line out and see if you can log in when 2FA is disabled. From memory it won't work

.env.example
```diff
@@ -2,7 +2,7 @@ APPLICANTS_APP_HOST=bops-care.link:3001
 PAAPI_HOST=paapi.services
 OS_VECTOR_TILES_API_KEY=xxxxx
 2FA_ENABLED=true
```
benbaumann95, line 4: just need to delete this line

benbaumann95 (approved):

EGiataganas (approved):

PR #919, opened 2023-01-24
Allow cloning of planning applications for development / staging

benbaumann95:
### Description of change

**Allow cloning of planning applications for development / staging**
- This clone feature just replicates the POST request that PlanX makes into our API
- Extract the planning application API create action into a service so that we can repurpose this for both cloning internally and external API requests
- This is only supported on development and staging environments
- Cloning will only work on planning applications where there is an audit_log value storing the params we get from the PlanX post request. This is used to mimic the same request when cloning
- If a planning application has been processed further along within BOPS then the cloned application will not have these updates

### Story Link

https://trello.com/c/ckdkEHkj/1427-add-the-ability-to-clone-a-planning-application-staging-only

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -51,6 +51,11 @@ Lint/AmbiguousBlockAssociation:
   Exclude:
     - spec/models/review_policy_class_spec.rb
 
+Naming/VariableNumber:
+  Exclude:
+    - spec/**/*
+    - app/services/planning_application_creation_service.rb
+
 # we tend to use update_all in migrations
 Rails/SkipsModelValidations:
   Exclude:
--- a/app/views/planning_applications/show.html.erb
+++ b/app/views/planning_applications/show.html.erb
@@ -13,13 +13,21 @@
       sections: %i[key_application_dates contact_information audit_log notes]
     )
   ) %>
-  <% if @planning_application.in_progress? %>
-    <%= link_to(
-      t(".close_or_cancel"),
-      close_or_cancel_confirmation_planning_application_path(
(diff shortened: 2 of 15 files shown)


.rubocop.yml
```diff
@@ -51,6 +51,9 @@ Lint/AmbiguousBlockAssociation:
   Exclude:
     - spec/models/review_policy_class_spec.rb
 
+Naming/VariableNumber:
+  Enabled: false
```
benjamineskola, line 55: Is it necessary to disable this completely? If we don't want to use `rubocop:disable` comments, perhaps only disable it for `spec/**/*`?

app/views/planning_applications/show.html.erb
```diff
@@ -13,13 +13,19 @@
...
+          @planning_application
+        ),
+        class: "govuk-button govuk-button--warning"
+      ) %>
+    <% end %>
+
+    <% if @planning_application.can_clone? %>
+      <%= link_to "Clone", clone_planning_application_path(@planning_application), class: "govuk-button govuk-button--secondary", method: :post, data: { confirm: "This will clone the planning application identical to how it was created via PlanX. Are you sure?" } %>
+    <% end %>
+  </div>
```
EGiataganas, line 30: I am wondering whether the `<div class="govuk-button-group">` could go inside the ifs.

Now if the planning application is not in progress and you're on production you will have the `  <div class="govuk-button-group">` which will add `margin-bottom: 15px;`

EGiataganas (approved):

PR #924, opened 2023-02-01
Switch from webpack+sassc to dartsass-rails

benjamineskola:
### Description of change

Build CSS from SCSS using dartsass instead of webpack+sassc.

Both libsass (used by sassc) and the webpacker gem are deprecated; this, along with a similar PR to replace the Javascript pipeline, will allow use to replace them.

### Story Link

https://trello.com/c/xNWm3kZI/1392-move-css-and-other-assets-from-webpacker-to-asset-pipeline-on-bops

### Further testing or sign off required [OPTIONAL]

e.g.

1. product manager to sign off view templates

PR #927, opened 2023-02-02
Switch from webpacker to jsbundling-rails

benjamineskola:
### Description of change

Replaces the webpacker gem with jsbundling-rails, both using webpack as a backend. Webpacker is deprecated and this along with #924 will allow us to remove it from our workflow.

### Story Link

https://trello.com/c/xNWm3kZI/1392-move-css-and-other-assets-from-webpacker-to-asset-pipeline-on-bops

### Decisions [OPTIONAL]

It's possible also to switch from webpack to esbuild or some other build tool; I believe esbuild is the jsbundling-rails default. It might be slightly less complex, webpack has a lot of features that we're not using, because it duplicates a lot of what is done by Rails' asset pipeline. However, the tradeoff would be increased complexity in the migration as we'd need to work out the correct configuration for esbuild.

The main issue I think is that esbuild needs platform-specific binaries so building `node_modules` on macOS and simply copying it into the docker container will not work. I can do a little more investigation into this but it may simply not be worth the effort.

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


EGiataganas (approved):

benjamineskola:
I've done a little more investigation into switching from webpack itself to esbuild. The problem appears to be that esbuild depends on native packages, but because `node_modules` gets shared between the host and the container because of the bind mount in `docker-compose.yml`, the container ends up with the wrong esbuild binary and it all becomes a bit of a mess.

I think that's fixable for the dev environment but then that adds extra complexity for production docker. It might be something we want to revisit once we've finished the infrastructure migration.

PR #933, opened 2023-02-14
Style 'restore document' button as a link

benjamineskola:
### Description of change

### Story Link

https://link-to-issue

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


benbaumann95:
I see you've mentioned in the ticket that this should probably be a button as it modifies something in our backend - I agree that's more conventional so maybe is worth pushing back on this one

cdccollins (approved):

PR #936, opened 2023-02-14
Upgrade to Rails 7.0.4.2

benbaumann95:
### Description of change

- See commits for details

### Story Link

https://trello.com/c/0G7Hmtsi/1397-upgrade-from-rails-617-to-704-for-bops

Requires https://github.com/unboxed/bops-terraform/pull/96 and ansible reapplied to pick up the env variable


package.json
```diff
@@ -4,8 +4,8 @@
   "dependencies": {
     "@hotwired/stimulus": "^3.2.1",
     "@opensystemslab/map": "^0.5.7",
-    "@rails/activestorage": "^6.0.0",
-    "@rails/ujs": "^6.0.0",
+    "@rails/activestorage": "^7.0.4-2",
+    "@rails/ujs": "^7.0.4-2",
```
benjamineskola, line 8: Do we need `@rails/ujs` at all? It seems like it's the predecessor to turbo, which we also don't use. (I have no idea about `@rails/activestorage`.)

config/initializers/devise.rb
```diff
@@ -11,10 +11,17 @@
 
   config.mailer_sender = "support@unboxed.co"
 
-  config.mailer = "DeviseMailer"
+  require "application_mailer"
```
benjamineskola, line 14: Can this be done in a `to_prepare` block instead of with the explicit requires?

spec/models/document_spec.rb
```diff
@@ -83,9 +83,9 @@
 
     it "is invalid for an unpermitted file content type" do
       document.file.attach(
-        io: Rails.root.join("spec/fixtures/images/bmp.bmp").open,
-        filename: "bmp.bmp",
-        content_type: "image/bmp"
+        io: Rails.root.join("spec/fixtures/images/image.gif").open,
```
benjamineskola, line 86: I think we should also remove `spec/fixtures/images/bmp.bmp`.

config/initializers/devise.rb
benbaumann95: Would you be able to pull this branch locally and give this a go? Think I tried but maybe was missing something

benjamineskola (approved):

EGiataganas (approved):

PR #946, opened 2023-02-21
Rename constraints.yml to avoid mixup

cdccollins:
### Description of change

Rename constraints in constraints.yml to avoid confusion when listing out the constraints on the constraint edit page. 

### Story Link

https://trello.com/c/91yYQEwg/1462-review-the-list-of-constraints

### Decisions [OPTIONAL]

"Constraints" are defined in two places in the various yml files (constraints.yml and en.yml), resulting in when we iterate through the list of constraints for the constraint edit page we get some weird results - the success/update messages are listed as actual constraints.

Rename the constraint.yml file so this doesn't happen. Thought that was easier than trying to override the constraints controller to know to look for something different.

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -364,7 +364,7 @@ class PlanningApplication < ApplicationRecord
   # around it), just assume plain-text constraints rather than a
   # key-value mapping.
   def defined_constraints
-    I18n.t("constraints").values.flatten
+    I18n.t("constraint_list").values.flatten
   end
 
   def custom_constraints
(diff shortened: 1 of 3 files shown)


cdccollins:
> Does it makes sense to add any specs to avoid whatever was happening before?

Not really... I guess I could add a spec saying "page does not contain" whatever was there before, but it would be so specific.

benbaumann95:
> > Does it makes sense to add any specs to avoid whatever was happening before?
> 
> Not really... I guess I could add a spec saying "page does not contain" whatever was there before, but it would be so specific.

Ah okay yeah if the initial specs were covering then no need

PR #947, opened 2023-02-21
Add feedback fish widget

benbaumann95:
### Description of change

Embed[ feedback fish](https://feedback.fish/) widget

- Send user's email in feedback request
- Split pid id per environment so we can have a specific project for staging / production

### Story Link

https://trello.com/c/s24alR4w/1467-can-we-use-feedback-fish-on-bops-as-an-interim-solution

Screenshot 2023-02-27 at 16 31 09
Screenshot 2023-02-27 at 16 31 19

### DIFF
--- a/spec/system/feedback_fish_spec.rb
+++ b/spec/system/feedback_fish_spec.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Feedback fish" do
+  let(:local_authority) { create(:local_authority, :default) }
+  let(:assessor) { create(:user, :assessor, local_authority: local_authority) }
+
+  before do
+    sign_in assessor
+  end
+
+  context "when FEEDBACK_FISH_ID env variable is set" do
+    before do
+      ENV["FEEDBACK_FISH_ID"] = "randomid123"
+    end
+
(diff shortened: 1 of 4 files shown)


spec/system/feedback_fish_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    it "is displayed with a reference to the user's email" do
+      visit root_path
+
+      within find("a[data-feedback-fish][data-feedback-fish-userid='#{assessor.email}']") do
+        expect(page).to have_content "feedback"
+      end
+    end
+  end
```
cdccollins, line 25: I don't understand this - could it not be like the test below and just check for an email? I might be misunderstanding

cdccollins (commented):
Could you add a screenshot please?

spec/system/feedback_fish_spec.rb
benbaumann95: Attached some screenshots.

The email address is just sent as metadata so not actually testing visually for this

benbaumann95: Just want to test that the correct html is formed for this widget

cdccollins (approved):

PR #955, opened 2023-02-27
Filter all my applications

cdccollins:
### Description of change

Changing how we use the tabs on the frontend 
- One tab for now "Your applications" when looking at only your applications, another for closed applications
- You can filter on this tab based on status
- Reviewer has a different view which only has two filters

### Story Link

https://trello.com/c/DBXZ8Omv/1298-1-2-create-a-filter-for-the-all-my-application-tab-to-reduce-the-number-of-tabs-on-officer-view

### Screenshots

Screenshot 2023-03-08 at 16 11 25

Screenshot 2023-03-08 at 16 11 32

### DIFF
--- a/app/services/planning_application_filter.rb
+++ b/app/services/planning_application_filter.rb
@@ -3,14 +3,18 @@
 class PlanningApplicationFilter
   include ActiveModel::Model
 
-  attr_accessor :filter_options, :planning_applications
+  attr_accessor :filter_options, :planning_applications, :user
 
   def results
-    records_matching_query || []
+    records_matching_reference || []
   end
 
   def filter_types
     filter_options.reject(&:empty?)
+  end
+
+  def filtered_filter_types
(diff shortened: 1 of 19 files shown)


app/services/planning_application_filter.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  private
+
+  def records_matching_query
+    records_matching_reference.presence
+  end
+
+  def records_matching_reference
+    filtered_filter_types.map { |filter| planning_applications.send(filter) }.flatten
```
benbaumann95, line 27: Is this doing more queries than necessary?

Could we not load all the planning applications in memory and then filter them from this point. I think this will send a separate query for each different status

cdccollins: Yeah think this was a left over from me trying to do it another stupider way

benbaumann95 (approved):

EGiataganas (approved):

PR #956, opened 2023-02-28
WIP

benbaumann95:
### Description of change

### Story Link

https://link-to-issue

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


app/models/audit.rb
```diff
@@ -7,6 +7,13 @@ class Audit < ApplicationRecord
 
   scope :by_created_at, -> { order(created_at: :asc) }
   scope :with_user_and_api_user, -> { preload(:user, :api_user) }
+  scope :with_planning_application, -> { includes(:planning_application) }
+  scope :not_by_assigned_officer, lambda {
+    joins(:planning_application).where("audits.user_id != planning_applications.user_id OR planning_applications.user_id IS NULL")
+  }
+  scope :most_recent_for_planning_applications, lambda {
```
benbaumann95, line 14: This should just create the one query.

Essentially we want to grab the audit with the max created at date (i.e. most recent) per planning application and then group these by the id so we don't duplicate planning applications in the scope. We also only care about audits created by someone other than the assigned officer so we require the scope `not_by_assigned_officer`

app/assets/stylesheets/_main.scss
```diff
@@ -307,3 +307,7 @@ html * {
 .italic {
   font-style: italic;
 }
+
+.govuk-table__cell.bottom-border-none {
+  border-bottom: none;
+}
```
benjamineskola, line 313: This is a pretty minor point, but I'm not a huge fan of having to add a class to the cells based on the contents of the next row. How about using `:has()` to detect whether it's followed by a particular element?

```suggestion
.govuk-table__row:has(+ .audit) .govuk-table__cell {
  border-bottom: none;
}
```

And then instead of adding the `bottom-border-none` class on L3 of `row_component`, we could add an `audit` class (or whatever name) on L14 instead, inside the `if audit` conditional.

The downside is this might be a little more complex to read and browser support isn't 100% yet, so feel free to say no to this.

benbaumann95: Yeah I agree with where you're coming from. I'm actually thinking whether I just do:

```
.bottom-border-none {
  border-bottom: none;
}
```
since this css rule doesn't necessarily have to follow a `govuk-table__cell` and should be able to be reused anywhere

cdccollins (approved):

PR #957, opened 2023-03-06
Allow requesting replacement documents at assessment stage

benjamineskola:
### Description of change

Currently, replacement documents can only be requested at the validation stage. This adds UI for creating them from the assessment stage and alters the database validation to permit that.

### Story Link

https://trello.com/c/KnZEB2t8/1470-request-replacement-document-at-assessment-stage

### Screenshots

Screenshot 2023-03-06 at 16 20 37
Screenshot 2023-03-06 at 16 20 26

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

There's a couple of things that it'd be good to check with someone more familiar:

- [x]  I'm passing `post_validation: true` when creating the request, and this is being reflected in the document, but I hadn't found where this affects the UI (a list of open post-validation requests somewhere?). Presumably it does but I haven't visually confirmed this.

- [x] I need to double check whether the changes I've made to the document management views affect the UI in the validation stage as well as the assessment stage — it's possible I'm passing `post_validation: true` when it shouldn't be.

e.g.

### DIFF
--- a/spec/models/replacement_document_validation_request_spec.rb
+++ b/spec/models/replacement_document_validation_request_spec.rb
@@ -101,11 +101,11 @@ RSpec.describe ReplacementDocumentValidationRequest do
           create(:replacement_document_validation_request, planning_application: planning_application)
         end
 
-        it "prevents a replacement_document_validation_request from being created" do
+        it "allows a replacement_document_validation_request to be created" do
           expect do
             replacement_document_validation_request
-          end.to raise_error(ValidationRequestable::ValidationRequestNotCreatableError,
-                             "Cannot create Replacement Document Validation Request when planning application has been validated")
+          end.not_to raise_error(ValidationRequestable::ValidationRequestNotCreatableError,
+                                 "Cannot create Replacement Document Validation Request when planning application has been validated")
         end
       end
     end
--- a/spec/system/planning_applications/replacement_document_validation_request_spec.rb
+++ b/spec/system/planning_applications/replacement_document_validation_request_spec.rb
@@ -501,6 +501,19 @@ RSpec.describe "Requesting document changes to a planning application" do
       end
     end
 
+    it "sends an email notification" do
+      delivered_emails = ActionMailer::Base.deliveries.count
+
+      click_link "Check and assess"
+      click_button "Documents"
+      click_link "Request replacement"
+
+      fill_in "List all issues with the document.", with: "This is very invalid"
+      click_button "Send request"
+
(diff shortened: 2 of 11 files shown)


benjamineskola:
And the edit/cancel flow works as expected:
Screenshot 2023-03-07 at 14 30 54

benbaumann95:
If you actually respond as an applicant with a replacement document does this appear in the UI and what happens with the replaced document?

benjamineskola:
Uploading a replacement document works:

Screenshot 2023-03-08 at 10 17 37

Screenshot 2023-03-08 at 10 16 20

spec/models/replacement_document_validation_request_spec.rb
```diff
@@ -101,11 +101,11 @@
           create(:replacement_document_validation_request, planning_application: planning_application)
         end
 
-        it "prevents a replacement_document_validation_request from being created" do
```
benjamineskola, line 104: This means there's now no test for `ensure_planning_application_not_validated!`. Can someone suggest a good alternative example of something that should be possible only before validation? @benbaumann95 @EGiataganas

benbaumann95: Other change validation requests should not be possible after validation

benbaumann95 (commented):
We should have a system end to end test here mimicking the applicant response.
We also want to test if post_validation has been set to true when a request is made this way

spec/system/planning_applications/replacement_document_validation_request_spec.rb
```diff
@@ -466,6 +470,68 @@
...
+
+      expect(page).to have_content("Replacement document validation request successfully created.")
+      document1.reload
+
+      document2 = create(:document, :with_file, planning_application: planning_application)
+      request = planning_application.replacement_document_validation_requests.last
+      request.new_document = document2
+      request.state = "closed"
+      request.save!
+      request.old_document.archive("replaced by new document")
```
benjamineskola, line 520: I'm a little wary of this section: I'm not sure if there's a better way to generate a response, but I'm concerned that manipulating the objects directly might just bypass any issues that might occur in real usage. However, this seems to be what other tests do already, and I've verified as best I can that it's not just working by accident. Since we can't AFAIK access bops-applicants directly during these tests I think this is the best we can do.

benbaumann95 (approved):

benbaumann95 (approved):

PR #958, opened 2023-03-07
Move secure password generator to it's own class

EGiataganas:
### Description of change

Move secure password generator to it's own class


benbaumann95 (approved):

benjamineskola (approved):
Looks good! Much clearer and ensures that all of the requirements are definitely met (my implementation just relied on selecting enough characters that it would almost certainly have some from each type).

PR #968, opened 2023-03-09
Fix assessment against legislation intermittent spec

benbaumann95:
### Description of change

The intermittent failure was due to a Selenium::WebDriver::Error::UnexpectedAlertOpenError caused by navigating away from the page due to the unsaved_changes_controller where it checks if you will leave the page with any unsaved content. Without adding this rescue block, occasionally the current_path remains on the policy classes page and therefore it cannot find the link "Part 1, Class D". This might actually be a bug with Selenium as the failed screenshot suggests this shouldn't be the case

This fix just splits the test where the important thing to check is that the updated comment and previous comments are visible

### Story Link

https://trello.com/c/aFoNhyjg/1479-fix-flaky-test-for-adding-and-updating-comments

### DIFF
--- a/spec/system/planning_applications/assessment_against_legislation_spec.rb
+++ b/spec/system/planning_applications/assessment_against_legislation_spec.rb
@@ -139,7 +139,7 @@ RSpec.describe "assessment against legislation" do
       end
     end
 
-    it "lets the user add and update comments", skip: "TOFIX: intermittent failure" do
+    it "lets the user add comments" do
       travel_to(Time.zone.local(2022, 9, 1))
       click_link("Check and assess")
       add_policy_classes(["Class D - porches"])
@@ -160,11 +160,6 @@ RSpec.describe "assessment against legislation" do
 
         expect(page).not_to have_content("Previous comments")
       end
-
-      expect(row_with_content("D.1a")).to have_field(
-        "Comment added on 01 Sep 2022 by Alice Smith",
-        with: "New comment"
-      )
 
       expect(row_with_content("D.1b")).to have_field("Add comment", with: "")
 
@@ -194,30 +189,42 @@ RSpec.describe "assessment against legislation" do
           with: ""
         )
       end
+    end
 
-      within(row_with_content("D.1a")) do
-        fill_in(
-          "Comment added on 01 Sep 2022 by Alice Smith",
-          with: "Updated comment"
-        )
+    context "when user updates a comment" do
+      let(:policy_class) { create(:policy_class, planning_application:) }
+      let(:policy) { create(:policy, policy_class:) }
+      let(:comment1) { create(:comment, text: "Original comment") }
+      let(:comment2) { create(:comment, text: "Updated comment") }
+
+      before do
+        Current.user = assessor
+
+        travel_to(Time.zone.local(2022, 9, 1))
+        policy.comments << comment1
+
+        travel_to(Time.zone.local(2022, 10, 1))
+        policy.comments << comment2
       end
 
-      click_button("Save and come back later")
-      click_link("Part 1, Class D")
+      it "shows the updated comment and previous comments" do
+        click_link("Check and assess")
+        click_link("Part 1, Class A")
 
-      within(row_with_content("D.1a")) do
-        expect(page).to have_field(
-          "Comment updated on 02 Sep 2022 by Bella Jones",
-          with: "Updated comment"
-        )
+        within(row_with_content("A.1A")) do
+          expect(page).to have_field(
+            "Comment updated on 01 Oct 2022 by Alice Smith",
+            with: "Updated comment"
+          )
 
-        find("span", text: "Previous comments").click
+          find("span", text: "Previous comments").click
 
-        expect(page).to have_content(
-          "Comment added on 01 Sep 2022 by Alice Smith"
-        )
+          expect(page).to have_content(
+            "Comment added on 01 Sep 2022 by Alice Smith"
+          )
 
-        expect(page).to have_content("New comment")
+          expect(page).to have_content("Original comment")
+        end
       end
     end
 


spec/system/planning_applications/assessment_against_legislation_spec.rb
```diff
@@ -161,11 +161,6 @@
         expect(page).not_to have_content("Previous comments")
       end
 
-      expect(row_with_content("D.1a")).to have_field(
```
benbaumann95, line 164: This is duplicate to https://trello.com/c/aFoNhyjg/1479-fix-flaky-test-for-adding-and-updating-comments

spec/system/planning_applications/assessment_against_legislation_spec.rb
```diff
@@ -193,16 +188,18 @@
...
-      within(row_with_content("D.1a")) do
         fill_in(
           "Comment added on 01 Sep 2022 by Alice Smith",
           with: "Updated comment"
         )
       end
 
       click_button("Save and come back later")
+
+    rescue Selenium::WebDriver::Error::UnexpectedAlertOpenError
```
benjamineskola, line 200: I wonder if there's a different way of structuring this. If I'm understanding it correctly, there are different `expect` statements depending on whether the exception is raised or not?

Or is it that the exception will _always_ fire and so L182–188 _and_ L206–217 will have effect?

I wonder if we could limit the `rescue` block to only around the line that's expected to raise the exception, and/or use an `ensure` block so that it's clear what will happen if the exception is/isn't raised.

benbaumann95: My thinking was just to rescue this error and do nothing as it's not relevant to the test itself.

Basically navigating away from that page i.e. doing `visit planning_application_assessment_tasks_path(planning_application)` or `click_link("Part 1, Class D")` intermittently causes that error to raise when it shouldn't.  

I've spent a bit too much time on this though so if you could pull it locally and give it a whirl to your suggestion? I've tried wrapping it in a begin block but i think more than one action actually triggers this weird error so it's hard to isolate. I'm not sure it's worth the pain/effort to go deeper in a rabbit hole though.

benbaumann95: A good metric was to see if this would all be green `repeat 10 { bundle exec rspec spec/system/planning_applications/assessment_against_legislation_spec.rb:142 }`

benjamineskola: > My thinking was just to rescue this error and do nothing as it's not relevant to the test itself.

That's what confused me — it seems like actually the error is being rescued and then part of the test is done in the rescue block.

But if the exception can be thrown at multiple points then that might complicate matters.

I'd be happy to take a look at it myself this morning.

spec/system/planning_applications/assessment_against_legislation_spec.rb
```diff
@@ -193,16 +188,18 @@
...
+      visit planning_application_assessment_tasks_path(planning_application)
+
       click_link("Part 1, Class D")
```
benjamineskola, line 204:
```suggestion
      visit planning_application_assessment_tasks_path(planning_application)

      begin
        click_link("Part 1, Class D")
      rescue Selenium::WebDriver::Error::UnexpectedAlertOpenError
      end
```

I think this is the only line that strictly needs to be wrapped.

benjamineskola: Oh, having said that, I've just gotten it to fail again after it passed ten times in a row.

spec/system/planning_applications/assessment_against_legislation_spec.rb
benjamineskola: I had thought perhaps `accept_alert { … }` might get us somewhere but even when the code is wrapped in that block the exception is thrown.

spec/system/planning_applications/assessment_against_legislation_spec.rb
benbaumann95: It's an absolute 🤯 

I don't think it's worth more of our time on this but appreciate you looking into it.
Either we just drop that part of the spec and do it as a new one or just keep that ambiguous ` rescue Selenium::WebDriver::Error::UnexpectedAlertOpenError` in.

What you reckon

benbaumann95: Either way let's not give this more time since it's not a priority

benbaumann95: I'm going to extract this into a new test

cdccollins (commented):
Can you explain what the problem was and how this helps please 😂

benjamineskola:
@cdccollins The basic problem is that this test fails intermittently. It seems like that's because sometimes there's a popup opening — I think warning that the form data hasn't been saved? On a deeper level, it's not clear to me _why_ that's happening.

benbaumann95:
> Can you explain what the problem was and how this helps please 😂

The `click_link("Part 1, Class D")` (originally https://github.com/unboxed/bops/blob/main/spec/system/planning_applications/assessment_against_legislation_spec.rb#L206)

would sometimes cause a `Selenium::WebDriver::Error::UnexpectedAlertOpenError` error to raise. You can see this in the spec logs if you run it locally.

I think this is because we are navigating away from this page when there is an alert open which I think the unsaved_changes controller is the culprit. We have some code here that means you get a pop up alert if you navigate away from a page with any unsaved changes. For example, go to this url https://southwark.bops-staging.services/planning_applications/1610/policy_classes/260/edit and try adding a comment (without saving) and then leaving the page. You should see an alert pop up `Changes that you made may not be saved.`

In some instances, for whatever reason I think the test at this point thinks we are navigating away without saving these changes and therefore raises this error.

I could be completely wrong though 😄 

Anyway, the fix I've made is just to isolate showing the updated comments, which is the important bit, as a separate test so we don't run into that issue.

benjamineskola:
>I think this is because we are navigating away from this page when there is an alert open which I think the unsaved_changes controller is the culprit. We have some code here that means you get a pop up alert if you navigate away from a page with any unsaved changes.

What I don't understand is why the changes aren't being saved — either they should always be saved or never be saved.

It does seem like the test would be better off split into two, since it seems like it's testing two different but related things; but it does make me wonder if the issue could still occur in real usage if the user happens to do both of those things in a row.

benbaumann95:
> > I think this is because we are navigating away from this page when there is an alert open which I think the unsaved_changes controller is the culprit. We have some code here that means you get a pop up alert if you navigate away from a page with any unsaved changes.
> 
> What I don't understand is why the changes aren't being saved — either they should always be saved or never be saved.
> 
> It does seem like the test would be better off split into two, since it seems like it's testing two different but related things; but it does make me wonder if the issue could still occur in real usage if the user happens to do both of those things in a row.

It's been split into two which essentially tests the same thing. Just setting up in the backend rather than user interaction before checking the updated comment and previous comments. I think this is fine, we can reassess if we actually run into an issue with a user experiencing this

benjamineskola (approved):

PR #972, opened 2023-03-13
Change references to RIPA to PlanX

cdccollins:
### Description of change

Change references in the code from "RIPA" to "PlanX"

### Story Link

https://trello.com/c/fkups3td/1357-sweep-for-and-remove-references-to-ripa

### Decisions [OPTIONAL]

There was some question whether this should be submission agnostic, but might be a question for further down the line when we have more idea of the integration plans

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -21,24 +21,28 @@ class LocalAuthority < ApplicationRecord
   end
 
   def council_name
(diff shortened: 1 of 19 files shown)


benbaumann95 (approved):
The only issue I can think of is in some migration files we reference `ripa` but I think we don't define this in our seeds file anyway so hopefully shouldn't be an issue for anyone doing the initial db:setup step.

Might be worth double checking staging/prod db to see if the planx value is set properly

PR #975, opened 2023-03-14
Do not audit when saved changes are all either nil, blank or zero

benbaumann95:
### Description of change

Do not audit when saved changes are all either nil, blank or zero

- There was an issue where if we saved a nil payment amount as 0 it would create an audit with "payment amount changed from £0.00 to £0.00" which is obviously not helpful
- This change just means we ignore any non meaningful value change

### DIFF
--- a/spec/system/planning_applications/updated_tab_spec.rb
+++ b/spec/system/planning_applications/updated_tab_spec.rb
@@ -79,6 +79,31 @@ RSpec.describe "Planning application updated tab spec" do
     # Now perform an update to a planning application
     visit edit_planning_application_path(planning_application2)
 
+    within(find(:fieldset, text: "Agent information")) do
+      fill_in("Email address", with: "new_agent_email@example.com")
+    end
+
+    click_button("Save")
+    visit(root_path)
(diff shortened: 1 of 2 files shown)


spec/system/planning_applications/updated_tab_spec.rb
```diff
@@ -79,6 +79,31 @@
...
+    end
+
+    click_button("Save")
+    visit(root_path)
+    click_link "Updated"
+
+    last_audit = Audit.last
+    within("#audit_#{last_audit.id}") do
+      within(".govuk-inset-text") do
+        expect(page).to have_content(
```
benbaumann95, line 93: This part of the test will fail without the `next if saved_change_to_attribute(attribute_name).all? { |value| value.blank? || value.try(:zero?) }` changes

benbaumann95: It will expect the audit entry `"payment amount changed from £0.00 to £0.00"` instead which is what this PR is trying to get eliminate

cdccollins (approved):

PR #980, opened 2023-03-16
Label application types with their work status

benjamineskola:
### Description of change

- Add the `work_status` (proposed, existing) to the "application type" column.
- Abbreviate the type but show the full name in a tooltip.
- Default to full name of type if no abbreviation defined.

### Story Link

https://trello.com/c/LRheh9Cn/1176-change-how-application-type-is-labelled-on-dashboard

### Screenshots

Screenshot 2023-03-16 at 15 15 11

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.

### DIFF
--- a/app/presenters/planning_application_presenter.rb
+++ b/app/presenters/planning_application_presenter.rb
@@ -9,6 +9,7 @@ class PlanningApplicationPresenter
   include ProposalDetailsPresenter
   include ValidationTasksPresenter
   include AssessmentTasksPresenter
+  include ActionView::Helpers::SanitizeHelper
 
   def initialize(template, planning_application)
     @template = template
@@ -21,6 +22,18 @@ class PlanningApplicationPresenter
 
   def application_type_name
     I18n.t("application_types.#{application_type}")
+  end
+
+  def application_type_abbreviation
+    I18n.t("application_types.#{application_type}_abbr", default: application_type_name)
+  end
+
+  def application_type_with_status
+    status_title = work_status.titlecase
+    sanitize(
+      "<abbr title=\"#{application_type_name} #{status_title}\">" \
+      "#{application_type_abbreviation} #{status_title}</abbr>"
+    )
   end
 
   %i[awaiting_determination_at expiry_date outcome_date].each do |date|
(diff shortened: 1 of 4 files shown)


app/presenters/planning_application_presenter.rb
```diff
@@ -23,6 +23,16 @@ def application_type_name
...
   end
 
+  def application_type_abbreviation
+    I18n.t("application_types.#{application_type}_abbr", default: application_type_name)
+  end
+
+  def application_type_with_status
+    status_title = work_status.titlecase
+    "<abbr title=\"#{application_type_name} #{status_title}\">" \
+    "#{application_type_abbreviation} #{status_title}</abbr>".html_safe # rubocop:disable Rails/OutputSafety
```
benbaumann95, line 33: Can we use `sanitize` instead here?

I tend to try and avoid `html_safe` if possible. 
You might need to `include ActionView::Helpers::SanitizeHelper` if the method isn't available.

benjamineskola: Perhaps I've misunderstood — I thought `html_safe` was necessary in order for the returned string to be displayed as HTML. If not `html_safe` then the HTML in this string would be escaped.

`sanitize` does almost the opposite, removing untrusted HTML from the string. Wouldn't it still need to be marked as safe in order to be rendered correctly?

And I do agree that it's good to avoid using `html_safe` — but in this case, everything that goes into making up the HTML is from a trusted source, I believe. And, because of the way the table view component simply loops over a list of the model's attributes, it's difficult to treat just one of the attributes differently at the view level. It would mean an awkward special case was needed to add the abbreviation here.

spec/presenters/planning_application_presenter_spec.rb
```diff
@@ -253,4 +253,24 @@
...
+    let(:presenter) { described_class.new(view, planning_application) }
+
+    context "when the status is proposed" do
+      it "is reflected in the type" do
+        expect(presenter.application_type_with_status).to include("LDC Proposed")
+        expect(presenter.application_type_with_status).to include("Lawful Development Certificate Proposed")
+      end
+    end
+
+    context "when the status is in progress" do
```
benbaumann95, line 268: tiny nitpick/typo
```suggestion
    context "when the work status is existing" do
```

benbaumann95 (approved):
Could you please hold off merging this until after the deploy to prod first thing tomorrow 🙏 
Thanks!

PR #981, opened 2023-03-21
Copy ruby version to Dockerfile.production

benbaumann95:


Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN apt-get install -y --no-install-recommends poppler-utils
 
 ## Install gems in a separate Docker fs layer
 WORKDIR /gems
-COPY Gemfile Gemfile.lock ./ .ruby-version ./
```
EGiataganas, line 17: COPY Gemfile Gemfile.lock .ruby-version /app/

PR #984, opened 2023-03-23
Show a warning when constraints have changed

benjamineskola:
### Description of change

Currently a warning is displayed if certain fields change, notifying the user that it may no longer be valid.

This extends that to include the constraints field.

### Story Link

https://trello.com/c/5SgiA53g/1169-show-application-has-been-updated-warning-in-accordion-after-any-constraints-have-been-changed

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_04_12_102829) do
+ActiveRecord::Schema[7.0].define(version: 2023_04_13_104408) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
@@ -340,6 +340,7 @@ ActiveRecord::Schema[7.0].define(version: 2023_04_12_102829) do
     t.datetime "received_at"
     t.boolean "from_production", default: false
     t.string "review_documents_for_recommendation_status", default: "not_started", null: false
+    t.text "changed_constraints", array: true
     t.index "lower((reference)::text)", name: "ix_planning_applications_on_lower_reference"
     t.index "to_tsvector('english'::regconfig, description)", name: "index_planning_applications_on_description", using: :gin
     t.index ["api_user_id"], name: "ix_planning_applications_on_api_user_id"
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -19,7 +19,7 @@
     t.string "record_type", null: false
     t.bigint "record_id", null: false
     t.bigint "blob_id", null: false
-    t.datetime "created_at", null: false
+    t.datetime "created_at", precision: nil, null: false
```
cdccollins, line 22: A load of these kind of things have changed in the schema, missing precisions etc

benjamineskola: Oh hang on, I saw that had changed (no idea why) but I thought I'd excluded it from the commit. Will fix.

db/migrate/20230323110408_add_changed_constraints_to_planning_application.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddChangedConstraintsToPlanningApplication < ActiveRecord::Migration[7.0]
+  def change
+    add_column :planning_applications, :changed_constraints, :text, array: true
```
cdccollins, line 5: Can you explain why you've done it this way? I've not seen this kind of pattern before and it looks a bit strange to me to have this stored doubly, but I'm not sure!

benjamineskola: I based this on how `updated_address_or_boundary_geojson` is tracked and stored. In that case, a boolean is set if certain fields are changed; in this case I'm storing an array of the filenames — currently however we only care whether it's empty or not.

So I don't think it's storing doubly (unless I've misunderstood you) — at the moment, we store the new values when an application is modified, but we don't track the fact that they're different from the original values. This will track which values are updated, but not what they've changed from or to.

If we wanted to go a bit further, something like papertrail would be a lot more powerful, but that feels like overkill for now.

benbaumann95: Maybe a spec would help illustrate this?

i.e. make an update to constraints and then check what `planning_application.changed_constraints` is `[....]`. Then another update and check again etc.

cdccollins (approved):

PR #986, opened 2023-03-23
Show correct source of update in audits

benbaumann95:
### Description of change

We were incorrectly displaying that an update from BoPS applicants was via PlanX

### Screenshots

Screenshot 2023-03-23 at 16 22 41

### DIFF
--- a/app/helpers/audit_helper.rb
+++ b/app/helpers/audit_helper.rb
@@ -112,7 +112,9 @@ module AuditHelper
   end
 
   def audit_user_name(audit)
-    if audit.api_user.present?
+    if applicant_activity_types.include?(audit.activity_type)
+      t("audit_user_name.bops_applicants")
+    elsif audit.api_user.present?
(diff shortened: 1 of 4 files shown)


app/helpers/audit_helper.rb
```diff
@@ -112,7 +112,9 @@ def activity(type_of_activity, args = nil)
   end
 
   def audit_user_name(audit)
-    if audit.api_user.present?
+    if applicant_activity_types.include?(audit.activity_type)
+      t("audit_user_name.bops_applicants")
+    elsif audit.api_user.present?
```
cdccollins, line 117: This whole file is kind of wild with all the if/else statements. Feels like there's maybe a better way to be doing it, not that I've thought about it that much.

benbaumann95: There definitely is but just wanted to get a fix out without spending too much time on this

EGiataganas (approved):

PR #998, opened 2023-03-31
Dockerfile which will be in rails 7.1

EGiataganas:
### Description of change

Dockerfile which will be in rails 7.1


benjamineskola:
What would it take to get this unblocked? Does it just need rebasing? Can we pair on it sometime?

EGiataganas:
@benjamineskola we don't progress as we didn't had capacity, happy to book some time to pair and try out 😄

config/application.rb
```diff
@@ -25,6 +25,13 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 7.0
 
+    # New Rails 7.1 default
+    config.secret_key_base = if ENV["SECRET_KEY_BASE_DUMMY"]
```
benjamineskola, line 29: I think this is a definite improvement but I think it might also lead to `SECRET_KEY_BASE` being checked for even more frequently, which then causes rake tasks to fail unless it or the dummy is set. (See the changes to the GH Workflows I made in this PR.)

I wonder if we can improve it further? Maybe only check it for production/staging for example. We definitely need to ensure it's never blank in production though. Perhaps we can just leave it for now and try to improve it later if it turns out to be really annoying.

PR #1001, opened 2023-04-03
Post prod applications to staging

cdccollins:
### Description of change

When an application comes into production from Plan X, post it to the staging environment for testing purposes

### Story Link

https://trello.com/c/GC9MeMg6/1428-post-production-data-as-it-arrives-from-planx-into-our-staging-api

Screenshot 2023-04-12 at 11 56 16

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,6 @@ gem "faker", require: false
 gem "faraday", require: false
 gem "govuk_design_system_formbuilder"
 gem "grover"
-gem "httparty"
 gem "i18n", "< 1.9"
 gem "i18n-tasks"
 gem "image_processing", "~> 1.12"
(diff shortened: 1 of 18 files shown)


Gemfile
```diff
@@ -19,6 +19,7 @@ gem "faker", require: false
 gem "faraday", require: false
 gem "govuk_design_system_formbuilder"
 gem "grover"
+gem "httparty"
```
EGiataganas: we already have [faraday](https://github.com/unboxed/bops/blob/main/Gemfile#L19) in the project. How about trying to use Faraday to make the `POST` request?

benbaumann95 (approved):
woo this looks good 🎉 

Did we want to distinguish whether a planning application has originated from production? Or ping us a notification etc when this has been triggered. Just for future proofing but it's not a blocker

cdccollins:
> woo this looks good 🎉
> 
> Did we want to distinguish whether a planning application has originated from production? This could be in the sense of just a boolean flag we add on the db Or ping us a notification etc when this has been triggered. Just for future proofing but it's not a blocker

Oh yes forgot about that. Probably a good idea!

PR #1010, opened 2023-04-05
Use output command

EGiataganas:
### Description of change

Use output command


.github/workflows/deploy-environment.yml
```diff
@@ -123,11 +123,19 @@ jobs:
           aws ecs describe-task-definition --task-definition bops-${{ matrix.service_type }}-${{ inputs.environment-name }} --query taskDefinition | \
           jq -r 'del(.compatibilities, .taskDefinitionArn, .requiresAttributes, .revision, .status, .registeredAt, .registeredBy)' > ${{ matrix.service_type }}.json
 
-      - name: Generate task definition for ${{ matrix.service_type }}
-        id: steps[format('task-def-{0}', matrix.service_type)]
+      - name: Generate task definition for console
+        id: task-def-console
```
benjamineskola, line 127: Although, actually, I wonder if we're overthinking this.

If the task is run twice, once for each item in the matrix, but the id does not have a variable in it, doesn't GHA have a way of distinguishing between those?

Let's go with this PR for now, and I'll read up more on matrix builds and dependencies and we can see about improving it afterwards.

PR #1026, opened 2023-04-11
Apply autocorrectable Rubocop changes from TODO

benjamineskola:
### Description of change

Apply autocorrectable changes currently ignored in rubocop_todo.yml, as well as some one-off exclusions in rubocop.yml — either fixed or switched to using directives in comments rather than ignoring the whole file.


app/helpers/audit_helper.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-module AuditHelper
-  def activity(type_of_activity, args = nil)
+module AuditHelper # rubocop:disable Metrics/ModuleLength
+  def activity(type_of_activity, args = nil) # rubocop:disable Metrics/CyclomaticComplexity
```
cdccollins, line 4: I guess it's good having it in the code so we know that there's a problem, but would also be good to think about how we might fix it 😂

cdccollins: Though this is def a step towards that!

cdccollins: Sorry, one final thing, do you need to enable it again or is it just on a file by file basis?

benjamineskola: Yeah, my thinking in these changes is that `.rubocop_todo.yml` is a bit too out-of-sight, out-of-mind and if we disable things in there we may never think about it again. Especially since that disables it for the whole file, not just individual lines/methods that are an issue.

The `disable` comments apply for the current line if they're at the end of the line, or they apply to a range of lines if they're on their own (and then they need an `enable` comment later. I usually prefer the single-line form, but in a couple of places here I've used the multiline form, because the single-line form got flagged as unnecessary, then when it was removed it showed the error I was trying to disable again — thanks rubocop!

benjamineskola:
I also wonder if, for some of the checks that we've disabled a larger number of times, whether we should just decide that we don't care about it in the first place. `Layout/LineLength`, for example.

benbaumann95:
> I also wonder if, for some of the checks that we've disabled a larger number of times, whether we should just decide that we don't care about it in the first place. `Layout/LineLength`, for example.

I'm definitely in favour of getting rid of that cop. A pain when you want to work with a long string.

benjamineskola:
> A pain when you want to work with a long string.

Yeah it'd be nice if you could configure it, for example, to ignore very long strings. Actually, there is an option, I'll see if I can find a regex that works for it.

cdccollins (approved):

PR #1032, opened 2023-04-12
Update bops-applicants git submodule ref

cdccollins:


benjamineskola:
If I try to rebase or anything it insists that the commit is already in `main`, which it's obviously not. Putting it down to submodules just being fundamentally weird.

benbaumann95 (commented):
Don't know what i'm reviewing here as no files are being changed

benjamineskola:
Looks like it was also changed in 59ec55d1

PR #1038, opened 2023-04-14
Rename permitted dev to immunity/permitted dev if app is immune

cdccollins:
### Description of change

Change relevant titles etc to Immunity/permitted development rights on the permitted development rights page if the planning application is immune

### Story Link

https://trello.com/c/BKImPjtw/1541-change-title-of-permitted-development-rights-to-immunity-permitted-development-rights

### DIFF
--- a/app/controllers/planning_application/review_permitted_development_rights_controller.rb
+++ b/app/controllers/planning_application/review_permitted_development_rights_controller.rb
@@ -10,6 +10,8 @@ class PlanningApplication
     before_action :ensure_planning_application_is_validated
     before_action :set_permitted_development_right, only: %i[show edit update]
     before_action :set_permitted_development_rights, only: %i[show edit]
+    before_action :set_page_title, only: %i[show edit]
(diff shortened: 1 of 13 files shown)


app/controllers/planning_application/review_permitted_development_rights_controller.rb
```diff
@@ -50,5 +52,23 @@ def permitted_development_right_params
...
+    end
+
+    def set_page_heading
+      @page_heading =
+        if @planning_application.possibly_immune?
+          "Review immunity/permitted development rights"
+        else
+          "Check permitted development rights"
+        end
+    end
```
benbaumann95, line 72: Could we not just put this in a helper instead, why instantiate variables ?

cdccollins: I guess the only thing about using a helper instead is it will have to calculate it three times on the page rather than once in the controller. But I can see the benefit of not having loads of logic in hte controller

benbaumann95 (approved):

PR #1040, opened 2023-04-17
Allow reviewer to assess Evidence of immunity

cdccollins:
### Description of change

Allow reviewers to accept evidence of immunity or return it to officer with comment

### Story Link

https://trello.com/c/ls77xVBv/1540-allow-reviewers-to-accept-or-return-to-officer-with-comment-on-the-review-evidence-of-immunity-page

### DIFF
--- a/app/controllers/planning_application/review_immunity_details_controller.rb
+++ b/app/controllers/planning_application/review_immunity_details_controller.rb
@@ -7,6 +7,7 @@ class PlanningApplication
 
     before_action :set_planning_application
     before_action :ensure_planning_application_is_validated
+    before_action :ensure_user_is_reviewer
     before_action :set_immunity_detail, only: %i[show edit update]
     before_action :set_review_immunity_detail, only: %i[show edit update]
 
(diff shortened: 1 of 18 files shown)


app/controllers/planning_application/review_immunity_details_controller.rb
```diff
@@ -22,35 +23,50 @@ def edit
...
     def update
-      @immunity_detail.assign_attributes(
-        review_status: status
-      )
+      @immunity_detail.assign_attributes(review_status:)
 
       respond_to do |format|
-        if @immunity_detail.update(immunity_detail_params)
+        if @immunity_detail.update(status: immunity_detail_status) &&
+           @review_immunity_detail.update(review_immunity_detail_params)
```
benbaumann95, line 30: Does this need to be in a transaction?
I guess the first update could still happen and be persisted

app/controllers/planning_application/review_immunity_details_controller.rb
```diff
@@ -22,35 +23,50 @@ def edit
     end
 
     def update
```
benbaumann95, line 25: Thinking we could add a `before_action` to check that the current user is a reviewer. Definitely don't want an assessor being able to review and approve their own assessment.

benbaumann95 (commented):
Lookin good.
I think we might need to coordinate on this one to get the assessor part in at the same time https://github.com/unboxed/bops/pull/1049

benbaumann95 (approved):

PR #1042, opened 2023-04-18
Add ReviewImmunityDetails

benbaumann95:
### Description of change
Add ReviewImmunityDetails table

- Create immunity_detail has_many review_immunity_details association so we can capture any decision and reasons/comments made by both assessor and reviewer. We have a has_many relation so we can keep track of historical exchanges and display this in the frontend if an assessor needs to create a new decision and reason again.


cdccollins (approved):
LGTM!
 
Even if it's not perfect, will do for now 👍

PR #1048, opened 2023-04-24
Add document evidence to immunity page

cdccollins:
### Description of change

Add accordion of document evidence to immunity page 

### Story Link

https://trello.com/c/LFI7E28g/1532-add-list-of-documents-to-evidence-of-immunity-page-assessment

### Screenshots
screencapture-southwark-southwark-localhost-3000-planning-applications-65-immunity-details-1-edit-2023-04-24-15_39_49

### DIFF
--- a/app/controllers/planning_application/immunity_details_controller.rb
+++ b/app/controllers/planning_application/immunity_details_controller.rb
@@ -62,8 +62,18 @@ class PlanningApplication
 
     def immunity_details_params
       params.require(:immunity_detail)
-        .permit(evidence_groups_attributes: [:id, :start_date, :end_date, :missing_evidence, :missing_evidence_entry, { comments_attributes: [:text] }])
-        .merge(status:)
+            .permit(
+              evidence_groups_attributes:
+                [
+                  :id,
+                  :start_date,
+                  :end_date,
+                  :missing_evidence,
+                  :missing_evidence_entry,
+                  { comments_attributes: [:text] }
+                ]
+            )
+            .merge(status:)
     end
 
     def status
--- a/app/controllers/planning_application/immunity_details_controller.rb
+++ b/app/controllers/planning_application/immunity_details_controller.rb
@@ -29,7 +29,7 @@ class PlanningApplication
 
     def update
       respond_to do |format|
-        if @immunity_detail.update!(immunity_details_params)
+        if @immunity_detail.update(immunity_details_params)
           format.html do
             redirect_to planning_application_assessment_tasks_path(@planning_application),
                         notice: I18n.t("assessment_details.immunity_detail_successfully_updated")
--- a/app/controllers/planning_application/immunity_details_controller.rb
+++ b/app/controllers/planning_application/immunity_details_controller.rb
@@ -29,7 +29,7 @@ class PlanningApplication
 
     def update
       respond_to do |format|
-        if @immunity_detail.update(immunity_details_params)
+        if @immunity_detail.update!(immunity_details_params)
           format.html do
             redirect_to planning_application_assessment_tasks_path(@planning_application),
                         notice: I18n.t("assessment_details.immunity_detail_successfully_updated")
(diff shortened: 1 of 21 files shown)


app/services/immunity_details_creation_service.rb
```diff
@@ -22,4 +23,27 @@ def call
...
+    @planning_application.immunity_detail.evidence_groups.each do |eg|
+      Document::EVIDENCE_QUESTIONS[eg.tag.to_sym].each do |question|
+        if question.include? "show"
+          eg.applicant_comment = @planning_application.find_proposal_detail(question).first.response_values.first
+        elsif (question.include?("start") || question.include?("issued"))
+          eg.start_date = @planning_application.find_proposal_detail(question).first.response_values.first
+        elsif question.include? "run"
+          eg.end_date = @planning_application.find_proposal_detail(question).first.response_values.first
+        end
+        eg.save
```
benbaumann95, line 45: Do we wanna use `save!` here since we want to raise an error instead of silently failing

app/controllers/planning_application/immunity_details_controller.rb
```diff
@@ -29,7 +29,7 @@ def edit
 
     def update
       respond_to do |format|
-        if @immunity_detail.update(immunity_details_params)
+        if @immunity_detail.update!(immunity_details_params)
```
benbaumann95, line 32: I think this means we won't reach the `else` block as it'll either raise an error or evaluate as `nil` if nothing is raised (even if the update is successful).

So probs need to rescue any error if we wanted to present the edit page again

app/views/shared/_submit_buttons.html.erb
```diff
@@ -1,15 +1,33 @@
...
+      class: "govuk-button",
+      data: { module: "govuk-button" },
+      disabled: local_assigns.fetch(:disabled, false)
+    ) %>
+    <%= form.submit(
+      t("form_actions.save_and_come_back_later"),
+      class: "govuk-button govuk-button--secondary",
+      data: { module: "govuk-button" },
+      disabled: local_assigns.fetch(:disabled, false),
+      form: id
```
EGiataganas, line 29: is this something missed?

benbaumann95 (commented):
Can add this to the next PR, but we probably want to validate that the end date (ends until) cannot be before the start date (starts from)

cdccollins:
> Can add this to the next PR, but we probably want to validate that the end date (ends until) cannot be before the start date (starts from)

I think this might be a little bit tricky. I don't think PlanX does any of that validation so it could come through to us invalid. And then any time the planning officer saved the page it would be invalid which might be annoying if they don't want to check that at that moment? Are there other places where we're basically validating planx data?

app/views/shared/_submit_buttons.html.erb
```diff
@@ -1,15 +1,32 @@
...
+      data: { module: "govuk-button" },
+      disabled: local_assigns.fetch(:disabled, false)
+    ) %>
+    <%= form.submit(
+      t("form_actions.save_and_come_back_later"),
+      class: "govuk-button govuk-button--secondary",
+      data: { module: "govuk-button" },
+      disabled: local_assigns.fetch(:disabled, false)
+    ) %>
+  <% end %>
```
EGiataganas, line 30: we can't do the following:
```ruby
<%= form.submit(
    t("form_actions.save_and_mark_as_complete"),
    class: "govuk-button",
    data: { module: "govuk-button" },
    disabled: local_assigns.fetch(:disabled, false),
    form: local_assigns.fetch(:id, nil)
  ) %>
  <%= form.submit(
    t("form_actions.save_and_come_back_later"),
    class: "govuk-button govuk-button--secondary",
    data: { module: "govuk-button" },
    disabled: local_assigns.fetch(:disabled, false),
    form: local_assigns.fetch(:id, nil)
  ) %>
  ```
  Why do we need the `id`?

benbaumann95:
> > Can add this to the next PR, but we probably want to validate that the end date (ends until) cannot be before the start date (starts from)
> 
> I think this might be a little bit tricky. I don't think PlanX does any of that validation so it could come through to us invalid. And then any time the planning officer saved the page it would be invalid which might be annoying if they don't want to check that at that moment? Are there other places where we're basically validating planx data?

Ah good point

benbaumann95 (approved):

PR #1049, opened 2023-04-24
Allow assessor to make a decision whether an application is immune from enforcement

benbaumann95:
### Description of change

Allow assessor to make a decision whether an application is immune from enforcement

- If decision is "Yes" then a decision reason and summary must be provided but no permitted development right response is required
- If decision is "No" then only a reason is required but the assessor must respond also to the permitted development rights

### Story Link

https://trello.com/c/DOzhSrrk/1543-allow-assessor-to-decide-whether-application-is-immune-or-not

### Screenshots

Screenshot 2023-04-24 at 17 38 26
Screenshot 2023-04-24 at 17 39 02
Screenshot 2023-04-24 at 17 58 03

### Decisions [OPTIONAL]

The way this has been implemented is a bit unideal: 

- This feature probably should have split the update of review immunity detail and permitted development rights into two stages rather than using conditional form rendering (and JS to hide/display fields). I've created a form object to handle the updates to both models and tried to cover as best as possible with tests.
- If this becomes an issue, we can revisit this functionality and try to split the steps into two different forms which may be preferable.

### DIFF
--- a/app/controllers/planning_application/permitted_development_rights_controller.rb
+++ b/app/controllers/planning_application/permitted_development_rights_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class PlanningApplication
-  class PermittedDevelopmentRightsController < AuthenticationController # rubocop:disable Metrics/ClassLength
+  class PermittedDevelopmentRightsController < AuthenticationController
     include CommitMatchable
     include PlanningApplicationAssessable
     include PermittedDevelopmentRights
--- a/app/controllers/planning_application/permitted_development_rights_controller.rb
+++ b/app/controllers/planning_application/permitted_development_rights_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class PlanningApplication
-  class PermittedDevelopmentRightsController < AuthenticationController
+  class PermittedDevelopmentRightsController < AuthenticationController # rubocop:disable Metrics/ClassLength
     include CommitMatchable
     include PlanningApplicationAssessable
     include PermittedDevelopmentRights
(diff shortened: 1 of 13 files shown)


app/controllers/planning_application/permitted_development_rights_controller.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class PlanningApplication
-  class PermittedDevelopmentRightsController < AuthenticationController
+  class PermittedDevelopmentRightsController < AuthenticationController # rubocop:disable Metrics/ClassLength
```
benjamineskola, line 4: I wonder if we should just remove this check (and related ones) — it seems like we're just going to disable it whenever it crops up. (And if I'm honest that seems like the most reasonable response anyway.)

app/models/review_immunity_detail.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ReviewImmunityDetail < ApplicationRecord
-  DECISIONS = %w[yes no].freeze
+  DECISIONS = %w[Yes No].freeze
```
benjamineskola, line 4: Very minor, but is it possible for these to be symbols? It might not be worth the effort to convert them, but it seems like they're internal identifiers rather than user-visible text, which would make symbols the technically-correct option I think.

Having to convert the case in order to check seems slightly suboptimal.

benjamineskola (approved):

app/controllers/planning_application/permitted_development_rights_controller.rb
benbaumann95: Yeah agreed, will add a commit to remove it

EGiataganas, line 4: I’m wondering whether is worth of moving the `immunity  details` on their own controller

benbaumann95: I've created a new controller in my next PR as I'll need another show view so it makes sense 👍

PR #1051, opened 2023-04-26
Show evidence groups on Review Evidence of Immunity

cdccollins:
### Description of change

- Show the evidence group accordion on the Review Evidence of Immunity page 
- Reviewers can "edit" assessor's comments, same as policy classes. "Edit" actually means create a new one

### Story Link

https://trello.com/c/Dziq7qQv/1538-add-list-of-documents-to-evidence-of-immunity-page-reviewer

### Screenshots
screencapture-southwark-southwark-localhost-3000-planning-applications-82-review-immunity-details-2-edit-2023-04-26-15_34_28


app/controllers/comments_controller.rb
```diff
@@ -1,28 +1,50 @@
...
         policy_class:,
         policy:,
-        comment: policy.comment,
+        comment: @comment_type.comment,
+        new_comment:
+      }
+    )
+  end
+
+  def create_evidence_group_comment_partial(new_comment)
```
benbaumann95, line 41: It's probably not a bug introduced by this PR, but when i'm editing a comment (i.e. creating a new comment), the new comment doesn't immediately appear in the previous comments list so it's hard to know whether the update was successful or not. I have to refresh the page to see the new records. I'm not sure if this is because the collection is being memoized or that the rendering json only updates the comment partial view rather than the previous comments too. We probs want to pass it in the updated list so we can see the update straight away.

cdccollins: Yes I noticed this and then was lazy and said that's how it works currently so I don't want to fix it 😂 
I will now try and fix it 😬

benbaumann95: Ah ok well maybe it is out of scope for this PR then.
Let's add a bug ticket for it perhaps as it might not be so straightforward and shouldn't be a blocker

cdccollins: It looks like there aren't that many dev tickets coming up, so may have time before everything's UAT-ed to have a look for this PR

cdccollins: Having done an initial think/chat I think this is way too complicated for any gain we might get for now

EGiataganas (approved):

PR #1052, opened 2023-04-26
Add immunity data for planning application

EGiataganas:
### Description of change

Add immunity data for planning application


cdccollins (approved):
You could add in even more evidence types if we want even more data to come through? But this looks great, thanks!

PR #1057, opened 2023-04-28
Show evidence groups on permitted development rights page (assessor)

benjamineskola:
### Description of change

Add a list of documents and their dates to permitted development rights page.

### Story Link

https://trello.com/c/AsMtAw87/1542-include-immunity-information-in-the-permitted-development-rights-page

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


app/models/immunity_detail.rb
```diff
@@ -49,4 +48,30 @@ def accepted?
...
+    group = evidence_groups.order(end_date: :asc, start_date: :asc).last
+    group.end_date || group.start_date
+  end
+
+  def evidence_gaps? # rubocop:disable Metrics/AbcSize
+    return if evidence_groups.blank?
+
+    # https://stackoverflow.com/a/8765467
+    date_ranges = evidence_groups.map do |g|
+      Range.new(g.start_date.to_datetime.to_date, g.end_date.to_datetime.to_date)
```
cdccollins, line 66: I've just pulled this locally, and it's occasionally blowing up if there isn't a start date or an end date present, which is possible

cdccollins: (think this also applies in other places in the code where it's assuming they are both present)

cdccollins: Sorry, one more comment. I had to do something like this in my PR, can't you just do 
```    
evidence_groups.any?(&:missing_evidence)
```
because the assessor marks whether there are any missing gaps, I don't think the code has to be clever enough to work it out

benjamineskola: Ah that's a good point, I'm not sure how this should account for missing dates. Exclude them? Or set the end date to today?

cdccollins: So turns out our cards cover similar things... 😬 I did 
```
      <% if immunity_detail.earliest_evidence_cover.present? && immunity_detail.latest_evidence_cover.present? %>
        from <%= immunity_detail.earliest_evidence_cover %> to <%= immunity_detail.latest_evidence_cover %>
      <% else %>
        <%= immunity_detail.earliest_evidence_cover || immunity_detail.latest_evidence_cover %>
      <% end %>
```
but not sure if that covers it all

benjamineskola: > `evidence_groups.any?(&:missing_evidence)`

I'm not sure if that would be enough — it seems to me that it's possible for all of the evidence groups to be complete for the range each one covers separately, _but_ for there to be too few evidence groups to cover the total range. But if we think that's sufficient then we should definitely go with the simpler code.

app/views/planning_application/permitted_development_rights/_immunity_details.erb
```diff
@@ -7,17 +7,26 @@
...
+     </li>
+    <li><strong>Missing evidence (gap in time):</strong> <%= planning_application.immunity_detail.evidence_gaps? ? "Yes" : "No" %> </li>
   </ul>
+
+  <div class="govuk-body">
+    <hr>
+    <% for evidence_group in @evidence_groups %>
+    <h3>
+      <%= link_to sanitize("#{evidence_group.name} (#{evidence_group.documents.count})<br>#{evidence_group.date_range}"),
+                  new_planning_application_immunity_detail_path(@planning_application) + "#accordion-default-heading-#{evidence_group.id}",
```
cdccollins, line 26: Another thing I noticed was this links to the section, but doesn't open the accordion. I think it's supposed to be open - what I think Raffaella meant by "link to page/open tab"

benjamineskola: I'll try to dig into it and see if we can open the accordions by default.

app/models/immunity_detail.rb
```diff
@@ -49,4 +48,19 @@ def accepted?
   def current_review_immunity_detail
     review_immunity_details.last
   end
+
+  def start_date
+    evidence_groups.order(start_date: :asc).first&.start_date
+  end
+
+  def end_date
```
cdccollins, line 56: NB immunity detail already has an end_date column defined

app/views/planning_application/permitted_development_rights/_immunity_details.erb
```diff
@@ -7,17 +7,32 @@
...
-        <%= proposal_detail.question %> 
-        <% proposal_detail.response_values.each do |response_value| %>
-          <strong>
-            <%= response_value.to_time.nil? ? response_value : response_value.to_time.to_formatted_s(:day_month_year_slashes) %>
-          </strong>
+    <li><strong>Evidence cover:</strong>
+      <% if  planning_application.immunity_detail.start_date.present? %>
+        <%= planning_application.immunity_detail.start_date.to_fs(:day_month_year_slashes) %>
+        <% if  planning_application.immunity_detail.end_date.present? %>
+          to <%= planning_application.immunity_detail.end_date.to_fs(:day_month_year_slashes) %>
```
cdccollins, line 18: Sorry, I think my last comment about end dates might have confused this. I think we still need to find the latest date that the evidence covers. The end date on the immunity detail is when the applicant claims that the work was finished.

benjamineskola: Oh I see, yeah — it's that the names conflict, not that the method isn't needed. (Now I realise why you'd used different names.)

app/models/immunity_detail.rb
```diff
@@ -49,4 +48,19 @@ def accepted?
...
     review_immunity_details.last
   end
+
+  def earliest_evidence_cover
+    evidence_groups.order(start_date: :asc).first&.start_date
+  end
+
+  def latest_evidence_cover
+    group = evidence_groups.order(end_date: :asc, start_date: :asc).last
+    group&.end_date || group&.start_date
```
cdccollins, line 58: When I wrote some specs for this method (having copied it off you) I found there were some cases where it wasn't working. It's something like if one evidence group has a nil end date then that one gets returned, even if the start date is earlier. 

Might be worth either copying what I did or copying over my unit tests and seeing if you can come up with a better way than me.

app/controllers/planning_application/permitted_development_rights_controller.rb
```diff
@@ -73,6 +73,9 @@ def set_planning_application
       planning_application = planning_applications_scope.find(planning_application_id)
 
       @planning_application = PlanningApplicationPresenter.new(view_context, planning_application)
+      @evidence_groups = planning_application&.immunity_detail&.evidence_groups&.map do |group|
+        EvidenceGroupPresenter.new(view_context, group)
+      end
```
cdccollins, line 78: Wondering if this might be nicer in it's own method, kind of like `set_review_immunity_detail_and_form` that's in this file. Seems a bit cleaner and if you return if it's not immune then we don't need all the "&"s

app/models/immunity_detail.rb
benjamineskola: Ah that makes sense, because `NULL` will come last when ordering. Should be simple to fix, but I'll copy your unit tests to be sure.

benjamineskola: >Should be simple to fix

Famous last words. I hadn't realised that a start date that comes after the last end date should be returned instead. I can't figure out how to do that with just an ActiveRecord query so it seems like your way is the best.

cdccollins (approved):
I think when @benbaumann95 's PR goes in you might need to do some re-jigging as he's adding a show and edit page which will need this element too. And he adds a new controller which may also be relevant 😬

spec/system/planning_applications/permitted_development_right_spec.rb
```diff
@@ -361,6 +361,61 @@
         expect(page).to have_content("When were the works completed? 01/02/2015")
         expect(page).to have_content("Has anyone ever attempted to conceal the changes? No")
         expect(page).to have_content("Has enforcement action been taken about these changes? No")
+
+        expect(page).to have_content("Assessment summary — Evidence of immunity")
+      end
+
+      context "when it has no evidence attached" do
```
benbaumann95, line 368: possibly worth copying all these specs over to `review_immunity_detail_permitted_development_right_spec.rb`

benbaumann95 (approved):

PR #1060, opened 2023-04-28
Show immunity details on draft recommendation

cdccollins:
### Description of change

Show immunity details on the draft recommendation

### Story Link

https://trello.com/c/TZGZlAjj/1545-display-immunity-information-in-make-draft-recommendation-page-assessor

### Screenshots

screencapture-southwark-southwark-localhost-3000-planning-applications-76-recommendations-new-2023-04-28-15_50_11


spec/models/immunity_detail_spec.rb
```diff
@@ -44,4 +44,76 @@
       expect(immunity_detail.evidence_groups.first.tag).to eq("council_tax_document")
     end
   end
+
+  describe "#evidence_gaps?" do
```
benjamineskola, line 48: Might be overkill but would a test for the case where there's no evidence at all be useful? (I think it returns nil then.)

benjamineskola (approved):

app/components/planning_applications/immunity_details_assessment_component.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module PlanningApplications
+  class ImmunityDetailsAssessmentComponent < ViewComponent::Base
+    def initialize(planning_application:, evidence_groups:)
+      @planning_application = planning_application
+      @evidence_groups = evidence_groups.map do |group|
+        EvidenceGroupPresenter.new(view_context, group)
+      end
+    end
```
EGiataganas, line 10: I think passing the `evidence_groups` separately might not be needed.  `evidence_groups` are just `planning_application.immunity_detail.evidence_groups` so bypassing only the `planning_application` is good enough, what do you think?

benjamineskola: Are Presenter classes instantiated automatically? Perhaps I added that unnecessarily in that case.

cdccollins: I just tried and I think I do need to instantiate the presenter, otherwise it can't find certain method names

EGiataganas (approved):

PR #1068, opened 2023-05-09
Immunity improvements

EGiataganas:
### Description of change

Immunity improvements:
- Fix heading tags
- Add missing word Review
- Consistency between routes: The pattern we follow is that usually on the assessment we have only the route and on the review we are adding the review_ACTION, thus I've renamed it and thought of adding assess_ACTION just to have a better division of routes
- Rearrange text


EGiataganas:
@cdccollins Yeah I totally agree, is something we can definitely improve most probably once we finish the reviewer journey 😄

PR #1071, opened 2023-05-10
Receive Prior Approval applications from PlanX

cdccollins:
### Description of change

Receive prior approval applications from PlanX, but don't do anything particularly special with them

### Story Link

https://trello.com/c/VcYFhO02/1593-receive-prior-approval-applications-from-planx

### DIFF
--- a/app/services/planning_application_creation_service.rb
+++ b/app/services/planning_application_creation_service.rb
@@ -65,6 +65,8 @@ class PlanningApplicationCreationService
   end
 
   def planning_application_params
+    params[:application_type] = "prior_approval" if params[:application_type] == "Apply for prior approval"
+
     permitted_keys = [:application_type,
                       :description,
                       :applicant_first_name,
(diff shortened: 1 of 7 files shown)


app/services/planning_application_creation_service.rb
```diff
@@ -125,4 +127,15 @@ def payment_amount_in_pounds(amount)
   def possibly_immune?(planning_application)
     planning_application.immune_proposal_details.count > 1
   end
+
+  def check_for_prior_approval
+    return unless params[:application_type] == "Apply for prior approval"
+    raise CreateError, "BoPS does not accept this Prior Approval type" unless permitted_prior_approval_type?
+
+    params[:application_type] = "prior_approval"
```
benjamineskola, line 135: I hate that this is necessary, but fair enough.

config/locales/en.yml
```diff
@@ -212,10 +212,13 @@ en:
     lawfulness_certificate:
       existing: LDCE
       proposed: LDCP
+    prior_approval:
+      proposed: ''
```
benjamineskola, line 216: If only lawfulness certificates have an existing/proposed distinction then we might want to refactor this in future so we don't have to define meaningless translation keys for everything. But that's probably out of scope for now.

benjamineskola (approved):

PR #1077, opened 2023-05-12
Reduce duplication in deploy job

benjamineskola:
### Description of change

The deploy-web job is almost identical to the deploy-services job, it should be possible to reduce duplication by including it in the matrix.

Additionally, the generation of task definitions is identical but for the service name. It's possible to use one step for this; the ID `task-def` will refer to the instance of the task within _this_ branch of the matrix, so it'll get the right one.

### Story Link

https://link-to-issue

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.

### DIFF
--- a/.github/workflows/deploy-environment.yml
+++ b/.github/workflows/deploy-environment.yml
@@ -181,30 +181,11 @@ jobs:
           aws ecs describe-task-definition --task-definition bops-${{ matrix.service_type }}-${{ inputs.environment-name }} --query taskDefinition | \
           jq -r 'del(.compatibilities, .taskDefinitionArn, .requiresAttributes, .revision, .status, .registeredAt, .registeredBy)' > ${{ matrix.service_type }}.json
 
-      - name: Generate task definition for console
-        if: matrix.service_type == 'console'
-        id: task-def-console
+      - name: Generate task definition for ${{ inputs.environment-name }}
+        id: task-def
         uses: aws-actions/amazon-ecs-render-task-definition@v1
         with:
-          task-definition: console.json
-          container-name: bops
-          image: ${{ steps.ecr-image.outputs.uri }}
-
-      - name: Generate task definition for worker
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-environment.yml
```diff
@@ -189,66 +192,30 @@ jobs:
...
-          IMAGE_TAG: ${{ steps.github.outputs.sha }}
-        run: |
-          echo "uri=$ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG" >>$GITHUB_OUTPUT
-
-      - name: Download task definition for web and strip unused properties
-        run: |
-          aws ecs describe-task-definition --task-definition bops-web-${{ inputs.environment-name }} --query taskDefinition | \
-          jq -r 'del(.compatibilities, .taskDefinitionArn, .requiresAttributes, .revision, .status, .registeredAt, .registeredBy)' > web.json
-
       - name: Generate task definition for web
```
benjamineskola, line 195: I think we can probably also run this only once and use the `matrix.service-type` variable if we can figure out the right syntax for referring to the output. I'll experiment.

.github/workflows/deploy-environment.yml
```diff
@@ -138,9 +138,12 @@ jobs:
     needs: [build-image, deploy-db-migrate-service]
     strategy:
       matrix:
-        service_type: [ 'console', 'worker' ]
+        service_type: ['console', 'worker', 'web']
 
     steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
```
EGiataganas, line 145: No that matters, but I would also add `if: matrix.service_type == 'web'` as this action is only necessary for web 😄

benjamineskola: I had wondered if I should do that — done now

EGiataganas (approved):

PR #1081, opened 2023-05-17
Improve performance

cdccollins:
### Description of change

Performance on BoPS is getting pretty slow, around 3 seconds for home page to load. I've install the gem bullet (https://github.com/flyerhzm/bullet) which tells you when you're loading things inefficiently. Using this and looking at AppSignal logs have made some improvements

I'm seeing this as a beginning to improve queries on BoPS, if we have it on we can improve as we go rather than me just trying to fix everything in one PR

### DIFF
--- a/app/controllers/concerns/permitted_development_rights.rb
+++ b/app/controllers/concerns/permitted_development_rights.rb
@@ -12,6 +12,6 @@ module PermittedDevelopmentRights
   end
 
   def planning_applications_scope
-    current_local_authority.planning_applications.includes(:permitted_development_rights)
+    current_local_authority.planning_applications
   end
 end
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -25,7 +25,7 @@ class PlanningApplication < ApplicationRecord
 
   with_options dependent: :destroy do
     has_many :audits, -> { by_created_at }, inverse_of: :planning_application
-    has_many :documents, -> { with_file_attachment.by_created_at }, inverse_of: :planning_application
+    has_many :documents, -> { by_created_at }, inverse_of: :planning_application
 
     has_many(
       :recommendations,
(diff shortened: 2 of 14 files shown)


config/environments/test.rb
```diff
@@ -9,7 +9,7 @@
   config.after_initialize do
     Bullet.enable        = true
     Bullet.bullet_logger = true
-    Bullet.raise         = true # raise an error if n+1 query occurs
+    Bullet.raise         = false # raise an error if n+1 query occurs
```
benjamineskola, line 12:
```suggestion
    Bullet.raise         = false # do not raise an error if n+1 query occurs
```

app/models/planning_application.rb
```diff
@@ -25,7 +25,7 @@ class WithdrawOrCancelError < RuntimeError; end
 
   with_options dependent: :destroy do
     has_many :audits, -> { by_created_at }, inverse_of: :planning_application
-    has_many :documents, -> { with_file_attachment.by_created_at }, inverse_of: :planning_application
```
benbaumann95, line 28: Ah did this default scope not work when calling the association?

cdccollins: It was doing some good things in some places, and in some places was loading it unnecessarily

benbaumann95: Yeah makes sense

app/controllers/concerns/permitted_development_rights.rb
```diff
@@ -12,6 +12,6 @@ def set_permitted_development_rights
   end
 
   def planning_applications_scope
-    current_local_authority.planning_applications.includes(:permitted_development_rights)
```
benbaumann95, line 15: This not needed?

Thought it would help loading all the previous responses because we need to iterate through this collection

PR #1083, opened 2023-05-18
Ensure docker versions match production

benjamineskola:
### Description of change

Ensure we use the same versions of Postgres and Redis as in production

Currently in development we're using an older version of Postgres which could lead to issues. We're also pinning to the `latest` version of Redis, which is currently the same as used in production but could change without our intending it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 ---
 services:
   db:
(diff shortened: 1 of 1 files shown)


benbaumann95 (approved):
🎉

PR #1088, opened 2023-05-19
Show basic map of neighbours on consultations page

cdccollins:
### Description of change

- Creates "Consultation" task in application workflow
- Creates "Send letters to neighbours" task under consultation
- Has basic map we're using currently on it

### Story Link

https://trello.com/c/pt0xSQyb/1615-see-the-application-site-on-a-map

### Screenshots
Screenshot 2023-05-19 at 16 01 00

screencapture-southwark-southwark-localhost-3000-planning-applications-1-consultations-new-2023-05-19-16_01_14

### Decisions [OPTIONAL]

Ideally, the map would show neighbours addresses on it - i.e. click a pin and it gives you the exact one. This would mean extending our current map or making a PR against https://github.com/theopensystemslab/map. I've separated into a different story as I'm expecting that's quite a lot of work and seems better to get this started quickly

### DIFF
--- a/app/views/planning_application/consultations/new.html.erb
+++ b/app/views/planning_application/consultations/new.html.erb
@@ -1,0 +1,30 @@
+<% content_for :page_title do %>
+  Send letters to neighbours - <%= t('page_title') %>
+<% end %>
+
+<% add_parent_breadcrumb_link "Home", root_path(filter_options: PlanningApplication::FILTER_OPTIONS) %>
+<% add_parent_breadcrumb_link "Application", planning_application_path(@planning_application) %>
+<% add_parent_breadcrumb_link "Consultation", planning_application_path(@planning_application) %>
+<% content_for :title, "Send letters to neighbours" %>
+
+<%= render(
+  partial: "shared/proposal_header",
+  locals: { heading: "Send letters to neighbours" }
+) %>
+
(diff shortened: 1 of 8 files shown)


app/views/planning_application/consultations/new.html.erb
```diff
@@ -0,0 +1,32 @@
...
+<% add_parent_breadcrumb_link "Application", planning_application_path(@planning_application) %>
+<% add_parent_breadcrumb_link "Consultation", planning_application_path(@planning_application) %>
+<% content_for :title, "Send letters to neighbours" %>
+
+<%= render(
+  partial: "shared/proposal_header",
+  locals: { heading: "Send letters to neighbours" }
+) %>
+
+<%= render "shared/dates_and_assignment_header" %>
```
EGiataganas, line 15: Can you please delete this so it follows the same pattern as other views?

When I was reviewing the content and design of BOPS with Raffaella few months ago we said that is best to have the contents of `<%= render "shared/dates_and_assignment_header" %>` only in the parent directory such as `validation, assessment and review`.

Since you're adding a new step `Consultation` should definitely be there but not in the child. I know that some designs are not updated with this, so we should update them

benbaumann95 (approved):

PR #1089, opened 2023-05-22
Create database structure for constraints

benbaumann95:
### Description of change

Create [database structure](https://miro.com/app/board/uXjVMZhLyKE=/) for constraints

- Adds Constraints table
- Adds PlanningApplicationConstraintsQuery table
- Adds PlanningApplicationConstraint table

Currently we have constraints as an array field on planning applications. However, since we will be making an API request to retrieve the updated constraints from planx/planning data, we need to modify the schema to allow a relation with the query made and constraints returned.

This commit does not yet migrate the existing constraints array on planning applications to the new association and the constraints_list.yml has been kept in for now.

Since custom constraints can be added, we need to iterate and migrate the existing items in all the planning application constraints array that include a name that hasn't been seeded and create a new Constraint entry in the database. This needs to be done before we can switch the code to use the new has_many constraints association for planning applications.

I've not added any indexes for performance (better searching) as this can be added when needed after

### Story Link

https://trello.com/c/BIRiwVeV/1664-create-constraints-database-structure
https://trello.com/c/tu2RHkaV/1610-update-constraints-when-red-line-boundary-is-re-drawn

### DIFF
--- a/app/models/constraint.rb
+++ b/app/models/constraint.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+class Constraint < ApplicationRecord
+  validates :category, :name, presence: true
+
+  enum category: {
+    flooding: "flooding",
+    military_and_defence: "military_and_defence",
+    ecology: "ecology",
+    heritage_and_conservation: "heritage_and_conservation",
+    general_policy: "general_policy",
+    tree: "tree",
+    other: "other",
+    local: "local"
+  }
+
+  has_many :planning_application_constraints, dependent: :destroy
+end
--- a/app/models/constraint.rb
+++ b/app/models/constraint.rb
@@ -2,6 +2,7 @@
 
 class Constraint < ApplicationRecord
   validates :category, :name, presence: true
+  validates :name, uniqueness: { scope: :local_authority }
 
   belongs_to :local_authority, optional: true
 
--- a/app/models/constraint.rb
+++ b/app/models/constraint.rb
@@ -3,6 +3,8 @@
 class Constraint < ApplicationRecord
   validates :category, :name, presence: true
 
+  belongs_to :local_authority, optional: true
+
   enum category: {
     flooding: "flooding",
     military_and_defence: "military_and_defence",
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -47,6 +47,8 @@ class PlanningApplication < ApplicationRecord
     has_many :requests, class_name: "ValidationRequest"
     has_many :assessment_details, -> { by_created_at_desc }, inverse_of: :planning_application
     has_many :permitted_development_rights, -> { order :created_at }, inverse_of: :planning_application
+    has_many :planning_application_constraints
+    has_many :constraints, through: :planning_application_constraints, source: :constraint
     has_one :immunity_detail, required: false
     has_one :consultation, required: false
 
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -48,6 +48,7 @@ class PlanningApplication < ApplicationRecord
     has_many :assessment_details, -> { by_created_at_desc }, inverse_of: :planning_application
     has_many :permitted_development_rights, -> { order :created_at }, inverse_of: :planning_application
     has_many :planning_application_constraints
+    has_many :planning_application_constraints_queries
     has_many :constraints, through: :planning_application_constraints, source: :constraint
     has_one :immunity_detail, required: false
     has_one :consultation, required: false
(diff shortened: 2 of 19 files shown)


cdccollins (approved):

app/models/constraint.rb
```diff
@@ -0,0 +1,19 @@
...
+    ecology: "ecology",
+    heritage_and_conservation: "heritage_and_conservation",
+    general_policy: "general_policy",
+    tree: "tree",
+    other: "other",
+    custom: "custom"
+  }
+
+  has_many :planning_application_constraints, dependent: :destroy
+end
```
EGiataganas, line 19: Add
```
validates :name, uniqueness: {scope: :local_authority}

belongs_to :local_authority, optional: true
```

Also change `custom` to `local`

EGiataganas: On `local_authority` add `has_many :constraints`

benjamineskola: ☝🏻 this came out of a discussion with Andrew. We've realised that custom constraint ought to be scoped to the local authority, so that e.g. bucks and lambeth could both have a constraint with the same name, but without allowing a local authority to have multiple constraints with the same name, and only showing them their own custom constraints.

app/models/planning_application.rb
```diff
@@ -47,6 +47,9 @@ class WithdrawOrCancelError < RuntimeError; end
     has_many :requests, class_name: "ValidationRequest"
     has_many :assessment_details, -> { by_created_at_desc }, inverse_of: :planning_application
     has_many :permitted_development_rights, -> { order :created_at }, inverse_of: :planning_application
+    has_many :planning_application_constraints
+    # To change to "constraints" when we update the code to use the new association and are able to drop the "constraints" array field
+    has_many :new_constraints, through: :planning_application_constraints, source: :constraint
```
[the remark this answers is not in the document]
EGiataganas: We also need to add
```
has_many :planning_application_constraint_queries
```

app/models/planning_application.rb
```diff
@@ -47,6 +47,9 @@ class WithdrawOrCancelError < RuntimeError; end
     has_many :requests, class_name: "ValidationRequest"
     has_many :assessment_details, -> { by_created_at_desc }, inverse_of: :planning_application
     has_many :permitted_development_rights, -> { order :created_at }, inverse_of: :planning_application
+    has_many :planning_application_constraints
```
benjamineskola, line 50: We think that according to the original diagram this should be

```
has_many :planning_application_constraint_queries
```

However, we also think that we're going to need to support custom constraints that aren't associated with a query. So in fact we'll want:

```suggestion
    has_many :planning_application_constraint_queries # for constraints from planx
    has_many :planning_application_constraints # for manually-added constraints
```

app/models/planning_application.rb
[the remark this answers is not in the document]
EGiataganas: > TODO: After dropping text field array constraints, we should rename the association new_constraints to constraints


We are going to rename the current `constraints` field to `old_constraints` so we don't need to do any further renaming after. This will need to be in 2 prs. 1pr will add an `alias_attribute` and 2nd pr will be the one of doing the renaming.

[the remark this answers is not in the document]
EGiataganas: 1st pr: https://github.com/unboxed/bops/pull/1097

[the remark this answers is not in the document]
EGiataganas: 2nd PR: Rename constraints to old_constraints #1099

app/models/planning_application.rb
EGiataganas: `has_many :planning_application_constraints_queries`

app/models/constraint.rb
benbaumann95: Yeah had this in mind for the subsequent PR as I didn't have time to finish but thanks for adding here

db/migrate/20230602181048_create_constraints.rb
```diff
@@ -0,0 +1,78 @@
...
+
+class CreateConstraints < ActiveRecord::Migration[7.0]
+  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/AbcSize
+  def change
+    create_table :constraints do |t|
+      t.string :name, null: false
+      t.string :category, null: false
+
+      t.index [:name], unique: true
```
benbaumann95, line 11: Could remove this if we are adding a unique index for the combination of local authority and name instead

EGiataganas: yes good shout!

PR #1082, opened 2023-05-17
Add review type to review immunity detail

benbaumann95:
### Description of change

- This change fixes a bug where review evidence of immunity and review immunity enforcement were updating the same instance instead of storing the reviewer comment separately
- Adding a review type means we can distinguish between the different types of review (evidence / enforcement)

- This also adds a complete e2e test for assessing and reviewing an immunity case going back and forth between the assessor and reviewer

### Story Link

https://trello.com/c/FzhpuW0E/1599-end-to-end-automated-test-for-immunity-assessment
https://trello.com/c/f2sRt6yX/1601-reviewing-evidence-of-immunity-overwrites-the-review-immunity-enforcement-response

PR #1090, opened 2023-05-23
Fix review immunity details to be specific for review_type

EGiataganas:
### Description of change

After the #1082 pr, we introduced the `review_type` on the `review_immunity_details` table. But we miss fixing these page to be based on the `review_type`.


benbaumann95 (approved):
Could we please add a spec for this, otherwise looks good.

PR #1091, opened 2023-05-23
Add LetterSendingService

benjamineskola:
### Description of change

- Add notify API key and letter template to local authority model
- Add LetterSendingService
- Add NeighbourLetter model to record sending of letter

### Story Link

https://trello.com/c/KhmtfhhN/1630-send-the-letter-to-the-selected-neighbours

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.

### DIFF
--- a/app/services/letter_sending_service.rb
+++ b/app/services/letter_sending_service.rb
@@ -5,22 +5,33 @@ require "notifications/client"
 class LetterSendingService
   DEFAULT_NOTIFY_TEMPLATE_ID = "701e32b3-2c8c-4c16-9a1b-c883ef6aedee"
 
-  attr_reader :address, :message
+  attr_reader :neighbour, :message
 
-  def initialize(local_authority, address, message)
-    @local_authority = local_authority
-    @address = address
+  def initialize(neighbour, message)
+    @local_authority = neighbour.consultation.planning_application.local_authority
+    @neighbour = neighbour
     @message = message
   end
 
   def deliver!
-    client.send_letter(
-      template_id: notify_template_id,
-      personalisation: {
-        address:,
-        message:
-      }
-    )
+    letter_record = NeighbourLetter.new(neighbour:, text: message)
+    letter_record.save!
+
+    begin
+      response = client.send_letter(
+        template_id: notify_template_id,
+        personalisation: {
+          address: neighbour.address,
+          message:
+        }
+      )
+    rescue Notifications::Client::RequestError
+      return
+    end
+
+    letter_record.sent_at = Time.zone.now
+    letter_record.notify_response = response
+    letter_record.save!
   end
 
   private
--- a/app/services/letter_sending_service.rb
+++ b/app/services/letter_sending_service.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require "notifications/client"
+
+class LetterSendingService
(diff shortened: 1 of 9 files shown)


cdccollins:
I think it'd be useful to create a record of the letter on our side for auditing purposes and so we can track it. Or is that for a later PR that's coming?

cdccollins:
> I think the Notify web interface has this feature built in — but it shouldn't be difficult to do on our side either, I'll add that.

In another app we create the "message" object and then update that with whatever Notify sends us back which works pretty nicely

db/migrate/20230526133904_create_neighbour_letters.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class CreateNeighbourLetters < ActiveRecord::Migration[7.0]
+  def change
+    create_table :neighbour_letters do |t|
+      t.references :planning_application, null: false
+      t.string :address
+      t.string :text
+      t.string :sent_at
+      t.json :response
```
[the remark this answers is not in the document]
benjamineskola: That's a good idea, it needs to be associated with _something_, but perhaps the consultation is more useful.

I suppose it won't have more than one consultation for a given planning application, so this will be unambiguous either way? But referencing the consultation might at least be more convenient.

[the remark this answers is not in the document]
benjamineskola: Oh, actually, hang on, you've got a `neighbours` table, so presumably I can just reference that instead.

cdccollins (approved):

db/migrate/20230526133904_create_neighbour_letters.rb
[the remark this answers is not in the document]
cdccollins: Ah yes forgot the code that I wrote 4 days ago 😂

app/services/letter_sending_service.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def deliver!
+    letter_record = NeighbourLetter.new(neighbour:, text: message)
+    letter_record.save!
+
+    begin
+      response = client.send_letter(
+        template_id: notify_template_id,
+        personalisation: {
+          address: neighbour.address,
```
cdccollins, line 24: Just reading the docs for my story and it seems you need to split the address into lines https://docs.notifications.service.gov.uk/rest-api.html#send-a-letter, at least 3 lines needed

cdccollins: Don't think it'll ever be non-UK so shouldn't have to worry about that? 

Do you mean adding postcode, name to the neighbour model and then passing those as further address lines in this method?

benjamineskola: Currently the address is a comma-separated string (or maybe that's just what factory_bot gives us) but I think that's ambiguous, since some people put a comma after a house number. I suppose it's not the end of the world if the house number ends up on a separate line but I think it'd be better to treat it as a newline-separated string?

cdccollins: Might be good to have a look at what the API we're using gives us - you can test it on staging. Generally seems to be a comma after the house number, but yeah don't think it's end of the world

cdccollins (approved):

PR #1094, opened 2023-05-25
Fetch documents from production planx if it is copied from prod

cdccollins:
### Description of change

When a planning application is copied over to staging from production, it tries to fetch the documents from PlanX but doesn't have the right API key. This ensures it's using the right key

### DIFF
--- a/app/services/upload_documents_service.rb
+++ b/app/services/upload_documents_service.rb
@@ -9,7 +9,8 @@ class UploadDocumentsService
   def call
     files&.each do |file_params|
       filename = file_params[:filename]
-      file = URI.parse(filename).open("api-key" => ENV.fetch("PLANX_FILE_API_KEY", nil))
+
+      file = fetch_file(filename)
 
       raise Api::V1::Errors::WrongFileTypeError.new(nil, filename) if forbidden?(file.content_type)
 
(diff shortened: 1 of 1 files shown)


app/services/upload_documents_service.rb
```diff
@@ -6,10 +6,15 @@ def initialize(files:, planning_application:)
...
+  def call # rubocop:disable Metrics/AbcSize
     files&.each do |file_params|
       filename = file_params[:filename]
-      file = URI.parse(filename).open("api-key" => ENV.fetch("PLANX_FILE_API_KEY", nil))
+
+      file = if @planning_application.from_production?
+               URI.parse(filename).open("api-key" => ENV.fetch("PLANX_FILE_PRODUCTION_API_KEY", nil))
+             else
+               URI.parse(filename).open("api-key" => ENV.fetch("PLANX_FILE_API_KEY", nil))
+             end
```
EGiataganas, line 17: How about moving this functionality on a private method?

app/services/upload_documents_service.rb
```diff
@@ -37,4 +38,12 @@ def attach_files(planning_application, file, file_params)
...
     end
   end
+
+  def fetch_file(filename)
+    if @planning_application.from_production?
+      URI.parse(filename).open("api-key" => ENV.fetch("PLANX_FILE_PRODUCTION_API_KEY", nil))
+    else
+      URI.parse(filename).open("api-key" => ENV.fetch("PLANX_FILE_API_KEY", nil))
+    end
+  end
```
EGiataganas, line 48: how about something like that
```
  def fetch_file(filename)
    URI.parse(filename).open("api-key" => planx_file_api_key)
  end

  def planx_file_api_key
    @planning_application.from_production? ? ENV.fetch("PLANX_FILE_PRODUCTION_API_KEY", nil) : ENV.fetch("PLANX_FILE_API_KEY", nil)
  end
```

cdccollins: 👍 rubocop corrected `planx_file_api_key` to an if/else statement

EGiataganas (approved):

PR #1099, opened 2023-05-31
Rename constraints to old_constraints

EGiataganas:
### Description of change

Rename constraints to old_constraints

### DIFF
--- a/app/views/constraints/edit.html.erb
+++ b/app/views/constraints/edit.html.erb
@@ -24,7 +24,7 @@
           </legend>
           <div class="form-group">
             <div class="govuk-checkboxes govuk-checkboxes--small">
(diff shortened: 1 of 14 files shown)


app/views/constraints/edit.html.erb
benjamineskola, line 1: I wonder if this should remain as `constraints`, as it's externally-visible? Although I think it probably doesn't matter very much since it won't cause any kind of API breakage.

EGiataganas: Yes it should stay as `constraints` so we won't cause any kind of API breakage. Plus this is temporary as we will soon rebase & refactor as we discussed the PR #1089

benjamineskola (approved):

PR #1100, opened 2023-05-31
Add button to send letters to neighbours

benjamineskola:
### Description of change

A previous PR added the backend service for sending a letter. This adds a controller action to invoke the service and a button on the frontend.

It also adds a field to the Consultation model to record the text of the letter to be sent to the neighbours.

### Story Link

https://trello.com/c/KhmtfhhN/1630-send-the-letter-to-the-selected-neighbours

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.

### DIFF
--- a/app/controllers/planning_application/consultations_controller.rb
+++ b/app/controllers/planning_application/consultations_controller.rb
@@ -58,7 +58,7 @@ class PlanningApplication
     def send_neighbour_letters
       return if @consultation.blank?
 
-      @consultation.neighbours.each do |neighbour|
+      @consultation.neighbours.reject(&:letter_created?).each do |neighbour|
         LetterSendingService.new(neighbour, @consultation.neighbour_letter_text).deliver!
       end
     end
--- a/app/controllers/planning_application/consultations_controller.rb
+++ b/app/controllers/planning_application/consultations_controller.rb
@@ -63,7 +63,10 @@ class PlanningApplication
       end
 
       respond_to do |format|
-        format.html { redirect_to edit_planning_application_consultation_path(@planning_application, @consultation) }
+        format.html do
+          redirect_to planning_application_consultation_path(@planning_application, @consultation),
+                      flash: { sent_neighbour_letters: true }
+        end
       end
     end
 
--- a/app/controllers/planning_application/consultations_controller.rb
+++ b/app/controllers/planning_application/consultations_controller.rb
@@ -55,6 +55,14 @@ class PlanningApplication
       end
     end
 
+    def send_neighbour_letters
+      return if @consultation.blank?
+
+      @consultation.neighbours.each do |neighbour|
+        LetterSendingService.new(neighbour, @consultation.neighbour_letter_text).deliver!
+      end
+    end
+
     private
 
     def set_planning_application
--- a/app/controllers/planning_application/consultations_controller.rb
+++ b/app/controllers/planning_application/consultations_controller.rb
@@ -61,6 +61,10 @@ class PlanningApplication
       @consultation.neighbours.reject(&:letter_created?).each do |neighbour|
         LetterSendingService.new(neighbour, @consultation.neighbour_letter_text).deliver!
       end
+
+      respond_to do |format|
+        format.html { redirect_to edit_planning_application_consultation_path(@planning_application, @consultation) }
+      end
     end
 
     private
--- a/app/views/planning_application/consultations/show.html.erb
+++ b/app/views/planning_application/consultations/show.html.erb
@@ -31,6 +31,11 @@
     <%= render "contacted_list", planning_application: @planning_application, consultation: @consultation %>
 
     <div class="govuk-button-group govuk-!-margin-top-5">
+      <%= form_with model: [@planning_application, @consultation], url:
+  planning_application_consultation_send_neighbour_letters_path(planning_application: @planning_application, consultation_id: @consultation.id), method: 'post' do |form| %>
+        <%= form.submit "Print and send letters", class: "govuk-button govuk-button--primary", style: "margin-inline-end: 2rem;" %>
+      <% end %>
+
       <%= form_with(
           model: [@planning_application, @consultation],
           builder: GOVUKDesignSystemFormBuilder::FormBuilder,
--- a/app/views/planning_application/consultations/show.html.erb
+++ b/app/views/planning_application/consultations/show.html.erb
@@ -12,6 +12,24 @@
   locals: { heading: "Send letters to neighbours" }
 ) %>
 
+<% if flash[:sent_neighbour_letters].present? %>
+  <div class="govuk-notification-banner govuk-notification-banner--success" role="alert"
+    aria-labelledby="govuk-notification-banner-title"
+    data-module="govuk-notification-banner">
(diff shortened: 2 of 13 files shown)


benjamineskola:
[Send the letter to the selected neighbours](https://trello.com/c/KhmtfhhN/1630-send-the-letter-to-the-selected-neighbours)

app/controllers/planning_application/consultations_controller.rb
```diff
@@ -46,6 +46,14 @@ def destroy
       end
     end
 
+    def send_neighbour_letters
+      return if consultation.blank?
+
+      consultation.neighbours.reject(&:letter_created?).each do |neighbour|
+        LetterSendingService.new(neighbour, consultation.neighbour_letter_text).deliver!
+      end
+    end
```
EGiataganas, line 55: Shall we move it on a Job?

benjamineskola: I think `deliver_later!` would work here, right? I'm not super familiar with ActiveJob but yes, we're likely to want this to send quite a few at once and so we don't want it delaying the page load.

app/views/planning_application/consultations/show.html.erb
```diff
@@ -26,6 +26,24 @@
 
     <%= render "form", planning_application: @planning_application, consultation: @consultation %>
 
+    <% if flash[:sent_neighbour_letters].present? %>
```
cdccollins, line 29: I'm wondering if this should be at the top of the page so it's more obvious?

benjamineskola: That's what the [GOV.UK Design System](https://design-system.service.gov.uk/components/notification-banner/#how-it-works) says, and it would match how we normally use flash messages; but the [design](https://www.figma.com/proto/l6yzC3qHvmgeYuSFzC6YWp/Consultations?type=design&node-id=239-15949&scaling=min-zoom&page-id=239%3A15793) says it should be just before the list.

cdccollins: Ah I see. But that design doesn't have the map on it, so I think it'd end up being quite far down the page? Maybe worth asking a designer

benjamineskola: Ah, I see — because the design doesn't have the map, I've interpreted it as "right above the list" because that's the same as "right below the header details". I think if I put it right below the header but above the map it makes sense in the context of the design _and_ the added map.

cdccollins (approved):

cdccollins:
This occurred to me last night 😂 have you set up a template to use in Notify?

benjamineskola:
> This occurred to me last night 😂 have you set up a template to use in Notify?

I have not! Currently I've just defined it to use the SMS template, which definitely won't work in real life. There's a card to design a template, so we can swap out the ID once that's done. We won't be able to do a real-world test until then.

I suppose we could create a blank-ish template just to have the ID of it, and edit it later.

cdccollins:
> > This occurred to me last night 😂 have you set up a template to use in Notify?
> 
> I have not! Currently I've just defined it to use the SMS template, which definitely won't work in real life. There's a card to design a template, so we can swap out the ID once that's done. We won't be able to do a real-world test until then.
> 
> I suppose we could create a blank-ish template just to have the ID of it, and edit it later.

My other late night thought that I just remembered is that we should probably have a switch for this to be on/off and disable this in staging?

benjamineskola:
> we should probably have a switch for this to be on/off and disable this in staging?

One option would be to have a separate Notify service for staging, set to testing only — I'm not sure how that works for letters though. It might also be more complicated if we want to use different accounts for every council anyway, so maybe a switch is the easiest option.

(Or just only define the API key in production?)

cdccollins:
> > we should probably have a switch for this to be on/off and disable this in staging?
> 
> One option would be to have a separate Notify service for staging, set to testing only — I'm not sure how that works for letters though. It might also be more complicated if we want to use different accounts for every council anyway, so maybe a switch is the easiest option.
> 
> (Or just only define the API key in production?)

I think I looked and there isn't a test for letters, annoyingly but maybe unsurprisingly. Worth double checking though

app/controllers/planning_application/consultations_controller.rb
benbaumann95: Oh actually that's for mailer classes.
I think in the job class you can just add `queue_as :low_priority`

db/migrate/20230601161704_make_neighbour_letter_neighbour_unique.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class MakeNeighbourLetterNeighbourUnique < ActiveRecord::Migration[7.0]
+  def up
+    remove_index :neighbour_letters, :neighbour_id
+    change_column :neighbour_letters, :neighbour_id, :bigint, unique: true
+    add_index :neighbour_letters, :neighbour_id, unique: true, name: "ix_neighbour_letters_on_neighbour_id"
```
benbaumann95, line 7: Will this be an issue when we get round to being able to resend letters to the same neighbour?

benjamineskola: I was thinking in terms of a resent letter being the same NeighbourLetter, but perhaps that wouldn't actually make sense.

If we did it this way, then when resending we'd throw away the original Notify ID in favour of the new one, which might be OK but if we want to keep track of the fact that it was resent then perhaps it would be better to have a one-to-many relationship here (one neighbour can have multiple letters).

In which case, we could still check that a letter hasn't been sent by looking at the one-to-many, so that part would still work.

benbaumann95: Yeah I reckon it'll have to be one neighbour having multiple letters. In addition to resending, it's unlikely but possible that there are several planning applications surrounding that neighbour too which they'd get a separate letter for

benjamineskola: >it's unlikely but possible that there are several planning applications surrounding that neighbour too which they'd get a separate letter for

This wouldn't be an issue. A Neighbour is unique per Consultation, so the same address might be represented as more than one Neighbour if there are several Consultations close enough to each other, and their status would be tracked individually. I don't think there would be any need for us to know/care that two instances of 60 Commercial Street are actually the same.

benjamineskola: OK, I've removed this migration, but for now the logic about when to send doesn't account for this — we can fix it once we actually add the resending functionality.

config/locales/en.yml
```diff
@@ -896,6 +896,7 @@ en:
     not_started: Not started
     permanent_failure: Permanent failure
     posted: Posted
+    printing: Printing
```
benjamineskola, line 899: @cdccollins it looks like in the NeighbourLetter model the `accepted` status is also called `printing` and then it looks that value up in the translations, not `accepted` — can I just double-check with you that that makes sense?

PR #1101, opened 2023-05-31
Show letter status

cdccollins:
### Description of change

Show the status of letters sent by notify

### Story Link

https://trello.com/c/Ow38QnS9/1634-show-a-record-of-when-letters-were-actually-sent-by-printer-and-to-who

### Screenshot

screencapture-southwark-southwark-localhost-3000-planning-applications-1-consultations-4-2023-06-02-14_04_42

### DIFF
--- a/app/controllers/planning_application/consultations_controller.rb
+++ b/app/controllers/planning_application/consultations_controller.rb
@@ -102,7 +102,7 @@ class PlanningApplication
     end
 
     def status
-      (save_progress? || add_neighbour?) ? :in_progress : :complete
+      save_progress? || add_neighbour? ? :in_progress : :complete
     end
   end
 end
--- a/app/controllers/planning_application/consultations_controller.rb
+++ b/app/controllers/planning_application/consultations_controller.rb
@@ -7,7 +7,13 @@ class PlanningApplication
     before_action :set_planning_application
     before_action :set_consultation, except: %i[new create]
     before_action :assign_params, only: %i[update create]
-    before_action :update_letter_statuses, only: %i[show]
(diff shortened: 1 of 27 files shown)


app/controllers/planning_application/consultations_controller.rb
```diff
@@ -81,5 +90,19 @@ def assign_params
         @attributes = consultation_params
       end
     end
+
+    def update_letter_statuses
+      return if @consultation.neighbour_letters.none?
+
+      # This is not ideal for now as will block the page loading, if it becomes a problem this would be
+      # a good place for optimisation
+      @consultation.neighbour_letters.each do |letter|
```
benjamineskola, line 99: Evangelos mentioned a background job for the letter sending; perhaps that would work here too (in a later PR).

Or alternatively (or additionally?) perhaps just doing it through AJAX and updating the status asynchronously that way. Either way this is fine for now, of course.

cdccollins: Yeah I would just worry about the status changing while the people were looking at the page and then them missing it. Or them looking at the page and it showing an incorrect status because the background job is out of date. 

Basically I kind of want to see how much of a problem this is before committing either way

benjamineskola: Yes, that's true. We'd want to show some sort of indication rather than just showing an old status and silently replacing it. (A loading icon instead of the status at first perhaps?) It might also work to run all of the fetches in parallel within the controller method, with threads.

But absolutely, yeah, if it turns out we can fetch dozens of statuses without significantly impacting the page load then we shouldn't worry about it.

benjamineskola (approved):

PR #1102, opened 2023-06-02
Use parallel_tests gem to manage test parallelisation

benjamineskola:
### Description of change

Currently we define the parallel groups manually but several of these are significantly longer than others and they are likely to change over time.

The parallel-tests gem should handle this for us.

### Known issues

It would be possible to put each test in a separate matrix group but I'm not sure we'd get much value out of that; it also has the downside that we'd go back to having to install the PDF tools for all test groups even when they aren't needed.

This will by default estimate which files to run in which group as it won't have any data about which are slowest. It's possible to tune that but probably not worth it to begin with.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,29 +41,9 @@ jobs:
 
   specs:
     uses: ./.github/workflows/testing.yml
-    strategy:
-      matrix:
-        specs:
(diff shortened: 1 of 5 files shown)


benjamineskola:
I haven't been able to get this working in a matrix (it runs all the tests on each), and running it in parallel on a single instance is slower than what we currently have (because they're all single-core) so I don't think this is worthwhile for now.

PR #1106, opened 2023-06-09
Update all constraints logic to use new db associations

benbaumann95:
### Description of change

- Manually add/update/remove constraints using new associations
- Update PlanningApplicationCreationService to use the new db structure
- Add audit creation callback for adding/removing constraints

Todo: Once this is up and working, we can remove all references to the `old_constraints` array field and drop this column

### Story Link

https://trello.com/c/aQagsxTg/1742-use-new-constraints-database-structure-for-manually-adding-constraints
https://trello.com/c/J9djmvZq/1738-update-planning-application-creation-service-to-use-new-constraints-database-structure

### DIFF
--- a/app/services/constraints_creation_service.rb
+++ b/app/services/constraints_creation_service.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+class ConstraintsCreationService
+  def initialize(planning_application:, constraints_params:)
+    @planning_application = planning_application
+    @constraints_params = constraints_params
+  end
+
+  def call # rubocop:disable Metrics/AbcSize
+    constraints.each do |constraint|
+      existing_constraint = Constraint.options_for_local_authority(planning_application.local_authority_id)
+                                      .find_by("LOWER(name)= ?", constraint.downcase)
+
+      if existing_constraint
+        planning_application.planning_application_constraints.create!(constraint_id: existing_constraint.id)
+      else
+        planning_application.constraints.find_or_create_by!(
+          name: constraint.titleize, category: "local", local_authority_id: planning_application.local_authority_id
+        )
+      end
+    end
+  rescue ActiveRecord::RecordInvalid, NoMethodError => e
+    Appsignal.send_error(e)
+  end
+
+  private
+
+  attr_reader :planning_application, :constraints_params
+
+  def constraints
+    if constraints_params.present?
+      constraints_params.to_unsafe_hash.filter_map do |key, value|
+        key.humanize if value
+      end
+    else
+      []
+    end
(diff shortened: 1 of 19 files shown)


app/services/constraints_creation_service.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class ConstraintsCreationService
+  def initialize(planning_application:, constraints_params:)
+    @planning_application = planning_application
+    @constraints_params = constraints_params
```
benjamineskola, line 6: What about calling this `constraints_ids` or something like that? I think `params` is a little ambiguous here.

benbaumann95: This actually comes in from PlanX like https://github.com/unboxed/bops/pull/1106/files#diff-b615619cfe0d1ee1441e3489b67ccaae21f57a485a460bdda1ea00152f171d13R11-R20

```ruby
ActionController::Parameters.new(
    {
      "conservation_area" => true,
       "protected_trees" => true,
       "national_park" => true,
        "listed_building" => true
     }
  )
```

benbaumann95: I think you'll be able to see this in swagger too for the Post request

app/views/constraints/edit.html.erb
```diff
@@ -56,8 +54,8 @@
       <% end %>
 
       <div class="form-group">
-        <%= form.label :old_constraints, "Add a local constraint", class: 'govuk-heading-s govuk-!-margin-top-6' %>
-        <%= form.text_area :old_constraints, class: "govuk-textarea", rows: "3", "aria-describedby": "constraints-hint", multiple: true, value: "" %>
+        <%= form.label :name, "Add a local constraint", class: 'govuk-heading-s govuk-!-margin-top-6' %>
+        <%= form.text_area :name, class: "govuk-textarea", rows: "3", "aria-describedby": "constraints-hint" %>
```
EGiataganas, line 58: I am wondering whether we should add send everything as `capitalize` or `titleize` here. For a example if someone adds `Flood zone 4` or `Flood Zone 4` or even `Flood zone 4  ` it should be stored it as `Flood zone 4`. I know that's an improvement to the current flow and maybe not part of this. What do you think?

benbaumann95: Added here - https://github.com/unboxed/bops/pull/1106/files#diff-56352b080a99ea83abab463a40b35fc4f34425f8e84ec8c1f7117e22d2783b62R21

EGiataganas (approved):

PR #1114, opened 2023-06-13
Add date and status to letter

cdccollins:
### Description of change

Show date letter was attempted to be sent if it was rejected, and show it in the audit log


app/services/letter_sending_service.rb
```diff
@@ -37,6 +38,7 @@ def deliver! # rubocop:disable Metrics/AbcSize
     letter_record.notify_id = response.id
     letter_record.status = "accepted"
     letter_record.save!
+    update_audit("sent")
```
benbaumann95, line 41: possibly out of scope but should we put these multi record saves in a transaction?

benbaumann95 (approved):
This is gold dust 🤩

PR #1116, opened 2023-06-14
Order first Prior Approval application type

EGiataganas:
### Description of change

Order first Prior Approval application type

### Story Link

https://trello.com/c/YhhrFPKb

### DIFF
--- a/app/services/planning_application_search.rb
+++ b/app/services/planning_application_search.rb
@@ -7,7 +7,7 @@ class PlanningApplicationSearch
   STATUSES = %w[not_started invalidated in_assessment awaiting_determination to_be_reviewed].freeze
   REVIEWER_STATUSES = %w[awaiting_determination to_be_reviewed].freeze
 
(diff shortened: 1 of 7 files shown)


app/services/planning_application_search.rb
```diff
@@ -83,7 +83,7 @@ def my_applications
   end
 
   def all_applications
-    @all_applications ||= local_authority.planning_applications.includes([:application_type]).by_created_at_desc
+    @all_applications ||= local_authority.planning_applications.includes([:application_type])
```
pixeltrix, line 86: Are you expecting the planning applications to be sorted by application type? The `includes` will most likely do a second query to load the application type and the applications will be unsorted

EGiataganas: Funny enough, I am getting Bullet Warnings to delete the `includes` and when I am deleting it I am getting another warning to add it back

benjamineskola (approved):

PR #1125, opened 2023-06-21
Drop filter options from root_path, Home

EGiataganas:
### Description of change

Drop filter options from root_path, Home

### DIFF
--- a/app/views/audits/index.html.erb
+++ b/app/views/audits/index.html.erb
@@ -2,7 +2,7 @@
   Audit log - <%= t('page_title') %>
 <% end %>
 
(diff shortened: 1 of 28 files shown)


cdccollins (approved):

pixeltrix:
> Will this work? Will the filter options all be automatically selected? I feel like I'm missing something

The routes.rb has this definition:
https://github.com/unboxed/bops/blob/6af61c9a42b62adf5f386b24ff5b4e447006ddbf/config/routes.rb#L8-L12

This means the params has two copies of the filter options - one at the top level and one inside `:planning_application_search_filter`. Seems like this one is what's affecting whether the checkboxes are selected.

cdccollins:
Ah yes me being silly, got it

PR #1132, opened 2023-06-23
Add endpoints for bops applicants

cdccollins:
### Description of change

Add more information to the BoPS API so that BoPS applicants can display all the information needed for a planning application page


app/controllers/concerns/active_storage/set_disk_blob.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def decrypted_hash
+      # rubocop:disable Security/MarshalLoad
+      Marshal.load(Base64.decode64(JSON.parse(Base64.decode64(encrypted_hash)).dig("_rails", "message")))
+      # rubocop:enable Security/MarshalLoad
+    end
+
+    def verify_request
+      ActiveStorage.verifier.valid_message?(secret) || raise_bad_request
```
pixeltrix, line 33: The `verify` method on the verifier will raise for us so no need for the `|| raise` if you use that method

app/controllers/concerns/active_storage/set_disk_blob.rb
```diff
@@ -0,0 +1,48 @@
...
+    def set_disk_blob
+      return unless @blob.nil?
+
+      @blob = ActiveStorage::Blob.find_by(key: decrypted_hash[:key])
+    rescue StandardError, ActiveSupport::MessageVerifier::InvalidSignature
+      head :not_found
+    end
+
+    def encrypted_hash
+      request.path_info[%r{(?<=disk/).*?(?=-)}]
```
pixeltrix, line 23: Is this not just the `:encoded_key` path param?

app/controllers/concerns/active_storage/set_disk_blob.rb
```diff
@@ -0,0 +1,48 @@
...
+      before_action :set_disk_blob
+    end
+
+    private
+
+    def set_disk_blob
+      return unless @blob.nil?
+
+      @blob = ActiveStorage::Blob.find_by(key: decrypted_hash[:key])
+    rescue StandardError, ActiveSupport::MessageVerifier::InvalidSignature
```
pixeltrix, line 18: You can use `||=` to avoid the nil check, e.g.

``` ruby
@blob ||= begin
  ActiveStorage::Blob.find_by(key: decrypted_hash[:key])
rescue StandardError
  head :not_found
end
```

The exception `ActiveSupport::MessageVerifier::InvalidSignature` is a [descendant][1] of `StandardError` and will match the rescue anyway.

[1]: https://github.com/rails/rails/blob/b9ca94caea2ca6a6cc09abaffaad67b447134079/activesupport/lib/active_support/message_verifier.rb#L104

app/controllers/concerns/active_storage/set_disk_blob.rb
```diff
@@ -0,0 +1,48 @@
...
+      head :not_found
+    end
+
+    def encrypted_hash
+      request.path_info[%r{(?<=disk/).*?(?=-)}]
+    end
+
+    def decrypted_hash
+      # rubocop:disable Security/MarshalLoad
+      Marshal.load(Base64.decode64(JSON.parse(Base64.decode64(encrypted_hash)).dig("_rails", "message")))
```
pixeltrix, line 28: I'd probably not do all this in one line so that any errors are more visible when you look at it

app/controllers/concerns/active_storage/set_disk_blob.rb
```diff
@@ -0,0 +1,48 @@
...
+      # rubocop:enable Security/MarshalLoad
+    end
+
+    def verify_request
+      ActiveStorage.verifier.valid_message?(secret) || raise_bad_request
+    end
+
+    def secret
+      request.path_info[%r{(?<=disk/).*?(?=/)}]
```
pixeltrix, line 37: As above you should be able to take the `:encoded_key` param and split it on `--` rather than using a regex

app/controllers/concerns/active_storage/set_disk_blob.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def raise_bad_request
+      raise ActiveSupport::MessageVerifier::InvalidSignature
+    end
+
+    def verification_token
+      Rails.application.credentials.config[:active_record_encryption][:primary_key]
+    end
```
pixeltrix, line 46: Is this method used?

app/controllers/api/v1/local_authorities_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+          respond_to(:json)
+        else
+          send_not_found_response
+        end
+      end
+
+      def send_not_found_response
+        render json: { message: "Unable to find record" },
+               status: :not_found
+      end
```
pixeltrix, line 23: Do you know that Rails will try to render a JSON error if that was the original format? This is what you'll get for an ActiveRecord::RecordNotFound error: 
``` json
{"status":404,"error":"Not Found"}
```

This would allow you to change the controller to:
``` ruby
module Api
  module V1
    class LocalAuthoritiesController < Api::V1::ApplicationController
      before_action :set_cors_headers, if: :json_request?
      skip_before_action :authenticate

      def show
        @local_authority = LocalAuthority.find_by!(subdomain: params[:subdomain])
        respond_to(:json)
      end
    end
  end
end
```

app/controllers/api/v1/planning_applications_controller.rb
```diff
@@ -16,7 +16,7 @@ def index
 
       def show
         @planning_application = current_local_authority.planning_applications.where(id: params[:id]).first
-        if @planning_application
+        if @planning_application&.validated?
```
pixeltrix, line 19: Same as above - if you use a finder method that raises you can avoid all the logic

pixeltrix (approved):
Looks like the spec failures are related to test setup as it's saying it can't create a representation of the file attached to the document. Otherwise it's fine - the find_by! stuff can be cleaned up in another PR as I suspect that there's a pattern to refactor throughout the app.

PR #1134, opened 2023-06-27
Show both added and removed constraints

benjamineskola:
### Description of change

Track when existing constraints are removed by a new redline query, and display those removed constraints alongside the existing constraints.

### Story Link

https://trello.com/c/JBSziiHI/1756-update-constraints-when-red-line-boundary-is-re-drawn-frontend

### Screenshots

Screenshot 2023-06-28 at 8 55 06

### Known issues

UI could use improvement; in particular, we could display human-readable names once we start getting them from Planx.

e.g.

### DIFF
--- a/app/models/planning_application_constraint.rb
+++ b/app/models/planning_application_constraint.rb
@@ -13,6 +13,9 @@ class PlanningApplicationConstraint < ApplicationRecord
   delegate :name, to: :constraint
   delegate :audits, to: :planning_application
 
+  scope :active, -> { where({ removed_at: nil }) }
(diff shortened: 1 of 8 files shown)


cdccollins (commented):
Looks good! I have some thoughts on the UI but I'm guessing that's not the focus now. Think the database makes sense 👍

benjamineskola:
> I have some thoughts on the UI

Yep, so far I've only just reused the same component for both lists, and it's not really ideal when duplicated.

cdccollins (approved):
Looks good, could you add some specs?

PR #1139, opened 2023-06-29
Fix flaky tests

benjamineskola:
The helper searches within any govuk tab panel, but appears to be intended only to search within the current (i.e., not hidden) one.


cdccollins (approved):
🎉 🎉 🎉

PR #1141, opened 2023-06-29
Make letter template editable

cdccollins:
### Description of change

Make the letter template editable in case any of the measurements are wrong

Screenshot 2023-06-29 at 16 39 27


app/services/letter_sending_service.rb
```diff
@@ -7,23 +7,24 @@ class LetterSendingService
...
     @neighbour = neighbour
     @consultation = neighbour.consultation
+    @letter_content = letter_content
   end
 
   def deliver!
     return if NeighbourLetter.find_by(neighbour:).present?
 
-    letter_record = NeighbourLetter.new(neighbour:, text: message)
+    letter_record = NeighbourLetter.new(neighbour:, text: @letter_content)
```
benbaumann95, line 20: nitpick but shall we add `@letter_content` as an `attr_reader` for consistency?

benbaumann95 (commented):
Does this letter being changed also need to reflect in the email we send to the applicant/agent?

I think these need to be identical right

cdccollins:
> Does this letter being changed also need to reflect in the email we send to the applicant/agent?
> 
> I think these need to be identical right

Ah forgot about that, thanks

benjamineskola (approved):

PR #1150, opened 2023-07-06
Change portal_name to section_name

cdccollins:
PlanX have changed the name because it was buggy on their end.

They want to switch off `portal_name` altogether so we need to change it on our end

https://trello.com/c/WVyDS17G/1761-chore-change-sectionname-to-portalname-in-proposaldetails

### DIFF
--- a/app/components/proposal_details/groupable.rb
+++ b/app/components/proposal_details/groupable.rb
@@ -12,7 +12,7 @@ module ProposalDetails
 
     attr_reader :group
 
-    delegate :portal_name, :proposal_details, to: :group
+    delegate :section_name, :proposal_details, to: :group
 
     def auto_answered?
       proposal_details.all?(&:auto_answered?)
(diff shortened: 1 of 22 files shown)


app/components/proposal_details/groupable.rb
```diff
@@ -27,13 +27,13 @@ def title
     end
 
     def name
-      case portal_name
+      case section_name
       when "_root"
```
benjamineskola, line 31: Is this identifier still going to be the same under the new structure?

cdccollins: lol I have no idea, I just did a replace all to see how many tests would break. Amazingly none. Will now try and work out what's going on in a bit more detail 😂

benjamineskola: I guess they haven't actually given us an example of what the new structure will look like in practice. (Unless they've already started sending it and we can look at that.)

cdccollins: They have! It's alongside the old one. It looks pretty similar but the names are a bit "cleaner"

benjamineskola (approved):

PR #1154, opened 2023-07-07
Don't try rendering a nonexistent template

benjamineskola:
### Description of change

### Story Link

https://link-to-issue

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


app/controllers/sitemaps_controller.rb
```diff
@@ -41,7 +41,7 @@ def update
                       notice: t(".success")
         end
       else
-        format.html { render :edit }
+        redirect_to edit_planning_application_validation_tasks_path(@planning_application)
```
cdccollins, line 44: is this an existing path? I just tried searching for it in the routes but couldn't find it.

benjamineskola: No, that was total guesswork 😭

app/controllers/sitemaps_controller.rb
```diff
@@ -41,7 +41,7 @@ def update
                       notice: t(".success")
         end
       else
-        format.html { render :edit }
+        redirect_to edit_planning_application_red_line_boundary_change_path(@planning_application)
```
cdccollins, line 44: I found edit_planning_application_red_line_boundary_change_validation_request_path but not that one? You could just re-render the page it was on, which I guess is the show page, with a flash message?

cdccollins (approved):
The message could be a little bit more helpful 😂

benjamineskola:
> The message could be a little bit more helpful 😂

Not my proudest moment in terms of content design but I'm not sure how to get a list of what the actual errors are — I presume it's possible because Rails does it automatically for ordinary forms.

app/controllers/sitemaps_controller.rb
```diff
@@ -41,7 +41,8 @@ def update
                       notice: t(".success")
         end
       else
-        format.html { render :edit }
+        flash.now[:alert] = "Invalid parameter"
```
benbaumann95, line 44: Can we dynamically show what the error message was here ? I'm not sure if anything gets added to `@planning_application.errors` for example

benbaumann95: Unless it should never reach this block and in that case a "please contact support" message might be helpful to indicate an action can be taken by the user

PR #1158, opened 2023-07-11
Merge Apis::PlanX::Client and Apis::PlanX::Query

benjamineskola:
### Description of change

I originally created this based on the pattern we've used with the other APIs, but I don't think there's actually much value in splitting this into two classes. It's overcomplicated and isn't clear where the responsibilities of one end and the other begins.

At least currently, an instance of `Query` is only ever used to make a single query — so there can't be any benefit of storing an instance of `Client` to reuse it, we'll always have to reinstantiate the `Client` object.

(Using a module therefore might actually be more resource-efficient, because AIUI it won't be instantiated each time.)

### Story Link

https://link-to-issue

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


cdccollins (approved):

benbaumann95:
Would you mind adding a spec to check if there is an error with saving. I think similar has been done here,

https://github.com/unboxed/bops/blob/main/spec/services/constraints_creation_service_spec.rb#L110-L131

EGiataganas (approved):

PR #1165, opened 2023-07-13
Link status panels to appropriate filters

benjamineskola:
### Description of change

Link the status panels to the appropriate filter, or to the unfiltered view if the filter is already enabled.

### Story Link

https://trello.com/c/3J0a9MnX/1772-smart-filtering-filter-by-clicking-on-status-bar

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


app/views/planning_applications/index.html.erb
```diff
@@ -35,13 +35,17 @@
...
+      <a class="status-filter-link" href="<%= planning_applications_path %><% unless params[:status]&.include?('to_be_reviewed') && params[:application_type].blank?  %>?status%5B%5D=to_be_reviewed<% end %>">
+        <h3 class="govuk-heading-s"><%= t ".reviewer_requests.title" %></h3>
+        <p class="govuk-heading-m"><%= t ".reviewer_requests", count: @search.current_planning_applications.to_be_reviewed.count %></p>
+      </a>
     </div>
-    <div class="status-panel" id="not-started">
-      <h3 class="govuk-heading-s"><%= t ".prior_approvals.title" %></h3>
-      <p class="govuk-heading-m"><%= t ".prior_approvals", count: @search.current_planning_applications.prior_approvals.not_started.count %></p>
+    <div class="status-panel<% if params[:application_type]&.include?('prior_approval') && params[:status].blank? %> status-panel--highlighted<% end %>" id="not-started">
+      <a class="status-filter-link" href="<%= planning_applications_path %><% unless params[:application_type]&.include?('prior_approval') && params[:status].blank?  %>?application_type%5B%5D=prior_approval<% end %>">
```
benbaumann95, line 45: Could we not just do something like

```<a class="status-filter-link" href="<%= planning_applications_path(status: ["not_started"], application_type: ["prior_approval"]) %>">```

**Review by benbaumann95 — COMMENTED** [review]
Looks good but could we please just add a system test to this (check the expected planning applications are showing and also the relevant filter checkboxes are selected, with the others unselected.

Or possibly at least unit testing the `filter_enabled_uniquely?` with a couple examples

**Comment by benjamineskola** [comment]
I was going to say that because all this does is use the normal search filters it's probably fine to just rely on the existing search tests, but it seems like the tests for which applications are being shown when searching/filtering are some of the flaky ones. I've checked that the panel is highlighted when selecting a relevant filter, and I think we can rely on the tests for the `SearchFilterService` if not the Capybara search tests.

File: app/views/planning_applications/index.html.erb [hunk 2]
```diff
@@ -35,13 +35,17 @@
...
-      <h3 class="govuk-heading-s"><%= t ".prior_approvals.title" %></h3>
-      <p class="govuk-heading-m"><%= t ".prior_approvals", count: @search.current_planning_applications.prior_approvals.not_started.count %></p>
+    <div class="status-panel<% if filter_enabled_uniquely?(application_type: 'prior_approval') %> status-panel--highlighted<% end %>" id="not-started">
+      <a class="status-filter-link" href="<%= planning_applications_path(filter_enabled_uniquely?(application_type: "prior_approval") ? {} : { application_type: [:prior_approval] }) %>">
+        <h3 class="govuk-heading-s"><%= t ".prior_approvals.title" %></h3>
+        <p class="govuk-heading-m"><%= t ".prior_approvals", count: @search.current_planning_applications.prior_approvals.not_started.count %></p>
+      </a>
     </div>
   </div>
 </div>
```
EGiataganas, line 51: As an improvement I think if we don't have any `planning_applications.to_be_reviewed` and `prior_approvals. not-started`

EGiataganas: Also I think it would make sense to move it in a component and we won't have the method in helpers

benjamineskola: > As an improvement I think if we don't have any `planning_applications.to_be_reviewed` and `prior_approvals. not-started`

I'm not sure what you mean here?

> Also I think it would make sense to move it in a component and we won't have the method in helpers

I thought it might be better to do this only when we need to dynamically create the filter links. Right now there will only ever be precisely two, statically defined.

EGiataganas (approved):

PR #1168, opened 2023-07-13
Filter by planning application type

benbaumann95:
### Description of change

- Filter by planning application type
- Also adds indexes for better search/filtering optimisation

### Story Link

https://trello.com/c/DDBkmVaD/1744-add-filters-for-planning-application-type

### Screenshots

Screenshot 2023-07-13 at 12 11 52

### DIFF
--- a/app/components/planning_applications/search_component.html.erb
+++ b/app/components/planning_applications/search_component.html.erb
@@ -3,7 +3,7 @@
   scope: '',
   builder: GOVUKDesignSystemFormBuilder::FormBuilder,
   method: :get,
-  url: planning_applications_path(anchor: panel_type, status: selected_status),
+  url: planning_applications_path(anchor: panel_type, status: selected_status, application_type: selected_application_type),
   data: { controller: "clear-form" }
 ) do |form| %>
   <%= form.govuk_error_summary(presenter: ErrorPresenter) %>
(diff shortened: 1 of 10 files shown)


app/services/planning_application_search.rb
```diff
@@ -143,4 +143,16 @@ def query_terms
...
+
+  def status_type
+    status&.reject(&:empty?)
+  end
+
+  def filtered_scope(scope = current_planning_applications)
+    scope.where(status: [status_type], application_type: [selected_application_type_ids])
+  end
+
+  def selected_application_type_ids
```
benbaumann95, line 155: Since it's an association, have to search by id rather than name

benjamineskola (approved):

app/services/planning_application_search.rb
```diff
@@ -7,7 +7,10 @@ class PlanningApplicationSearch
   STATUSES = %w[not_started invalidated in_assessment awaiting_determination to_be_reviewed].freeze
   REVIEWER_STATUSES = %w[awaiting_determination to_be_reviewed].freeze
 
+  APPLICATION_TYPES = %w[lawfulness_certificate prior_approval].freeze
+
   attribute :view, :enum, values: %w[all mine], default: "mine"
+  attribute :application_type, :list
```
EGiataganas, line 13: how about adding the default here?
```
attribute :application_type, :list, default: APPLICATION_TYPES
```

also after merging the PR #1116 we can do `ApplicationType::NAME_ORDER`

app/services/planning_application_search.rb
```diff
@@ -16,10 +19,11 @@ class PlanningApplicationSearch
 
   define_model_callbacks :initialize, only: :after
 
-  after_initialize :init_status
+  after_initialize :init_filter_options
 
-  def init_status
+  def init_filter_options
     self.status ||= statuses
+    self.application_type ||= application_types
```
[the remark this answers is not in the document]
EGiataganas: I've had to do the assignment of status like that as it depends on a method and at the moment it was failing using lamda

app/components/planning_applications/search_component.html.erb
```diff
@@ -34,13 +34,25 @@
         <div class="govuk-accordion__section-header">
           <h3 class="govuk-accordion__section-heading">
             <button type="button" class="govuk-accordion__section-button">
-              <%= t(".status_filter", selected: selected_status_count, total: total_status_count) %>
+              Filters
```
EGiataganas, line 37: How about improving and showing how many application types filters are selected as well?

db/migrate/20230713085643_add_indexes_to_planning_applications.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddIndexesToPlanningApplications < ActiveRecord::Migration[7.0]
+  def change
+    add_index :planning_applications, %i[status application_type_id]
+    add_index :planning_applications, [:status]
+  end
+end
```
EGiataganas, line 8: 🎉

app/services/planning_application_search.rb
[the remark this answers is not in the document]
benbaumann95: Yeah just tried to be consistent and thought we might want to show different application types based on the user or add some condition in the future. And i'm calling `application_types` in the component rather than calling the constant like `PlanningApplicationSearch::APPLICATION_TYPES`

app/components/planning_applications/search_component.html.erb
benbaumann95: I think this was no longer needed so i've removed it but can confirm with the POs

app/services/planning_application_search.rb
benbaumann95: Happy for you to add this change in your PR 👍

PR #1203, opened 2023-08-04
Make 'send letters' status in list reflect state of letters

benjamineskola:
### Description of change

"Send letters to neighbours" step status previously reflected the state of the consultation as a whole: it would only show as complete when the consultation was marked as complete. (I don't know when this actually happens, incidentally.)

It is now marked as complete when any letters have been sent.

### Story Link

https://trello.com/c/G1ONu15P/1840-bt-when-letters-sent-show-send-letters-to-neighbours-task-as-complete-not-in-progress

### Decisions

The actual state I've used for the letters here is that they are _created_, i.e., sent _to Notify_ — this is regardless of whether Notify has posted them, because that seems like something the user has no control over and thus irrelevant.

There's no really practical way of distinguishing between "some letters have been sent" and "all letters have been sent", because we don't have a concept of what "all letters" means. If the user sends a letter to only one neighbour it will be marked as complete, even if they need to send letters to more neighbours later. This is the user's responsibility, I think.

### DIFF
--- a/app/components/task_list_items/send_letters_to_neighbours_component.rb
+++ b/app/components/task_list_items/send_letters_to_neighbours_component.rb
@@ -31,7 +31,13 @@ module TaskListItems
 
     def status_tag_component
       StatusTags::BaseComponent.new(
-        status: @planning_application.consultation&.status || "not_started"
+        status: if @planning_application.consultation&.neighbour_letters_failed?
+                  "failed"
+                elsif @planning_application.consultation&.neighbour_letters_sent?
+                  "complete"
+                else
+                  "not_started"
+                end
       )
     end
   end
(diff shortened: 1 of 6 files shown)


EGiataganas (approved):

app/components/task_list_items/send_letters_to_neighbours_component.rb
```diff
@@ -31,7 +31,13 @@ def link_path
 
     def status_tag_component
       StatusTags::BaseComponent.new(
-        status: @planning_application.consultation&.status || "not_started"
+        status: if @planning_application.consultation&.neighbour_letters_failed?
+                  "failed"
+                elsif @planning_application.consultation&.neighbour_letters_sent?
```
benbaumann95, line 36: same

benbaumann95 (approved):
👍  Just a minor comment

PR #1207, opened 2023-08-07
Neighbours edit

EGiataganas:
### Description of change

- Show neighbours submitted by the applicant on the consultations edit page when another neighbour is removed
- Added sample of proposal details questions with neighbours

### Story Link

https://trello.com/c/VtgduOqH

### Screenshots

neighbours added

### DIFF
--- a/app/views/planning_application/consultations/edit.html.erb
+++ b/app/views/planning_application/consultations/edit.html.erb
@@ -24,6 +24,8 @@
       }
     ) %>
 
+    <%= render "neighbours_added", planning_application: @planning_application, consultation: @consultation %>
+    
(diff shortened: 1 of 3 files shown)


app/views/planning_application/consultations/edit.html.erb
```diff
@@ -24,6 +24,8 @@
       }
     ) %>
 
+    <%= render "neighbours_added", planning_application: @planning_application, consultation: @consultation %>
```
benjamineskola, line 27: Ah so this was only in the `new` and not `edit` pages? Good catch!

benjamineskola (approved):

PR #1210, opened 2023-08-07
Add warning message if consultation is still ongoing

benbaumann95:
### Story Link

https://trello.com/c/awbPSSDk/1842-bt-warn-assessor-reviewer-if-recommendation-is-before-21-days-publicity-time-expired

### Screenshots

Screenshot 2023-08-07 at 14 39 44

### DIFF
--- a/spec/models/consultation_spec.rb
+++ b/spec/models/consultation_spec.rb
@@ -138,4 +138,45 @@ RSpec.describe Consultation do
       expect(consultation.neighbour_responses_by_summary_tag).to eq({ "objection" => 1, "supportive" => 2, "neutral" => 1 })
     end
   end
+
+  describe "publicity_active?" do
+    let(:consultation) { create(:consultation, end_date: Time.zone.local(2023, 9, 23, 13)) }
+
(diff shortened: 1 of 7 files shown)


spec/models/consultation_spec.rb
```diff
@@ -138,4 +138,37 @@
       expect(consultation.neighbour_responses_by_summary_tag).to eq({ "objection" => 1, "supportive" => 2, "neutral" => 1 })
     end
   end
+
+  describe "publicity_active?" do
```
benjamineskola, line 142: Would a test for the behaviour when the end date is today be useful?

benbaumann95: I think we'd still want this warning if the end date is today as technically it should go on until the end of day right

benjamineskola: Yeah, I think the behaviour seems correct: "greater than today".

benbaumann95: Added a spec to compare the actual day rather than time in this instance.

PR #1217, opened 2023-08-09
Show document count on site visit

cdccollins:
### Description of change

- Show count of documents on site visit element
- Remove time from when site visit happened on show page. User doesn't input this so it always shows as midnight which isn't very helpful

### Story Link

https://trello.com/c/RPMVpWMw/1852-bt-make-it-clear-if-there-are-photographs-associated-with-a-site-visit-response

### Screenshots

Screenshot 2023-08-09 at 11 20 28

### DIFF
--- a/app/views/planning_application/site_visits/show.html.erb
+++ b/app/views/planning_application/site_visits/show.html.erb
@@ -18,7 +18,7 @@
 <p class="govuk-body">Response created by: <%= @site_visit.created_by.name %></p>
 <p class="govuk-body">Response created at: <%= @site_visit.created_at.to_fs %></p>
 <% if @site_visit.decision? %>
-  <p class="govuk-body">Visited at: <%= @site_visit.visited_at&.to_fs %></p>
+  <p class="govuk-body">Visited at: <%= @site_visit.visited_at&.to_date.to_fs %></p>
 <% end %>
 <p class="govuk-body">Comment: <%= @site_visit.comment %></p>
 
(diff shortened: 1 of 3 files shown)


app/views/planning_application/site_visits/show.html.erb
```diff
@@ -18,7 +18,7 @@
 <p class="govuk-body">Response created by: <%= @site_visit.created_by.name %></p>
 <p class="govuk-body">Response created at: <%= @site_visit.created_at.to_fs %></p>
 <% if @site_visit.decision? %>
-  <p class="govuk-body">Visited at: <%= @site_visit.visited_at&.to_fs %></p>
+  <p class="govuk-body">Visited at: <%= @site_visit.visited_at&.strftime("%-d %B %Y") %></p>
```
benjamineskola, line 21: How about this instead:
```suggestion
  <p class="govuk-body">Visited at: <%= @site_visit.visited_at&.to_date.to_fs %></p>
```

This way it's using the configured formats in `config/initializers/date_formatter.rb` and we can ensure consistency.

benjamineskola: I just looked and we're using at least 17 different date formats on top of configured ones, so maybe I'm worrying unnecessarily.

cdccollins: Ah didn't realise that was what was happening 👍

benjamineskola (approved):

PR #1222, opened 2023-08-14
Create endpoint for neighbour responses

cdccollins:
### Description of change

BoPS end of direct uploading of neighbour responses. We need an endpoint that they can be posted to!

### Story Link

https://trello.com/c/QCtuDUfm/1875-see-the-comments-received-from-neighbours-via-a-form

### Screenshots

Screenshot 2023-08-30 at 12 15 46


app/services/neighbour_response_creation_service.rb
```diff
@@ -0,0 +1,53 @@
...
+    options.each { |k, v| instance_variable_set("@#{k}", v) unless v.nil? }
+  end
+
+  def call
+    save_neighbour_response!(build_neighbour_response)
+  end
+
+  private
+
+  attr_reader :local_authority, :params, :api_user
```
benbaumann95, line 17: Are you using `local_authority` or `api_user` ?
And could add planning_application to drop the @ below for consistency

app/services/neighbour_response_creation_service.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    (neighbour.presence || @planning_application.consultation.neighbours.build(
+      address: neighbour_response_params[:address], selected: false
+    ))
+  end
+
+  def save_neighbour_response!(neighbour_response)
+    NeighbourResponse.transaction do
+      neighbour_response.save!
+    end
```
benbaumann95, line 43: Not sure you need this transaction if all you are doing is calling save! on neighbour_response

app/services/neighbour_response_creation_service.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    neighbour_response
+  rescue ActiveRecord::RecordInvalid, ArgumentError, NoMethodError => e
```
benbaumann95, line 46: Not sure we'd ever see a `ArgumentError` or `NoMethodError` being raised here

app/services/neighbour_response_creation_service.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+class NeighbourResponseCreationService
+  class CreateError < StandardError; end
+
+  def initialize(planning_application:, **options)
+    @planning_application = planning_application
+    options.each { |k, v| instance_variable_set("@#{k}", v) unless v.nil? }
```
benbaumann95, line 8: I know I did something similar with the `PlanningApplicationCreationService` but I think we are only ever passing `params` and `planning_application` to initialize this service right?

app/views/planning_application/neighbour_responses/new.html.erb
```diff
@@ -48,16 +47,17 @@
   <% end %>
   <%= form.govuk_text_field :name %>
   <%= form.govuk_text_field :email %>
-  <%= form.govuk_select :address, @consultation.neighbours.select(&:persisted?).map(&:address), options: { include_blank: true }, label: { text: "Select an existing neighbour address" } %>
+  <%= form.govuk_select :address, @consultation.neighbours.select(&:persisted?).select(&:selected?).map(&:address), options: { include_blank: true }, label: { text: "Select an existing neighbour address" } %>
```
benbaumann95, line 50: Is this query something we could add to the model as a class method perhaps

`@consultation.neighbours.select(&:persisted?).select(&:selected?).map(&:address)`

benbaumann95 (approved):

PR #1224, opened 2023-08-15
Sort dashboard by created date

benjamineskola:
### Description of change

Order search results by created date, which should approximate the time remaining to their deadline.

### Story Link

https://trello.com/c/1dhrEPZm/1831-dashboard-showing-cases-out-of-date-order

### DIFF
--- a/app/services/planning_application_search.rb
+++ b/app/services/planning_application_search.rb
@@ -149,7 +149,7 @@ class PlanningApplicationSearch
   end
 
   def filtered_scope(scope = current_planning_applications)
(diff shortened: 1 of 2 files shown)


benbaumann95 (approved):
Could we please just add a tiny spec for this with something that would have failed prior to this change 🙏

benjamineskola:
> something that would have failed prior to this change

Setting `received_at` to nil in the test data, and then checking for precise ordering rather than presence, seems to fulfil that.

PR #1230, opened 2023-08-17
Fix bops-applications docker

benbaumann95:
### Description of change

### Story Link

https://link-to-issue

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -64,8 +64,10 @@ services:
         source: .
         target: /app
   applicants:
-    build: ./bops-applicants
-    entrypoint: ./docker-entrypoint.sh
+    build:
+      context: ./bops-applicants/docker/ruby
+      dockerfile: Dockerfile
+    entrypoint: /home/rails/bops-applicants/docker/ruby/docker-entrypoint.sh
     stdin_open: true
     tty: true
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -64,8 +64,10 @@ services:
         source: .
         target: /app
   applicants:
-    build: ./bops-applicants
-    entrypoint: ./docker-entrypoint.sh
+    build:
+      context: ./bops-applicants/docker/ruby
+      dockerfile: Dockerfile
+    entrypoint: /home/rails/bops-applicants/docker/ruby/docker-entrypoint.sh
```
benjamineskola, line 70: Looking at the bops-applicants docker file https://github.com/unboxed/bops-applicants/blob/main/docker/ruby/Dockerfile#L31 `/bin/docker-entrypoint.sh` might work here too?

benbaumann95: This makes sense but doesn't work for me hm

benjamineskola: I don't think it's worth trying to figure out why, in that case; let's just go with this as it is.

benjamineskola (approved):

PR #1233, opened 2023-08-22
Allow marking applications as CIL-liable

benjamineskola:
### Description of change

Add database field and form for editing CIL liability status.

### Story Link

https://trello.com/c/bJFCoQ00/1803-record-an-application-if-its-cil-liable

### Screenshots

Screenshot 2023-08-30 at 16 53 54

Screenshot 2023-08-30 at 16 54 38

### Known issues [OPTIONAL]

The "Save and come back later" button might actually be meaningless for this.

### DIFF
--- a/app/components/task_list_items/cil_liability_component.rb
+++ b/app/components/task_list_items/cil_liability_component.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module TaskListItems
+  class CilLiabilityComponent < TaskListItems::BaseComponent
+    def initialize(planning_application:)
+      @planning_application = planning_application
+    end
+
+    private
+
+    attr_reader :planning_application
+
+    delegate(:cil_liable, to: :planning_application)
+
+    def link_text
+      t(".link_text")
+    end
+
+    def link_path
+      edit_planning_application_cil_liability_path(
+        planning_application
+      )
+    end
+
+    def status
+      case cil_liable
+      when TrueClass
+        :cil_liable
+      when FalseClass
+        :not_cil_liable
+      when NilClass
+        :not_started
+      else
+        raise "Unexpected value for `cil_liable': #{cil_liable.inspect}"
+      end
+    end
+  end
+end
--- a/app/controllers/planning_application/cil_liability_controller.rb
+++ b/app/controllers/planning_application/cil_liability_controller.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+class PlanningApplication
+  class CilLiabilityController < ApplicationController
+    include PlanningApplicationAssessable
+    before_action :set_planning_application
+    before_action :ensure_planning_application_is_validated
+
+    def edit
+      respond_to do |format|
+        format.html
+      end
+    end
+
+    def update
+      if @planning_application.update(cil_liability_params)
+        redirect_to planning_application_assessment_tasks_path(@planning_application), notice: t(".success")
+      else
+        render :edit
+      end
+    end
+
+    private
+
+    def planning_application_id
+      Integer(params[:planning_application_id])
+    end
+
+    def planning_applications_scope
+      current_local_authority.planning_applications
+    end
+
+    def set_planning_application
+      planning_application = planning_applications_scope.find(planning_application_id)
+
+      @planning_application = PlanningApplicationPresenter.new(view_context, planning_application)
+    end
+
+    def cil_liability_params
+      params.require(:planning_application).permit([:cil_liable])
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -145,6 +145,7 @@ Rails.application.routes.draw do
       resources :review_immunity_details, only: %i[edit update show]
 
       resources :permitted_development_rights, only: %i[new create edit update show]
+      resource :cil_liability, only: %i[edit update], controller: :cil_liability
 
       resources :review_immunity_enforcements, only: %i[show edit update]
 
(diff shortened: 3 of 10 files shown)


app/controllers/planning_application/cil_liability_controller.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class PlanningApplication
+  class CilLiabilityController < ApplicationController
+    include PlanningApplicationAssessable
+    before_action :set_planning_application
+    before_action :ensure_planning_application_is_validated
+
+    def new
+      warn "in new"
```
EGiataganas, line 10: Do we need this warn?

app/controllers/planning_application/cil_liability_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+      respond_to do |format|
+        format.html
+      end
+    end
+
+    def edit
+      warn "in edit"
```
EGiataganas, line 17: Do we need this warn?

app/controllers/planning_application/cil_liability_controller.rb
benjamineskola: This was due to me trying to understand `config/routes.rb` 😅

config/routes.rb
```diff
@@ -145,6 +145,9 @@
       resources :review_immunity_details, only: %i[edit update show]
 
       resources :permitted_development_rights, only: %i[new create edit update show]
+      get "/cil_liability/edit", action: :edit, controller: "cil_liability"
+      post "/cil_liability", action: :update, controller: "cil_liability"
+      patch "/cil_liability", action: :update, controller: "cil_liability"
```
benbaumann95, line 150: Can we not use the regular `resources` style of routing here

Perhaps need to do `resource` if the edit/update actions aren't actually modifying a new instance of cil_liability

benjamineskola: I couldn't figure out the syntax for `resources` or `resource` — rails seems to be trying to be too clever. (Perhaps because `cil_liability` isn't a model, just a property of `PlanningApplication`. Rails wanted to add an ID to the url for the particular instance of `cil_liability`, which of course doesn't exist.)

benbaumann95: Yeah I think that's why I added before something like `resource :legislation, only: %i[show update]` because legislation isn't actually a model and using resources was trying to add a legislation_id in the routing

benbaumann95: Maybe you've already seen but this gives a quick breakdown - https://www.rubyinrails.com/2019/04/16/rails-routes-difference-between-resource-and-resources/

benjamineskola: >Moreover, it still directs requests to pluralized controller name.

Ah this was part of the issue I was having: it was pluralising even when I didn't want it to. But maybe I can use `resource` and only override the controller name.

app/components/task_list_items/cil_liability_component.rb
```diff
@@ -0,0 +1,38 @@
...
+        planning_application
+      )
+    end
+
+    def status
+      if cil_liable.nil?
+        :not_started
+      else
+        "cil_liable_#{cil_liable}".to_sym
+      end
```
pixeltrix, line 30: Since `cil_liable` is a boolean, dynamically creating strings which will need to be GC'd and converting them to symbols doesn't seem the best way to implement this, how about:

```
def status
  case cil_liable
  when TrueClass
    :cil_liable_true
  when FalseClass
    :cil_liable_false
  when NilClass
    :not_started
  else
    raise RuntimeError, "Unexpected value for `cil_liable': #{cil_liable.inspect}"
  end
end
```

Also, would the statuses not be better as `not_started`, `cil_liable`, `not_cil_liable` ?

app/components/task_list_items/cil_liability_component.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    private
+
+    attr_reader :planning_application
+
+    delegate(:cil_liable, to: :planning_application)
+
+    def link_text
+      "CIL Liability"
```
pixeltrix, line 16: Should we not be extracting UI text to a locale file? Understand we don't want to spend time in a sprint to go back and change things but we probably want to be proactive about not making things worse.

app/controllers/planning_application/cil_liability_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    def edit
+      respond_to do |format|
+        format.html
+      end
+    end
+
+    def update
+      @planning_application.cil_liable = cil_liability_params[:cil_liable]
+      @planning_application.save!
```
benbaumann95, line 17: Could we not do the conventional

```
if @planning_application.update(cil_liability_params)
  redirect_to planning_application_assessment_tasks_path(@planning_application)
else
  render :edit
end
```

 Or we should handle the save! since it will just leave the user with a broken page

benbaumann95: And we'd probably want to add a success message similar to how we do in `en.yml` with

```
update:
  success:
```

benbaumann95 (approved):

app/components/task_list_items/cil_liability_component.rb
```diff
@@ -0,0 +1,38 @@
...
+      if cil_liable.nil?
+        :not_started
+      else
+        "cil_liable_#{cil_liable}".to_sym
+      end
+    end
+
+    def to_be_reviewed?
+      planning_application.recommendation&.rejected? &&
+        permitted_development_right.update_required?
```
pixeltrix, line 35: I notice there's a few of these `&.` for the `recommendation` across the application - fine to leave this for now but something to revisit everywhere on another day as it's not good OOP pattern to have business logic in a view component.

PR #1234, opened 2023-08-23
Query neighbour addresses by polygon search

benbaumann95:
### Description of change

- Polygon geosjon used to query OS places API and return all addresses (up to a limit of 100) within that defined area
- The returned addresses can be reviewed and individually removed before they are persisted to the database
- The drawn polygon is also saved in a different colour so on subsequent visits you can see this on the map
- Since the geojsonChange event triggers for every coordinate change, a debounce function has been added so we only trigger an API request at most once per second
- A unique constraint is added on a combination of neighbour address and consultation id so we do not have any duplicates

For PR review, I would advise pulling the branch down locally and testing it around for yourself.

### Story Link

https://trello.com/c/HrEFGrSL/1884-search-neighbour-addresses-using-polygon

### Screenshots

> ### **Loading Spinner**

Screenshot 2023-08-24 at 17 27 43

> ### **List of returned addresses**

Screenshot 2023-08-24 at 17 25 00

> ### **Error if address already added**
Screenshot 2023-08-24 at 11 46 55

> ### **Previously drawn polygon**
Screenshot 2023-08-24 at 17 29 19


app/javascript/controllers/polygon_search_controller.js
```diff
@@ -0,0 +1,215 @@
...
+  onGeojsonChange = ({ detail: geoJSON }) => {
+    const geoJsonFeatures = geoJSON.features
+    if (geoJsonFeatures === undefined) {
+      return
+    }
+
+    this.handleGeojsonChange(geoJsonFeatures[0])
+  }
+
+  clearAddresses = () => {
```
benjamineskola, line 84: Really minor point overall, but would we be better off using the same function syntax where possible? I think these are equivalent in this context.

```suggestion
  clearAddresses() {
```

app/javascript/controllers/polygon_search_controller.js
```diff
@@ -0,0 +1,215 @@
...
+
+  hideLoadingSpinner() {
+    const spinner = document.querySelector(".loading-spinner")
+
+    if (spinner) {
+      spinner.remove()
+    }
+  }
+
+  removeHiddenAddressInputs = () => {
```
benjamineskola, line 200:
```suggestion
  removeHiddenAddressInputs() {
```

app/javascript/controllers/polygon_search_controller.js
```diff
@@ -0,0 +1,215 @@
...
+    const hiddenInputs = document.querySelectorAll(
+      'input[type="hidden"][name="addresses[]"]',
+    )
+
+    hiddenInputs.forEach((input) => input.parentNode.removeChild(input))
+  }
+
+  debounce = (func, wait) => {
```
benjamineskola, line 208:
```suggestion
  debounce(func, wait) {
```

app/javascript/controllers/polygon_search_controller.js
```diff
@@ -0,0 +1,215 @@
...
+    })
+
+    if (!response.ok) {
+      throw new Error(`HTTP error with status: ${response.status}`)
+    }
+
+    return response
+  }
+
+  onGeojsonChange = ({ detail: geoJSON }) => {
```
benjamineskola, line 75:
```suggestion
  onGeojsonChange({ detail: geoJSON }) {
```

app/javascript/controllers/polygon_search_controller.js
benbaumann95: Ha didn't even notice. Yeah better to keep consistent.
Is this something we could add to our js linter 
 Rome ?

benjamineskola: Just looked and I couldn't find a linter rule for it. (We currently use their recommended set, which expands occasionally when new rules are stable enough, but there doesn't look like any of the experimental rules check for this yet. It may be something they add in the future.)

benbaumann95: Hm the specs fail when I change to this.

I think it's to do with the interaction with the this keyword.
i think for now I'll leave as is since it works as expected but can make a note to refactor

EGiataganas (approved):

app/javascript/controllers/polygon_search_controller.js
benjamineskola: Oh yeah, that makes sense — arrow functions don't get a new `this`. We'd definitely be able to rewrite it but it wouldn't be as simple as I had expected.

PR #1243, opened 2023-09-04
Improve layout for querying neighbour addresses by polygon search

benbaumann95:
### Description of change

This PR allows multiple geojson features to be stored on a single map as part of a "FeatureCollection". This enables us to have the consultation polygon overlay the planning application red line boundary. In theory, if we allowed multiple geojson features for the polygon search we could combine this in one map too. We are also able to draw in a specified colour using drawColor and drawFillColor props.

The colour is now set differently for the polygon boundary to distinguish from the red line boundary. This has been done by setting a colour property attribute on the feature which the map component will dynamically set.

Some updates were necessary on the [map](https://github.com/theopensystemslab/map) component to achieve the above.
See: https://github.com/theopensystemslab/map/pull/381 & https://github.com/theopensystemslab/map/pull/379

### Story Link

https://trello.com/c/IQsb3WNF/1908-improve-content-and-layout-for-querying-neighbour-addresses-by-polygon-search

### **Drawn polygon boundary is saved in another colour**

Screenshot 2023-09-10 at 21 06 37

### **Drawing the consultation polygon in another colour**

Screenshot 2023-08-31 at 16 27 09


app/models/consultation.rb
```diff
@@ -87,6 +91,38 @@ def add_neighbour_addresses!(addresses, geojson)
     raise AddNeighbourAddressesError, e.message
   end
 
+  def polygon_fill_colour
+    "#{polygon_colour}20"
+  end
+
+  def concatenate_geojsons(*geojsons)
```
benbaumann95, line 98: Need this to represent one geosjon FeatureCollection with all the geosjon features (planning application red line boundary, consultation boundary etc) so we can present this on one map.

If the column type is polygon_geojson we insert the polygon colour so that the map component can distinguish the type of feature

app/models/concerns/geojson_formattable.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module GeojsonFormattable
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def format_geojson_epsg(column_name)
```
benbaumann95, line 7: the map component introduces a breaking change here - https://github.com/theopensystemslab/map/pull/256/files

This is to ensure compatibility with the previous way we store geojson

app/controllers/planning_application/consultations_controller.rb
```diff
@@ -150,5 +151,12 @@ def redirect_path
         planning_application_consultation_path(@planning_application, @consultation)
       end
     end
+
+    def set_geosjon_features
```
benjamineskola, line 155:
```suggestion
    def set_geojson_features
```

app/controllers/planning_application/consultations_controller.rb
```diff
@@ -150,5 +151,12 @@ def redirect_path
...
+      return unless (consultation_geosjon = @consultation.polygon_geojson)
```
benjamineskola, line 156:
```suggestion
      return unless (consultation_geojson = @consultation.polygon_geojson)
```

app/views/planning_application/consultations/new.html.erb
```diff
@@ -14,24 +14,16 @@
...
-        }
-      }
-    ) %>
+    <%= render "select_neighbours", geojson: @planning_application.boundary_geojson, consultation: @consultation %>
 
     <%= render "neighbour_addresses_form", planning_application: @planning_application, consultation: @consultation %>
 
-    <%= render "neighbours_added", planning_application: @planning_application, consultation: @consultation %>
-    
+    <p class="govuk-body">or</p>
```
benjamineskola, line 21: Is this meant to say just "or" on its own? It's not very clear in template form (although it might make sense in the context of the partials around it).

app/views/planning_application/consultations/show.html.erb
```diff
@@ -32,30 +32,15 @@
...
 
     <%= render "neighbour_addresses_form", planning_application: @planning_application, consultation: @consultation %>
 
-    <% if geojson = @consultation.polygon_geojson %>
-      <h2 class="govuk-heading-m govuk-!-padding-top-6">
-        Previously drawn area for neighbour addresses search
-      </h2>
-      <%= render "shared/location_map", locals: { geojson: geojson, geojson_color: @consultation.polygon_colour } %>
-    <% end %>
+    <p class="govuk-body">or</p>
```
benjamineskola, line 39: Same here.

app/views/planning_application/consultations/new.html.erb
benbaumann95: Yeah it's just an "or" 😄

app/controllers/planning_application/consultations_controller.rb
```diff
@@ -10,6 +10,7 @@ class ConsultationsController < AuthenticationController
     before_action :assign_params, only: %i[update create]
     before_action :update_letter_statuses, only: %i[show edit]
     before_action :ensure_public_portal_is_active, only: :send_neighbour_letters
+    before_action :set_geosjon_features, only: %i[show edit update]
```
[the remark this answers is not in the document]
benbaumann95: ahh silly me, good spot!

app/javascript/controllers/polygon_search_controller.js
```diff
@@ -73,8 +73,9 @@ export default class extends Controller {
   }
 
   onGeojsonChange({ detail: geoJSON }) {
-    const geoJsonFeatures = geoJSON.features
+    const geoJsonFeatures = geoJSON["EPSG:27700"].features
     if (geoJsonFeatures === undefined) {
+      console.log("No geojson features were found")
```
benjamineskola, line 78: Do we want to be printing output to the console or is this just for debugging?

app/models/consultation.rb
```diff
@@ -87,6 +91,38 @@ def add_neighbour_addresses!(addresses, geojson)
...
+    features = []
+
+    geojsons.compact.each do |geojson|
+      parsed = JSON.parse(geojson)
+
+      # Extract features based on whether the parsed object is a feature or a feature collection
+      if parsed["type"] == "Feature"
+        features << parsed
+      elsif parsed["type"] == "FeatureCollection" && parsed["features"].is_a?(Array)
```
benjamineskola, line 107: I'm a little wary of using `is_a?` here, but I don't really have a better suggestion so I don't think it's a blocker.

Could `parsed["features"]` ever be present but not an array? Maybe `parsed["features"].present?` would be sufficient?

benjamineskola (approved):

app/javascript/controllers/polygon_search_controller.js
benbaumann95: Was left in for debugging so I'll remove

app/models/consultation.rb
benbaumann95: Method has been updated to `polygon_search_and_boundary_geojson`

config/database.yml
```diff
@@ -1,7 +1,7 @@
 default: &default
   adapter: postgis
   encoding: unicode
-  url: <%= ENV["DATABASE_URL"] %>
+  url: <%= ENV.fetch('DATABASE_URL', '').sub(/^postgres/, 'postgis') %>
```
benjamineskola, line 4: How about this:

```suggestion
  url: <%= ENV.fetch('DATABASE_URL', '').sub(%r{^postgres://}, 'postgis://') %>
```

Just makes it a little more obvious that it's the protocol being replaced.

benjamineskola (approved):

PR #1244, opened 2023-09-05
Allow editing consultee responses

benjamineskola:
### Description of change

In order to be able to view consultee responses we need to be able to track them, not just a summary of them.

### Story Link

https://trello.com/c/KYztoXq0/1888-view-all-comments-received-from-consultees

### Screenshots

Added "edit response" link to list of consultees:
Screenshot 2023-09-06 at 10 11 38

Which links to a form:
Screenshot 2023-09-06 at 10 12 10

### DIFF
--- a/app/controllers/consultees_controller.rb
+++ b/app/controllers/consultees_controller.rb
@@ -2,11 +2,26 @@
 
 class ConsulteesController < AuthenticationController
   before_action :set_planning_application
+  before_action :set_consultee, only: %i[edit update]
+
+  def edit
+    respond_to do |format|
+      format.html
+    end
+  end
 
   def create
     consultee = planning_application.consultees.new(consultee_params)
     consultee = planning_application.consultees.new if consultee.save
     render(json: { partial: consultees_partial(consultee) })
+  end
+
+  def update
+    if @consultee.update(consultee_params)
+      redirect_to after_update_consultees_path, notice: t(".success")
+    else
+      render :edit
+    end
   end
 
   def destroy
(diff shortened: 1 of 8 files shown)


app/controllers/consultees_controller.rb
```diff
@@ -3,12 +3,38 @@
 class ConsulteesController < AuthenticationController
   before_action :set_planning_application
 
+  def edit
+    @consultee = planning_application.consultees.find(params[:id])
```
pixeltrix, line 7: If `params[:id]` is an integer we should use `Integer(params[:id])` to ensure that's the case otherwise it may be possible to inject hash arguments to `find`.

app/controllers/consultees_controller.rb
```diff
@@ -3,12 +3,38 @@
...
+      if (consultation_summary = planning_application.consultation_summary)
+        redirect_to edit_planning_application_assessment_detail_path(planning_application, consultation_summary)
+      else
+        # it's possible that the list of consultees is being edited before the
+        # consultation summary has been saved and so we can't link to the edit path
+        redirect_to new_planning_application_assessment_detail_path(
+          planning_application,
+          category: "consultation_summary"
+        )
+      end
```
pixeltrix, line 32: I'd extract out the redirect path into a private helper method to make the logic clearer, e.g.

``` ruby
def update
  @consultee = planning_application.consultees.find(params[:id])

  if @consultee.update(consultee_params)
    redirect_to after_update_consultees_path
  else
    render :edit
  end
end

private

  def after_update_consultees_path
    if consultation_summary = planning_application.consultation_summary
      edit_planning_application_assessment_detail_path(planning_application, consultation_summary)
    else
      # it's possible that the list of consultees is being edited before the
      # consultation summary has been saved and so we can't link to the edit path
      new_planning_application_assessment_detail_path(planning_application, category: "consultation_summary")
    end
  end
```

Also are we not doing 'update successful' type flash messages?

benjamineskola: > Also are we not doing 'update successful' type flash messages?
yes but I keep forgetting them 🤦🏻

app/controllers/consultees_controller.rb
benjamineskola: Doesn't look like `find` takes hash arguments, but it does take array arguments and also string arguments, and string arguments could either be IDs in string format or arbitrary SQL conditions 😱, so… your point stands.

app/controllers/consultees_controller.rb
```diff
@@ -3,12 +3,29 @@
 class ConsulteesController < AuthenticationController
   before_action :set_planning_application
 
+  def edit
+    @consultee = planning_application.consultees.find(Integer(params[:id]))
```
benbaumann95, line 7: nitpick: since this is the same as finding the consultee in the update action we could put this into a private method and have a before_action ?

benbaumann95 (approved):

EGiataganas (approved):

PR #1247, opened 2023-09-07
Add planning_permission full householder application type

benbaumann95:
### Description of change

Add support for planning permission full householder applications.
Currently we receive `"application_type": "Apply for planning permission"` from PlanX so this is a temporary solution until they can send us the proper format 

### Story Link

https://trello.com/c/q26RKA7c/1986-add-householder-application-type-so-we-can-receive-this-from-planx

### DIFF
--- a/app/models/application_type.rb
+++ b/app/models/application_type.rb
@@ -3,7 +3,7 @@
 class ApplicationType < ApplicationRecord
   APPLICATION_STEPS = %i[validation publicity assessment review].freeze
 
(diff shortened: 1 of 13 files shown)


cdccollins (approved):
Do we have JSON we could add to the swagger?

EGiataganas (approved):
I agree with @cdccollins adding a sample json would be great

benbaumann95:
> Do we have JSON we could add to the swagger?

I'll make a card for it. Will ask Holly to submit from PlanX again when this actually works and then we can copy that json over

PR #1248, opened 2023-09-07
Constraints proposed

EGiataganas:
### Description of change

- We need to support the new field `constraints_proposed` from PlanX on our application
- Add and use `constraints_proposed` field on specs and swaggers to create new planning application
- For now we won't allow any constraint creation, until we make a proper data mapping
- Add a new logic and UI for check constraints

### Story Link

https://trello.com/c/yL8eKT6l

### Screenshots

Screenshot 2023-09-28 at 18 57 04
Screenshot 2023-09-28 at 19 02 06


app/services/constraints_creation_service.rb
```diff
@@ -9,26 +9,31 @@ def initialize(planning_application:, constraints_params:, constraints_query: ni
...
+        planx_query: constraint[:planxRequest],
+        planning_data_query: constraint[:sourceRequest]
+      )
+
+      constraint[:constraints].each do |k, v|
+        pa_constraint = planning_application.constraints.find_or_create_by!(
+          type: k.to_s,
+          category: v[:category].parameterize.underscore,
+          local_authority_id: planning_application.local_authority_id
+        )
```
benjamineskola, line 24:
```suggestion
        pa_constraint = planning_application.constraints.find_or_create_by!(
          type: k.to_s,

        ) do |pac|
        pac.category = v[:category].parameterize.underscore,
           pac.local_authority_id = planning_application.local_authority_id
        end
```

app/controllers/constraints_controller.rb
```diff
@@ -2,36 +2,36 @@
...
-    @constraints = @planning_application.constraints
-    planning_application_constraints = @planning_application.planning_application_constraints.includes(:constraint)
-    @constraint_ids = planning_application_constraints.pluck(:constraint_id)
-  end
-
   def update
     ActiveRecord::Base.transaction do
-      constraint_ids.each do |constraint_id|
-        @planning_application_constraints.find_or_create_by!(constraint_id:)
+      @planning_application_constraints.each do |pac|
```
benbaumann95, line 12: Tiny nitpick but is using `constraint` rather than `pac` a bit more readable here

app/models/constraint.rb
```diff
@@ -20,9 +22,27 @@ def type_code
     end
   end
 
+  def start_date
+    nil
+  end
+
+  def description
+    nil
+  end
```
benbaumann95, line 31: Might need to explain this logic to me 😄

app/controllers/constraints_controller.rb
EGiataganas: It's named as `pac`, as is an object of `planning_application_constraints` and not `constraints`

benbaumann95 (approved):

PR #1249, opened 2023-09-07
Enable postGIS extension

benbaumann95:
https://github.com/rgeo/activerecord-postgis-adapter

### Story Link

https://trello.com/c/zVgHb15e/1990-add-postgis-extension-to-bops-for-better-spatial-data-support

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -40,7 +40,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres
+        image: postgis/postgis
         ports: ["5432:5432"]
         env:
           POSTGRES_PASSWORD: postgres
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ source "https://rubygems.org"
 gem "dotenv-rails", require: "dotenv/rails-now"
 
 gem "aasm"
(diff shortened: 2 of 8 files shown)


.github/workflows/testing.yml
```diff
@@ -40,7 +40,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres
+        image: postgis/postgis
         ports: ["5432:5432"]
         env:
           POSTGRES_PASSWORD: postgres
```
EGiataganas, line 46: Previously we had to include the following step after ruby's setup. Not sure if is still needed, good to check 

```
- name: Enable postgis extension
  uses: huaxk/postgis-action@v1
  with:
    postgresql version: '11'
```
cc https://github.com/unboxed/bops/pull/8

benbaumann95: Hey thanks.
CI seems to be running ok though so maybe not needed

benjamineskola: All the `huaxk/postgis-action` seems to do is call docker-run with the `postgis/postgis` image. I'm not sure that we'd get anything from it that setting the image directly doesn't do.

benjamineskola (approved):

EGiataganas (approved):

PR #1250, opened 2023-09-08
Add file upload

cdccollins:
### Description of change

Add files to neighbour responses

### Story Link

https://trello.com/c/YGJ8s8Qe/1938-send-files-in-via-bops-applicants-for-neighbour-responses


app/models/neighbour.rb
```diff
@@ -5,7 +5,7 @@ class Neighbour < ApplicationRecord
   has_one :neighbour_letter, dependent: :destroy
   has_many :neighbour_responses, dependent: :destroy
 
-  validates :address, presence: true, unless: :response_present?
+  validates :address, presence: true, unless: :not_selected?
```
benjamineskola, line 8: Can I just check I understand this logic?

It used to be that they had to have an address unless they had any responses. (Actually I've just realised that I don't understand why we did that.)

Now they need an address if they are selected. I guess this relates to the planning officer being able to deselect some neighbours when searching/adding them?

cdccollins: Hahah the reason it's confusing is that logic that I wrote before didn't actually work. 

This means that a neighbour has to have an address if the planning officer is adding them to the consultation on the "select neighbours" page, but otherwise they don't have to have one. i.e. if a neighbour submits a comment but doesn't want to give their address

benjamineskola: OK, that makes sense and it matches my guess about why it was being done that way around in the first place. 👍🏻

benjamineskola (approved):

PR #1253, opened 2023-09-11
List consultee responses if any exist

benjamineskola:
### Description of change

If any consultee responses exist with a response text attached, list them for ease of writing a summary.

### Story Link

https://trello.com/c/KYztoXq0/1888-view-all-comments-received-from-consultees

### Screenshots

Screenshot 2023-09-11 at 15 21 41

### DIFF
--- a/app/views/planning_application/assessment_details/consultation_summary/_information.html.erb
+++ b/app/views/planning_application/assessment_details/consultation_summary/_information.html.erb
@@ -6,3 +6,24 @@
     read_only: read_only
   }
 ) %>
+
+<% if @planning_application.consultees.with_response.present? %>
+  <h3>Consultation responses</h3>
+  <div class="govuk-accordion" data-module="govuk-accordion">
+    <% @planning_application.consultees.with_response.each do |consultee| %>
+      <div class="govuk-accordion__section">
+        <div class="govuk-accordion__section-header">
+          <h3 class="govuk-accordion__section-heading">
+            <button type="button"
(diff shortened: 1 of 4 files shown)


app/views/planning_application/assessment_details/consultation_summary/_information.html.erb
```diff
@@ -6,3 +6,24 @@
     read_only: read_only
   }
 ) %>
+
+<% if @planning_application.consultees.where.not(response: nil).present? %>
```
benbaumann95, line 10: Can we add this as a scope on the consultee model i.e 

`scope :with_response, -> { where.not(response: nil) }`

guess you can even store this as an instance variable if calling multiple times

app/views/planning_application/assessment_details/consultation_summary/_information.html.erb
```diff
@@ -6,3 +6,24 @@
...
+  <h3>Consultation responses</h3>
+  <div class="govuk-accordion" data-module="govuk-accordion">
+    <% @planning_application.consultees.where.not(response: nil).each do |consultee| %>
```
benbaumann95, line 13: same as above

benbaumann95 (commented):
Could we also add a spec for this pls, thanks

benbaumann95:
> Could we also add a spec for this pls, thanks

Have the POs suggested that this was all that was required for this task?

EGiataganas (approved):

PR #1257, opened 2023-09-12
Remove planx_data from planning applications

benbaumann95:
### Description of change

Remove planx_data from planning applications

Needs to be merged/deployed after https://github.com/unboxed/bops/pull/1256


pixeltrix (approved):
I don't think we need to worry about it but have you investigated what PostgreSQL will do on dropping the column and whether it'll have any performance impact like locking a table? Either way lets deploy this early in the morning just in case.

benjamineskola (approved):

PR #1271, opened 2023-09-19
Resend letters to neighbours

benjamineskola:
### Description of change

Add an option to the "send letters to neighbours page" to resend letters to all existing neighbours rather than add new neighbours. Require a reason to be specified when doing so.

### Story Link

https://trello.com/c/csQTNIyO/1829-repeat-previous-publicity-tasks-at-any-point-of-the-planning-process

### Screenshots

Screenshot_2023-10-04_at_14 55 57

### Known issues

Currently this doesn't prevent the user from both adding new neighbours _and_ choosing to resend all. In this case it'll probably resend to existing neighbours _and_ send to the new neighbours, the latter of whom will be told it's a resend when it isn't (for them).

However it seems like this is an unlikely scenario. In the future we can either prevent the user from trying to do both, _or_ make both work (which would actually be quite easy in terms of the code but not preferable from a UX perspective I think).

### DIFF
--- a/app/controllers/planning_applications/consultations_controller.rb
+++ b/app/controllers/planning_applications/consultations_controller.rb
@@ -22,7 +22,7 @@ module PlanningApplications
     private
 
     def consultation_params
-      params.require(:consultation).permit(:neighbour_letter_text)
+      params.require(:consultation).permit(:neighbour_letter_text, :resend_existing, :resend_reason)
     end
   end
 end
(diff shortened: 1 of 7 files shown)


benjamineskola:
[Repeat previous Publicity tasks at any point of the planning process](https://trello.com/c/csQTNIyO/1829-repeat-previous-publicity-tasks-at-any-point-of-the-planning-process)

EGiataganas (approved):

pixeltrix (approved):

EGiataganas (approved):

PR #1274, opened 2023-09-20
Run apt-get upgrade with chromium

benbaumann95:


Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN apt-get update -y
 RUN apt-get install -y ca-certificates curl gnupg
 
 # Install Chromium for the feature tests
-RUN apt-get install -y --no-install-recommends chromium chromium-driver
+RUN apt-get update && apt-get install -y --no-install-recommends chromium chromium-driver
```
pixeltrix, line 11: Why do we need to run it again since we already run it on line 7?

PR #1276, opened 2023-09-22
Add site notice

cdccollins:
### Description of change

Create site notices in BoPS

### Story Link

https://trello.com/c/oUttvKzu/1868-print-off-the-site-notice-to-publicise-the-proposal

Screenshot 2023-09-29 at 11 44 31
Screenshot 2023-09-29 at 11 50 05

### DIFF
--- a/app/components/task_list_items/confirm_site_notice_component.rb
+++ b/app/components/task_list_items/confirm_site_notice_component.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module TaskListItems
+  class ConfirmSiteNoticeComponent < TaskListItems::BaseComponent
+    def initialize(planning_application:)
+      @planning_application = planning_application
+      @site_notice = @planning_application.site_notices&.last
+    end
+
+    private
+
+    def link_text
+      "Confirm site notice is in place"
+    end
+
+    def link_path
+      if @site_notice.displayed_at.nil?
+        edit_planning_application_site_notice_path(@planning_application, @planning_application.site_notices.last)
+      else
+        planning_application_site_notice_path(@planning_application, @planning_application.site_notices.last)
+      end
+    end
+
+    def status_tag_component
+      if @site_notice.displayed_at.nil?
+        StatusTags::BaseComponent.new(
+          status: "not_started"
+        )
+      elsif @site_notice.displayed_at.present?
+        StatusTags::BaseComponent.new(
+          status: "complete"
+        )
+      end
+    end
+  end
+end
--- a/app/components/task_list_items/confirm_site_notice_component.rb
+++ b/app/components/task_list_items/confirm_site_notice_component.rb
@@ -10,7 +10,7 @@ module TaskListItems
     private
 
     def link_text
-      "Confirm site notice is in place"
+      "Confirm site notice"
     end
 
     def link_path
--- a/app/components/task_list_items/confirm_site_notice_component.rb
+++ b/app/components/task_list_items/confirm_site_notice_component.rb
@@ -9,15 +9,17 @@ module TaskListItems
 
     private
 
+    attr_reader :planning_application, :site_notice
+
     def link_text
-      "Confirm site notice"
+      t(".link_text")
     end
 
     def link_path
       if @site_notice.displayed_at.nil?
-        edit_planning_application_site_notice_path(@planning_application, @planning_application.site_notices.last)
+        edit_planning_application_site_notice_path(planning_application, site_notice)
       else
-        planning_application_site_notice_path(@planning_application, @planning_application.site_notices.last)
+        planning_application_site_notice_path(planning_application, site_notice)
       end
     end
 
@@ -26,7 +28,7 @@ module TaskListItems
         StatusTags::BaseComponent.new(
           status: "not_started"
         )
-      elsif @site_notice.displayed_at.present?
+      else
         StatusTags::BaseComponent.new(
           status: "complete"
         )
--- a/app/components/task_list_items/site_notice_component.rb
+++ b/app/components/task_list_items/site_notice_component.rb
@@ -10,15 +10,11 @@ module TaskListItems
     private
 
     def link_text
-      "Create site notice"
+      "Send site notice"
     end
 
     def link_path
-      if @site_notice.present?
-        planning_application_site_notice_path(@planning_application, @planning_application.site_notices.last)
-      else
-        new_planning_application_site_notice_path(@planning_application)
-      end
+      new_planning_application_site_notice_path(@planning_application)
     end
 
     def status_tag_component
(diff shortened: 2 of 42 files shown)


app/components/task_list_items/confirm_site_notice_component.rb
```diff
@@ -0,0 +1,36 @@
...
+  class ConfirmSiteNoticeComponent < TaskListItems::BaseComponent
+    def initialize(planning_application:)
+      @planning_application = planning_application
+      @site_notice = @planning_application.site_notices&.last
+    end
+
+    private
+
+    def link_text
+      "Confirm site notice"
```
benjamineskola, line 13: Should we put this in the translations file instead?

app/components/task_list_items/confirm_site_notice_component.rb
```diff
@@ -0,0 +1,36 @@
...
+        planning_application_site_notice_path(@planning_application, @planning_application.site_notices.last)
+      end
+    end
+
+    def status_tag_component
+      if @site_notice.displayed_at.nil?
+        StatusTags::BaseComponent.new(
+          status: "not_started"
+        )
+      elsif @site_notice.displayed_at.present?
```
benjamineskola, line 29: I think this can only be `nil` or present, right?

```suggestion
      else
```

app/components/task_list_items/site_notice_component.rb
```diff
@@ -0,0 +1,31 @@
...
+module TaskListItems
+  class SiteNoticeComponent < TaskListItems::BaseComponent
+    def initialize(planning_application:)
+      @planning_application = planning_application
+    end
+
+    private
+
+    def link_text
+      "Send site notice"
```
benjamineskola, line 12: Same here

app/models/planning_application.rb
```diff
@@ -337,6 +338,21 @@ def secure_change_url
...
+  def site_notice_link
+    protocol = Rails.env.production? ? "https" : "http"
+    path = "planning_applications/#{id}/site_notices/download"
+
+    if Bops.env.production?
+      "#{protocol}://planningapplications.#{local_authority.subdomain}.gov.uk/#{path}"
+    else
+      "#{protocol}://#{local_authority.subdomain}.bops-applicants.services/#{path}"
+    end
+  end
```
benjamineskola, line 350: Would it be worth having a variable somewhere for the bops-applicants URL, to ensure we're being consistent about it? I know there's an `APPLICANTS_APP_HOST` variable that's used in a couple of places but not sure if that relates to this (or should).

app/views/planning_application/consultations/index.html.erb
```diff
@@ -55,16 +55,18 @@
...
-          <span class="app-task-list__task-name">
-            <a class="govuk-link" href="#">Confirm site notice</a>
-          </span>
-        </li>
+        <%= render(
+          TaskListItems::SiteNoticeComponent.new(
+            planning_application: @planning_application
+          )
+        ) %>
+        <% if @planning_application.site_notices.any? && @planning_application.site_notices.last.required %>
```
benjamineskola, line 63: Would safe navigation work here?

```suggestion
        <% if @planning_application.site_notices&.last&.required %>
```

app/components/task_list_items/confirm_site_notice_component.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    private
+
+    def link_text
+      "Confirm site notice"
+    end
+
+    def link_path
+      if @site_notice.displayed_at.nil?
+        edit_planning_application_site_notice_path(@planning_application, @planning_application.site_notices.last)
```
benbaumann95, line 18: Looks like you're already setting `@planning_application.site_notices.last` as `@site_notice` above.

Also maybe worth adding `@planning_application` and `@site_notice` as an attr_reader

app/controllers/planning_application/site_notices_controller.rb
```diff
@@ -0,0 +1,132 @@
...
+    def create
+      @site_notice = SiteNotice.new(site_notice_params
+                                    .except(:method, :internal_team_email)
+                                    .merge(planning_application_id: @planning_application.id))
+
+      if @site_notice.save
+        @site_notice.update(content: @site_notice.preview_content(@planning_application))
+
+        if params[:commit] == "Email site notice and mark as complete"
+          if site_notice_params[:internal_team_email].presence
```
benbaumann95, line 36: Is this something they need to fill out each time or would it be consistent enough to add a field on the local_authority table?

app/controllers/planning_application/site_notices_controller.rb
```diff
@@ -0,0 +1,132 @@
...
+      @site_notice = SiteNotice.new
+    end
+
+    def edit
+      respond_to do |format|
+        format.html
+      end
+    end
+
+    def create
```
benbaumann95, line 27: Possibly a lot happening and with several db updates in this controller action so thinking if we could maybe simplify and move things to the model, especially the email sending / auditing?
I'll have a go locally

app/models/document.rb
```diff
@@ -62,6 +63,10 @@ class NotArchiveableError < StandardError; end
     "Site Visit"
   ].freeze
 
+  SITE_NOTICE_TAGS = [
```
benbaumann95, line 66: Had to do similar with Press notice, reckon we can just whack these both in an `OTHER_TAGS` or would you prefer listing it separately

app/views/planning_application/site_notices/new.html.erb
```diff
@@ -0,0 +1,94 @@
...
+    model: [@planning_application, @site_notice],
+    builder: GOVUKDesignSystemFormBuilder::FormBuilder,
+    class: "govuk-!-margin-top-5",
+    method: :post
+  ) do |form| %>
+    <div data-controller="show-hide-form">
+      <div class="display-none">
+        <div id="site-notice-content">
+          <div style="width: 430px;">
+            <%= @site_notice.preview_content(@planning_application.planning_application).html_safe %>
```
benbaumann95, line 44: Ha what is `@planning_application.planning_application`

cdccollins: When we set @planning_applcation in views, it's usually the planning application view, which seems to only load some of the attributes. If you want the full planning application you have to do @planning_application.planning_application 😂

benbaumann95: Ahh is it because we are initialising like 
```
PlanningApplicationPresenter.new(view_context, planning_application)
```

app/components/task_list_items/site_notice_component.rb
```diff
@@ -0,0 +1,31 @@
...
+    def status_tag_component
+      if @planning_application.site_notices.empty?
+        StatusTags::BaseComponent.new(
+          status: "not_started"
+        )
+      else
+        StatusTags::BaseComponent.new(
+          status: "complete"
+        )
+      end
```
benjamineskola, line 28:
```suggestion
      StatusTags::BaseComponent.new(
        status: @planning_application.site_notices.empty? ? "not_started" : "complete"
      )
```

benjamineskola (approved):

PR #1280, opened 2023-09-26
Treat consultations with no end date as not started

benjamineskola:
A consultation doesn't immediately get an end_date set in every case, so we should treat nil values as not started.

### DIFF
--- a/app/components/task_list_items/neighbour_responses_component.rb
+++ b/app/components/task_list_items/neighbour_responses_component.rb
@@ -36,10 +36,10 @@ module TaskListItems
     end
 
     def status
-      if @planning_application.consultation.nil? || @planning_application.consultation.neighbour_responses.none?
+      if @planning_application.consultation.nil? ||
+         @planning_application.consultation.not_started?
         "not_started"
-      elsif @planning_application.consultation.neighbour_responses.any? &&
-            @planning_application.consultation.end_date < Time.zone.now
+      elsif @planning_application.consultation.complete?
         "complete"
       else
         "in_progress"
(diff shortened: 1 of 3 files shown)


EGiataganas:
@benjamineskola How about adding a spec for this failure?

app/components/task_list_items/neighbour_responses_component.rb
```diff
@@ -36,7 +36,9 @@ def status_tag_component
     end
 
     def status
-      if @planning_application.consultation.nil? || @planning_application.consultation.neighbour_responses.none?
+      if @planning_application.consultation.nil? ||
+         @planning_application.consultation.neighbour_responses.none? ||
+         @planning_application.consultation.end_date.nil?
```
EGiataganas, line 41: How about wrapping these conditions on a method under `consultation`?
Maybe it could be `@planning_application.consultation.not_started?`

benjamineskola:
It seems like this would be unlikely to happen in real-world use (it can only happen if there are responses before the end date is set, which means responses before letters are sent), so the test case was a little awkward to figure out, but the status methods should be a bit of an improvement anyway.

app/models/consultation.rb
```diff
@@ -127,6 +127,18 @@ def polygon_search_and_boundary_geojson
...
+  def not_started?
+    !started?
+  end
+
+  def started?
+    neighbour_responses.present? && end_date.present?
+  end
+
+  def complete?
+    neighbour_responses.present? && end_date.present? && end_date < Time.zone.now
```
benbaumann95, line 139: Could you do `started? && end_date < Time.zone.now`

PR #1281, opened 2023-09-27
Assess against policy and guidance

cdccollins:
### Description of change

Officers can assess against policy and guidance relevant to householder planning permission

### Decisions
Planning permission is discretionary and judged against local plans and guidance. Instead of trying to import all the different councils' local policies into BoPS, we've made the format a bit looser. But this means trying to put it into the current policy class won't really work. Hence why I've created a slightly different model to capture this. 

### Story Link

https://trello.com/c/QY5Sitc2/1897-document-my-assessment-of-local-regional-and-national-policies-and-any-local-guidance

Screenshot 2023-10-03 at 15 27 01


app/components/planning_applications/assessment_report_component.html.erb
```diff
@@ -66,14 +66,29 @@
 <% end %>
 <section id="policy-classes-section" class="govuk-!-padding-bottom-5">
   <h3 class="govuk-heading-s"><%= t(".legislation") %></h3>
-  <% if policy_classes.any? %>
-    <% policy_classes.each do |policy_class| %>
-      <%= render(
-        PolicyClasses::SummaryComponent.new(policy_class: policy_class)
-      ) %>
+  <% if @planning_application.application_type.name == "planning_permission" %>
```
benjamineskola, line 69: I don't know how policy classes versus policy areas work precisely, but I wonder if we could do without checking the application type here.

Something like (pseudocode)

```
if policy_area.present?
  considerations.each
    …
  end
elif policy_classes.present?
  policy_classes.each
    …
  end
end
```

cdccollins: Mm it's a bit of a nightmare. Essentially they're the same thing but we can't load all of the law into bops so we've got a half way house. Not sure the naming/structure is the right way to go tbh 😬

app/components/planning_applications/assessment_report_component.html.erb
```diff
@@ -66,14 +66,29 @@
...
+      <% @planning_application.local_policy.local_policy_areas.each do |local_policy_area| %>
+        <p class="govuk-body"><strong><%= local_policy_area.area.humanize %></strong></p>
+        <p class="govuk-body"><strong>Policies:</strong> <%= local_policy_area.policies %></p>
+        <% if local_policy_area.guidance.present? %>
+          <p class="govuk-body"><strong>Guidance</strong> <%= local_policy_area.guidance %></p>
+        <% end %>
+        <p class="govuk-body"><strong>Assessment:</strong> <%= local_policy_area.assessment %></p>
+      <% end %>
+    <% else %>
+      <p class="govuk-body"><%= t(".no_legislation_assessed") %></p>
```
pixeltrix, line 80: The logic for this should really be:

``` erb
<% if local_policy_areas.any? %>
  <% local_policy_areas.each do |local_policy_area| %>
    <p class="govuk-body"><strong><%= local_policy_area.area.humanize %></strong></p>
    <p class="govuk-body"><strong>Policies:</strong> <%= local_policy_area.policies %></p>
    <% if local_policy_area.guidance.present? %>
      <p class="govuk-body"><strong>Guidance</strong> <%= local_policy_area.guidance %></p>
    <% end %>
    <p class="govuk-body"><strong>Assessment:</strong> <%= local_policy_area.assessment %></p>
  <% end %>
<% else %>
  <p class="govuk-body"><%= t(".no_legislation_assessed") %></p>
<% end %>
```

Move the logic for `local_policy_areas` into the component class and return an empty array for situations where there shouldn't be any. You could also move the block inside the `each` into its own component but probably that's too much abstraction.

app/views/planning_application/local_policies/_comment.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if @local_policy.review_local_policies.where.not(reviewer_comment: "").any? %>
```
pixeltrix, line 1: This shouldn't be in a view - it should be an extension on the review_local_policies association or a method on local_policy. Also is the query `reviewer_comment = '' OR reviewer_comment IS NULL` ? If so and the association is already loaded it might be quicker to do in memory by `review_local_policies.any?(&:present?)`.

app/views/planning_application/local_policies/_comment.html.erb
```diff
@@ -0,0 +1,10 @@
...
+  <div class="govuk-inset-text" id="reviewer_comment">
+    <p class="govuk-body govuk-!-font-weight-bold govuk-!-margin-bottom-0">
+      Reviewer comment:
+    </p>
+    <% @comment = @local_policy.review_local_policies.where.not(reviewer_comment: nil).order(:created_at).last %>
```
pixeltrix, line 6: Similar issue here - we've got `nil` instead of `""` so unclear whether it's going to hit the same set of records. Again might be quicker to do in memory if the comments are already loaded.

pixeltrix (changes requested):

pixeltrix (approved):

app/controllers/planning_applications/local_policies_controller.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+module PlanningApplication
```
pixeltrix, line 3: This needs to be `PlanningApplications`

app/controllers/planning_applications/policy_areas_controller.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
+
+module PlanningApplication
```
pixeltrix, line 3: This also needs to be `PlanningApplications`

app/controllers/planning_applications/policy_guidances_controller.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+class PlanningApplication
```
pixeltrix, line 3: This also needs to be `PlanningApplications`

app/controllers/planning_applications/review_local_policies_controller.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+module PlanningApplication
```
pixeltrix, line 3: Again here

app/controllers/planning_applications/review_policy_areas_controller.rb
```diff
@@ -0,0 +1,92 @@
+# frozen_string_literal: true
+
+module PlanningApplication
```
pixeltrix, line 3: Final one

pixeltrix (changes requested):
You'll need to move the view directories as well

pixeltrix (approved):

PR #1282, opened 2023-09-27
Confirm press notices

benbaumann95:
### **Confirm press notices**

- Add press_notice_email to local authorities which can be edited in the admin area
- Add press sent at and published at dates
- Ability to upload documents in support of the press notice
- Add updates to audit log

### Story Link

https://trello.com/c/VBzgVc6o/2016-send-press-notice-response

### Screenshots

Screenshot 2023-09-29 at 13 34 16

Screenshot 2023-09-29 at 13 41 46

Screenshot 2023-09-29 at 13 36 11

Screenshot 2023-09-29 at 13 50 52

### DIFF
--- a/app/assets/stylesheets/_main.scss
+++ b/app/assets/stylesheets/_main.scss
@@ -325,6 +325,10 @@ html * {
   display: flex;
 }
 
+.justify-content-space-between {
+  justify-content: space-between;
+}
+
 .align-items-center {
   align-items: center;
 }
@@ -404,4 +408,8 @@ body {
   margin-top: 1rem;
   margin-bottom: 1rem;
   padding: 2rem;
+}
+
+.background-light-grey {
+  background-color: govuk-colour("light-grey", $legacy: "grey-3");
 }
--- a/app/controllers/planning_application/confirm_press_notices_controller.rb
+++ b/app/controllers/planning_application/confirm_press_notices_controller.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+class PlanningApplication
+  class ConfirmPressNoticesController < AuthenticationController
+    before_action :set_planning_application
+    before_action :set_press_notice, only: %i[edit update]
+    before_action :ensure_press_notice_is_editable, only: %i[edit update]
+
+    def edit
+      respond_to do |format|
+        format.html
+      end
+    end
+
+    def update
+      respond_to do |format|
+        if @press_notice.update(press_notice_params)
+          format.html do
+            redirect_to planning_application_consultations_path(@planning_application), notice: t(".success")
+          end
+        else
+          format.html { render :edit }
+        end
+      end
+    end
+
+    private
+
+    def press_notice_params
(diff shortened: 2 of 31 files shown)


app/controllers/planning_application/confirm_press_notices_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def edit
+      respond_to do |format|
+        format.html
+      end
+    end
+
+    def update
+      respond_to do |format|
+        if @press_notice.update(press_notice_params)
```
cdccollins, line 17: doesn't this normally go on the outside of the respond_to block?

cdccollins (approved):

app/assets/stylesheets/_main.scss
```diff
@@ -325,6 +325,10 @@ html * {
   display: flex;
 }
 
+.justify-content-space-between {
+  justify-content: space-between;
+}
+
```
benjamineskola, line 331: Just a nitpick really but I don't much like the style of using many CSS utility classes. Better to do something like

```suggestion
.press-notice-status {
  display: flex;
  justify-content: space-between;
}
```

and then use the more specific class name rather than having to do `class="display-flex justify-content-space-between"` everywhere.

benjamineskola (approved):

app/controllers/planning_application/confirm_press_notices_controller.rb
benbaumann95: It's inconsistent as I think we do both styles in the app but this way enables us to just have the one block and also have the else to use the format
```
else
    format.html { render :edit }
 end
```

app/assets/stylesheets/_main.scss
benbaumann95: I think this would be a good topic for debate! 
I personally prefer the descriptive and reusable nature of these css classes that are agnostic to the component but as you say some people prefer the other way you suggested.
Using `.press-notice-status` might not be appropriate for another element on another page that requires these css classes. Maybe if we had a more generic name i.e. `status-panel` then we can reuse without being too specific.

Happy to discuss as there's never just the one best answer for these type of things

benjamineskola: > Maybe if we had a more generic name 

Yeah definitely — I haven't looked through everything to see if there is an appropriate generic component (and I acknowledge that's part of the problem). We do have other classes that use the same combination of attributes, but it's not immediately clear whether it makes sense for them to be shared in these specific cases.

I think at the moment we have a mixture of both ways of doing things, which isn't ideal either. Worth thinking about at some point — maybe something to discuss at dev club.

benbaumann95: Yep agreed.
You're probably not a fan of tailwind css then 😆

benjamineskola: I'm not, ideally — but I do think that if we want to primarily use utility classes then we should consider it. It might reduce the hassle of creating the classes, at least.

PR #1283, opened 2023-09-28
Update neighbour_letter model to permit resending

benjamineskola:
### Description of change

- Make neighbour->neighbour_letter a one-to-many relationship
- Add resend reason to neighbour letter
- Add last_letter_sent_at to neighbour, and update it from LetterSendingService

### Story Link

https://trello.com/c/csQTNIyO/1829-repeat-previous-publicity-tasks-at-any-point-of-the-planning-process

### DIFF
--- a/app/models/neighbour_letter.rb
+++ b/app/models/neighbour_letter.rb
@@ -3,6 +3,9 @@
 class NeighbourLetter < ApplicationRecord
   belongs_to :neighbour
 
+  validates :resend_reason, absence: true, unless: :resend?
+  validates :resend_reason, presence: true, if: :resend?
+
   STATUSES = {
     technical_failure: "technical failure",
     permanent_failure: "permanent failure",
@@ -36,5 +39,9 @@ class NeighbourLetter < ApplicationRecord
 
   def notify_api_key
     neighbour.consultation.planning_application.local_authority.notify_api_key || ENV.fetch("NOTIFY_API_KEY")
+  end
+
+  def resend?
+    neighbour.last_letter_sent_at.present?
   end
 end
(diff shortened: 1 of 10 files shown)


benjamineskola:
[Repeat previous Publicity tasks at any point of the planning process](https://trello.com/c/csQTNIyO/1829-repeat-previous-publicity-tasks-at-any-point-of-the-planning-process)

app/models/neighbour_letter.rb
```diff
@@ -3,6 +3,9 @@
 class NeighbourLetter < ApplicationRecord
   belongs_to :neighbour
 
+  validates :resend_reason, absence: true, on: :create, if: -> { neighbour.last_letter_sent_at.nil? }
+  validates :resend_reason, presence: true, on: :create, if: -> { neighbour.last_letter_sent_at.present? }
+
```
pixeltrix, line 8: Could do these as:

``` ruby
validates :resend_reason, absence: true, on: :create, unless: :resend?
validates :resend_reason, presence: true, on: :create, if: :resend?

private

  def resend?
    neighbour.last_letter_sent_at?
  end
```

Should the validations be `on: :create` or do they need to be on `update` as well?

pixeltrix (approved):
Just a couple of suggestions

app/models/neighbour_letter.rb
benjamineskola: I'm not sure if we're ever likely to modify the letter object after creation but it can't hurt to make certain.

pixeltrix: you can just remove `on`

PR #1285, opened 2023-10-02
Update docker compose setup

pixeltrix:
- Mirror directory changes for bops-applicant
- Remove broken cucumber e2e setup and feature
- Remove redundant sidekiq service as included in Procfile.dev
- Restrict exposed ports to 127.0.0.1 to improve security
- Use tmpfs to prevent stuck pidfiles
- Use volumes for gems/npm packages so images don't need rebuilding
- Update Makefile to remove redundant commands and fix broken ones
- Disable shared memory to make chromium more stable in docker


cdccollins (approved):
Works on my local machine 👍

benbaumann95 (approved):

PR #1287, opened 2023-10-02
Use titleize for council_name/formatted_subdomain

pixeltrix:
When the subdomain has a hypen like great-yarmouth then capitalize will produce an incorrect result so use titleize which will capitalize all words and remove the hyphen.

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -23,11 +23,11 @@ class LocalAuthority < ApplicationRecord
   end
 
   def council_name
-    plan_x? ? subdomain.capitalize : "#{subdomain.capitalize} Council"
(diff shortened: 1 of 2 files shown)


app/models/local_authority.rb
```diff
@@ -23,11 +23,11 @@ def signatory
   end
 
   def council_name
-    plan_x? ? subdomain.capitalize : "#{subdomain.capitalize} Council"
```
pixeltrix, line 26: For PlanX I switched this to use `council_code` rather than `capitalize` or `titleize` since `Planx` would've been the result.

benbaumann95 (approved):

PR #1288, opened 2023-10-02
Fix a missed apt-get clean in Dockerfile.production

benjamineskola:
### Description of change

Missing a call to apt-get clean etc means those files will continue to exist in one layer even if removed in later layers.

### DIFF
--- a/Dockerfile.production
+++ b/Dockerfile.production
@@ -58,7 +58,9 @@ RUN bash -c "export DEBIAN_FRONTEND=noninteractive && \
 RUN bash -c "export DEBIAN_FRONTEND=noninteractive && \
     apt-get update -qq && \
     apt-get upgrade -y && \
-    apt-get install -y imagemagick libmagickwand-dev ghostscript poppler-utils xvfb"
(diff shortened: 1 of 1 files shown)


Dockerfile.production
```diff
@@ -58,7 +58,9 @@ RUN bash -c "export DEBIAN_FRONTEND=noninteractive && \
 RUN bash -c "export DEBIAN_FRONTEND=noninteractive && \
     apt-get update -qq && \
     apt-get upgrade -y && \
-    apt-get install -y imagemagick libmagickwand-dev ghostscript poppler-utils xvfb"
+    apt-get install -y imagemagick libmagickwand-dev ghostscript poppler-utils xvfb && \
```
pixeltrix, line 61: What do we use ghostscript for? Image thumbnails? Wonder if we can switch to libvips which will reduce the layer size quite a bit - something for another PR though.

pixeltrix (approved):

Dockerfile.production
benjamineskola: I'm not completely sure — seems maybe to be related to file uploads, possibly for filetype detection?

I also wonder if there's any real benefit to having so many separate calls to apt-get.

PR #1293, opened 2023-10-03
Add GTM and update content security policy

benbaumann95:
### Description of change

Add [GTM](https://tagmanager.google.com/#/home) and update the content security policy

- Existing script tags have had to be updated to use a nonce
- Don't generate a nonce for inline style (the <my-map> is a third party component we use which needs this)
- Remove bullet config that injects inline scripts

### Story Link

https://trello.com/c/lz3eGcB1/1991-add-google-analytics-to-bops

### Screenshots

Screenshot 2023-10-03 at 15 51 22


app/views/application/_head.html.erb
```diff
@@ -20,12 +20,34 @@
     rel: "mask-icon"
   ) %>
   <link rel="stylesheet" href="https://use.typekit.net/zmv8alo.css">
-  <script>L_DISABLE_3D = true;</script>
```
benbaumann95, line 23: I think this is some legacy code for Leaflet? No longer needed anyway

config/environments/development.rb
```diff
@@ -4,9 +4,7 @@
   config.after_initialize do
     Bullet.enable        = true
     Bullet.bullet_logger = true
-    Bullet.console       = true
     Bullet.rails_logger  = true
-    Bullet.add_footer    = true
```
benbaumann95, line 9: Couldn't find an obvious answer to have this inline script permitted by our csp
https://github.com/flyerhzm/bullet/issues/478

I think having rails logs is enough anyway unless someone strongly disagrees and wants to find a non hacky solution for this

app/views/application/_head.html.erb
```diff
@@ -20,12 +20,34 @@
...
   <%= stylesheet_link_tag 'application', 'data-turbolinks-track': 'reload', media: 'all' %>
   <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload' %>
 
-  <% if ENV["FEEDBACK_FISH_ID"].present? %>
-    <script defer src="https://feedback.fish/ff.js?pid=<%= ENV['FEEDBACK_FISH_ID'] %>"></script>
+  <% if feedback_fish_id = ENV["FEEDBACK_FISH_ID"] %>
+    <%= tag.script(src: "https://feedback.fish/ff.js?pid=#{feedback_fish_id}", defer: "defer", nonce: request.content_security_policy_nonce) %>
+  <% end %>
+
+  <% if gtmid = ENV["GOOGLE_TAG_MANAGER_ID"] %>
```
benjamineskola, line 31: Could we put these environment variables into configuration settings instead of reading `ENV` directly from a view?

config/environments/development.rb
benjamineskola: I think that's right — we'll always be able to see that in dev anyway.

benjamineskola (approved):

app/views/application/_head.html.erb
```diff
@@ -20,12 +20,34 @@
...
+
+        var f = d.getElementsByTagName(s)[0],
+            j = d.createElement(s),
+            dl = l != 'dataLayer' ? '&l=' + l : '';
+
+        j.async = true;
+        j.src = 'https://www.googletagmanager.com/gtm.js?id=' + i + dl;
+        f.parentNode.insertBefore(j, f);
+      })(window, document, 'script', 'dataLayer', '<%= gtmid %>');
+    <% end %>
```
EGiataganas, line 49: how about wrapping that in a separate file under `app/javascript /analytics.js`

PR #1295, opened 2023-10-04
Fix site notice

cdccollins:
### Description of change

- Make the PDF look a bit better
- Fix a bug so that the confirm page shows when there isn't a document showing the site notice in place attached
- Rework the flow for downloading a site notice. First input the date it will be displayed, create the site notice, then download it

### DIFF
--- a/app/components/task_list_items/confirm_site_notice_component.rb
+++ b/app/components/task_list_items/confirm_site_notice_component.rb
@@ -16,23 +16,17 @@ module TaskListItems
     end
 
     def link_path
-      if @site_notice.displayed_at.nil?
-        edit_planning_application_site_notice_path(planning_application, site_notice)
-      else
+      if @site_notice.document.present?
         planning_application_site_notice_path(planning_application, site_notice)
+      else
+        edit_planning_application_site_notice_path(planning_application, site_notice)
       end
     end
 
     def status_tag_component
-      if @site_notice.displayed_at.nil?
-        StatusTags::BaseComponent.new(
-          status: "not_started"
-        )
-      else
-        StatusTags::BaseComponent.new(
-          status: "complete"
-        )
-      end
+      StatusTags::BaseComponent.new(
+        status: @site_notice.document.present? ? "complete" : "not_started"
+      )
     end
   end
 end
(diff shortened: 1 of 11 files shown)


app/components/task_list_items/confirm_site_notice_component.rb
```diff
@@ -16,21 +16,21 @@ def link_text
...
+        edit_planning_application_site_notice_path(planning_application, site_notice)
       end
     end
 
     def status_tag_component
-      if @site_notice.displayed_at.nil?
+      if @site_notice.document.present?
         StatusTags::BaseComponent.new(
-          status: "not_started"
+          status: "complete"
```
benjamineskola, line 29: Perhaps

```suggestion
          status: @site_notice.document.present? ? "complete" : "not_started"
```

instead of a conditional?

app/controllers/planning_application/site_notices_controller.rb
```diff
@@ -101,7 +106,9 @@ def send_mail
       if site_notice_params[:internal_team_email].presence
         @planning_application.send_internal_team_site_notice_mail(site_notice_params[:internal_team_email])
       else
-        @planning_application.send_site_notice_mail(@planning_application.applicant_email)
+        @planning_application.send_site_notice_mail(
+          @planning_application.applicant_email || @planning_application.agent_email
```
benjamineskola, line 110: I'm not completely sure here but should the agent email come first? Use the agent if one exists otherwise use the applicant.

If having an agent always means there's no applicant email then this doesn't matter anyway.

app/javascript/controllers/pdf_controller.js
```diff
@@ -2,15 +2,15 @@ import { Controller } from "@hotwired/stimulus"
...
-    event.preventDefault()
+  handleClick(_event) {
     const doc = new jsPDF("p", "px", "a4")
     const applicationReference = document.getElementById(
       "application-reference",
     ).innerText
     const pdfjs = document.getElementById("site-notice-content")
     doc.html(pdfjs, {
       callback: function (doc) {
+        doc.setFont("Helvetica")
```
benjamineskola, line 13: Do Windows machines have Helvetica? I thought not but my knowledge is probably very out of date.

benjamineskola (approved):

PR #1303, opened 2023-10-06
BUG FIX Consultation summary review

cdccollins:
### Description of change

Make sure the reviewer sees the right assessment details per application type when they're reviewing the assessor's work

### Story Link

https://trello.com/c/rASG8D9t/2017-consultation-review-summary-cant-be-blank-but-theres-no-option-to-complete-it

### DIFF
--- a/app/views/planning_applications/_decision_notice.html.erb
+++ b/app/views/planning_applications/_decision_notice.html.erb
@@ -17,7 +17,7 @@
           "Applicant" => planning_application.applicant_name,
           "Application number" => planning_application.reference,
           "Application received" => planning_application.received_at.to_fs,
-          "Decision date" => planning_application.determination_date.to_fs || "TBD",
+          "Decision date" => planning_application.determination_date&.to_fs || "TBD",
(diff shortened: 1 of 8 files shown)


app/views/planning_applications/_decision_notice.html.erb
```diff
@@ -16,8 +16,8 @@
         {
           "Applicant" => planning_application.applicant_name,
           "Application number" => planning_application.reference,
-          "Application received" => planning_application.received_at,
-          "Decision date" => planning_application.determination_date || "TBD",
+          "Application received" => planning_application.received_at.to_fs,
+          "Decision date" => planning_application.determination_date.to_fs || "TBD",
```
benjamineskola, line 20: I think if the determination date is nil this won't behave as intended — the `||` works for the previous code being nil, but here the `to_fs` will blow up instead. Safe navigation should do the trick?

```suggestion
          "Decision date" => planning_application.determination_date&.to_fs || "TBD",
```

benjamineskola (approved):

PR #1304, opened 2023-10-06
Move controllers to PlanningApplications module

pixeltrix:


cdccollins (approved):

benjamineskola (approved):

config/routes.rb
```diff
@@ -118,7 +118,7 @@
 
     resource :legislation, only: %i[show update]
 
-    scope module: :planning_application do
+    scope module: :planning_applications do
```
benjamineskola, line 121: Do we need to worry about any existing links that this might break? (probably not)

EGiataganas (approved):

config/routes.rb
pixeltrix: The `:module` option doesn't affect either the url generated or the name of the url helper method, it just controls the namespace of the controller so unless there's somewhere we're using `url_for` with `:controller` and `:action` options then it should be fine (I search for `"planning_application/` across the project so should've picked it up anyway).

PR #1307, opened 2023-10-09
Make assignment of planning application in controllers more DRY

pixeltrix:
Uses the path parameter keys to determine whether we're in a nested controller.

### DIFF
--- a/app/controllers/planning_applications/notes_controller.rb
+++ b/app/controllers/planning_applications/notes_controller.rb
@@ -30,18 +30,8 @@ module PlanningApplications
 
     private
 
-    def set_planning_application
-      planning_application = planning_applications_scope.find(planning_application_id)
-
-      @planning_application = PlanningApplicationPresenter.new(view_context, planning_application)
-    end
-
     def planning_applications_scope
-      current_local_authority.planning_applications.includes(:notes)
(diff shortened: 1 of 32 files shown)


pixeltrix:
Dependent on #1306 being merged first

app/controllers/planning_applications/notes_controller.rb
```diff
@@ -30,20 +30,6 @@ def create
...
     private
 
-    def set_planning_application
-      planning_application = planning_applications_scope.find(planning_application_id)
-
-      @planning_application = PlanningApplicationPresenter.new(view_context, planning_application)
-    end
-
-    def planning_applications_scope
-      current_local_authority.planning_applications.includes(:notes)
```
benbaumann95, line 40: Did we want to do similar to how you did with recommendations for the planning application scope i.e. `super.includes(:notes)`

benbaumann95: I think there's a few other examples of this too (unless you deemed that the includes wasn't necessary)

cdccollins (approved):

benbaumann95 (approved):

EGiataganas (approved):

PR #1312, opened 2023-10-09
Recommendations more dynamic

EGiataganas:
### Description of change

- Make state the reason more dynamic
- Make is the use more dynamic
- Make decision more dynamic
- Additional locales for planning_permission application type

### Story Link

https://trello.com/c/lkTb7kEH


app/form_models/recommendation_form.rb
benjamineskola, line 1: I don't really understand what this is doing based on the commit messages/PR comment. It seems to be replacing LDC with PA and I'm not sure how this relates to making it more dynamic?

benjamineskola (approved):

PR #1314, opened 2023-10-09
Return JSON from OS Places APIs

pixeltrix:
By default Rails will return JS to non-HTML requests so we need to use the router's defaults feature to ensure that the response is application/json.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -23,8 +23,15 @@ Rails.application.routes.draw do
     get "resend_code", to: "users/sessions#resend_code", as: "resend_code"
   end
 
-  resources :os_places_api, only: %i[index]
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -23,8 +23,15 @@
...
-  post "search_addresses_by_polygon", to: "os_places_api#search_addresses_by_polygon"
+  defaults format: "json" do
+    get "/os_places_api",
+      to: "os_places_api#index",
+      as: "os_places_api_index"
+
+    post "/search_addresses_by_polygon",
+      to: "os_places_api#search_addresses_by_polygon",
+      as: "search_addresses_by_polygon"
+  end
```
pixeltrix, line 34: These routes are terrible but I'm going 🙈 for now

benbaumann95 (approved):

PR #1318, opened 2023-10-10
Handle when GOOGLE_TAG_MANAGER_ID is set to empty string

benbaumann95:
Our aws parameter store will expect a mapping that this env exists, and we do not have this set on production

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -53,7 +53,7 @@ module Bops
 
     config.os_vector_tiles_api_key = ENV.fetch("OS_VECTOR_TILES_API_KEY", nil)
     config.feedback_fish_id = ENV.fetch("FEEDBACK_FISH_ID", nil)
-    config.google_tag_manager_id = ENV.fetch("GOOGLE_TAG_MANAGER_ID", nil)
+    config.google_tag_manager_id = ENV.fetch("GOOGLE_TAG_MANAGER_ID", " ").presence
   end
 
   def self.env


config/application.rb
```diff
@@ -53,7 +53,8 @@ class Application < Rails::Application
 
     config.os_vector_tiles_api_key = ENV.fetch("OS_VECTOR_TILES_API_KEY", nil)
     config.feedback_fish_id = ENV.fetch("FEEDBACK_FISH_ID", nil)
-    config.google_tag_manager_id = ENV.fetch("GOOGLE_TAG_MANAGER_ID", nil)
+    config.google_tag_manager_id = ENV["GOOGLE_TAG_MANAGER_ID"] unless ["false",
```
benbaumann95, line 56: rubocop..

config/application.rb
```diff
@@ -53,7 +53,8 @@ class Application < Rails::Application
...
+                                                                        nil].include?(ENV["GOOGLE_TAG_MANAGER_ID"])
```
benbaumann95, line 57: rubocop..

config/application.rb
```diff
@@ -53,7 +53,9 @@ class Application < Rails::Application
 
     config.os_vector_tiles_api_key = ENV.fetch("OS_VECTOR_TILES_API_KEY", nil)
     config.feedback_fish_id = ENV.fetch("FEEDBACK_FISH_ID", nil)
-    config.google_tag_manager_id = ENV.fetch("GOOGLE_TAG_MANAGER_ID", nil)
+    config.google_tag_manager_id = ENV.fetch("GOOGLE_TAG_MANAGER_ID", nil) unless ["false"].include?(
+      ENV.fetch("GOOGLE_TAG_MANAGER_ID", nil)
+    )
```
pixeltrix, line 58: How about this?

``` ruby
if ENV.fetch("GOOGLE_TAG_MANAGER_ID", "false") != "false"
  config.google_tag_manager_id = ENV.fetch("GOOGLE_TAG_MANAGER_ID")
end
```

(Undoubtedly rubocop will complain 😂 )

PR #1325, opened 2023-10-11
Consult the right people

pixeltrix:


app/controllers/planning_applications/assessment_details_controller.rb
```diff
@@ -8,7 +8,8 @@ class AssessmentDetailsController < AuthenticationController
     before_action :set_planning_application
     before_action :ensure_planning_application_is_validated
     before_action :set_assessment_detail, only: %i[show edit update]
-    before_action :set_category, :set_rejected_assessment_detail, only: %i[new create]
+    before_action :set_category, :set_rejected_assessment_detail, only: %i[new create edit update show]
```
benjamineskola, line 11: `only:` might be unnecessary here?

app/jobs/notify_email_job.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
+
+require "notifications/client"
+
+class NotifyEmailJob < ApplicationJob
+  class NotConfiguredError < RuntimeError; end
+
+  NOTIFY_TEMPLATE_ID = "7cb31359-e913-4590-a458-3d0cefd0d283"
```
benjamineskola, line 8: Should we put this in a configuration variable or something? We might want to use a different notify account per local authority, for example. (Though I'm not sure if it matters so much for emails.)

app/views/planning_applications/consultee/emails/_consultees.html.erb
benjamineskola, line 1: I hate this HTML structure so much. But I'm not sure there's much we can do about that if that's how the gov.uk accordions work.

EGiataganas (approved):

EGiataganas (approved):

PR #1326, opened 2023-10-12
Add warning if site notice or press notice not complete

benbaumann95:
### Description of change

- If no site notice displayed at or press notice published at dates have been set, display a warning as this has an implication on the consultation end date
- Display warning to reviewer when determining the application if these dates have not been set

### Story Link

https://trello.com/c/VcpHL603/2026-make-sure-a-press-notice-and-site-notice-has-been-displayed-for-three-weeks-before-making-a-decision-and-update-the-consultation

### DIFF
--- a/app/models/press_notice.rb
+++ b/app/models/press_notice.rb
@@ -12,9 +12,11 @@ class PressNotice < ApplicationRecord
 
   validates :required, inclusion: { in: [true, false] }
 
+  after_update :update_consultation_end_date!
   after_save :audit_press_notice!
 
   delegate :audits, to: :planning_application
+  delegate :consultation, to: :planning_application
   delegate :press_notice_email, to: "planning_application.local_authority", allow_nil: true
 
   scope :required, -> { where(required: true) }
@@ -65,5 +67,12 @@ class PressNotice < ApplicationRecord
 
   def joined_reasons
     reasons.values.join(", ")
+  end
+
+  def update_consultation_end_date!
+    return unless consultation
+    return unless saved_changes.include? "published_at"
+
+    consultation.update!(end_date: published_at + 21.days)
   end
 end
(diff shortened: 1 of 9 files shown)


app/models/press_notice.rb
```diff
@@ -66,4 +68,11 @@ def audit_press_notice!
   def joined_reasons
     reasons.values.join(", ")
   end
+
+  def update_consultation_end_date!
+    return unless consultation
+    return unless saved_changes.include? "published_at"
+
+    consultation.update!(end_date: published_at + 21.days)
```
cdccollins, line 76: I'm just thinking - for the site notice we give some extra wiggle room and add a couple more days, in case the applicant prints it and then doesn't put it up until the next day. 

In which case, it could be possible that the consultation date might be 23 days ahead, and here we're reducing it by two days or so. 

Could we check whether this date is less or more than the current date, and then set it?

benjamineskola: Somewhere we do "next business day + 21 days", for example?

cdccollins (approved):

app/views/shared/_dates_and_assignment_header.html.erb
```diff
@@ -1,39 +1,58 @@
...
-              <%= t("planning_applications.days_left", count: @planning_application.consultation.days_left) %>
-            <% else %>
-              <%= t("planning_applications.overdue", count: @planning_application.consultation.days_left.abs) %>
-            <% end %>
+          <% unless display_or_publish_required?(@planning_application, :site) || display_or_publish_required?(@planning_application, :press) %>
+            <%= time_tag @planning_application.consultation.end_date %>
+            <span class="govuk-!-margin-left-1">
+              <span class="govuk-tag govuk-tag--#{status_consultation_date_tag_colour}">
+                <% days_left = @planning_application.consultation.days_left %>
+                <%= days_left > 0 ? t("planning_applications.days_left", count: days_left) : t("planning_applications.overdue", count: days_left.abs) %>
```
benjamineskola, line 29: This might be clearer in a helper.

(I think we have similar code elsewhere for showing days left/overdue, but not sure if it's easily generalisable.)

PR #1334, opened 2023-10-23
Do not duplicate list of assessment details per application type

benjamineskola:
### Description of change

This has now (5599c39) been moved to a database field, but some code still uses the static lists. This then prevents changes to the assessment detail task list being applied correctly.

### Story Link

https://trello.com/c/wZIjZwLe/2045-remove-amenity-in-task-list-for-householder-assessment

### DIFF
--- a/app/models/assessment_detail.rb
+++ b/app/models/assessment_detail.rb
@@ -58,25 +58,6 @@ class AssessmentDetail < ApplicationRecord
     scope :"#{category}", -> { where(category:) }
   end
 
-  class << self
-    def category_keys
-      categories = AssessmentDetail.categories.keys.excluding("past_applications")
-
-      categories.partition { |category| category != "additional_evidence" }.sum([])
-    end
-
-    def categories_for(application_type)
-      case application_type
-      when :lawfulness_certificate
-        category_keys - %w[publicity_summary amenity]
-      when :prior_approval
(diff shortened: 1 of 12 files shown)


app/models/assessment_detail.rb
```diff
@@ -71,6 +71,8 @@ def categories_for(application_type)
         category_keys - %w[publicity_summary amenity]
       when :prior_approval
         category_keys - %w[consultation_summary]
+      when :planning_permission
+        category_keys - %w[amenity]
```
pixeltrix, line 75: Shouldn't this be driven by this field on the assessment type?

https://github.com/unboxed/bops/blob/fa908c3cf0560d5dfd9c618c3673eb0b9b885aeb/db/schema.rb#L80

pixeltrix: Added by Celia a couple of weeks ago:
https://github.com/unboxed/bops/commit/5599c399df820da7356bf1eab227ca249652fbe6

benjamineskola: Good point! I remember that being added but I assumed without double-checking that it was for something different. So presumably these lists were actually duplicated previously, and this one didn't get updated. Will fix now.

pixeltrix (approved):

PR #1338, opened 2023-10-24
Use glob routing for the planning guides

pixeltrix:
The template_exists? method is restricted to the configured view paths so there is no risk of escaping to render files outside of the application.

### DIFF
--- a/spec/system/public/planning_guides_spec.rb
+++ b/spec/system/public/planning_guides_spec.rb
@@ -159,6 +159,13 @@ RSpec.describe "Planning guides" do
       expect(page).to have_image_displayed("use_plans/proposed")
       expect(page).to have_link("Back", href: public_planning_guides_path)
     end
+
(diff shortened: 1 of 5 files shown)


benjamineskola (approved):

benjamineskola (approved):

spec/system/public/planning_guides_spec.rb
```diff
@@ -159,6 +159,13 @@
       expect(page).to have_image_displayed("use_plans/proposed")
       expect(page).to have_link("Back", href: public_planning_guides_path)
     end
+
+    it "returns 404 for guides that don't exist" do
+      expect do
+        visit "/planning_guides/does/not/exist"
+        expect(page).to have_selector("h1", text: "Does not exist")
```
pixeltrix, line 166: Note that the spec fails without this expectation even though it can never be true because the example doesn't wait for the response and the exception occurs outside of the `expect` block.

PR #1339, opened 2023-10-24
Remove broken grover 'fix'

benjamineskola:
Hardcoding the path to chromium breaks production where we use chrome instead.

### DIFF
--- a/config/initializers/grover.rb
+++ b/config/initializers/grover.rb
@@ -1,9 +1,0 @@
-# frozen_string_literal: true
-
-Grover.configure do |config|
-  if Gem::Platform.local.os != "darwin"
-    config.options = {
-      executable_path: "/usr/bin/chromium"
-    }
-  end
-end


docker/ruby/Dockerfile
```diff
@@ -35,7 +35,13 @@ ARG BUNDLER_VERSION=2.3.26
...
 
 # Install Chromium and Chromium-Driver
-RUN bash -c "apt-get install -y --no-install-recommends chromium chromium-driver"
+RUN export DEBIAN_FRONTEND=noninteractive && \
+    wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | gpg --dearmor -o /usr/share/keyrings/googlechrome-linux-keyring.gpg && \
+    bash -c 'echo "deb [arch=amd64 signed-by=/usr/share/keyrings/googlechrome-linux-keyring.gpg] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' && \
+    apt-get update && \
+    apt-get install -y chromedriver google-chrome-stable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-khmeros fonts-kacst fonts-freefont-ttf libxss1 --no-install-recommends && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
```
pixeltrix, line 44: AFAIK Google doesn't ship arm64 binaries for chromedriver - is there an issue where Chrome/Chromium don't match?

benjamineskola: I think it's simply that it's looking for the wrong paths. Configuring it to check both would also work, or presumably so would using chromium rather than chrome in production.

pixeltrix (approved):

PR #1343, opened 2023-10-25
Review conditions

EGiataganas:
### Description of change

Adding ability to reviewer for reviewing conditions.
Also introducing review polymorphic table

### Story Link

https://trello.com/c/QLegnd93

### Screenshots

Screenshot 2023-11-02 at 04 51 50


pixeltrix:
Maybe extract the `Review` -> `Reviewing` change into a separate PR and get that merged?

EGiataganas:
@pixeltrix I was thinking the same thing! That's why I've made it in a separate commit. Pushing now new branch and PR #1344

EGiataganas:
Adding #1351 to be able to review overall conditions

app/controllers/planning_applications/review/conditions_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+        respond_to do |format|
+          format.html
+        end
+      end
+
+      def edit
+        respond_to do |format|
+          format.html
+        end
+      end
```
pixeltrix, line 25: The `show` and `editable` views are the same apart from `editable` being passed to the `form` partial being `true/false` - we should only have the `show` action and use the status to determine the value of `editable`.

app/controllers/planning_applications/review/conditions_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+
+      def return_to_officer?
+        params.dig(:condition_set, :action) == "rejected"
+      end
+
+      def ensure_user_is_reviewer
+        current_user.reviewer?
+      end
```
pixeltrix, line 109: This appears to be all for setting the status on the `ConditionSet` depending on what the review is - in which case it should move to callbacks on the `Reviewable` concern and this controller should only be updating the review directly.

app/models/concerns/reviewable.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Reviewable
+  extend ActiveSupport::Concern
+
+  included do
+    has_one :review, as: :reviewable, dependent: :destroy
```
pixeltrix, line 7: Not for this PR but what should happen if there are multiple rounds of reviews where the conditions are sent back for amendments by the reviewer? We should probably track that history (especially for minors/majors) so we might need to change this to a `has_many` and have a setup where we only append new reviews and show the latest.

pixeltrix (changes requested):

app/models/condition.rb
```diff
@@ -13,6 +13,10 @@ def sort_key
     [title_key, timestamp_key, Float::INFINITY].compact.first
   end
 
+  def review_title
+    title.blank? ? "Other" : title
```
benjamineskola, line 17:
```suggestion
    title.presence || "Other"
```

> `object.presence` is equivalent to `object.present? ? object : nil`
[\[docs\]](https://api.rubyonrails.org/classes/Object.html#method-i-presence)

app/controllers/planning_applications/review/tasks_controller.rb
```diff
@@ -5,12 +5,24 @@ module Review
...
+
+      private
+
+      def set_condition_set
+        @condition_set = @planning_application.condition_set || @planning_application.create_condition_set!
+      end
+
+      def set_condition_set_review
+        @condition_set_review = @condition_set.review || @condition_set.create_review!
+      end
```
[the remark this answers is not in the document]
benjamineskola: Yes, I think it would be good to avoid a situation like we had with consultations, where it was possible for the object not to have been created when already trying to display the status of it.

app/models/concerns/reviewable.rb
benbaumann95: Agreed, needs to be a `has_many` reviews as there can be multiple rounds

app/components/task_list_items/reviewing/condition_component.rb
```diff
@@ -0,0 +1,40 @@
...
+      def status_tag_component
+        StatusTags::BaseComponent.new(status:)
+      end
+
+      def status
+        if condition_set.review.present?
+          condition_set.review.status
+        else
+          :in_progress
+        end
```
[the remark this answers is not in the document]
EGiataganas: Double checked it and no need for this condition, will leave `condition_set.review.status`

app/models/concerns/reviewable.rb
EGiataganas: Agree, but as I didn't want to expand it more I've left it as `has_one :review` which is what is needed for this

app/controllers/planning_applications/review/tasks_controller.rb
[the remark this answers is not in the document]
benjamineskola: Yes, I think it would be good to avoid a situation like we had with consultations, where it was possible for the object not to have been created when already trying to display the status of it.

(comment moved to correct place 🤦🏻‍♂️)

app/controllers/planning_applications/review/conditions_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      def condition_set_review_params
+        params.require(:condition_set)
+          .permit(review_attributes: %i[action comment],
+            conditions_attributes: %i[_destroy id standard title text reason])
+          .to_h
+          .deep_merge(
+            status: condition_set_status,
+            review_attributes: {reviewed_at: Time.current, reviewer: current_user, status: status}
+          )
```
pixeltrix, line 47: These really should be set in the model via callbacks or arguments to a custom method but I think we agreed to leave it to a separate PR, right?

pixeltrix (approved):
Looks good 👍🏻

app/controllers/planning_applications/review/conditions_controller.rb
EGiataganas: Yes, I agree! We will make a follow up pr to also replace `CommitMatchable`

PR #1349, opened 2023-10-27
Re-consult Consultees when changes are made to application

pixeltrix:
### Description of change

Add the ability to reconsult existing consultees with a reasons message added to the top of the new email.

### Story Link

https://trello.com/c/GUacbpjT

### Screenshots

image

### DIFF
--- a/app/jobs/resend_consultee_emails_job.rb
+++ b/app/jobs/resend_consultee_emails_job.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+class ResendConsulteeEmailsJob < NotifyEmailJob
+  queue_as :high_priority
+
+  def perform(consultation, consultees, message, date, subject, body)
+    planning_application = consultation.planning_application
+    local_authority = planning_application.local_authority
+
+    defaults = {
+      signatory_name: local_authority.signatory_name,
+      signatory_job_title: local_authority.signatory_job_title,
+      local_authority: local_authority.council_name,
+      reference: planning_application.reference,
+      description: planning_application.description,
+      address: planning_application.address,
+      link: consultation.application_link,
+      closing_date: date.to_fs
+    }
+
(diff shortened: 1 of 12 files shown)


config/locales/en.yml
```diff
@@ -1164,6 +1171,15 @@ en:
       update_assessment: Update assessment
     update:
       success: Recommendation was successfully reviewed.
+  reconsult_consultees_component:
+    date_label: Consultee reponse required by
+    formatting_hint_html: The message is formatted using <a href="https://www.notifications.service.gov.uk/using-notify/formatting">GOV.UK Notify formatting rules</a>. Placeholders like <nobr><code>%{closing_date}</code></nobr> will be substituted when the email is sent.
```
benjamineskola, line 1176: "Placeholders like" — should we make sure there's an actual list of the supported placeholders somewhere? Not a blocker, though.

app/jobs/resend_consultee_emails_job.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+class ResendConsulteeEmailsJob < NotifyEmailJob
```
benjamineskola, line 3: This looks very similar to `SendConsulteeEmailsJob`. Am I right in thinking that the only differences are the `message` and `date` parameters and the logic around them? I wonder if it's worth just adding these as optional parameters to the existing job?

benjamineskola (approved):

EGiataganas (approved):

config/locales/en.yml
pixeltrix: Ideally, we'd link 'placeholders' to a documentation page but I don't think we have any in-app documentation at the moment so didn't want to expand the story.

benjamineskola: Yeah, it's more something to follow up on later.

PR #1351, opened 2023-10-31
Condition set

EGiataganas:
### Description of change

### Story Link

https://link-to-issue

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


app/components/task_list_items/conditions_component.rb
```diff
@@ -2,28 +2,29 @@
...
+    attr_reader :condition_set
+    delegate :planning_application_id, to: :condition_set
 
     def link_text
       "Add conditions"
     end
 
     def link_path
-      planning_application_conditions_path(@planning_application)
+      planning_application_conditions_path(planning_application_id)
```
benjamineskola, line 19: Can't we use `planning_application` here instead of `planning_application_id`? (With the delegate changed to match, of course.)

Or doesn't delegate work for activerecord relations like that?

app/models/condition.rb
```diff
@@ -2,6 +2,9 @@
 
 class Condition < ApplicationRecord
   belongs_to :planning_application
```
benjamineskola, line 4: Should we now also remove `belongs_to :planning_application`?

Or can a condition belong to both an application and a condition set? (It looks like planning application no longer `has_many` conditions.)

app/models/condition.rb
```diff
@@ -2,6 +2,9 @@
...
+  belongs_to :condition_set
+
+  validates :text, :reason, presence: true
```
benjamineskola, line 7: Duplicated?

```suggestion
```

EGiataganas: Ok, I've definitely mixed up things on rebase 😂

app/models/condition.rb
EGiataganas: You're right, I remember deleting that line! Maybe mixed it up on rebase 😄

benjamineskola (approved):

db/migrate/20231027102432_migrate_conditions_to_condition_set.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
EGiataganas: I wanted to have this as 2 separate migrations just to be on a safe side! Another PR will follow to drop `:planning_application_id` from `:conditions`

PR #1354, opened 2023-11-03
Redact and publish consultee responses

benjamineskola:
- Catch a potential nil
- Add barebones consultee list just so there's somewhere to link from
- Add redact neighbour response controller
- Remove leftover response field on consultee


cdccollins (commented):
I see that you can redact comments, but does this actually publish them as well? I may have misunderstood what this story is about

benjamineskola:
> I see that you can redact comments, but does this actually publish them as well?

No, I've made this just the UI for marking them as redacted/published, based on the neighbour response classes — still figuring out how the actual publication is meant to happen.

cdccollins (approved):

cdccollins (approved):

PR #1362, opened 2023-11-10
Manage consultee responses

pixeltrix:
### Description of change

* Add a view consultee responses page
* Shows an overview of the consultees status
* Shows any existing responses
* Allows uploading of new responses

### Story Link

https://trello.com/c/c8TFiZSO
https://trello.com/c/SeLx4U4x

### Screenshots

southwark bops localhost_3000_planning_applications_2064_consultee_responses(iPad Air)


cdccollins (commented):
I was just playing around with this locally, is there a way to make the dev experience a bit easier? I "sent" an email to a consultee, but because it's a test notify key the status hung at "sending", so I had to go in the db to update it so I could go to the next stage. Appreciate it may not be a huge priority.

cdccollins:
> I was just playing around with this locally, is there a way to make the dev experience a bit easier? I "sent" an email to a consultee, but because it's a test notify key the status hung at "sending", so I had to go in the db to update it so I could go to the next stage. Appreciate it may not be a huge priority.

Apparently it's just my machine that this doesn't work on, ignore!

app/models/consultation.rb
```diff
@@ -105,22 +105,22 @@ def send_consultee_emails(attributes)
   end
 
   def start_deadline
-    update!(end_date: end_date_from_now, start_date: start_date || 1.business_day.from_now)
+    update!(end_date: end_date_from_now, start_date: start_date || default_start_date)
   end
 
   def extend_deadline(new_date)
-    update!(end_date: [end_date, new_date].max)
+    update!(end_date: [end_date, new_date.end_of_day].max)
```
benjamineskola, line 112: I wonder if some of these date times would be better as dates, and then treat the time of day as implicit? I don't know if that's even possible or if we'd always just get a date time back from postgres.

app/models/consultee.rb
```diff
@@ -37,7 +37,7 @@ def period(now = Time.current)
   end
 
   def consulted?
-    awaiting_response? || responses?
+    !not_consulted?
```
benjamineskola, line 48: I don't like the "not-not-consulted", but I presume that as `not_consulted` comes from the enum there's nothing much to do about this.

benjamineskola (approved):

PR #1363, opened 2023-11-10
Arrange validation task routes together

benjamineskola:
### Description of change

Arrange the validation-related pages into a single namespace

### Story Link

https://trello.com/c/i0MSozBu/2080-improve-validation-routes


app/services/constraint_query_update_service.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class ConstraintQueryUpdateService
+  # TODO not sure why restoring this is necessary, removed by eg in 257a4bec
+  # but breaks ./spec/system/planning_applications/sitemap_spec.rb:38
+  class SaveError < StandardError; end
```
benbaumann95, line 6: I've removed this locally and the test `spec/system/planning_applications/sitemap_spec.rb:38` still seems to run ok?

benbaumann95 (approved):
Looks good, just the one comment about the need for the `SaveError`

app/services/constraint_query_update_service.rb
pixeltrix: It's referred to in the SitemapsController so it may cause a NameError but I don't see it being raised anywhere:
https://github.com/unboxed/bops/blob/c9a0618284a33b22e69a54ad8f64ac3912e2c5b8/app/controllers/sitemaps_controller.rb#L30-L53

benbaumann95: The code to raise the `SaveError` was removed in https://github.com/unboxed/bops/commit/257a4bec8702125004a0f20a6bc5bd1921046367 along with the specs. So if this exception handling isn't required anymore we can probably just remove the `rescue ConstraintQueryUpdateService::SaveError => e` block

benbaumann95: Might be worth checking if we are handling errors here properly as part of that update to the constraint query though

benjamineskola: I can't reproduce this now so I've removed it again.

Specifically the error I was getting was that `SaveError` was undefined, and I'm not sure where/why it was being referenced now.

PR #1370, opened 2023-11-14
Switch away from native interpolation and validate fields

pixeltrix:
### Description of change

The built-in Ruby `format` method is inappropriate for use on user-supplied data as it can use a variety of formats and generate exceptions when substitutions are improperly formed. Use `gsub` instead which will ignore anything that doesn't match the pattern and validate the message fields on submission for any placeholders that don't match expected values.

### Story Link

https://trello.com/c/hIn32jDk

### Screenshots

image
image


cdccollins (commented):
There should probably be something like this on the neighbour letter too, and the next PR too...

benbaumann95 (approved):

EGiataganas (approved):

PR #1372, opened 2023-11-15
Allow consultee email content to be reset

pixeltrix:
Now that the message content is only set to the default on record creation add an explicit mechanism to reset the message as blanking out the content no longer works and wasn't particularly user-friendly.

### Story Link

https://trello.com/c/YXWqoPzz

### Screenshots

image

### DIFF
--- a/app/javascript/controllers/consultee_email_controller.js
+++ b/app/javascript/controllers/consultee_email_controller.js
@@ -1,0 +1,19 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["subject", "body"]
+
+  reset() {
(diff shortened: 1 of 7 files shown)


app/javascript/controllers/consultee_email_controller.js
```diff
@@ -0,0 +1,19 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["subject", "body"]
+
+  reset() {
+    console.log("Reset consultee email")
```
benbaumann95, line 7: Is this still required or just for debugging?

benbaumann95 (approved):

PR #1374, opened 2023-11-16
Prevent the use of path helpers in specs

pixeltrix:
When writing specs the expected values should not be changed by changes in the application.


benbaumann95 (approved):
🎉 😍

PR #1375, opened 2023-11-16
Step to redact documents

cdccollins:
### Description of change

A step in validation to ensure planning officers redact documents

### Story Link

https://trello.com/c/JwHe5mwE/1883-able-to-add-a-copy-of-a-redacted-document

### Screenshots

screencapture-southwark-bops-localhost-3000-planning-applications-3-document-redactions-2023-11-16-10_54_31
screencapture-southwark-bops-localhost-3000-planning-applications-4-document-redactions-2023-11-16-10_54_52


app/controllers/planning_applications/validation/document/redactions_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+        def assign_params
+          documents_attributes = planning_application_params[:documents_attributes].select do |_key, value|
+            value[:file].present?
+          end
+
+          new_params = planning_application_params
+
+          new_params[:documents_attributes] = documents_attributes
+          new_params
+        end
```
benbaumann95, line 40: I think we can also do this like:

```ruby
# models/planning_application.rb
accepts_nested_attributes_for :documents, reject_if: proc { |attributes| attributes["file"].blank? }

# controllers/planning_applications/validation/document/redactions_controller.rb
  params.require(:planning_application)
        .permit(documents_attributes: documents_params)
end

def documents_params
  %i[file redacted publishable]
end
```

which means we don't have to do with `assign_params` stuff.
But not sure if that will mess with other ways we update documents

app/views/planning_applications/validation/document/redactions/_form.html.erb
```diff
@@ -0,0 +1,39 @@
...
+) do |form| %>
+  <%= form.govuk_error_summary(presenter: ErrorPresenter) %>
+  <table class="govuk-table redacted-document-table">
+    <thead class="govuk-table__head">
+      <tr class="govuk-table__row">
+        <th class="govuk-table__header govuk-!-width-one-half">Download original</th>
+        <th class="govuk-table__header govuk-!-width-one-half">Upload redacted</th>
+      </tr>
+    </thead>
+  <%= form.fields_for :documents, @planning_application.documents.where(redacted: false) do |ff| %>
```
benbaumann95, line 15: Could we use a scope instead of querying here?

benbaumann95 (approved):

app/controllers/planning_applications/validation/document/redactions_controller.rb
cdccollins: ooh that's cool, haven't seen that before! i'll give it a go 😄

PR #1378, opened 2023-11-16
Ensure address is valid when uploading neighbour response

benbaumann95:
### Description of change

- Ensure address is valid when uploading neighbour response
- Wrap related updates to neighbour, neighbour response in a transaction so we don't have unexpected updated results for one record without the other
- Since we need to send letters to these addresses, we need to validate the address here too

### Story Link

https://trello.com/c/Yu9iwQwv/2078-when-address-isnt-three-lines-we-should-probably-catch-it-earlier

### Screenshots

Screenshot 2023-11-16 at 13 49 45

### DIFF
--- a/app/controllers/planning_applications/neighbour_responses_controller.rb
+++ b/app/controllers/planning_applications/neighbour_responses_controller.rb
@@ -18,49 +18,46 @@ module PlanningApplications
     end
 
     def create
-      @neighbour_response = @consultation.neighbour_responses
-        .build(neighbour_response_params.except(:address, :new_address, :files))
+      @neighbour_response = @consultation.neighbour_responses.new(neighbour_response_params.except(:address, :new_address, :files))
 
-      @neighbour_response.neighbour = find_neighbour
+      ActiveRecord::Base.transaction do
+        @neighbour_response.neighbour = find_or_build_neighbour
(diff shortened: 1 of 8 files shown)


app/controllers/planning_applications/neighbour_responses_controller.rb
```diff
@@ -18,49 +18,44 @@ def edit
...
-          end
-          create_audit_log(@neighbour_response, "uploaded")
-        end
-      else
-        render :new
+        @neighbour_response.save!
+        create_audit_log(@neighbour_response, "uploaded")
       end
+
+      redirect_to planning_application_consultation_neighbour_responses_path(@planning_application)
```
cdccollins, line 32: Is there a success message?

cdccollins, line 32: Is there a success message?

cdccollins: Also missing the format.html?

cdccollins (approved):

app/controllers/planning_applications/neighbour_responses_controller.rb
benbaumann95: I shall add in both, thanks

PR #1380, opened 2023-11-17
Add session_id and params fields to planx_planning_data

benbaumann95:
### Description of change

Adding session_id and param fields to planx_planning_data in preparation for the new ODP schema submission.
We want a relation between the old and new using the unique session_id. 
As part of this, we also want to drop the bulky `audit_log` field on planning applications and have it live in `params_v1` instead and this is the first necessary step towards that

### Story Link

https://trello.com/c/FM5ycejN/2159-receive-planning-applications-using-new-odp-submission-schema

### DIFF
--- a/app/services/planning_application_creation_service.rb
+++ b/app/services/planning_application_creation_service.rb
@@ -16,6 +16,7 @@ class PlanningApplicationCreationService
       @send_email = false
     else
       options.each { |k, v| instance_variable_set("@#{k}", v) unless v.nil? }
+      @session_id = params.dig(:planx_debug_data, :session_id)
     end
   end
 
@@ -25,7 +26,7 @@ class PlanningApplicationCreationService
 
   private
 
-  attr_reader :local_authority, :params, :api_user
+  attr_reader :local_authority, :params, :api_user, :session_id
 
   def build_planning_application
     planning_application = PlanningApplication.new(
@@ -57,7 +58,10 @@ class PlanningApplicationCreationService
         end
         if params[:planx_debug_data].present?
           PlanxPlanningData.create!(
-            planning_application:, entry: params[:planx_debug_data].to_json
+            planning_application:,
+            entry: params[:planx_debug_data].to_json,
+            session_id:,
+            params_v1: params.to_json
           )
         end
         ConstraintsCreationService.new(planning_application:,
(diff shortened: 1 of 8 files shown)


db/migrate/20231117134418_add_param_fields_to_planx_planning_data.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class AddParamFieldsToPlanxPlanningData < ActiveRecord::Migration[7.0]
+  def up
+    add_column :planx_planning_data, :params_v1, :jsonb
+    add_column :planx_planning_data, :params_v2, :jsonb
+
+    execute <<-SQL
+      UPDATE planx_planning_data
+      SET params_v1 = planning_applications.audit_log
```
benbaumann95, line 10: Have to make the code changes to use this new `params_v1` field on `planx_planning_data` and then eventually drop the audit_log on planning applications

app/services/planning_application_creation_service.rb
```diff
@@ -57,7 +57,10 @@ def save_planning_application!(planning_application)
         end
         if params[:planx_debug_data].present?
           PlanxPlanningData.create!(
-            planning_application:, entry: params[:planx_debug_data].to_json
+            planning_application:, 
+            entry: params[:planx_debug_data].to_json, 
+            session_id: params[:planx_debug_data][:session_id],
```
benbaumann95, line 62: This comes in like
```
"planx_debug_data": {
    "session_id": "21161b70-0e29-40e6-9a38-c42f61f25ab9",
    "passport": {
      "data": {}
    }
  },
  ```

File: app/services/planning_application_creation_service.rb [hunk 3]
```diff
@@ -14,6 +14,7 @@ def initialize(planning_application: nil, **options)
       @local_authority = planning_application.local_authority
       @api_user = planning_application.api_user
       @send_email = false
+      @session_id = SecureRandom.uuid
```
benbaumann95, line 17: Just generating a new session_id when a planning application is cloned. Can't use the same one due to the unique constraint and we don't care about the session_id for cloned applications

pixeltrix: Maybe make the column nullable? That way you can have the unique index but don't have fake session ids floating around.

cdccollins (approved):

PR #1382, opened 2023-11-20
Produce doc tags based on application type

cdccollins:
### Description of change

Produce document tags dynamically based on application type

### Story Link

https://trello.com/c/bu6RZstR/2153-show-relevant-tags-for-full-planning-applications

screencapture-southwark-bops-localhost-3000-planning-applications-2204-documents-7594-edit-2023-12-11-15_27_19
screencapture-southwark-bops-localhost-3000-planning-applications-2204-documents-7594-edit-2023-12-11-15_27_32

### DIFF
--- a/app/helpers/document_helper.rb
+++ b/app/helpers/document_helper.rb
@@ -17,6 +17,10 @@ module DocumentHelper
     Document::SUPPORT_TAGS.include?(tag)
   end
 
+  def irrelevant_tags(planning_application, key)
+    Document.send("#{key}_tags") - planning_application.application_type.document_tags[key]
+  end
+
   def created_by(document)
     if document.user.present?
       "This document was manually uploaded by #{document.user.name}."
--- a/app/helpers/document_helper.rb
+++ b/app/helpers/document_helper.rb
@@ -14,11 +14,7 @@ module DocumentHelper
   end
 
   def supporting_document_tag?(tag)
-    Document::SUPPORT_TAGS.include?(tag)
-  end
-
-  def irrelevant_tags(planning_application, key)
-    Document.send("#{key}_tags") - planning_application.application_type.document_tags[key]
+    Document::SUPPORTING_DOCUMENT_TAGS.include?(tag)
   end
 
   def created_by(document)
@@ -39,15 +35,5 @@ module DocumentHelper
 
   def reference_or_file_name(document)
     document.numbers.presence || document.name
-  end
-
-  def tags_first_half(tags)
-    count = (tags.count.round / 2.to_f).ceil
-    tags.in_groups_of(count).first
-  end
-
-  def tags_second_half(tags)
-    count = (tags.count.round / 2.to_f).ceil
-    tags.in_groups_of(count).second
   end
 end
--- a/app/helpers/document_helper.rb
+++ b/app/helpers/document_helper.rb
@@ -13,6 +13,10 @@ module DocumentHelper
     Document::EVIDENCE_TAGS.include?(tag)
   end
 
+  def supporting_document_tag?(tag)
+    Document::SUPPORT_TAGS.include?(tag)
+  end
+
   def created_by(document)
     if document.user.present?
       "This document was manually uploaded by #{document.user.name}."
@@ -31,5 +35,15 @@ module DocumentHelper
 
   def reference_or_file_name(document)
     document.numbers.presence || document.name
+  end
+
+  def tags_first_half(tags)
+    count = (tags.count.round / 2.to_f).ceil
+    tags.in_groups_of(count).first
+  end
+
+  def tags_second_half(tags)
+    count = (tags.count.round / 2.to_f).ceil
+    tags.in_groups_of(count).second
   end
 end
--- a/app/views/documents/edit.html.erb
+++ b/app/views/documents/edit.html.erb
@@ -50,25 +50,25 @@
       <strong>File name:</strong> <%= @document.name %>
     </p>
 
-    <p class="govuk-body govuk-body govuk-!-margin-bottom-0">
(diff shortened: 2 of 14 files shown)


app/helpers/document_helper.rb
```diff
@@ -32,4 +40,14 @@ def titled_reference_or_file_name(document)
...
+
+  def tags_first_half(tags)
+    count = (tags.count.round / 2.to_f).ceil
+    tags.in_groups_of(count).first
+  end
+
+  def tags_second_half(tags)
+    count = (tags.count.round / 2.to_f).ceil
+    tags.in_groups_of(count).second
+  end
```
pixeltrix, line 52: You don't need these methods - you can just use `in_groups(2, false)` to split the array into two halves (the `false` prevents the second array being filled with `nil` when there's an odd number).

app/views/documents/edit.html.erb
```diff
@@ -46,29 +46,29 @@
     <% end %>
   </div>
   <div class="govuk-grid-column-one-half">
-    <p class="govuk-body">
+    <p class="govuk-body govuk-!-margin-bottom-0">
       <strong>File name:</strong> <%= @document.name %>
     </p>
 
-    <p class="govuk-body">
+    <p class="govuk-body govuk-body govuk-!-margin-bottom-0">
```
benjamineskola, line 53: Some duplicated css classes in this file: `govuk-body` twice here and below.

app/helpers/document_helper.rb
```diff
@@ -13,6 +13,14 @@ def evidence_tag?(tag)
     Document::EVIDENCE_TAGS.include?(tag)
   end
 
+  def supporting_document_tag?(tag)
+    Document::SUPPORT_TAGS.include?(tag)
+  end
+
+  def irrelevant_tags(planning_application, key)
+    Document.send("#{key}_tags") - planning_application.application_type.document_tags[key]
```
pixeltrix, line 21: Lets try to avoid `send` - at a minimum it should be `public_send` and preferably not used at all. Not that performance is an issue with this call site but it prevents the JIT compiler from getting any performance benefit and bypasses the Ruby method cache.

pixeltrix (changes requested):

benjamineskola (approved):

app/views/documents/form_partials/_tags.html.erb
```diff
@@ -13,35 +14,69 @@
...
-              <%= b.label class: "govuk-label govuk-checkboxes__label tag_checkbox_label" %>
+          </div>
+          <div class="govuk-grid-column-one-half">
+            <div class="govuk-checkboxes govuk-checkboxes--small">
+              <%= form.collection_check_boxes :tags, tags_second_half(@planning_application.application_type.document_tags[key]), :itself, :itself do |b| %>
+                <div class="govuk-checkboxes__item">
+                  <%= b.check_box class: "govuk-checkboxes__input" %>
+                  <%= b.label class: "govuk-label govuk-checkboxes__label tag_checkbox_label" %>
+                </div>
+              <% end %>
```
pixeltrix, line 40: Sorry, wasn't clear - you can just use `in_groups` directly with a block, e.g.

``` erb
<% @planning_application.application_type.document_tags[key].in_groups(2, false) do |tags| %>
  <div class="govuk-grid-column-one-half">
    <div class="govuk-checkboxes govuk-checkboxes--small">
      <%= form.collection_check_boxes :tags, tags, :itself, :itself do |b| %>
        <div class="govuk-checkboxes__item">
          <%= b.check_box class: "govuk-checkboxes__input" %>
          <%= b.label class: "govuk-label govuk-checkboxes__label tag_checkbox_label" %>
        </div>
      <% end %>
    </div>
  </div>
<% end %>
```

You can do the same below with `irrelevant_tags`

cdccollins: ahhh gotcha.

app/helpers/document_helper.rb
```diff
@@ -13,6 +13,14 @@ def evidence_tag?(tag)
     Document::EVIDENCE_TAGS.include?(tag)
   end
 
+  def supporting_document_tag?(tag)
+    Document::SUPPORTING_DOCUMENT_TAGS.include?(tag)
+  end
+
+  def irrelevant_tags(planning_application, key)
+    Document.const_get("#{key.singularize.upcase}_TAGS") - planning_application.application_type.document_tags[key]
+  end
```
pixeltrix, line 22: Better to use a `case` statement here, e.g.

``` ruby
def irrelevant_tags(planning_application, key)
  document_tags = planning_application.application_type.document_tags[key]

  case key
  when "plans"
    Document::PLAN_TAGS - document_tags
  when "evidence"
    Document::EVIDENCE_TAGS - document_tags
  when "supporting_documents"
    Document::SUPPORTING_DOCUMENTS - document_tags
  else
    raise ArgumentError, "Unexpected document tag type: #{key}"
  end
end
```

That way it's obvious where any error is from the Appsignal notification

pixeltrix (dismissed):

app/helpers/document_helper.rb
pixeltrix: Maybe better to place this method on the model: ``` ruby
class ApplicationType < ApplicationRecord
  def irrelevant_tags(key)
    case key
    when "plans"
      Document::PLAN_TAGS - document_tags[key]
    when "evidence"
      Document::EVIDENCE_TAGS - document_tags[key]
    when "supporting_documents"
      Document::SUPPORTING_DOCUMENTS - document_tags[key]
    else
      raise ArgumentError, "Unexpected document tag type: #{key}"
    end
  end
end
```

cdccollins: yep that looks much better

PR #1383, opened 2023-11-20
Further routes improvements

EGiataganas:
### Description of change

- Move legislation under validation
- Move cancel_validation_requests member
- Move validation_tasks under validation

### Story Link

https://trello.com/c/MqPXLI3h


app/components/task_list_items/validating/legislation_component.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module TaskListItems
+  module Validating
```
benjamineskola, line 4: Should we be using `Validating` or `Validation`?

EGiataganas: The reason I did it `Validating` is because we already have `Reviewing`. By doing that, it will allow us to have a `validation` model in the future and not clash

config/routes.rb
```diff
@@ -172,6 +162,16 @@
...
 
+        resource :legislation, only: %i[show update]
+
+        concern :cancel_validation_requests do
+          member do
+            get :cancel_confirmation
+
+            patch :cancel
+          end
+        end
```
pixeltrix, line 173: Not for this PR but it should be a nested singleton resource like this:

``` ruby
concern :cancellable do
  resource :cancellation, only: %i[new show create]
end
```

The `CancellationsController` would still work with the validation request model, i.e. we wouldn't add a `Cancellation` model.

pixeltrix (approved):

PR #1389, opened 2023-11-21
Add initial skeleton for BOPS API engine

pixeltrix:
TODO: There's no authentication at the moment.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -7,7 +7,8 @@ require "sidekiq/web"
 Rails.application.routes.draw do
   root to: "planning_applications#index"
 
-  mount Rswag::Ui::Engine => "api-docs"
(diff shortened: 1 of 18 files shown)


config/routes.rb
```diff
@@ -7,7 +7,8 @@
 Rails.application.routes.draw do
   root to: "planning_applications#index"
 
-  mount Rswag::Ui::Engine => "api-docs"
+  mount BopsApi::Engine, at: "/api"
+  mount Rswag::Ui::Engine, at: "/api-docs"
```
pixeltrix, line 11: Eventually we should be able to mount the api docs in the BopsApi engine

benbaumann95 (approved):

EGiataganas (approved):

PR #1390, opened 2023-11-22
Use params_v1 instead of audit_log

benbaumann95:
### Description of change

- Use params_v1 instead of audit_log
- Since we are moving the params to the planx_planning_data table in this params_v1 field, we can remove calls to the audit_log. This field will be removed in a following PR


app/services/planning_application_creation_service.rb
```diff
@@ -56,14 +55,11 @@ def save_planning_application!(planning_application)
         if planning_application.from_production?
           PlanningApplicationAnonymisationService.new(planning_application:).call!
         end
-        if params[:planx_debug_data].present?
-          PlanxPlanningData.create!(
-            planning_application:,
-            entry: params[:planx_debug_data].to_json,
```
benbaumann95, line 62: We don't need this

app/models/planning_application.rb
```diff
@@ -78,6 +78,8 @@ class WithdrawOrCancelError < RuntimeError; end
     delegate :required?, to: :press_notice
     delegate :required?, to: :site_notice
   end
+  delegate :params_v1, to: :planx_planning_data, allow_nil: true
+  delegate :params_v2, to: :planx_planning_data, allow_nil: true
```
EGiataganas, line 82: as we don't have any `params_v2` related in this PR, I wouldn't add it. I would add it in the next PR which will use it

EGiataganas (approved):

PR #1391, opened 2023-11-22
Refactor validation requests to use one model and one controller

cdccollins:
### Description of change

### Story Link

https://link-to-issue

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


app/models/validation_request.rb
```diff
@@ -1,23 +1,362 @@
 # frozen_string_literal: true
 
 class ValidationRequest < ApplicationRecord
```
benjamineskola, line 3: I wonder if renaming this to `PlanningApplication::Validation::Request` would make sense? But that would end up being a fairly sizeable piece of work in itself, so probably better in another PR (that someone else can do).

app/views/planning_applications/validation/validation_requests/new.html.erb
```diff
@@ -0,0 +1 @@
+<%= render "planning_applications/validation/#{@type.underscore}s/new" %>
```
pixeltrix, line 1: Use the validation subclass itself to get the name, e.g.

``` irb
>> OtherChangeValidationRequest.model_name.plural
=> "other_change_validation_requests"
```

This should fix one of the brakeman errors

app/components/accordion_sections/site_map_component.rb
```diff
@@ -17,7 +17,8 @@ def change_request_link_path
         )
       else
         new_planning_application_validation_red_line_boundary_change_validation_request_path(
-          planning_application
+          planning_application,
+          type: "red_line_boundary_change"
```
benjamineskola, line 21: How come this needs the type of validation request both in the method name and as a parameter? Can we avoid that somehow?

app/models/document.rb
```diff
@@ -277,8 +276,7 @@ def audit_date_comment
 
   def reset_replacement_document_validation_request_update_counter!
     return unless validated? || archived?
-
-    return unless (request = ReplacementDocumentValidationRequest.find_by(new_document_id: id))
+    return unless (request = owner)
 
     request.reset_update_counter!
```
benjamineskola, line 281: A little concerned that the assignment here is too easily mistaken for a comparison (since assignment in a condition is unusual).

Maybe more like:

```ruby
    return if owner.nil?

    owner.reset_update_counter!
```

benjamineskola (approved):

PR #1399, opened 2023-11-23
Add creation service to parse new odp schema

benbaumann95:
### Description of change

- Create new creation service for handling new ODP submission schema. Validate the submission params against the JSON schema using json_schemer
- Only parses the planning application information so far.
- Create new planning application in a "pending" state which shouldn't reflect in the UI
- Add error handling for API requests
- Add swagger docs and example body

**Still to do**
- Parsing and creating **documents**, **constraints** and **immunity details**
- Ability to clone applications on staging
- Add better error handling messaging
- Improve and add to swagger docs

**Next steps**
- Request that PlanX send both submissions (we will create 2 planning applications per submission, using the old and new schemas). From this we can compare and add incrementally to the new approach until we are happy that there is parity with the applications and associations that we are creating

### Story Link

https://trello.com/c/W0pFoKHJ/2182-create-new-service-to-handle-new-odp-submission-schema

### Screenshots

Screenshot 2023-11-24 at 15 24 31


app/models/planning_application.rb
```diff
@@ -92,6 +93,7 @@ class WithdrawOrCancelError < RuntimeError; end
   scope :with_user, -> { preload(:user) }
   scope :for_user_and_null_users, ->(user_id) { where(user_id: [user_id, nil]) }
   scope :prior_approvals, -> { joins(:application_type).where(application_type: {name: :prior_approval}) }
+  scope :non_pending, -> { where.not(status: "pending") }
```
benjamineskola, line 96: Really tiny point but I wonder if there's a better name for this than `non_pending`? Perhaps something like `accepted`, for example. But I'm not completely certain what name would be clear without being potentially confused with a later step (like approval for example), and it's unlikely ever to be a big deal anyway.

engines/bops_api/app/services/bops_api/application/creation_service.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+module BopsApi
+  module Application
+    class CreationService
+      class CreateError < StandardError; end
+
+      def initialize(**options)
+        raise CreateError, "This service is not permitted in production" if Bops.env.production?
```
benjamineskola, line 9: If this error might be seen by end users then referring to a 'service' might not make sense. Maybe e.g.:

```suggestion
        raise CreateError, "Creating planning applications by this method is not permitted in production" if Bops.env.production?
```

app/models/planning_application.rb
benbaumann95: Hm, active might not be appropriate for closed cases though.

So accepted might work

engines/bops_api/app/services/bops_api/application/creation_service.rb
benbaumann95: An end user wouldn't ever see this on the frontend and we'd get an error come through on AppSignal. We'd send a response back to PlanX in an error json so we can still make the message clearer I think

benbaumann95: Update the message and added a spec for this too

engines/bops_api/app/services/bops_api/application/creation_service.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+module BopsApi
+  module Application
+    class CreationService
+      class CreateError < StandardError; end
+
+      def initialize(**options)
+        raise CreateError, "Creating planning applications using this endpoint is not permitted in production" if Bops.env.production?
```
pixeltrix, line 9: Seems like this is a different category of error to one where we capture an active record error?

engines/bops_api/app/controllers/bops_api/v2/planning_applications_controller.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module BopsApi
+  module V2
+    class PlanningApplicationsController < AuthenticatedController
+      def create
+        @planning_application = BopsApi::Application::CreationService.new(
+          local_authority: @local_authority, params:, api_user: @current_user
+        ).call
```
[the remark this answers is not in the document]
pixeltrix: We should also bring up with Jess about what's the expected return from a service that accepts submissions - seems like that should be part of the standard

engines/bops_api/app/services/bops_api/application/creation_service.rb
```diff
@@ -0,0 +1,93 @@
...
+
+        options.each { |k, v| instance_variable_set("@#{k}", v) unless v.nil? }
+      end
+
+      def call
+        planning_application = build_planning_application
```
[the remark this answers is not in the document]
pixeltrix: Actually, json-schema doesn't support the draft 07 format that Jess is using so it'll have to be the json_schemer gem

engines/bops_api/app/services/bops_api/application/parsers/address_parser.rb
```diff
@@ -0,0 +1,34 @@
...
+            lonlat: lonlat(params[:longitude], params[:latitude])
+          }
+        end
+
+        private
+
+        def lonlat(longitude, latitude)
+          return unless longitude.present? && latitude.present?
+
+          "POINT(#{longitude} #{latitude})"
```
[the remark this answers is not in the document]
pixeltrix: You need a factory to make the point, e.g:

``` ruby
factory = RGeo::Geographic.spherical_factory(srid: 4326)
factory.point(longitude, latitude)
```

engines/bops_api/app/controllers/concerns/bops_api/error_handler.rb
```diff
@@ -0,0 +1,53 @@
...
+      "NotPermittedError" => {status: 403, message: "Forbidden"},
+      "ActiveRecord::RecordNotFound" => {status: 404, message: "Cannot find record"},
+      "InternalServerError" => {status: 500, message: "Internal Server Error"}
+    }.freeze
+
+    class BaseError < StandardError; end
+
+    EXCEPTIONS.each_key do |exception_name|
+      const_set(exception_name, Class.new(BaseError)) unless const_defined?(exception_name)
+    end
```
pixeltrix, line 19: This looks like you're defining these errors inside `ErrorHandler`? Probably better to define them inside an Errors constant and statically in a file in app/lib/bops_api/errors.rb, e.g:

``` ruby
module BopsApi
  module Errors
    class BaseError < ::StandardError; end
    class InvalidSchemaError < BaseError; end
    class NotPermittedError < BaseError; end
    class InternalServerError < BaseError; end
  end
end
```

engines/bops_api/app/controllers/bops_api/application_controller.rb
```diff
@@ -1,7 +1,10 @@
 # frozen_string_literal: true
 
 module BopsApi
-  class ApplicationController < ActionController::Base
+  class ApplicationController < ActionController::API
+    include ActionController::MimeResponds
+    include BopsApi::ErrorHandler
```
pixeltrix, line 6: No need for the `BopsApi` prefix - Zeitwerk will look in the proper place first for `ErrorHandler`

engines/bops_api/app/controllers/concerns/bops_api/error_handler.rb
```diff
@@ -0,0 +1,53 @@
...
+      rescue_from StandardError do |exception|
+        handle_error(exception, EXCEPTIONS["InternalServerError"])
+      end
+
+      EXCEPTIONS.each do |exception_name, context|
+        exception_class = const_get(exception_name)
+        rescue_from exception_class do |exception|
+          handle_error(exception, context)
+        end
+      end
```
pixeltrix, line 32: There's no need to add multiple `rescue_from` bocks - just rescue `StandardError` and the mapping will do the rest - if there's no mapping fallback to 500 error

pixeltrix (approved):
The error handler stuff can be refactored in a later PR if necessary

EGiataganas (approved):

engines/bops_api/app/controllers/concerns/bops_api/error_handler.rb
benbaumann95: Ahh yes because we inherit from this anyway

benbaumann95:
> The error handler stuff can be refactored in a later PR if necessary

Thanks for the approval.
I've gone ahead and made those recommended changes to the errors 👍

PR #1400, opened 2023-11-24
Feature/email new users

Rhian:
### Description of change

The previous workflow required the admin to set a password, which would have enabled the admin to log in with the user's credentials. This change overrides some of the Devise `recoverable` methods so that when the user is created, they receive a tailored version of the reset password email welcoming them to the platform and asking them to set a password. I also needed to add the `confirmable` capabilities to ensure that the user is not able to log in until their account is confirmed.

When they arrive on the set password screen, they see a generic header 'Create your password' and when the password is saved, their status changes to `confirmed`.

The reset token expires after 6 hours and the user is prompted to request a new link.

### Story Link

https://trello.com/c/qSaxCTE6/1496-when-a-new-user-is-created-on-the-system-send-them-an-email-to-create-their-password
https://trello.com/c/SPBjc9AQ/1514-add-expiry-time-to-email-for-users-to-create-their-password

### Screenshots

create

### Decisions

My original idea was to extend the `confirmable` module with some entirely new methods but then I realised it was easier to simply update the reset password flow.

I have removed the 'Send Confirmation Instructions` screen to avoid confusion as the user now confirms themselves by setting a password.

I have not added a spec in `login_spec` to ensure successful login after confirmation as this is handled on line 85 of the `managing_users` spec.

### Further testing or sign off required 

1. Laura needs to look at the text for the Welcome email.
2. We need to agree expiry period with the POs


app/views/devise/passwords/edit.html.erb
```diff
@@ -2,7 +2,7 @@
   Change your password - <%= t('page_title') %>
 <% end %>
 
-<h1 class="govuk-heading-l">Change your password</h1>
+<h1 class="govuk-heading-l">Create your password </h1>
```
benjamineskola, line 5: Really tiny point but I wonder if this should say either "change" or "create" depending on whether this is new or a reset? "Create" does sort of work for a reset though so maybe it doesn't matter.

benjamineskola (approved):

app/views/devise/passwords/edit.html.erb
Rhian: I orignally had an `if` statement in the .erb to do exactly that, but then I decided the added complexity wasn't really worth it. Maybe I will reinstate, as you've also thought this :)

PR #1404, opened 2023-11-28
Add new cop for literal path and not use a url helper

EGiataganas:
### Description of change

Add new cop for literal path and not use a url helper

### DIFF
--- a/lib/bops/cop.rb
+++ b/lib/bops/cop.rb
@@ -6,4 +6,5 @@ module Bops
 end
 
 require_relative "cop/style/have_current_path_literal"
+require_relative "cop/style/have_link_href_literal"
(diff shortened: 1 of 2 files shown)


lib/bops/cop/style/have_literal_path.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module Bops
+  module Cop
+    module Style
+      class HaveLiteralPath < RuboCop::Cop::Base
```
pixeltrix, line 6: Would call it `HaveLinkLiteralPath`

pixeltrix: Actually, `HaveLinkHrefLiteral`

PR #1413, opened 2023-12-05
Improve API performance by caching schemas

pixeltrix:
Loading and parsing the schema takes in the order of 50ms so cache the JSONSchemer object for reuse. Also avoid re-parsing the parameters again.


benbaumann95 (approved):

.rubocop.yml
```diff
@@ -29,6 +29,9 @@ AllCops:
     - 'vendor/**/*'
     - 'engines/*/bin/*'
 
+Layout/EmptyLineBetweenDefs:
+  AllowAdjacentOneLineDefs: true
+
```
benjamineskola, line 34: I feel like it would be better to avoid overriding rules unless we absolutely have to. Any individual rule always seems fine but I wouldn't want to end up back where we were with the long list of special cases.

pixeltrix: The config `AllowAdjacentOneLineDefs` is on by default in rubocop 1.23 and later - not sure why it's not enabled here as it's a weird one for standard-rb to switch.

PR #1414, opened 2023-12-05
Remove duplicate and unused CSS classes

benjamineskola:
The problem with utility classes is that it's easy to end up with the same ones multiple times if you're not careful.

~~In this case `make-invisible` is older, but `display-none` is more widely-used and has a clearer name so I've kept that one.~~ Both of these duplicate an existing override class in the GOV.UK design system.

### DIFF
--- a/app/views/planning_applications/validation/other_change_validation_requests/_form.html.erb
+++ b/app/views/planning_applications/validation/other_change_validation_requests/_form.html.erb
@@ -35,7 +35,7 @@
           class: "govuk-label govuk-label--s",
           text: (t(".labels.summary") if @validate_fee)
         },
-        hint: { class: ("make-invisible" if @validate_fee) },
+        hint: { class: ("display-none" if @validate_fee) },
         rows: 5,
         readonly: @planning_application.validated?
       ) %>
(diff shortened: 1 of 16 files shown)


app/views/planning_applications/validation/other_change_validation_requests/_form.html.erb
```diff
@@ -35,7 +35,7 @@
           class: "govuk-label govuk-label--s",
           text: (t(".labels.summary") if @validate_fee)
         },
-        hint: { class: ("make-invisible" if @validate_fee) },
+        hint: { class: ("display-none" if @validate_fee) },
```
pixeltrix, line 38: Should we be using `govuk-!-display-none`?

benjamineskola: Arguably, perhaps. I'm not very familiar with the gov.uk design system, but if we can remove more one-liner classes that seems good.

benjamineskola: That's prompted me to find a few other cases where we don't need to create our own utility classes.

pixeltrix (approved):

EGiataganas (approved):

PR #1415, opened 2023-12-05
Fix applicants URLs

pixeltrix:
The order of 'planningapplications' and the subdomain was reversed.


pixeltrix:
The data has been corrected manually on production so the migration is modified for historical purposes.

EGiataganas (approved):

PR #1417, opened 2023-12-05
Create ImmunityDetails in new API

benjamineskola:
### Description of change

Create ImmunityDetails when processing an application through the new API and ODP schema

### Story Link

https://trello.com/c/7GoEzWoN/2220-handle-immunity-detail-creation-using-new-odp-schema

### DIFF
--- a/engines/bops_api/app/services/bops_api/application/creation_service.rb
+++ b/engines/bops_api/app/services/bops_api/application/creation_service.rb
@@ -136,6 +136,10 @@ module BopsApi
           end
         end
       end
+
+      def possibly_immune?(planning_application)
+        planning_application.immune_proposal_details.many?
+      end
     end
   end
 end
--- a/engines/bops_api/app/services/bops_api/application/creation_service.rb
+++ b/engines/bops_api/app/services/bops_api/application/creation_service.rb
@@ -80,8 +80,8 @@ module BopsApi
             DocumentsService.new(planning_application:, user:, files:).call!
 
             # TODO: create constraints
-            # TODO: create immunity details
             process_ownership_certificate_details(planning_application)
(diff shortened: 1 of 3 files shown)


engines/bops_api/app/services/bops_api/application/creation_service.rb
```diff
@@ -136,6 +136,45 @@ def process_ownership_certificate_details(planning_application)
           end
         end
       end
+
+      def possibly_immune?(planning_application)
+        # TODO make planning_application an instance variable to simply this method et seq.
+        planning_application.immune_proposal_details.count > 1
```
pixeltrix, line 142: I'd use `many?` as the `count` made me think that it was doing an SQL on an association but that's not the case as it's a `select` over an array returned from a jsonb field.

engines/bops_api/app/services/bops_api/application/creation_service.rb
```diff
@@ -136,6 +136,45 @@ def process_ownership_certificate_details(planning_application)
...
+      end
+
+      def process_immunity_details(planning_application)
+        ActiveRecord::Base.transaction do
+          immunity_detail = ImmunityDetail.new(planning_application: planning_application)
+          immunity_detail.end_date = planning_application
+            .find_proposal_detail("When were the works completed?")
```
pixeltrix, line 149: Should we be suggesting adding something to the schema for this?

pixeltrix (approved):

engines/bops_api/app/services/bops_api/application/creation_service.rb
benjamineskola: Yeah, I think it would be great if they would provide a slightly more processed form of these questions. I also think it's not ideal that we have to hunt through an array for an English-language string, etc.

PR #1422, opened 2023-12-11
Check the ownership certificate

cdccollins:
### Description of change

Officer can check the ownership certificate, send a validation request to bops-applicants, and receive the updates from bops-applicants

bops-applicants part of this is https://github.com/unboxed/bops-applicants/pull/186

### Story Link

https://trello.com/c/T7hgEyWA/2173-officer-can-check-ownership-certificate-information

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -76,7 +76,6 @@ Rails/ThreeStateBooleanColumn:
     - db/migrate/2022*
     - db/migrate/20230*
     - db/migrate/202310*
-    - db/migrate/202312*
 
 # False positive: if being used in migrations, behaviour is different and intentional
 Rails/ApplicationRecord:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -76,6 +76,7 @@ Rails/ThreeStateBooleanColumn:
     - db/migrate/2022*
     - db/migrate/20230*
     - db/migrate/202310*
+    - db/migrate/202312*
 
 # False positive: if being used in migrations, behaviour is different and intentional
 Rails/ApplicationRecord:
--- a/app/components/status_tags/ownership_certificate_component.rb
+++ b/app/components/status_tags/ownership_certificate_component.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module StatusTags
+  class OwnershipCertificateComponent < StatusTags::BaseComponent
+    def initialize(planning_application:)
+      @planning_application = planning_application
+    end
+
+    private
+
+    attr_reader :planning_application
+
+    def status
+      if planning_application.valid_ownership_certificate.nil?
+        :not_started
+      elsif planning_application.valid_ownership_certificate?
+        :valid
+      else
+        :invalid
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/ownership_certificate_validation_requests_controller.rb
+++ b/app/controllers/api/v1/ownership_certificate_validation_requests_controller.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class OwnershipCertificateValidationRequestsController < Api::V1::ApplicationController
+      skip_before_action :verify_authenticity_token
+      before_action :check_token_and_set_application
+
+      def index
+        respond_to do |format|
+          format.json do
+            @ownership_certificate_validation_requests = @planning_application.ownership_certificate_validation_requests
+          end
+        end
+      end
+
+      def show
+        respond_to do |format|
+          if (@ownership_certificate_validation_request =
+                @planning_application.ownership_certificate_validation_requests.where(id: params[:id]).first)
+            format.json
+          else
+            format.json do
+              render json: {message: "Unable to find ownership certificate validation request with id: #{params[:id]}"},
+                status: :not_found
+            end
+          end
+        end
+      end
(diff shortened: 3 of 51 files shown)


.rubocop.yml
```diff
@@ -76,6 +76,7 @@ Rails/ThreeStateBooleanColumn:
     - db/migrate/2022*
     - db/migrate/20230*
     - db/migrate/202310*
+    - db/migrate/202312*
```
benjamineskola, line 79: I put these in as path exemptions because I thought we might want to avoid three-state booleans in future but that it wouldn't make sense to try to modify the existing migrations to fix this.

If we're happy to continue to have new nullable boolean columns, then I think we should just disable this rule entirely, rather than add new exemptions. Or, otherwise, perhaps only disable it with a comment in the file in question.

app/controllers/api/v1/ownership_certificate_validation_requests_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+              render json: {message: "Unable to find ownership certificate validation request with id: #{params[:id]}"},
+                status: :not_found
+            end
+          end
+        end
+      end
+
+      def update
+        @ownership_certificate_validation_request =
+          @planning_application.ownership_certificate_validation_requests.where(id: params[:id]).first
```
pixeltrix, line 33: Seems like this is the same finder as in `show`? Can we extract it to a before action hook and use `model.association.find(params[:id])`. Rather than using a conditional in the `respond_to` block you can use a `rescue_from` block to capture the Active Record exception and return the not found error JSON.

app/controllers/api/v1/ownership_certificate_validation_requests_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+          @planning_application.update(valid_ownership_certificate: true)
+
+          if @ownership_certificate_validation_request.approved?
+            OwnershipCertificateCreationService.new(
+              params: ownership_certificate_params[:params], planning_application: @planning_application
+            ).call
+          end
+
+          render json: {message: "Change request updated"}, status: :ok
+        else
```
pixeltrix, line 48: I would use the `update!` method and then use `rescue_from` block for the Active Record invalid record exception to render the error json here.

app/models/ownership_certificate_validation_request.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class OwnershipCertificateValidationRequest < ValidationRequest
+  validates :reason, presence: true
+  validate :rejected_reason_is_present?
```
pixeltrix, line 5: This feels like we could use the presence validator, e.g.

``` ruby
validates :rejected_reason, presence: true, if: :rejected?
```

app/components/status_tags/ownership_certificate_component.rb
```diff
@@ -0,0 +1,23 @@
...
+    def initialize(planning_application:)
+      @planning_application = planning_application
+    end
+
+    private
+
+    attr_reader :planning_application
+
+    def status
+      if planning_application.valid_ownership_certificate.nil?
```
benjamineskola, line 14: I think we could do this instead:

```suggestion
      if planning_application.ownership_certificate.nil?
```

Then that would allow us to avoid a nullable boolean and keep Rubocop happy.

pixeltrix (changes requested):

.rubocop.yml
benjamineskola, line 79:
```suggestion
```

db/migrate/20231212163939_add_valid_ownership_certificate_to_planning_application.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddValidOwnershipCertificateToPlanningApplication < ActiveRecord::Migration[7.0]
+  def change
+    add_column :planning_applications, :valid_ownership_certificate, :boolean
```
benjamineskola, line 5:
```suggestion
    add_column :planning_applications, :valid_ownership_certificate, :boolean # rubocop:disable Rails/ThreeStateBooleanColumn 
```

benjamineskola (changes requested):

benjamineskola (approved):

app/controllers/api/v1/ownership_certificate_validation_requests_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+        @ownership_certificate_validation_request.close!
+        @ownership_certificate_validation_request.create_api_audit!
+        @planning_application.send_update_notification_to_assessor
+        @planning_application.update(valid_ownership_certificate: true)
+
+        if @ownership_certificate_validation_request.approved?
+          OwnershipCertificateCreationService.new(
+            params: ownership_certificate_params[:params], planning_application: @planning_application
+          ).call
+        end
```
pixeltrix, line 39: This feels like a lot of stuff outside a transaction block - there's a bunch of methods there with `!` which suggests that any of them could raise an exception and leave things in an indeterminate state. It might be better to extract this to a separate class which can be tested in isolation and not just in a system test.

pixeltrix (changes requested):

pixeltrix (approved):

PR #1426, opened 2023-12-13
Authenticate with unique token per local authority

benbaumann95:
### Description of change

- Using one token per environment (staging/production) for a multi tenanted architecture is not the most secure approach and we should have individual and unique tokens for each local authority
- Add a unique token for swagger which will work with all instances (for staging only)
- Update local authority to have many api users and scope the tokens through this association. Therefore we drop the globally unique constraint on name and token and add a composite index including the local authority

We can give PlanX the relevant tokens and they will be able to start using them without any code changes from our end. Once we are happy that they can submit to all local authorities with the individual tokens, we can drop the global token associated with each local authority manually from the database

### Story Link

https://trello.com/c/6QvdcEul/2221-add-unique-token-per-local-authority-and-environment

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -2,54 +2,52 @@
 
 require "faker"
 
-lambeth = LocalAuthority.find_or_create_by!(
-  council_code: "LBH",
-  subdomain: "lambeth",
-  short_name: "Lambeth",
-  council_name: "Lambeth Council",
-  applicants_url: "http://lambeth.bops-applicants.localhost:3001",
-  signatory_name: "Christina Thompson",
-  signatory_job_title: "Director of Finance & Property",
-  enquiries_paragraph: "Planning, London Borough of Lambeth, PO Box 734, Winchester SO23 5DG",
-  email_address: "planning@lambeth.gov.uk",
-  feedback_email: "digitalplanning@lambeth.gov.uk",
-  press_notice_email: "digitalplanning@lambeth.gov.uk"
-)
-southwark = LocalAuthority.find_or_create_by!(
-  council_code: "SWK",
-  subdomain: "southwark",
-  short_name: "Southwark",
-  council_name: "Southwark Council",
-  applicants_url: "http://southwark.bops-applicants.localhost:3001",
-  signatory_name: "Stephen Platts",
-  signatory_job_title: "Director of Planning and Growth",
-  enquiries_paragraph: "Planning, London Borough of Southwark, PO Box 734, Winchester SO23 5DG",
-  email_address: "planning@southwark.gov.uk",
-  feedback_email: "digital.projects@southwark.gov.uk",
-  press_notice_email: "digital.projects@southwark.gov.uk"
-)
+LocalAuthority.find_or_create_by!(subdomain: "lambeth") do |la|
+  la.council_code = "LBH"
+  la.short_name = "Lambeth"
+  la.council_name = "Lambeth Council"
+  la.applicants_url = "http://lambeth.bops-applicants.localhost:3001"
+  la.signatory_name = "Christina Thompson"
+  la.signatory_job_title = "Director of Finance & Property"
+  la.enquiries_paragraph = "Planning, London Borough of Lambeth, PO Box 734, Winchester SO23 5DG"
+  la.email_address = "planning@lambeth.gov.uk"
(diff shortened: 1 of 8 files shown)


db/seeds.rb
```diff
@@ -2,55 +2,62 @@
 
 require "faker"
 
-lambeth = LocalAuthority.find_or_create_by!(
```
benbaumann95, line 5: This is far too specific as some fields below may be different in your local db so just check if council code or subdomain exist before attempting to create it

benjamineskola: You can still use `find_or_create_by!` for this, by passing a block: the parameters to the method are the ones you search by (e.g., just `subdomain` perhaps), and then the block sets the remaining parameters.

benjamineskola:
```ruby
lambeth = LocalAuthority.find_or_create_by!(subdomain: "lambeth") do |la|
   la.council_code = "LBH"
   la.short_name = "Lambeth"
   la.council_name = "Lambeth Council"
   la.applicants_url = "http://lambeth.bops-applicants.localhost:3001"
   la.signatory_name = "Christina Thompson"
   la.signatory_job_title = "Director of Finance & Property"
   la.enquiries_paragraph = "Planning, London Borough of Lambeth, PO Box 734, Winchester SO23 5DG"
   la.email_address = "planning@lambeth.gov.uk"
   la.feedback_email = "digitalplanning@lambeth.gov.uk"
   la.press_notice_email = "digitalplanning@lambeth.gov.uk"
 end

Comment (benbaumann95) [reply db/seeds.rb:5]: Oh nice!

File: db/seeds.rb [hunk 2]
```diff
@@ -2,55 +2,55 @@
...
-)
+LocalAuthority.find_or_create_by!(subdomain: "southwark") do |la|
+  la.council_code = "SWK"
+  la.subdomain = "southwark"
+  la.short_name = "Southwark"
+  la.council_name = "Southwark Council"
+  la.applicants_url = "http://southwark.bops-applicants.localhost:3001"
+  la.signatory_name = "Stephen Platts"
+  la.signatory_job_title = "Director of Planning and Growth"
+  la.enquiries_paragraph = "Planning London Borough of Southwark PO Box 734 Winchester SO23 5DG"
```
benjamineskola, line 26:
```suggestion
  la.enquiries_paragraph = "Planning, London Borough of Southwark, PO Box 734, Winchester SO23 5DG"
```

benbaumann95: ahh good spot, the dangers of cmd + d replacing blindly

db/seeds.rb
```diff
@@ -2,55 +2,55 @@
...
-  signatory_name: "Steve Bambick",
-  signatory_job_title: "Director of Planning",
-  enquiries_paragraph: "Planning, Buckinghamshire Council, Gatehouse Rd, Aylesbury HP19 8FF",
-  email_address: "planning@buckinghamshire.gov.uk",
-  feedback_email: "planning.digital@buckinghamshire.gov.uk",
-  press_notice_email: "planning.digital@buckinghamshire.gov.uk"
-)
+LocalAuthority.find_or_create_by!(subdomain: "southwark") do |la|
+  la.council_code = "SWK"
+  la.subdomain = "southwark"
```
benjamineskola, line 20:
```suggestion
```

Slight nitpick: I don't think we need this in the block if it's set in the formal parameters.

db/seeds.rb
```diff
@@ -2,55 +2,55 @@
...
+  la.enquiries_paragraph = "Planning, London Borough of Southwark, PO Box 734, Winchester SO23 5DG"
+  la.email_address = "planning@southwark.gov.uk"
+  la.feedback_email = "digital.projects@southwark.gov.uk"
+  la.press_notice_email = "digital.projects@southwark.gov.uk"
+end
 
-ApiUser.find_or_create_by!(name: "api_user", token: (ENV["API_TOKEN"] || "123"))
+LocalAuthority.find_or_create_by!(subdomain: "buckinghamshire") do |la|
+  la.council_code = "BUC"
+  la.subdomain = "buckinghamshire"
```
benjamineskola, line 34:
```suggestion
```

As above.

benjamineskola (approved):

PR #1432, opened 2023-12-18
Show fee calculations

benjamineskola:
### Description of change

Calculate fee changes based on different sources and display them in a table.

### Story Link

https://trello.com/c/LcUWqL8i/2148-show-how-the-application-fee-has-been-calculated

### Screenshots

### Decisions [OPTIONAL]

### Known issues

I'm not sure how well the workflow around altering the fee works when displaying calculations like this. We are expecting to receive details of the calculation from PlanX but these are the details provided by the user themselves, effectively; so if the officer corrects that then any claim of a discount and so on may be invalidated. In that case we might want to actually be able to show the _officer's_ calculation in full. That seems out of scope for this card; I've settled instead for showing the corrected fee with a note in the calculation table that it's set by the officer.

e.g.


app/views/planning_applications/validation/fee_items/show.html.erb
```diff
@@ -15,6 +15,61 @@
...
+          </tr>
+          <tr class="govuk-table__row">
+            <th scope="row" class="govuk-table__header">
+              Total
+            </th>
+            <td class="govuk-table__cell">
+              £<%= @planning_application.fee_calculation.requested_fee || @planning_application.fee_calculation.payable_fee || 0 %>
+            </td>
+            <% if  @planning_application.fee_calculation.requested_fee.present? %>
+              <td>(set by PO)</td>
```
benjamineskola, line 66: I'm not sure if this is strictly correct, as I think not every council has planning officers doing validation? But I don't know what the better generic term would be.

Really this field is just added as a last-minute thing because I think we overlooked the case where an officer has changed the fee and how to display that. Might be worth reviewing it with the designers after the holiday.

benbaumann95: Rather than set by PO it would probably be better to capture the signed in user? Guess we can do that by either adding a user association or from the relevant audit entry.

db/schema.rb
```diff
@@ -44,8 +44,8 @@
   end
 
   create_table "api_users", force: :cascade do |t|
-    t.string "name", null: false
```
benbaumann95, line 47: Was this not updated from a previous migration?

benjamineskola: When I ran the migrations I got a few unrelated changes to the schema, I'm not sure where from specifically. I'll remove them from this PR.

app/views/planning_applications/validation/fee_items/show.html.erb
benjamineskola: That's true. I'd thought of the audit but wondered if searching through the audit log might be a mess. I think adding a user field here might make sense, but I wonder if it would be better to wait for design feedback before modifying the database again?

benbaumann95 (approved):

PR #1434, opened 2023-12-19
Add Planx to local authorities list for migration

Rhian:
### Description of change

We need to add PlanX to the list of local authorities on staging.

### Story Link

https://trello.com/c/UUPtXcRf/989-admin-can-update-council-specific-content-information-in-a-profile-page


EGiataganas (approved):

Rhian:
The previous migration has run with no issues now, so no need to fix. I will do this directly in the console.

PR #1438, opened 2023-12-20
Redesign administrator users index

Rhian:
### Description of change

This PR updates the users index page seen by the admin. There are now tabs to separate confirmed and unconfirmed users, and the user info has been condensed into three columns. A status bar appears if there are unconfirmed users, prompting the admin to resend confirmation emails.

A new path has been added which calls the action to resend the confirmation email, and links have been added on the index table and edit user table to allow admins to do this.

### Story Link

https://trello.com/c/QAQlm4WO/2320-update-users-page-in-bops-admin

### Screenshots

new_users

### Decisions 

Made an additional decision (not on Figma designs) to additionally add the Resend Invite to the Edit User page and provide a feedback flash to let the admin know the action was successful. Omitted the Inactive tab from the Users index as we don't currently inactivate users.

### DIFF
--- a/app/controllers/administrator/users_controller.rb
+++ b/app/controllers/administrator/users_controller.rb
@@ -6,8 +6,8 @@ module Administrator
 
     def index
       @users = current_local_authority.users
-      @confirmed_users = current_local_authority.users.select { |u| u.confirmed? }
-      @unconfirmed_users = current_local_authority.users.select { |u| u.unconfirmed? }
+      @confirmed_users = current_local_authority.users.confirmed
+      @unconfirmed_users = current_local_authority.users.unconfirmed
     end
 
     def new
--- a/app/controllers/administrator/users_controller.rb
+++ b/app/controllers/administrator/users_controller.rb
@@ -2,10 +2,12 @@
 
 module Administrator
   class UsersController < ApplicationController
-    before_action :set_user, only: %i[edit update]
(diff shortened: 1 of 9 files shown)


app/controllers/administrator/users_controller.rb
```diff
@@ -2,10 +2,12 @@
 
 module Administrator
   class UsersController < ApplicationController
-    before_action :set_user, only: %i[edit update]
+    before_action :set_user, only: %i[edit update resend_invite]
 
     def index
       @users = current_local_authority.users
+      @confirmed_users = current_local_authority.users.select { |u| u.confirmed? }
+      @unconfirmed_users = current_local_authority.users.select { |u| u.unconfirmed? }
```
benbaumann95, line 10: I think we could add some scopes to the user model

`scope :confirmed, -> { where.not(confirmed_at: nil) }`
`scope :unconfirmed, -> { where(confirmed_at: nil) }`

Then use the scoped association rather than a select query

app/views/administrator/users/index.html.erb
```diff
@@ -6,23 +6,51 @@
 
 <% content_for :title, t(".users") %>
 
+<% if current_user.local_authority.users.any?(&:unconfirmed?) %>
```
benbaumann95, line 9: Think we can update this too now to just
`@unconfirmed_users.any?`

benbaumann95 (approved):

PR #1443, opened 2023-12-29
Fix proposal details to JSON migration

benbaumann95:
### Description of change

There is a custom method `proposal_details` in the planning application model but we need to update the db column value. This fixes the migration code so it can run properly


cdccollins (approved):
I was getting an error, found this PR, and it fixed it 🎉

PR #1451, opened 2024-01-04
Restrict access to planning applications for inactive councils

Rhian:
### Description of change

We do not want assessors or reviewers to see the index page containing planning applications before the local authority is activated, because missing information will lead to broken views. 

We discussed simply redirecting users to the administrator pages on login, but this does not work as the user is stuck in a redirect loop when the existing method tries to send them back to the root path.

Instead, I have gone for the simpler option of allowing them to see the index page (so they still have access to navigation and can log out) but with a custom message explaining why it is blank.

We do not need to worry about planners accessing other pages, as we are not going to allow active councils to be edited to be inactive, so this only applies to newly configured councils who are in the process of being activated.

### Story Link

https://trello.com/c/8NN0YtXu/2378-do-not-allow-assessors-and-reviewers-to-access-the-system-when-a-local-authority-is-inactive

### DIFF
--- a/app/views/planning_applications/index.html.erb
+++ b/app/views/planning_applications/index.html.erb
@@ -2,68 +2,74 @@
   Planning Applications List - <%= t('page_title') %>
 <% end %>
 
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <% if @search.exclude_others? %>
-      <h1 class="govuk-heading-l">
-        Your planning applications
-      </h1>
-      <p class="govuk-body-m">
-        <strong><%= current_user.name %>,</strong> <%= role_name %>
-      </p>
-      <p class="govuk-body">
-        <%= link_to "Add new application", new_planning_application_path, class: "govuk-link" %>
-      </p>
-      <p class="govuk-body"><%= link_to "View all applications", planning_applications_path(view: 'all'), class: "govuk-button" %></p>
-    <% else %>
-      <h1 class="govuk-heading-l">
-        All planning applications
-      </h1>
-      <p class="govuk-body-m">
-        <strong><%= current_user.name %>,</strong> <%= role_name %>
-      </p>
-      <p class="govuk-body">
-        <%= link_to "Add new application", new_planning_application_path, class: "govuk-link" %>
-      </p>
-      <p class="govuk-body">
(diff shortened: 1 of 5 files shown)


db/seeds.rb
```diff
@@ -39,6 +47,8 @@
   la.email_address = "planning@buckinghamshire.gov.uk"
   la.feedback_email = "planning.digital@buckinghamshire.gov.uk"
   la.press_notice_email = "planning.digital@buckinghamshire.gov.uk"
+  la.email_reply_to_id = "550e8400-e29b-41d4-a716-446655440000"
+  la.active = true
 end
 
 admin_roles = %i[assessor reviewer administrator]
```
EGiataganas, line 54: I think is better to move [this](https://github.com/unboxed/bops/blob/main/app/services/letter_sending_service.rb#L56-L70) logic on [local_autohity model](https://github.com/unboxed/bops/blob/main/app/models/local_authority.rb). By doing that we won't need to update it here.

app/views/planning_applications/index.html.erb
```diff
@@ -2,68 +2,74 @@
...
+          PlanningApplications::PanelsComponent.new(
+            planning_applications: @planning_applications,
+            search: @search,
+            local_authority: current_local_authority
+          )
+        ) %>
+      </div>
     </div>
   </div>
-</div>
```
EGiataganas, line 69: How about moving this logic on a `before_action` in the controller similar to [this](https://github.com/unboxed/bops/blob/main/app/controllers/application_controller.rb#L73-L77) ?

Rhian: I thought about this, but the issue is that this would prevent the Administrator from logging in because we can't detect their role until after they have logged in. So it would have to apply only after the user has logged in. And in that case they need to be able to log out, so they need to see the nav bar or else the sessions get messed up. I kind of get that it's technically possible, but not sure how to implement it.

EGiataganas: you could expand and add a specific method [here](https://github.com/unboxed/bops/blob/main/app/controllers/users/sessions_controller.rb#L13)

Rhian:
Closing because the POs inform us that this would not be a likely scenario (council admins would not create users before the local authority is active)

PR #1454, opened 2024-01-04
Migrate json string to json

EGiataganas:
### Description of change

- Migrate `boundary_geojson` sting data from PlanningApplication table to JSON
- Migrate `geojson` sting data from PlanningApplicationConstraintsQuery table to JSON
- Migrate `specific_attributes` string data format from ValidationRequests to JSON

### Story Link

https://trello.com/c/ixRIKlIy


benjamineskola (approved):

db/migrate/20240105140341_migrate_geojson_string_to_json.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class MigrateGeojsonStringToJson < ActiveRecord::Migration[7.0]
```
benbaumann95, line 3: Shall we rename this migration to be more specific ?

benbaumann95 (commented):
There's a `with_boundary_geojson` trait for planning application that we need to update to the json format. It's still a string now

app/services/planning_application_creation_service.rb
```diff
@@ -43,6 +42,7 @@ def build_planning_application
 
     planning_application.assign_attributes(site_params) if site_params.present?
     planning_application.assign_attributes(result_params) if result_params.present?
+    planning_application.assign_attributes(boundary_geojson: params[:boundary_geojson]) if params[:boundary_geojson].present?
```
benbaumann95, line 45: Since we are updating to use json instead of the string I think we need to update this to use permitted params, outlining the nested structure we are expecting

EGiataganas: I've first updated it and deleted this line. But the `if` statement questioned me whether I had to add it on `permitted params` since we don't require it. Added it back now and kept the `if` statement

db/migrate/20240105140341_migrate_geojson_string_to_json.rb
EGiataganas: I've followed the pattern with others, although happy to make it `MigrateGeojsonStringFromPlanningApplicationConstraintsQueryToJson`

app/models/red_line_boundary_change_validation_request.rb
```diff
@@ -19,6 +16,28 @@ def update_planning_application!(params)
     planning_application.update!(boundary_geojson: new_geojson)
   end
 
+  def new_geojson=(value)
+    if value.is_a?(String)
+      return if value.blank?
+      value.gsub!(/(?<!\\)"=>nil/, '":null')
+      value.gsub!(/(?<!\\)"=>/, '":')
```
pixeltrix, line 23: These shouldn't be necessary - we shouldn't have to deal with hash strings, just json strings.

spec/system/planning_applications/sitemap_spec.rb
```diff
@@ -156,7 +156,7 @@
       expect(map["showMarker"]).to eq("true")
 
       find(".govuk-visually-hidden",
-        visible: false).set '{"type":"Feature","properties":{},"geometry":{"type":"Polygon","coordinates":[[[-0.076715,51.501166],[-0.07695,51.500673],[-0.076,51.500763],[-0.076715,51.501166]]]}}'
+        visible: false).set({"type" => "Feature", "properties" => {}, "geometry" => {"type" => "Polygon", "coordinates" => [[[-0.076715, 51.501166], [-0.07695, 51.500673], [-0.076, 51.500763], [-0.076715, 51.501166]]]}})
```
pixeltrix, line 159: This is setting it to a hash string, not a json string so is probably why you need the `gsub` calls in the setter method

spec/system/planning_applications/sitemap_spec.rb
```diff
@@ -251,7 +251,7 @@
       end
 
       let(:new_geojson) do
-        {type: "FeatureCollection", features: [new_geojson_feature]}.to_json
+        {"type" => "FeatureCollection", "features" => [new_geojson_feature]}
```
pixeltrix, line 254: This is used on line 258 which will set a hash string - either call `to_json` here or on line 258

spec/system/planning_applications/sitemap_spec.rb
```diff
@@ -301,7 +301,7 @@
       expect(map["showMarker"]).to eq("true")
 
       # Draw proposed red line boundary
-      find(".govuk-visually-hidden", visible: false).set '{"type":"Feature","properties":{},"geometry":{"type":"Polygon","coordinates":[[[-0.076715,51.501166],[-0.07695,51.500673],[-0.076,51.500763],[-0.076715,51.501166]]]}}'
+      find(".govuk-visually-hidden", visible: false).set({"type" => "Feature", "properties" => {}, "geometry" => {"type" => "Polygon", "coordinates" => [[[-0.076715, 51.501166], [-0.07695, 51.500673], [-0.076, 51.500763], [-0.076715, 51.501166]]]}})
```
pixeltrix, line 304: Again it's a hash string - call `to_json` on it

pixeltrix (changes requested):

pixeltrix (approved):

PR #1457, opened 2024-01-05
Redo send letters

cdccollins:
### Description of change

Make the "send letters to neighbours" page better

### Story Link

https://trello.com/c/ZPr1uQ0s/2326-add-review-neighbours-to-send-letters-to-table

### Screenshots
screencapture-southwark-bops-localhost-3000-planning-applications-2103-consultation-neighbour-letters-2024-01-08-15_57_47


app/javascript/controllers/polygon_search_controller.js
```diff
@@ -31,7 +30,7 @@ export default class extends Controller {
     return document.getElementById("address-container")
   }
 
-  getConsultationNeighbourAddressesForm() {
+  getHiddenAddressesField() {
     return document.getElementById("addresses-hidden")
   }
```
benjamineskola, line 35: Probably not worth changing this here, but rather than looking up elements with `getElementById` we could also make them [Stimulus targets](https://stimulus.hotwired.dev/reference/targets), something like `data-polygon-search-target="hidden-addresses"`, which is then available as `this.hiddenAddressesTarget` and avoids having an extra function.

benjamineskola (approved):

PR #1460, opened 2024-01-08
Return more than 100 addresses for polygon search

benbaumann95:
### Description of change

- There is a hard limit of 100 results returned by the OS Places API - https://osdatahub.os.uk/docs/places/technicalSpecification when making a polygon search query. This is reflected in the maxresults parameter which takes a value between 1-100.
However, it might be the case that we want more than 100 results returned so we loop until a max total results set by us and keep making the api request whilst updating the offset parameter so we can return all the addresses in that range.

### Story Link

https://trello.com/c/ve4KXoVO/1902-allow-for-more-than-100-addresses-to-be-added-as-part-of-searching-neighbour-addresses-using-polygon


benjamineskola (approved):

app/services/apis/os_places/polygon_search_service.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+require "faraday"
+
+module Apis
+  module OsPlaces
+    class PolygonSearchService < Client
+      RESULTS_PER_PAGE = 100
+      MAX_TOTAL_RESULTS = 1000
```
benjamineskola, line 9: Can I just check where this number comes from? Is it that we know OS Places will never return more than 1000 results, or that we would never want to send more than 1000 letters?

cdccollins (approved):

EGiataganas (approved):

app/services/apis/os_places/polygon_search_service.rb
benbaumann95: An arbitrary number I just made up. Need to have a limit of some sort otherwise we could theoretically have a (near) infinite loop if the polygon covered a large enough area.

benjamineskola: Yeah, we don't want it completely unbounded. Let's keep an eye on it, it's _probably_ fine but I can imagine cases where there are a number of large blocks of flats nearby and we might actually legitimately reach that limit. (And although the user could then add more in a separate request it might be difficult to know which ones had been missed out. But this is definitely an edge case we might not ever need to be concerned with.)

PR #1463, opened 2024-01-09
Refactor the api/validation_requests_controllers

cdccollins:
### Description of change

There is now a ValidationRequestUpdateService, so the other validation requests should use it

I haven't done replacement_documents and additional_documents because they seemed too different?

### DIFF
--- a/app/services/validation_request_update_service.rb
+++ b/app/services/validation_request_update_service.rb
@@ -3,18 +3,20 @@
 class ValidationRequestUpdateService
   class UpdateError < StandardError; end
 
-  def initialize(validation_request:, params:, ownership_certificate: false, red_line_boundary_change: false)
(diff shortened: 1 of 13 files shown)


benjamineskola (approved):

app/services/validation_request_update_service.rb
```diff
@@ -16,26 +16,35 @@ def call!
...
     @params.require(:data).permit(
       params: [:certificate_type, land_owners: %i[name address_1 address_2 town city postcode notice_given notice_given_at]]
     )
   end
 
   def validation_request_params
-    @params.require(:data).permit(:approved, :rejection_reason)
+    @params.require(:data).permit(:approved, :rejection_reason, :response, supporting_documents: [])
   end
 end
```
EGiataganas, line 50: How about adding `attr_reader :planning_application, :validation_request, :params`

PR #1471, opened 2024-01-15
Added link to our notion page which shows guides for setting notify

EGiataganas:
### Description of change

Added a link to our notion page which shows guides for setting notify

### Story Link

https://trello.com/c/e4frF4I3

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -34,16 +34,16 @@ class LocalAuthority < ApplicationRecord
 
   before_update :set_active
 
+  def self.notify_notion_link
+    "https://oasis-marsupial-465.notion.site/Guide-to-Notify-set-up-7c18a8f3d43444d098c1f79eab48016c?pvs=74"
+  end
+
   def signatory
     "#{signatory_name}, #{signatory_job_title}"
   end
 
   def document_checklist?
     I18n.exists?("council_documents.#{subdomain}.document_checklist")
-  end
-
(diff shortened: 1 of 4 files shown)


app/models/local_authority.rb
```diff
@@ -34,6 +34,10 @@ class LocalAuthority < ApplicationRecord
 
   before_update :set_active
 
+  def self.notify_notion_link
```
benbaumann95, line 37: Why does this need to be a method?
Can't it just live directly in the yml file if it's not going to change.

benjamineskola: I don't think we're consistent about this. For example, we have links to GOV.UK advice, legislation, planning application fees, and Notify support in the views. We have other legislation, council documents, and Notify documentation in the locales.

benbaumann95: I'd still argue we should be putting static links or content that doesn't change based on application logic in the locale unless it's dynamic like `map_link` but yeah there doesn't seem to be all that much consistency.

benjamineskola: > I'd still argue we should be putting static links or content that doesn't change based on application logic in the locale

Yep — I didn't mean to sound like I was disagreeing with you, more that we should aim to increase consistency where possible.

Best to avoid having any content in the views really, we should try to put it all in the locales, and I think that should include URLs as well as the displayed text.

benjamineskola (approved):

PR #1474, opened 2024-01-16
Mark application as requiring an EIA

benbaumann95:
### Description of change

- If EIA (Environment Impact Assessment) is required, then extend the expiry date by another 8 weeks.
- Add this change to the audit log too.
- If application has been marked as requiring an EIA and then no longer needs it, set expiry date back to its original

### Story Link

https://trello.com/c/U6EFzljs/2171-able-to-recognise-an-eia-application-so-the-expiry-date-is-16-weeks

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -108,6 +108,7 @@ class PlanningApplication < ApplicationRecord
                 }, if: :valid_fee?
   before_update :audit_update_application_type!, if: :application_type_id_changed?
   before_update :create_proposal_measurement, if: :changed_to_prior_approval?
+  before_update :modify_expiry_date, if: :environment_impact_assessment_changed?
 
   after_update :audit_updated!
   after_update :update_constraints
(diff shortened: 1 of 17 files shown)


benjamineskola (approved):

app/models/planning_application.rb
```diff
@@ -935,4 +952,12 @@ def filter_documents_for_tab(documents, tab)
 
     documents.select { |document| (tags & document.tags).any? }
   end
+
+  def modify_expiry_date
+    if environment_impact_assessment
+      self.expiry_date += 56.days unless environment_impact_assessment_was
```
cdccollins, line 958: What is `environment_impact_assessment_was`?

benbaumann95: What the previous value of it was before the update. 
It will cover both false and nil cases. It effectively says, "if the previous value was not true (i.e., it was either false or nil), then update the expiry date."

PR #1476, opened 2024-01-17
Use value from ODP schema for immunity date

benjamineskola:
### Description of change

Rather than parsing the proposal-details array we can use the value in the schema as-is.

### Story Link

https://link-to-issue

### Decisions [OPTIONAL]

I've left the previous code (using proposal details) in place as a fallback. I'm not sure whether it will ever occur in practice but it seems to be the case for the test data.

### DIFF
--- a/engines/bops_api/app/services/bops_api/application/creation_service.rb
+++ b/engines/bops_api/app/services/bops_api/application/creation_service.rb
@@ -171,7 +171,9 @@ module BopsApi
       def process_immunity_details(planning_application)
         ActiveRecord::Base.transaction do
           immunity_detail = ImmunityDetail.new(planning_application: planning_application)
-          immunity_detail.end_date = planning_application
+
+          immunity_detail.end_date = params.dig("data", "proposal", "date", "completion")
+          immunity_detail.end_date ||= planning_application
             .find_proposal_detail("When were the works completed?")
             .first.response_values.first
           immunity_detail.save!
--- a/engines/bops_api/app/services/bops_api/application/creation_service.rb
+++ b/engines/bops_api/app/services/bops_api/application/creation_service.rb
@@ -95,7 +95,7 @@ module BopsApi
       def initialize_from_planning_application(planning_application)
         params_v2 = planning_application.params_v2 || raise_not_permitted_to_clone_error
 
-        @params = ActionController::Parameters.new(JSON.parse(params_v2))
+        @params = params_v2.with_indifferent_access
         @local_authority = planning_application.local_authority
         @user = planning_application.api_user
         @send_email = false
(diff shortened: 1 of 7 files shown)


engines/bops_api/app/services/bops_api/application/creation_service.rb
```diff
@@ -97,7 +97,10 @@ def send_email?(planning_application)
       def initialize_from_planning_application(planning_application)
         params_v2 = planning_application.params_v2 || raise_not_permitted_to_clone_error
 
-        @params = ActionController::Parameters.new(JSON.parse(params_v2))
+        # TODO not sure when this is needed: maybe only for test setup?
+        params_v2 = JSON.parse(params_v2) unless params_v2.respond_to?(:with_indifferent_access)
+
+        @params = ActionController::Parameters.new(params_v2)
```
pixeltrix, line 103: Do we need this to be `ActionController::Parameters`? Seems like it should be the same as submitted originally, i.e. a hash with indifferent access.

engines/bops_api/app/services/bops_api/application/creation_service.rb
```diff
@@ -144,7 +147,9 @@ def possibly_immune?(planning_application)
       def process_immunity_details(planning_application)
         ActiveRecord::Base.transaction do
           immunity_detail = ImmunityDetail.new(planning_application: planning_application)
-          immunity_detail.end_date = planning_application
+
+          immunity_detail.end_date = planning_application.params_v2.dig("data", "proposal", "date", "completion")
```
pixeltrix, line 151: Should this be accessing the params in `@params` ?

engines/bops_api/app/services/bops_api/application/creation_service.rb
```diff
@@ -144,7 +147,9 @@ def possibly_immune?(planning_application)
...
+          immunity_detail.end_date ||= planning_application
```
pixeltrix, line 152: Why are we setting the `end_date` to the planning application if the above line is nil?

spec/system/planning_applications/clone_spec.rb
```diff
@@ -117,7 +117,7 @@
         expect(page).to have_current_path("/planning_applications/#{cloned_planning_application.id}")
         expect(page).to have_content("Planning application was successfully cloned")
 
-        expect(JSON.parse(planning_application.params_v2)).to eq(JSON.parse(cloned_planning_application.params_v2))
+        expect(JSON.parse(planning_application.params_v2)).to eq(cloned_planning_application.params_v2)
```
pixeltrix, line 120: This seems to be messed up test setup - the params are a string instead of being a hash

engines/bops_api/app/services/bops_api/application/creation_service.rb
benjamineskola: We're not. We're setting it to the applicant's response from the proposal details, i.e., the previous behaviour, in case the `date` field is blank but the question exists — which seems to be the case in the test data.

cdccollins (approved):

PR #1477, opened 2024-01-17
Set font correctly

cdccollins:
### Description of change

We shouldn't be overwriting $govuk-font-family by setting !important on font-family. This then fixes the site notice spacing too.

### DIFF
--- a/app/views/planning_applications/site_notices/_status.html.erb
+++ b/app/views/planning_applications/site_notices/_status.html.erb
@@ -17,7 +17,7 @@
 <div class="govuk-!-display-none">
   <span id="application-reference"><%= @planning_application.reference %></span>
   <div id="site-notice-content">
-    <div style="width: 430px;">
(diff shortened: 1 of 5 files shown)


benjamineskola (approved):

app/views/planning_applications/site_notices/_status.html.erb
```diff
@@ -17,7 +17,7 @@
 <div class="govuk-!-display-none">
   <span id="application-reference"><%= @planning_application.reference %></span>
   <div id="site-notice-content">
-    <div style="width: 430px;">
+    <div style="width: 430px; font-family: Arial;">
```
benjamineskola, line 20: Putting Arial in the `govuk-font-family` instead of here might also work, as a fallback after open sans, but it occurred to me too late and it's probably not worth messing further with it for the time being.

PR #1479, opened 2024-01-17
Fix line height on PDFs

cdccollins:
### Description of change

Fix line height in the PDFs to see if that stops so much variation between machines


Rhian (approved):
LGTM!

PR #1481, opened 2024-01-18
Migrate polygon_geojson from consultations to json

EGiataganas:
### Description of change

Migrate `polygon_geojson` from consultations to json

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2024_01_16_115538) do
+ActiveRecord::Schema[7.0].define(version: 2024_01_18_114324) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
   enable_extension "postgis"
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2024_01_16_115538) do
+ActiveRecord::Schema[7.0].define(version: 2024_01_18_114324) do
```
benjamineskola, line 13: This isn't changing the schema so doesn't need to alter the schema version number. Or is there a missing schema change here?

benjamineskola (changes requested):

db/schema.rb
EGiataganas: By generating a new migration it automatically change that. Similarly [here](https://github.com/unboxed/bops/pull/1454/files#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3)

benjamineskola: Ah, it's just Rails being annoying then.

benjamineskola (approved):

PR #1485, opened 2024-01-18
Rubocop fixes/updates

benjamineskola:
Updating standardrb enables a lot of new rules, mainly rails-related. Most of these can be autocorrected. The major exception is `Rails/SaveBang`; there are a few false positives (where the value is being returned and handled elsewhere) and also a number of tests where we seem to depend on that behaviour.


app/models/api_user.rb
```diff
@@ -8,7 +8,7 @@ class ApiUser < ApplicationRecord
 
   has_secure_token :token, length: 36
 
-  validates :name, presence: true, uniqueness: true
+  validates :name, presence: true, uniqueness: true # rubocop:disable Rails/UniqueValidationWithoutIndex
```
pixeltrix, line 11: We should add the index and scope the uniqueness to the local authority.

app/models/condition_set.rb
```diff
@@ -4,7 +4,7 @@ class ConditionSet < ApplicationRecord
   include Reviewable
 
   belongs_to :planning_application
-  has_many :conditions, extend: ConditionsExtension
+  has_many :conditions, extend: ConditionsExtension # rubocop:disable Rails/HasManyOrHasOneDependent
```
pixeltrix, line 7: This should be `dependent: :destroy` as the planning application has `dependent: :destroy` on the `:condition_set`

app/models/ownership_certificate.rb
```diff
@@ -2,7 +2,7 @@
 
 class OwnershipCertificate < ApplicationRecord
   belongs_to :planning_application
-  has_many :land_owners
+  has_many :land_owners # rubocop:disable Rails/HasManyOrHasOneDependent
```
pixeltrix, line 5: Same here, planning application is `dependent: :destroy` on `has_one :ownership_certificate`

app/models/replacement_document_validation_request.rb
```diff
@@ -8,7 +8,7 @@ class ReplacementDocumentValidationRequest < ValidationRequest
   delegate :invalidated_document_reason, to: :old_document
   delegate :validated?, :archived?, to: :new_document, prefix: :new_document
 
-  has_one :new_document, as: :owner, class_name: "Document"
+  has_one :new_document, as: :owner, class_name: "Document" # rubocop:disable Rails/HasManyOrHasOneDependent
```
pixeltrix, line 11: This one as well - planning application has `dependent: :destroy` on `has_many :validation_requests`

db/migrate/20231124154049_migrate_validation_request_data.rb
```diff
@@ -76,7 +76,7 @@ def up
 
       vr.save(validate: false)
 
-      Document.where(additional_document_validation_request_id: request.id).each do |document|
+      Document.where(additional_document_validation_request_id: request.id).find_each do |document|
```
pixeltrix, line 79: This is fine here but it's an odd cop - `each` and `find_each` are not always directly replaceable as `find_each` will discard any `ORDER BY` clause to sort by id.

pixeltrix (changes requested):

pixeltrix (approved):

db/migrate/20240122111953_add_index_to_api_user_name.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddIndexToApiUserName < ActiveRecord::Migration[7.0]
+  def change
+    add_index :api_users, :name, unique: true
```
pixeltrix, line 5: This needs to be unique on `[:local_authority_id, :name]`

pixeltrix (changes requested):

app/models/user.rb
```diff
@@ -161,6 +161,6 @@ def legacy_otp_secret
     #    required for streaming cipher modes but including it is a best practice
     #    so that your code will continue to function correctly even if you later
     #    change to a block cipher mode.
-    cipher.update(cipher_text) + cipher.final
+    cipher.update!(cipher_text) + cipher.final
```
pixeltrix, line 164: This is a false positive - `cipher` isn't an Active Record instance

pixeltrix (changes requested):

app/models/user.rb
benjamineskola: Oh yeah, I've ignored that locally but apparently missed committing it.

db/schema.rb
```diff
@@ -53,6 +53,7 @@
     t.string "service"
     t.index ["local_authority_id"], name: "ix_api_users_on_local_authority_id"
     t.index ["name", "local_authority_id"], name: "index_api_users_on_name_and_local_authority_id", unique: true
```
pixeltrix, line 55: Looks like this index already exists but it's not using the prefix `ix_` so rubocop didn't pick it up - update the migration to drop this index if it exists.

db/schema.rb
```diff
@@ -51,9 +51,10 @@
     t.bigint "local_authority_id"
     t.jsonb "file_downloader"
     t.string "service"
+    t.index ["local_authority_id", "name"], name: "ix_api_users_on_local_authority_id__name", unique: true
+    t.index ["local_authority_id", "token"], name: "ix_api_users_on_local_authority_id__token", unique: true
     t.index ["local_authority_id"], name: "ix_api_users_on_local_authority_id"
-    t.index ["name", "local_authority_id"], name: "index_api_users_on_name_and_local_authority_id", unique: true
-    t.index ["token", "local_authority_id"], name: "index_api_users_on_token_and_local_authority_id", unique: true
+    t.index ["name"], name: "ix_api_users_on_name", unique: true
```
pixeltrix, line 57: Not sure where this index definition is from?

pixeltrix (approved):

PR #1490, opened 2024-01-22
Add fee and address to Environmental Assessment validation step

cdccollins:
### Description of change

Allow officers to add a fee and address where the public can view the Environmental Assessment to the validation step

### Story Link

This will allow the Site notice and Press notice work to go ahead, since both rely on this information


db/migrate/20240122104018_add_eia_information_to_planning_application.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddEiaInformationToPlanningApplication < ActiveRecord::Migration[7.0]
+  def change
+    add_column :planning_applications, :eia_address, :string
+    add_column :planning_applications, :eia_fee, :integer
+  end
```
benbaumann95, line 7: Might we want to create an EIA table at this rate? Only applies to certain application types too so might be worthwhile

Think there's 3 columns now associated with it

app/models/environment_impact_assessment.rb
```diff
@@ -0,0 +1,35 @@
...
+    audit!(
+      activity_type: "updated",
+      activity_information: "Environment impact assessment",
+      audit_comment: comment
+    )
+  end
+
+  def modify_expiry_date
+    planning_application.modify_expiry_date
+  end
```
benbaumann95, line 34: Think you could just delegate this to planning application instead so this method isn't needed

app/models/planning_application.rb
```diff
@@ -746,16 +745,22 @@ def generate_document_tabs(tabs = Document::DEFAULT_TABS)
...
+    end
+  end
+
+  def modify_expiry_date
+    if environment_impact_assessment.required?
+      self.expiry_date += 56.days
     else
-      raise ArgumentError, "Unexpected value for 'environment_impact_assessment': #{environment_impact_assessment.inspect}"
+      self.expiry_date = DAYS_TO_EXPIRE.days.after(validated_at || received_at)
+      self.target_date = 35.days.after(validated_at || received_at)
```
benbaumann95, line 760: Is this the same as `set_key_dates`. May want this to be consistent in case we change it in one place

app/views/planning_applications/validation/environment_impact_assessments/_form.html.erb
```diff
@@ -1,39 +1,109 @@
-<%= form_with model: [@planning_application], url: planning_application_validation_environment_impact_assessment_url(@planning_application) do |form| %>
-  <%= form.govuk_error_summary(presenter: ErrorPresenter) %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: [@environment_impact_assessment], url: planning_application_validation_environment_impact_assessment_url(@planning_application) do |form| %>
+      <% if @environment_impact_assessment.errors.any? %>
```
benbaumann95, line 4: Instead of this block does just adding 
```
  <%= form.govuk_error_summary(presenter: ErrorPresenter) %>
  ```
  do the same trick?

benbaumann95 (approved):

app/views/planning_applications/validation/environment_impact_assessments/_form.html.erb
cdccollins: Do you know how to do the different title errors vs the errors that get shown on the form? That's why I've got this separate block here

benbaumann95: Not off the top of my head, maybe just go ahead and merge and I can attempt as part of my PR

PR #1493, opened 2024-01-22
Allow overriding neighbour letter response period

benjamineskola:
### Description of change

Allow user to specify a deadline for neighbour responses, defaulting to the existing 21 days, and extend the consultation period as necessary if this is longer than the current consultation period.

### Story Link

https://trello.com/c/7gIhVzv1/2474-able-to-change-the-duration-of-neighbour-consultation-period

### Screenshots

Screenshot 2024-01-23 at 16 55 37

### Known issues

I'm not entirely happy with the way that the consultation is updated, it feels a little tangled due to the consultation potentially not having started when the letters are sent. I'm not sure if this is just an artefact of the test setup though.

### DIFF
--- a/app/controllers/planning_applications/neighbour_letters_controller.rb
+++ b/app/controllers/planning_applications/neighbour_letters_controller.rb
@@ -54,7 +54,9 @@ module PlanningApplications
         end
       end
     rescue Neighbour::AddressValidationError => e
-      redirect_address_validation_error(e)
+      redirect_after_rescue(e)
+    rescue ActiveRecord::RecordInvalid => e
+      redirect_after_rescue(e)
     end
 
     private
@@ -72,6 +74,7 @@ module PlanningApplications
         :neighbour_letter_text,
         :resend_existing,
         :resend_reason,
+        :deadline_extension,
         :polygon_geojson,
         neighbours_attributes: %i[id selected]
       )
@@ -114,7 +117,7 @@ module PlanningApplications
     end
 
     def update_consultation!
-      @consultation.update!(consultation_params.except(:resend_existing, :resend_reason).merge(status: "in_progress"))
+      @consultation.update!(consultation_params.except(:resend_existing, :resend_reason).merge(status: "in_progress"), :apply_deadline_extension)
     end
 
     def send_neighbour_consultation_letter_copy
@@ -143,9 +146,8 @@ module PlanningApplications
       consultation_params[:resend_reason] if resend_existing?
     end
 
-    def redirect_address_validation_error(error)
-      flash[:alert] = error
-      redirect_to planning_application_consultation_neighbour_letters_path(@planning_application)
+    def redirect_after_rescue(error)
+      redirect_to planning_application_consultation_neighbour_letters_path(@planning_application), alert: error
     end
 
     def record_audit_for_letters_sent!
--- a/app/models/consultation.rb
+++ b/app/models/consultation.rb
@@ -18,6 +18,7 @@ class Consultation < ApplicationRecord
   attribute :consultee_message_subject, :string, default: -> { default_consultee_message_subject }
   attribute :consultee_message_body, :string, default: -> { default_consultee_message_body }
 
+  attribute :deadline_extension, :integer
   attribute :email_reason, :string, default: "send"
   attribute :resend_message, :string
   attribute :reconsult_message, :string
@@ -84,6 +85,9 @@ class Consultation < ApplicationRecord
     end
   end
 
+  before_save :apply_deadline_extension, if: -> { deadline_extension.present? }
+  validates :deadline_extension, presence: true, on: :apply_deadline_extension
+
   accepts_nested_attributes_for :consultees, :neighbours
 
   enum status: {
@@ -147,11 +151,13 @@ class Consultation < ApplicationRecord
   end
 
   def start_deadline(now = Time.zone.now)
-    update!(end_date: end_date_from(now), start_date: start_date || default_start_date(now))
+    update!(end_date: [end_date, end_date_from(now)].compact.max, start_date: start_date || default_start_date(now))
   end
 
   def extend_deadline(new_date)
-    update!(end_date: [end_date, new_date.end_of_day].max)
+    return unless new_date > end_date
+
+    update!(end_date: new_date.end_of_day)
   end
 
   def end_date_from(now = Time.zone.now)
@@ -428,5 +434,13 @@ class Consultation < ApplicationRecord
       "type" => "FeatureCollection",
       "features" => features
     }
+  end
+
+  def apply_deadline_extension
+    unless start_date?
+      start_deadline
+    end
+
+    extend_deadline(deadline_extension.days.from_now)
   end
 end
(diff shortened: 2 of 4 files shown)


app/controllers/planning_applications/neighbour_letters_controller.rb
```diff
@@ -113,8 +115,16 @@ def require_reason_when_resending
...
+      return unless consultation_params[:deadline].blank?
+      flash.now[:alert] = "Deadline is required"
+      render :index and return
+    end
+
     def update_consultation!
-      @consultation.update!(consultation_params.except(:resend_existing, :resend_reason).merge(status: "in_progress"))
+      @consultation.update!(consultation_params.except(:resend_existing, :resend_reason, :deadline).merge(status: "in_progress"))
+
+      @consultation.extend_deadline(consultation_params[:deadline].to_i.days.from_now.end_of_day)
```
pixeltrix, line 127: Can we not push this onto the consultation model and use a custom validation context like we do for consultees with a virtual attribute

benjamineskola: By virtual attribute do you mean a `deadline=` method on Consultation and letting it be invoked through the `update!`? If so I can give that a go, yes.

pixeltrix: I mean using the Active Record Attributes API to define custom attributes on the model like we do here:

https://github.com/unboxed/bops/blob/df4af49c386f7384b9439060b2e6a229b1c95176/app/models/consultation.rb#L18-L21

pixeltrix: This is then used with a custom validation context (e.g `on: :send_consultee_emails`):

https://github.com/unboxed/bops/blob/df4af49c386f7384b9439060b2e6a229b1c95176/app/models/consultation.rb#L42-L50

It means you can use the standard form error message handling

app/controllers/planning_applications/neighbour_letters_controller.rb
```diff
@@ -114,7 +118,7 @@ def require_reason_when_resending
     end
 
     def update_consultation!
-      @consultation.update!(consultation_params.except(:resend_existing, :resend_reason).merge(status: "in_progress"))
+      @consultation.send_neighbour_emails(consultation_params.except(:resend_existing, :resend_reason).merge(status: "in_progress"))
```
cdccollins, line 121: I find this kind of confusing - a method that calls update_! actually does another action that doesn't really seem to have anything to do with updating. But then the update is in that action? And send_neighbour_emails doesn't seem to send anything? 

It seems like the methods don't do what I'm expecting them to

benjamineskola: I named it to be consistent with the `send_consultee_emails` method but you're right that it doesn't necessarily make sense to be called that. (And anyway should be letters, not emails.)

I originally had the logic just in `update_consultation!` but that seemed to require giving it too much knowledge of a consultation — e.g., if the consultation hasn't started yet, it must be started or `extend_deadline` will fail.

I think if I move the `start_deadline` into `extend_deadline` then the `update!` and `extend_deadline!` can be called from `update_consultation!`.

app/views/planning_applications/neighbour_letters/_letter_template.html.erb
```diff
@@ -37,3 +36,13 @@
...
+<div class="govuk-hint govuk-!-margin-bottom-0 govuk-!-margin-top-8">Step 4</div>
+<div class="govuk-form-group">
+  <%= form.label :deadline, "Set response period", class: "govuk-label govuk-label--m" %>
+  <p class="govuk-hint">Enter the number of days that neighbours have to respond.</p>
+
+ <div class="govuk-input__wrapper">
+    <%= form.govuk_text_field :deadline, placeholder: "", class: "govuk-input", label: -> {}, value: 21, required: true %>
+    <div class="govuk-input__suffix" aria-hidden="true">days</div>
+  </div>
+</div>
```
benbaumann95, line 48: I'm going to be adding very similar html so I may turn this into a component so we can reuse easily

benjamineskola: I wonder if the govuk-frontend component library that Andrew linked to the other day might already have one: https://govuk-components.netlify.app/

app/controllers/planning_applications/neighbour_letters_controller.rb
```diff
@@ -55,6 +55,9 @@ def send_letters
       end
     rescue Neighbour::AddressValidationError => e
       redirect_address_validation_error(e)
+    rescue ActiveRecord::RecordInvalid => e
+      flash[:alert] = e
+      redirect_to planning_application_consultation_neighbour_letters_path(@planning_application)
```
pixeltrix, line 60: This is the same behaviour as the `redirect_address_validation_error` method - I don't think it's a good idea to set the exception in the flash as it'll just use a `to_s` on the instance which isn't very user friendly. Would be better to do something like this:

``` ruby
rescue Neighbour::AddressValidationError
  redirect_after_rescue(t(".errors.address_validation"))
rescue ActiveRecord::RecordInvalid
  redirect_after_rescue(t(".errors.record_invalid"))
end

def redirect_after_rescue(alert)
  redirect_to planning_application_consultation_neighbour_letters_path(@planning_application), alert: alert
end
```

app/models/consultation.rb
```diff
@@ -18,6 +18,7 @@ class AddNeighbourAddressesError < StandardError; end
   attribute :consultee_message_subject, :string, default: -> { default_consultee_message_subject }
   attribute :consultee_message_body, :string, default: -> { default_consultee_message_body }
 
+  attribute :deadline, :string
```
pixeltrix, line 21: Firstly - is this ever set? In the controller we call `update!` but the params are filtered to remove `:deadline`. Then in `extend_deadline` we deal with the param from the hash directly and not from the model.

I think it'd be better to call the attribute something like `deadline_extension` and define it as an integer. We can then leave it in the assigned params and then use a `before_save` callback to set the new end date if the value is present and greater than one.

app/models/consultation.rb
```diff
@@ -84,6 +85,8 @@ class AddNeighbourAddressesError < StandardError; end
     end
   end
 
+  validates :deadline, presence: true, on: :send_neighbour_emails
```
pixeltrix, line 88: Is this validation context ever used?

pixeltrix (changes requested):

app/models/consultation.rb
benjamineskola: No — it was but I've broken it in the process of trying to address Celia's feedback 🤦🏻

app/views/planning_applications/neighbour_letters/_letter_template.html.erb
pixeltrix: It's in the form builder:
https://govuk-form-builder.netlify.app/form-elements/text-input/#inputs-with-prefixes-and-suffixes

benbaumann95 (approved):

app/views/planning_applications/neighbour_letters/_letter_template.html.erb
```diff
@@ -1,18 +1,17 @@
...
-    }
-  %>
+  <%= form.govuk_collection_select :template,
+        ["Notification", "Renotification"],
+        :itself,
+        :itself,
+        label: {text: "Choose which letter to send", size: "m"},
+        data: {
+          action: "change->neighbours#toggleTemplates"
+        } %>
```
cdccollins, line 10: nit pick - is this now misaligned?

benjamineskola: Autocorrected by my editor — reverted.

PR #1497, opened 2024-01-23
Update site notice content for EIA applications

cdccollins:
### Description of change

Site notice content needs to be slightly different for applications which are subject to EIA

### Story Link

https://trello.com/c/rSVR6kcI/2478-able-to-produce-a-site-notice-for-eia-application


app/models/site_notice.rb
```diff
@@ -43,7 +43,9 @@ def document
   def preview_content
     start_deadline unless consultation_started?
 
-    I18n.t("site_notice_template",
+    site_notice_type = consultation.planning_application.environment_impact_assessment&.required? ? "eia" : "general"
```
Rhian, line 46: It's a very minor point, but wondering if "default" might be better than "general" here, especially if we have to add more types of site notice later?

app/models/site_notice.rb
```diff
@@ -72,4 +75,11 @@ def new_consultation_end_date
   def extend_consultation!
     consultation.update!(end_date: new_consultation_end_date)
   end
+
+  def eia_statement
+    eia = planning_application.environment_impact_assessment
+    if eia&.required? && eia&.address_and_fee_present?
+      "<p>View a hard copy of the Environment Statement, free of charge at #{eia.address}. You can buy a hard copy for £#{eia.fee} at the same address</p>"
```
Rhian, line 82: Probably not for this PR, but it looks as though the Notify emails/letters for EIA applications may need to use exactly this wording, so it may be useful in the longer term to think about attaching it to the EIA rather than to the site notice so that it can be shared.

Rhian (approved):
Looks good to me... my comments were maybe things we could consider for the future rather than now.

app/models/consultation.rb
```diff
@@ -151,7 +151,11 @@ def end_date_from(now = Time.zone.now)
     # Letters are printed at 5:30pm and dispatched the next working day (Monday to Friday)
     # Second class letters are delivered 2 days after they’re dispatched.
     # Royal Mail delivers from Monday to Saturday, excluding bank holidays.
-    default_start_date(now).end_of_day + 21.days
+    if planning_application.environment_impact_assessment&.required?
+      default_start_date(now).end_of_day + 30.days
```
benbaumann95, line 155: As discussed, need to be adding this to the `extend_consultation` method too

benbaumann95: If you rebase I think this should be covered

PR #1499, opened 2024-01-24
Add email_address to EIA

benbaumann95:
### Description of change

Add email_address to EIA

Screenshot 2024-01-24 at 12 13 56


Rhian (approved):
LGTM. Just one query over whether we should validate presence of email address before fee is accepted, but that can be added later if required.

PR #1502, opened 2024-01-24
Update press notice when application requires an EIA

benbaumann95:
### Description of change

- Update press notice mail when application is marked as requiring an EIA.
- Extend consultation period for 30 days instead of 21

### Story Link

https://trello.com/c/FJVFDUq3/2480-update-press-notice-for-applications-requiring-an-eia

### DIFF
--- a/app/models/consultation.rb
+++ b/app/models/consultation.rb
@@ -12,16 +12,23 @@ class Consultation < ApplicationRecord
     signatory_name signatory_job_title local_authority
   ].freeze
 
+  DEFAULT_PERIOD = 21.days
+  EIA_PERIOD = 30.days
+
   attribute :consultee_message_subject, :string, default: -> { default_consultee_message_subject }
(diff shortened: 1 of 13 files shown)


app/models/consultation.rb
```diff
@@ -151,7 +158,7 @@ def end_date_from(now = Time.zone.now)
     # Letters are printed at 5:30pm and dispatched the next working day (Monday to Friday)
     # Second class letters are delivered 2 days after they’re dispatched.
     # Royal Mail delivers from Monday to Saturday, excluding bank holidays.
-    default_start_date(now).end_of_day + 21.days
+    default_start_date(now).end_of_day + DEFAULT_PERIOD
```
benjamineskola, line 161: I wonder if some of these uses of `DEFAULT_PERIOD` should be using `period_days` instead? As it is, they seem to assume that there isn't an EIA.

Rhian (approved):
All looks good!

benjamineskola (approved):

PR #1504, opened 2024-01-25
Show neighbour markers

cdccollins:
### Description of change

Show neighbours as points on a map so you can know whether you will be selecting them or not

### Story Link

https://trello.com/c/22suHpWg/2462-display-markers-on-neighbour-map-to-indicate-currently-selected-neighbours

### Screenshots

Screenshot 2024-01-25 at 15 38 52

### Decisions 
When a planning application gets created, a background job calls the OS places radius API to capture neighbours within 50 metres of their property and stores the result in a new field on planning application.


engines/bops_api/app/jobs/bops_api/create_neighbour_boundary_geojson_job.rb
```diff
@@ -0,0 +1,51 @@
...
+            coordinates: [
+              coordinates.first,
+              coordinates.second
+            ],
+            type: "Point"
+          }
+        }
+      end
+    end
+  end
```
benbaumann95, line 50: Could just do 

```ruby
results.map do |result|
      lng, lat = result["DPA"]["LNG"], result["DPA"]["LAT"]
      {
        type: "Feature",
        geometry: {
          coordinates: [lng, lat],
          type: "Point"
        }
      }
    end
 ```

File: engines/bops_api/app/jobs/bops_api/create_neighbour_boundary_geojson_job.rb [hunk 2]
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+module BopsApi
+  class CreateNeighbourBoundaryGeojsonJob < ApplicationJob
+    queue_as :low_priority
+    discard_on ActiveJob::DeserializationError
+
+    def perform(planning_application)
+      response = ::Apis::OsPlaces::Query.new.find_addresses_by_radius(planning_application.lonlat.latitude, planning_application.lonlat.longitude).to_json
```
benbaumann95, line 9: Not sure we need top be converting this `to_json` so we can avoid doing the double parsing `JSON.parse(JSON.parse(response)["body"])["results"]`

engines/bops_api/app/jobs/bops_api/create_neighbour_boundary_geojson_job.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      # API request fails
+      return if JSON.parse(response).empty?
+
+      results = JSON.parse(JSON.parse(response)["body"])["results"]
```
benbaumann95, line 14: Think we ought to add a `rescue JSON::ParserError` block to the method to handle any bad responses. Unless we want this to blow up but maybe we can trigger appsignal in that case

benbaumann95 (approved):

PR #1508, opened 2024-01-26
Extract common consultation logic for press/site notices

benbaumann95:
### Description of change

- Extract common consultation logic for press/site notices
- Use "period_days" to determine consultation period for site notice as this may be EIA applicable which has 30 days

### DIFF
--- a/app/models/site_notice.rb
+++ b/app/models/site_notice.rb
@@ -2,6 +2,7 @@
 
 class SiteNotice < ApplicationRecord
   include DateValidateable
+  include Consultable
 
   belongs_to :planning_application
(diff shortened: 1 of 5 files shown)


app/models/site_notice.rb
```diff
@@ -68,14 +62,6 @@ def application_link(planning_application)
     end
   end
 
-  def new_consultation_end_date
-    [displayed_at && (displayed_at + Consultation::DEFAULT_PERIOD).end_of_day, consultation_end_date].compact.max
```
benbaumann95, line 72: This is now using `period_days` to factor in when an EIA may be involved

benjamineskola (approved):

PR #1511, opened 2024-01-29
Refactor audit templates

cdccollins:
### Description of change
Remove custom audit templates unless absolutely necessary and just use the generic one

### Story Link

https://trello.com/c/kErViOxQ/2490-refactor-audit-templates

### DIFF
--- a/app/components/audits/activity_component.html.erb
+++ b/app/components/audits/activity_component.html.erb
@@ -1,1 +1,1 @@
-<%= render "audits/types/#{audit_template}", locals: { item: audit } %>
+<%= render "audits/types/#{audit_template}", audit: audit %>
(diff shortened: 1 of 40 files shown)


app/models/audit.rb
```diff
@@ -116,4 +116,11 @@ def validation_request
       sequence: activity_information
     )
   end
+
+  def valid_json?(comment)
+    JSON.parse(comment)
+    true
+  rescue JSON::ParserError, TypeError
+    false
+  end
```
benjamineskola, line 125: This could probably go in the helper, since it doesn't seem to relate to any of the instance variables.

Also I wonder if making it return the parsed JSON would make sense, that way we wouldn't have to parse it again in the view.

benjamineskola (approved):
Suggestion isn't critical so feel free to change or merge as it :)

Rhian (approved):
LGTM

PR #1512, opened 2024-01-29
Upgrade Rails to 7.1

benjamineskola:
https://trello.com/c/clqyU1qD/2522-upgrade-to-rails-71


engines/bops_admin/bops_admin.gemspec
benjamineskola, line 1: I'm not sure if we want to increment `spec.version` here and in the other gemspec? If it was for a public gem we'd need to but I don't know if it will ever matter for this case.

engines/bops_admin/bops_admin.gemspec
```diff
@@ -12,6 +12,6 @@ Gem::Specification.new do |spec|
     Dir["{app,config,db,lib}/**/*"]
   end
 
-  spec.add_dependency "rails", ">= 7.0.8", "< 7.1"
+  spec.add_dependency "rails", ">= 7.0.8", "< 7.2"
```
pixeltrix, line 15: This should be bumped to `>= 7.1.3`

engines/bops_admin/bops_admin.gemspec
pixeltrix: I don't think there's any need to

pixeltrix (changes requested):
I think you need to add the new defaults initializer from running the update task - we'll need to check those and then migrate them.

pixeltrix:
Rails Guides information about the update task:
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#the-update-task

pixeltrix:
Also what needs to be done for upgrading Rails 7.0 to 7.1:
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-7-0-to-rails-7-1

app/models/constraint.rb
```diff
@@ -37,11 +37,7 @@ def normalize_type(type)
   end
 
   def type_code
-    if I18n.t("constraint_type_codes.#{type}").include?("translation missing")
-      type.titleize.capitalize
-    else
-      I18n.t("constraint_type_codes.#{type}")
-    end
+    I18n.t("constraint_type_codes.#{type}", default: type.titleize.capitalize)
```
pixeltrix, line 40: Nice spot 👍🏻

config/application.rb
```diff
@@ -52,6 +57,15 @@ class Application < Rails::Application
     # Don't log certain requests that spam the log files
     config.middleware.insert_before Rails::Rack::Logger, QuietLogger, paths: ["/healthcheck"]
 
+    # don't fail tests in this case
+    config.active_record.raise_on_assign_to_attr_readonly = false
+
+    # use rails 7.0 encryption method
+    config.active_record.encryption.hash_digest_class = OpenSSL::Digest::SHA256
```
pixeltrix, line 64: I assume there's a migration path but good enough for now 👍🏻

config/application.rb
```diff
@@ -52,6 +57,15 @@ class Application < Rails::Application
     # Don't log certain requests that spam the log files
     config.middleware.insert_before Rails::Rack::Logger, QuietLogger, paths: ["/healthcheck"]
 
+    # don't fail tests in this case
+    config.active_record.raise_on_assign_to_attr_readonly = false
```
pixeltrix, line 61: This feels slightly dirty but I assume this is something new in Rails 7.1?

config/application.rb
```diff
@@ -52,6 +57,15 @@ class Application < Rails::Application
...
+
+    # use rails 7.0 encryption method
+    config.active_record.encryption.hash_digest_class = OpenSSL::Digest::SHA256
+
+    # changing this breaks some creation_service tests
+    config.active_record.run_commit_callbacks_on_first_saved_instances_in_transaction = true
```
pixeltrix, line 67: Might try a deeper investigation on this to see why it's happening and see if we can fix it in Rails

pixeltrix (approved):

config/application.rb
benjamineskola: 7.0 fails silently if readonly attributes are written; 7.1 raises an error. Some of our tests do write to readonly attributes. We should probably not do that, though, but presumably we expect the attribute not to change otherwise the tests would have been failing.

PR #1517, opened 2024-01-30
Enforce local authority scope for v1 API

Rhian:
### Description of change

### Story Link

https://link-to-issue

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


app/controllers/api/v1/application_controller.rb
```diff
@@ -63,6 +64,10 @@ def current_api_user
 
       protected
 
+      def set_local_authority
+        @local_authority = LocalAuthority.find_by!(subdomain: request.subdomain)
+      end
```
pixeltrix, line 69: There's no need for this method and the `before_action` - the base `ApplicationController` has a `current_local_authority` method

pixeltrix (changes requested):

benjamineskola (approved):

config/application.rb
```diff
@@ -63,6 +63,7 @@ class Application < Rails::Application
     config.planning_history_enabled = ENV["PLANNING_HISTORY_ENABLED"] == true
     config.planx_file_api_key = ENV["PLANX_FILE_API_KEY"]
     config.planx_file_production_api_key = ENV["PLANX_FILE_PRODUCTION_API_KEY"]
+    config.request_keys = [:email, :subdomain]
```
pixeltrix, line 66: You don't need `:email` here as it's part of `config.authentication_keys`

pixeltrix (changes requested):

app/models/user.rb
```diff
@@ -43,14 +43,9 @@ def by_name
...
 
-  def self.find_for_authentication(tainted_conditions)
-    if tainted_conditions[:subdomains].present?
-      local_authority = LocalAuthority.find_by(subdomain: tainted_conditions[:subdomains].first)
-      tainted_conditions.delete(:subdomains)
-      find_first_by_auth_conditions(tainted_conditions.merge(local_authority_id: local_authority.id))
-    else
-      find_first_by_auth_conditions(tainted_conditions)
-    end
+  def self.find_for_database_authentication(tainted_conditions)
```
pixeltrix, line 46: The default implementation of `find_for_database_authentication` just calls `find_for_authentication` so there shouldn't be a need to change this.

app/models/user.rb
```diff
@@ -43,14 +43,9 @@ def by_name
...
+    local_authority_id = LocalAuthority.find_by(subdomain: tainted_conditions[:subdomain]).id
+    find_first_by_auth_conditions(email: tainted_conditions[:email], local_authority_id: local_authority_id)
```
pixeltrix, line 48: The `find_first_by_auth_conditions` method is a class method so you can call it through the `users` association on the `LocalAuthority` model, e.g.

``` ruby
local_authority = LocalAuthority.find_by!(subdomain: tainted_conditions[:subdomain])
local_authority.users.find_first_by_auth_conditions(email: tainted_conditions[:email])
```

Note the use of the `find_by!` method to raise an error if the local authority isn't found.

pixeltrix (changes requested):

pixeltrix (approved):

PR #1519, opened 2024-01-30
Set duration of consultee consultation period

benbaumann95:
### Description of change

- Avoid N+1 queries when updating neighbour letter status
- Allow officer to set duration for consultee consultation period. Extend the consultation deadline if this exceed the current end date
- Allow officer to manually set the consultation end date

### Story Link

https://trello.com/c/DcoYObRH/2479-able-to-change-the-duration-of-internal-external-consultee-consultation-period

### DIFF
--- a/app/controllers/planning_applications/neighbour_letters_controller.rb
+++ b/app/controllers/planning_applications/neighbour_letters_controller.rb
@@ -85,12 +85,14 @@ module PlanningApplications
     end
 
     def update_letter_statuses
-      return if @consultation.neighbour_letters.none?
+      letters = @consultation.neighbour_letters.includes(:neighbour).where.not(status: "received")
+
+      notify_key = @planning_application.local_authority.notify_api_key || Rails.configuration.default_notify_api_key
 
       # This is not ideal for now as will block the page loading, if it becomes a problem this would be
       # a good place for optimisation
-      @consultation.neighbour_letters.each do |letter|
-        letter.update_status unless letter.status == "received"
+      letters.each do |letter|
+        letter.update_status(notify_key)
       end
     end
 
--- a/app/models/consultation.rb
+++ b/app/models/consultation.rb
@@ -12,8 +12,10 @@ class Consultation < ApplicationRecord
     signatory_name signatory_job_title local_authority
   ].freeze
 
-  DEFAULT_PERIOD = 21.days
-  EIA_PERIOD = 30.days
+  DEFAULT_PERIOD = 21
+  DEFAULT_PERIOD_DAYS = DEFAULT_PERIOD.days
+  EIA_PERIOD = 30
+  EIA_PERIOD_DAYS = EIA_PERIOD.days
 
   attribute :consultee_message_subject, :string, default: -> { default_consultee_message_subject }
   attribute :consultee_message_body, :string, default: -> { default_consultee_message_body }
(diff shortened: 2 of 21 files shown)


app/controllers/planning_applications/neighbour_letters_controller.rb
```diff
@@ -82,12 +82,14 @@ def neighbour_params
...
+      letters = @consultation.neighbour_letters.includes(:neighbour).where.not(status: "received")
+
+      notify_key = @planning_application.local_authority.notify_api_key || Rails.configuration.default_notify_api_key
 
       # This is not ideal for now as will block the page loading, if it becomes a problem this would be
       # a good place for optimisation
-      @consultation.neighbour_letters.each do |letter|
-        letter.update_status unless letter.status == "received"
+      letters.each do |letter|
+        letter.update_status(notify_key)
```
benbaumann95, line 92: We need to move this to a background job, it's very slow when there are many letters. (Out of scope for this PR but i'll add a chore in trello)

benbaumann95: Preloading neighbour for now to remove n+1 and setting the notify key once here instead of calling the `notify_api_key` method (which makes a query) for every iteration when updating the status

app/models/consultation.rb
```diff
@@ -12,6 +12,7 @@ class AddNeighbourAddressesError < StandardError; end
     signatory_name signatory_job_title local_authority
   ].freeze
 
+  DEFAULT_PERIOD_DAYS = 21
```
benjamineskola, line 15: How does this differ from `DEFAULT_PERIOD`?

app/models/consultation.rb
```diff
@@ -151,7 +161,7 @@ def send_consultee_emails(attributes)
   end
 
   def start_deadline(now = Time.zone.now)
-    update!(end_date: [end_date, end_date_from(now)].compact.max, start_date: start_date || default_start_date(now))
+    update!(end_date: [end_date_from(now), consultee_response_required_by(now)].compact.max, start_date: start_date || default_start_date(now))
```
benjamineskola, line 164: This behaviour isn't quite the same as before. The previous version is the maximum of the current end date or the new end date; i.e., it won't be updated if the new end date is earlier than the existing one. But it looks like the new version will always change it?

Ah, but `consultee_response_required_by` will always be later than today. That feels very similar to `end_date_from` though. Can't we just use `end_date_from` in this case? (i.e., leave this line as it was?)

app/models/consultation.rb
```diff
@@ -328,12 +339,16 @@ def application_link
...
     else
-      Consultation::DEFAULT_PERIOD
+      Consultation::DEFAULT_PERIOD_DAYS
     end
   end
 
+  def default_consultee_response_period
+    (environment_impact_assessment.try(:required?) && EIA_PERIOD) || DEFAULT_PERIOD
+  end
+
```
benjamineskola, line 351: I can't see if this is still being used anywhere?

app/views/planning_applications/consultee/emails/index.html.erb
```diff
@@ -32,6 +32,21 @@
...
+      <div class="govuk-form-group" id="response-period">
+        <h2 class="govuk-heading-m">
+          <div class="govuk-hint govuk-!-margin-bottom-0">Step <%= @consultees.consulted? ? "4" : "3" %></div>
+          <%= form.label :consultee_response_period, "Set response period", class: "govuk-label govuk-label--m" %>
+        </h2>
+
+        <p class="govuk-hint">Enter the number of days that consultees have to respond.</p>
+
+        <div class="govuk-input__wrapper">
+          <%= form.govuk_text_field :consultee_response_period, placeholder: "", class: "govuk-input", label: -> {}, required: true %>
```
benjamineskola, line 44: Would it be worth adding a default here? (i.e., prefilling it with 21) — especially since it's required.

app/models/consultation.rb
benbaumann95: Yes, forgot to delete thanks

benjamineskola (approved):

PR #1525, opened 2024-01-31
Display updated status before ownership certificate validated

Rhian:
### Description of change

Currently when an applicant or agent responds to an ownership certificate change request, the planning application is automatically updated with the ownership certificate and appears valid in the task list (see screenshot). This PR adds a condition that if the change request is not approved but the ownership certificate is present, then the task list shows 'Updated'. Note that this only applies if `approved` is `nil` rather than the change request being rejected.

### Story Link

https://trello.com/c/9gLgifuS/2524-improvements-to-ownership-certificate-flow

### Screenshots

change_request
change_request_2

### DIFF
--- a/app/components/status_tags/ownership_certificate_component.rb
+++ b/app/components/status_tags/ownership_certificate_component.rb
@@ -13,6 +13,8 @@ module StatusTags
     def status
       if planning_application.valid_ownership_certificate.nil?
         :not_started
+      elsif planning_application.valid_ownership_certificate? && planning_application.ownership_certificate_awaiting_validation?
+        :updated
       elsif planning_application.valid_ownership_certificate?
         :valid
       else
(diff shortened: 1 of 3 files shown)


benjamineskola (approved):

app/components/status_tags/ownership_certificate_component.rb
```diff
@@ -13,6 +13,8 @@ def initialize(planning_application:)
     def status
       if planning_application.valid_ownership_certificate.nil?
         :not_started
+      elsif planning_application.valid_ownership_certificate? && planning_application.ownership_certificate_awaiting_validation?
+        :updated
       elsif planning_application.valid_ownership_certificate?
         :valid
```
benjamineskola, line 19:
```suggestion
      elsif planning_application.valid_ownership_certificate?
        if planning_application.ownership_certificate_awaiting_validation?
          :updated
        else
          :valid
        end
```

PR #1526, opened 2024-02-01
Refactor consultable

benjamineskola:
### Description of change

A module shouldn't have to know which classes include it. This way, any class can include it just by implementing `consultable_event_at` rather than having to modify `Consultable`.


Rhian (approved):
Looks good to me.

pixeltrix (approved):

benbaumann95 (approved):

PR #1531, opened 2024-02-02
Use date type for immunity details end date

benjamineskola:
https://trello.com/c/mdt6JBfR/2500-use-date-columns-in-database-where-appropriate


pixeltrix (approved):
Obviously we wouldn't just change the column type like this for a proper live service but I think it's fine at this time

PR #1535, opened 2024-02-05
Use separate standardrb action

benjamineskola:
This should flag up linter errors in the github ui directly and so be a bit nicer to deal with

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,10 +19,18 @@ jobs:
     with:
       brakeman: true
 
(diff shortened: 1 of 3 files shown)


benjamineskola:
This works, with limitations: it's necessary to use the `.standard.yml` config file, not `.rubocop.yml`, and it can't use any extensions (like `rubocop-rails`/`standard-rails`). Unfortunately both of those are things we rely on, and while we could migrate to the other config format it doesn't solve the extension problem. Parking this for now.

PR #1537, opened 2024-02-06
Be more specific on h1

EGiataganas:
### Description of change

Be more specific on h1

### DIFF
--- a/spec/system/planning_applications/review/assessment_summaries_spec.rb
+++ b/spec/system/planning_applications/review/assessment_summaries_spec.rb
@@ -361,11 +361,12 @@ RSpec.describe "Reviewing assessment summaries" do
           choose("Accept")
         end
 
-        click_button("Save and mark as complete")
+        find_button("Save and mark as complete").click
 
         visit "/planning_applications/#{planning_application.id}/review/tasks"
+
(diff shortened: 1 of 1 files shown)


spec/system/planning_applications/review/assessment_summaries_spec.rb
```diff
@@ -361,10 +361,10 @@
           choose("Accept")
         end
 
-        find_button("Save and mark as complete").click
+        click_button("Save and mark as complete")
 
         within("#planning-application-details") do
-          expect(page).to have_selector("h1", text: "Review and sign-off")
+          expect(page).to have_selector("h1.govuk-heading-l", text: "Review and sign-off")
```
pixeltrix, line 367: Are there multiple `<h1>` elements? It's an accessibility fail if there are multiple.

EGiataganas: No, there's only 1 `h1` in the page and it this section. Although, there's the same text `Review and sign-off` at the breadcrumb which I am not sure if it thinks that is this one. Tbh this h1 check here I think it makes sense, maybe is better to check `current_path`

PR #1538, opened 2024-02-06
Remove consultation summary for ldc applications

benbaumann95:
### Story Link

https://trello.com/c/We1OUjeM/2537-consultations-arent-created-for-ldcs-but-we-still-have-consultation-summary


cdccollins:
Do we want to remove it? I thought they needed a way to summarise if anyone had written in about the LDC.

cdccollins (commented):
Could we add a test that tries to click on that link? Somehow that kept on passing even though it was completely broken

app/controllers/planning_applications/assessment/assessment_details_controller.rb
```diff
@@ -10,7 +10,7 @@ class AssessmentDetailsController < AuthenticationController
       before_action :ensure_planning_application_is_validated
       before_action :set_assessment_detail, only: %i[show edit update]
       before_action :set_category, :set_rejected_assessment_detail, only: %i[new create edit update show]
-      before_action :set_consultation, if: :consultation_summary?
+      before_action :set_consultation, if: -> { consultation_summary? && has_consultation? }
```
benjamineskola, line 13: Would it be better to make this a method rather than using a lambda for it?

app/views/planning_applications/assessment/assessment_details/consultation_summary/_information.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if @consultation.consultees.present? %>
+<% if @consultation && @consultation.consultees.present? %>
```
benjamineskola, line 1: How about just using safe navigation?

```suggestion
<% if @consultation&.consultees&.present? %>
```

benbaumann95:
> Could we add a test that tries to click on that link? Somehow that kept on passing even though it was completely broken

Now that i've properly set up the ldc application without a consultation, the spec will fail without the `@consultation&.consultees&.present?` change

cdccollins (approved):

PR #1542, opened 2024-02-08
Use GOVUK's old accordion style

cdccollins:
### Description of change

Users are finding GOVUK's new accordion confusing to use, this reverts back to the old GOVUK accordion

### Story Link

https://trello.com/c/4NbPu95x/2546-change-accordion-back-to-old-govuk-version

### DIFF
--- a/app/assets/stylesheets/_main.scss
+++ b/app/assets/stylesheets/_main.scss
@@ -4,6 +4,7 @@ $govuk-page-width: 1100px;
 
 @import "govuk-frontend/dist/govuk/all";
 @import "accessible-autocomplete/dist/accessible-autocomplete.min";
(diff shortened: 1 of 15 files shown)


app/assets/stylesheets/_main.scss
```diff
@@ -328,7 +329,7 @@ $govuk-page-width: 1100px;
 }
 
 .single-section-accordion {
-  .govuk-accordion__controls {
+  .bops-govuk-accordion__controls {
```
pixeltrix, line 332: I would use `bops-accordion` and not `bops-govuk-accordion` so that search a second time doesn't pick up the substring match

pixeltrix (approved):

PR #1543, opened 2024-02-08
List all neighbour responses and respond

cdccollins:
### Description of change

Allow officer to respond to each category of neighbour objections

### Story Link

https://trello.com/c/LFDH2ckS/2508-list-individual-neighbour-issues-and-address-each-one

### Screenshots

screencapture-southwark-bops-localhost-3000-planning-applications-2260-assessment-assessment-details-520-edit-2024-02-08-10_20_20

### DIFF
--- a/app/helpers/assessment_detail_helper.rb
+++ b/app/helpers/assessment_detail_helper.rb
@@ -4,10 +4,20 @@ module AssessmentDetailHelper
   def neighbour_responses_summary_text(neighbour_responses_by_summary_tag)
     return "There are no neighbour responses" unless neighbour_responses_by_summary_tag.any?
 
-    summary_text = neighbour_responses_by_summary_tag.map do |tag, count|
-      "#{count} #{tag}"
+    summary_text = neighbour_responses_by_summary_tag.map.with_index do |(tag, count), i|
+      i.zero? ? t(:neighbour_responses_by_summary_tag, tag:, count:) : "#{count} #{tag}"
     end
 
-    "There is #{summary_text.join(", ")}."
+    "#{summary_text.join(", ")}."
+  end
+
+  def updated_neighbour_responses_summary_text(responses, detail)
+    count = if detail.created_at.nil?
+      responses.length
+    else
+      responses.where("neighbour_responses.created_at > ?", detail.updated_at).length
+    end
+
+    t(:neighbour_responses_by_summary_tag, tag: "new neighbour response", count:)
   end
 end
--- a/app/models/neighbour_response.rb
+++ b/app/models/neighbour_response.rb
@@ -15,6 +15,8 @@ class NeighbourResponse < ApplicationRecord
   enum(summary_tag: {supportive: "supportive", neutral: "neutral", objection: "objection"})
 
   scope :redacted, -> { where.not(redacted_response: "") }
+  scope :with_tags, -> { where.not("tags = '[]'") }
+  scope :without_tags, -> { where("tags = '[]'") }
 
   TAGS = %i[design new_use privacy disabled_access noise traffic other].freeze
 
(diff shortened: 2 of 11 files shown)


app/helpers/assessment_detail_helper.rb
```diff
@@ -4,10 +4,24 @@ module AssessmentDetailHelper
   def neighbour_responses_summary_text(neighbour_responses_by_summary_tag)
     return "There are no neighbour responses" unless neighbour_responses_by_summary_tag.any?
 
-    summary_text = neighbour_responses_by_summary_tag.map do |tag, count|
-      "#{count} #{tag}"
+    summary_text = neighbour_responses_by_summary_tag.map.with_index do |(tag, count), i|
+      if i == 0
+        "There #{(count != 1) ? "are" : "is"} #{count} #{tag}#{(count != 1) ? "s" : ""}"
```
pixeltrix, line 9: You can use the I18n count feature to simplify this logic, e.g.

``` yaml
en:
  neighbour_responses_by_summary_tag:
    zero: "There are 0 %{tag}s"
    one: "There is 1 %{tag}"
    other: "There are %{count} %{tag}s"
```

``` ruby
summary_text = neighbour_responses_by_summary_tag.map.with_index do |(tag, count), i|
  i.zero? ? t(:neighbour_responses_by_summary_tag, tag:, count:) : "#{count} #{tag}"
end
```

app/helpers/assessment_detail_helper.rb
```diff
@@ -4,10 +4,24 @@ module AssessmentDetailHelper
...
+  end
+
+  def updated_neighbour_responses_summary_text(responses, detail)
+    count = if detail.created_at.nil?
+      responses.count
+    else
+      responses.where("neighbour_responses.created_at > ?", detail.updated_at).count
+    end
+
+    "There #{(count != 1) ? "are" : "is"} #{count} new neighbour response#{"s" if count != 1}"
```
pixeltrix, line 25: You could move this to i18n like the example above

app/helpers/assessment_detail_helper.rb
```diff
@@ -4,10 +4,24 @@ module AssessmentDetailHelper
...
-    "There is #{summary_text.join(", ")}."
+    "#{summary_text.join(", ")}."
+  end
+
+  def updated_neighbour_responses_summary_text(responses, detail)
+    count = if detail.created_at.nil?
+      responses.count
+    else
+      responses.where("neighbour_responses.created_at > ?", detail.updated_at).count
+    end
```
pixeltrix, line 23: Is the `responses` relation loaded at this point? If so it might be better to do this in memory and not use `count`

app/models/neighbour_response.rb
```diff
@@ -15,6 +15,8 @@ class NeighbourResponse < ApplicationRecord
   enum(summary_tag: {supportive: "supportive", neutral: "neutral", objection: "objection"})
 
   scope :redacted, -> { where.not(redacted_response: "") }
+  scope :with_tags, -> { where.not("tags = '[]'") }
+  scope :without_tags, -> { where("tags = '[]'") }
```
pixeltrix, line 19: I think you can just use the hash syntax here - Active Record should do the typecasting to JSON for you, e.g.

``` ruby
scope :with_tags, -> { where.not(tags: []) }
scope :without_tags, -> { where(tags: []) }
```

pixeltrix (changes requested):

app/models/neighbour_response.rb
cdccollins: That doesn't work but using the JSON does 🤷‍♀️

pixeltrix: Well, this feels like a bug to me 😂 

```
irb(main):001> NeighbourResponse.where.not(tags: []).to_sql
=> "SELECT \"neighbour_responses\".* FROM \"neighbour_responses\" WHERE 1=1"
irb(main):002> NeighbourResponse.where(tags: []).to_sql
=> "SELECT \"neighbour_responses\".* FROM \"neighbour_responses\" WHERE 1=0"
```

pixeltrix (approved):

PR #1552, opened 2024-02-12
Convert boundary geojson to JSON when drawing red line in validation

cdccollins:
### Description of change

Show the red line that was drawn in validation in the check red line step. The boundary geojson was being stored as a string, rather than json meaning that it didn't show on the map

### Story Link

https://trello.com/c/LkCILieQ/2545-example-bug-report


app/controllers/planning_applications/validation/sitemaps_controller.rb
```diff
@@ -33,7 +33,7 @@ def update
         audit_action = @planning_application.boundary_geojson.blank? ? "created" : "updated"
 
         respond_to do |format|
-          if @planning_application.update(boundary_geojson: boundary_geojson_params[:boundary_geojson],
+          if @planning_application.update(boundary_geojson: JSON.parse(boundary_geojson_params[:boundary_geojson]),
```
EGiataganas, line 36: This looks great

Although, I think it would be great to include the `JSON.parse` in the setter's model, similar to `new_geojson=(value)` on `RedLineBoundaryChangeValidationRequest` like [this](https://github.com/unboxed/bops/blob/main/app/models/red_line_boundary_change_validation_request.rb#L27-L34)

EGiataganas (approved):

PR #1555, opened 2024-02-12
Local authority refactor

EGiataganas:
### Description of change

Local authority refactor

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -47,8 +47,8 @@ class LocalAuthority < ApplicationRecord
     super || set_notify_api_key
   end
 
-  def notify_letter_template
-    super || Rails.configuration.default_notify_template_id
+  def letter_template_id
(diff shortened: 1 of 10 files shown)


app/models/local_authority.rb
```diff
@@ -68,6 +70,27 @@ def set_active
...
+    self.letter_template_id ||= set_letter_template_id
+  end
+
+  def set_notify_api_key
+    if Bops.env.production?
+      notify_api_key.presence || Rails.configuration.default_notify_api_key
+    else
+      Rails.configuration.notify_letter_api_key
+    end
+  end
```
pixeltrix, line 84: I think your problem is there's two Notify API keys to consider - there's the one for email and then there's the one for sending letters. The sending letters one is only ever a test api key in staging whereas the email one is live in staging so that we can actually send emails.

pixeltrix (approved):

PR #1562, opened 2024-02-14
Account for neighbour responses with multiple tags

cdccollins:
### Description of change

When listing the neighbour responses and the comment boxes for each, hadn't accounted for responses which had multiple tags

### Story Link

https://trello.com/c/LFDH2ckS/2508-list-individual-neighbour-issues-and-address-each-one


Rhian (approved):
Looks good!

PR #1565, opened 2024-02-15
Small fixes to changing duration of neighbour consultation

cdccollins:
### Description of change

- Don't replace placeholder values in neighbour letter to avoid confusion over setting new dates for when consultation will end - use consultee pattern

### Story Link

https://trello.com/c/7gIhVzv1/2474-able-to-change-the-duration-of-neighbour-consultation-period

### DIFF
--- a/app/models/consultation.rb
+++ b/app/models/consultation.rb
@@ -240,13 +240,20 @@ class Consultation < ApplicationRecord
   end
 
   def neighbour_letter_header
-    I18n.t("neighbour_letter_header.#{planning_application.application_type.name}",
+    body = I18n.t("neighbour_letter_header.#{planning_application.application_type.name}")
+
+    defaults = {
       closing_date: end_date_from_now.to_fs,
-      default: "")
+      default: ""
+    }
+
+    replace_placeholders(body, defaults)
   end
 
   def neighbour_letter_body
-    I18n.t("neighbour_letter_template.#{planning_application.application_type.name}",
+    body = I18n.t("neighbour_letter_template.#{planning_application.application_type.name}")
+
+    defaults = {
       expiry_date: planning_application.expiry_date.to_date.to_fs,
       address: planning_application.full_address,
       council: local_authority.short_name,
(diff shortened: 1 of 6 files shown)


app/models/consultation.rb
```diff
@@ -259,15 +266,22 @@ def neighbour_letter_body
...
     "# #{neighbour_letter_header}\n\n#{neighbour_letter_body}"
   end
 
   def neighbour_letter_text
-    super.presence || neighbour_letter_content
+    if super.presence&.include?("{{")
+      neighbour_letter_content
+    else
+      super.presence
+    end
```
cdccollins, line 284: Obviously not very happy with this. Can someone suggest a better way?

app/views/planning_applications/neighbour_letters/_letter_template.html.erb
```diff
@@ -33,7 +33,7 @@
         </p>
       <% end %>
     </div>
-    <%= form.text_area :neighbour_letter_text, rows: 25, class: "govuk-body neighbour-letter-template" %>
+    <%= form.text_area :neighbour_letter_text, value: "#{I18n.t("neighbour_letter_template.#{planning_application.application_type.name}")}", rows: 25, class: "govuk-body neighbour-letter-template" %>
```
cdccollins, line 36: Also not convinced this is the best way 😬

app/models/consultation.rb
benjamineskola: I wonder if it's worth just unconditionally using `neighbour_letter_content`? Is there any downside to running the `replace_placeholders` part if there are no placeholders in it? I assume there will be _some_ performance impact, but probably tiny, relatively speaking.

app/views/planning_applications/neighbour_letters/_letter_template.html.erb
benjamineskola: I don't think the string interpolation is needed, if we're interpolating the entire string.

```suggestion
    <%= form.text_area :neighbour_letter_text, value: I18n.t("neighbour_letter_template.#{planning_application.application_type.name}"), rows: 25, class: "govuk-body neighbour-letter-template" %>
```

benjamineskola (approved):

app/models/consultation.rb
cdccollins: The problem with that is it then overwrites any edits that the officer makes in the text box

PR #1570, opened 2024-02-19
Migrate jsonb tags to string array

EGiataganas:
### Description of change

- Migrate jsonb tags to string array


app/models/document.rb
```diff
@@ -169,11 +169,12 @@ class NotArchiveableError < StandardError; end
   scope :for_publication, -> { active.publishable }
   scope :for_display, -> { active.referenced_in_decision_notice }
 
-  scope :with_tag, ->(tag) { where("tags @> ?", "\"#{tag}\"") }
+  scope :with_tag, ->(tag) { where("? = ANY(tags)", tag) }
+  scope :with_plan_tags, -> { where(tags: [PLAN_TAGS]) }
```
pixeltrix, line 173: This doesn't seem like it would match anything as it would require `tags` to contain all of `PLAN_TAGS` - do you mean an overlap? e.g. tags matches any of `PLAN_TAG` ?

If so you can use `overlaps` similar to `contains`, e.g.

```
scope :with_plan_tags, -> { where(arel_table[:tags].overlaps(PLAN_TAGS)) }
```

db/migrate/20240219143649_change_document_tags.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class ChangeDocumentTags < ActiveRecord::Migration[7.1]
+  def change
+    add_column :documents, :string_tags, :string, array: true, default: []
```
pixeltrix, line 5: Might want to look at adding a GIN index to speed up queries based on tags:
https://www.postgresql.org/docs/current/indexes-types.html#INDEXES-TYPES-GIN

app/views/planning_applications/assessment/ownership_certificates/_relevant_documents_accordion.html.erb
```diff
@@ -11,7 +11,7 @@
     <div class="govuk-accordion__section-content">
       <div class="govuk-!-padding-left-2 govuk-!-padding-right-2">
         <div class="scroll-docs">
-          <% @planning_application.documents.where("tags ?| array[:tags]", tags: Document::PLAN_TAGS).each do |document| %>
```
pixeltrix, line 14: Not even sure how this worked 🤯 

According to the [docs][1], the `?|` operator does the following

```
Do any of these array strings exist as top-level keys?
```

PostgreSQL must be treating the array of strings as keys 🙈 

[1]: https://www.postgresql.org/docs/9.5/functions-json.html#FUNCTIONS-JSONB-OP-TABLE

db/migrate/20240219143649_change_document_tags.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    Document.find_each do |document|
+      document.tags.each { |tag| document.string_tags << tag }
+      document.save!(validate: false)
+    end
+
+    remove_column :documents, :tags, :jsonb, default: []
+    rename_column :documents, :string_tags, :tags
```
pixeltrix, line 13: Don't ever do this on a live app as it will die horribly 😂

pixeltrix (changes requested):

db/migrate/20240219143649_change_document_tags.rb
benjamineskola: Is there a better way of doing it in that case? Or would we just have to suffer through it?

I assume neither adding nor removing columns is the issue, so is the problem renaming an existing one?

pixeltrix (approved):

PR #1571, opened 2024-02-19
Avoid hardcoded date breaking spec

benjamineskola:
### Description of change

Ensure application creation date is before review date. Because the application was created outside the `travel_to`, it was getting a creation date based on the current date as the test was run. The review, however, happened at a specified date in the future: but once that date was passed, the review would again be happening before the application had been created and the test would fail. This led to 1e241ba7 where the date was updated to be in the future once more.

This could obviously be fixed simply by putting the `travel_to` date far enough in the future that we'll never have to worry about it, or by deriving the date from the current date so that it's always in the future, but creating the application in the past relative to the review seems cleanest.

This also reverts #1574, because this seemed to be ineffective: the times were being set outside of the `travel_to` block (in the test setup, not the test body), so they passed only because the tests were run in business hours ago. Instead the start and end dates are being calculated based on business hours.

### Story Link

https://trello.com/c/P50I0yiE/2535-remove-hard-coded-dates-from-specs

### DIFF
--- a/spec/system/planning_applications/consulting/send_letters_to_neighbours_spec.rb
+++ b/spec/system/planning_applications/consulting/send_letters_to_neighbours_spec.rb
@@ -345,11 +345,8 @@ RSpec.describe "Send letters to neighbours", js: true do
       fill_in("Resend reason",
         with: "Previous letter mistakenly listed applicant's address as Buckingham Palace.")
 
-      travel_to(Time.zone.now.change(hour: 16)) do
-        expect_any_instance_of(Notifications::Client).to receive(:send_letter).with(template_id: anything,
-          personalisation: hash_including(message: match_regex(/# Application updated\nThis application has been updated. Reason: Previous letter mistakenly listed applicant's address as Buckingham Palace.\n\n# Submit your comments by #{(1.business_day.from_now + 21.days).to_date.to_fs}\n\nDear Resident/))).and_call_original
-      end
-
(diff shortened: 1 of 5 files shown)


spec/system/planning_applications/consulting/send_letters_to_neighbours_spec.rb
```diff
@@ -367,11 +364,8 @@
         fill_in("Resend reason",
           with: "Previous letter mistakenly listed applicant's address as Buckingham Palace.")
 
-        travel_to(Time.zone.now.change(hour: 16)) do
-          expect_any_instance_of(Notifications::Client).to receive(:send_letter).with(template_id: anything,
-            personalisation: hash_including(message: match_regex(/# Application updated\nThis application has been updated. Reason: Previous letter mistakenly listed applicant's address as Buckingham Palace.\n\n# Submit your comments by #{(1.business_day.from_now + 21.days).to_date.to_fs}\n\nDear Resident/))).and_call_original
-        end
-
+        expect_any_instance_of(Notifications::Client).to receive(:send_letter).with(template_id: anything,
```
pixeltrix, line 367: I generally avoid `expect_any_instance_of` as it sometimes has unintended side effects where there could be multiple instances of a class, e.g. an Active Record model but it's probably okay here since there's unlikely to be a second instance and stubbing `new` on `Notifications::Client` seems to be just going through the motions for no real benefit.

pixeltrix (approved):

PR #1573, opened 2024-02-19
Propose heads of terms

cdccollins:
### Description of change

Propose Heads of Terms to the applicant 

### Story Link

https://trello.com/c/H4XOk2us/2513-when-i-am-proposing-heads-of-terms-i-need-to-get-the-applicants-agreement-so-that-we-both-understand-the-recommendation


app/models/heads_of_term.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+class HeadsOfTerm < ApplicationRecord
```
benjamineskola, line 3: This is kind of a minor thing, but I wonder if it might end up being confusing that this is called `HeadsOfTerm` rather than `HeadsOfTerms`. But it might be more trouble than it's worth to try to make it work properly with Rails' pluralisations. I think we can add a custom inflection but I don't know how well it works if we want to make the plural and the singular the same. (And having a `HeadsOfTermsesController` would be even worse.)

app/components/status_tags/heads_of_terms_component.rb
```diff
@@ -0,0 +1,25 @@
...
+      @heads_of_term = heads_of_term
+    end
+
+    private
+
+    attr_reader :heads_of_term
+
+    def status
+      if heads_of_term.current_review.present?
+        if heads_of_term.validation_requests.where("validation_requests.created_at >= ?", heads_of_term.current_review.created_at).any? { |validation_request| !validation_request.approved.nil? } && !heads_of_term.current_review.complete?
```
benbaumann95, line 15: Should this be a method or scope in the model?
Seems quite complicated to have here.

I think we were moving away from having the status tag as a component as it's not reusable but maybe for a future clean up

app/models/heads_of_term.rb
```diff
@@ -0,0 +1,47 @@
...
+  belongs_to :planning_application
+  has_many :reviews, as: :owner, dependent: :destroy, class_name: "Review"
+  has_many :terms, extend: TermsExtension, dependent: :destroy
+  has_many :validation_requests, through: :terms
+
+  accepts_nested_attributes_for :terms, allow_destroy: true
+  accepts_nested_attributes_for :reviews
+
+  after_update :maybe_create_review
```
benbaumann95, line 12: `create_review_if_review_required`
Or something a bit more descriptive than maybe

app/models/heads_of_term.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def maybe_create_review
+    return if current_review.nil?
+    return unless current_review.status == "updated" && current_review.review_status == "to_be_reviewed"
+
+    create_review
+  end
+
+  def create_review
+    Review.create!(assessor: Current.user, owner_type: "HeadsOfTerm", owner_id: id, status: "complete")
```
benbaumann95, line 45: Is this better to create directly through the association, then might not need the owner_id etc?
i.e. would doing `reviews.create ....` work?

db/migrate/20240226165816_create_heads_of_terms.rb
```diff
@@ -0,0 +1,17 @@
...
+class CreateHeadsOfTerms < ActiveRecord::Migration[7.1]
+  def change
+    create_table :heads_of_terms do |t|
+      t.references :planning_application
+      t.timestamps
+    end
+
+    create_table :terms do |t|
+      t.string :title
+      t.text :text
```
benbaumann95, line 12: do we want to add a `null: false` since you're validating that presence is true in the model

app/models/heads_of_term.rb
benjamineskola: Perhaps `after_update :create_review, if: :should_create_review`.

Then the `should_create_review` method only needs the boolean part, rather than also calling `create_review` itself.

app/models/term.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class Term < ApplicationRecord
+  has_many :validation_requests, as: :owner, class_name: "ValidationRequest", dependent: :destroy
+  belongs_to :heads_of_term
+
+  validates :text, :title, presence: true
+
+  after_create :create_validation_request
+  before_update :maybe_create_validation_request
```
benbaumann95, line 10: similarly here

app/models/heads_of_term.rb
cdccollins: Yeah this just seems like a big old can of worms 😬

app/controllers/api/v1/heads_of_terms_validation_requests_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+          params:
+        ).call!
+
+        render json: {message: "Change request updated"}, status: :ok
+      end
+
+      private
+
+      def set_heads_of_terms_validation_request
+        @heads_of_terms_validation_request = @planning_application.heads_of_terms_validation_requests.find(params[:id])
```
benjamineskola, line 45:
```suggestion
        @heads_of_terms_validation_request = @planning_application.heads_of_terms_validation_requests.find(Integer(params[:id]))
```

app/controllers/planning_applications/assessment/heads_of_terms_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+
+      private
+
+      def set_heads_of_terms
+        @heads_of_term = @planning_application.heads_of_term
+      end
+
+      def set_heads_of_term
+        @term = @heads_of_term.terms.find(params[:heads_of_term_id])
```
benjamineskola, line 47:
```suggestion
        @term = @heads_of_term.terms.find(Integer(params[:heads_of_term_id]))
```

benjamineskola (approved):
Just a couple of suggestions, otherwise looks good to me.

benbaumann95 (approved):

PR #1574, opened 2024-02-19
Fix failing spec after 5:30 pm

EGiataganas:
### Description of change

Fix failing specs after 5:30 pm

### DIFF
--- a/spec/mailer/planning_application_mailer_spec.rb
+++ b/spec/mailer/planning_application_mailer_spec.rb
@@ -797,9 +797,11 @@ RSpec.describe PlanningApplicationMailer, type: :mailer do
       expect(mail_body).to include(
         "This is a copy of your neighbour consultation letter."
       )
-      expect(mail_body).to include(
-        "Submit your comments by #{(1.business_day.from_now + 21.days).to_date.to_fs(:day_month_year)}"
-      )
+      travel_to(Time.zone.now.change(hour: 16)) do
+        expect(mail_body).to include(
+          "Submit your comments by #{(1.business_day.from_now + 21.days).to_date.to_fs(:day_month_year)}"
+        )
+      end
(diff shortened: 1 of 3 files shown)


spec/mailer/planning_application_mailer_spec.rb
```diff
@@ -797,9 +797,11 @@
...
         "This is a copy of your neighbour consultation letter."
       )
-      expect(mail_body).to include(
-        "Submit your comments by #{(1.business_day.from_now + 21.days).to_date.to_fs(:day_month_year)}"
-      )
+      travel_to(Time.zone.local(Time.zone.now.year, Time.zone.now.month, Time.zone.now.day, 16, 0, 0)) do
+        expect(mail_body).to include(
+          "Submit your comments by #{(1.business_day.from_now + 21.days).to_date.to_fs(:day_month_year)}"
+        )
+      end
```
pixeltrix, line 804: Why not just `travel_to` with a specific date/time rather than using the current date?

If you want to keep using the current date, you can use `change` to simplify it, e.g.

``` ruby
travel_to(Time.zone.now.change(hour: 16)) do
  # do stuff
end
```

benjamineskola:
It would be good to understand why these specs are failing. I can see a few possibilities:

- using `DateTime` columns when the data should be only a `Date`
- date arithmetic is converting a `Date` to a `DateTime`
- using the "next business day" calculation unnecessarily — when I converted some columns to `Date`s I wasn't sure about this, I think in some places it'll round it to the next day when now we might no longer need it to do so

If we can address the underlying cause we might not need to use `travel_to` at all. I'd be cautious of fixing tests simply by setting the time within working hours — yes, most people will only be working within working hours, but what if someone works late for some reason?

pixeltrix (approved):

PR #1576, opened 2024-02-20
Add pre-commencement conditions

cdccollins:
### Description of change

Allow officer to submit pre-commencement conditions to the applicant 

### Story Link

https://trello.com/c/EPTysRB6/2512-when-i-need-to-propose-a-pre-commencement-condition-i-need-to-give-the-applicant-at-least-10-days-notice-to-agree-or-object-so-i

### DIFF
--- a/app/controllers/planning_applications/assessment/conditions_controller.rb
+++ b/app/controllers/planning_applications/assessment/conditions_controller.rb
@@ -17,6 +17,10 @@ module PlanningApplications
         end
       end
 
+      def new
+        @condition_set.conditions.new
+      end
+
       def edit
         respond_to do |format|
           format.html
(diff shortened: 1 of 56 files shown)


app/controllers/planning_applications/assessment/conditions_controller.rb
```diff
@@ -37,21 +43,36 @@ def update
...
+        end
       end
 
       def set_conditions
         @conditions = @condition_set.conditions
       end
 
+      def set_pre_commencement
+        @pre_commencement = params[:pre_commencement] == "true"
+      end
```
benjamineskola, line 59: I think I would just make this a method rather than setting an instance variable. (And then no need to add a callback, etc.)

```suggestion
      def pre_commencement?
        params[:pre_commencement] == "true"
      end
```

latest.dump
benjamineskola, line 1: This probably ought to be removed.

benjamineskola (approved):

app/controllers/planning_applications/assessment/conditions_controller.rb
cdccollins: Not sure how that would work because I use the variable that gets set in the views?

PR #1578, opened 2024-02-21
Store/restore current user for Active Jobs

pixeltrix:


benjamineskola:
I wonder if it would be simpler just to pass in the user as an explicit parameter to any job that needs it, then instead of using `Current.user` we can use the parameter.

EGiataganas (approved):

PR #1583, opened 2024-02-22
Remove :only option from :verify_request callback

pixeltrix:
Rails 7.1 now raises an exception for actions that don't exist and `set_disk_blob` isn't defined yet and isn't an action.

Note this only affects development and test environments since that's the only place where the disk controller is used.

### DIFF
--- a/app/controllers/concerns/active_storage/set_disk_blob.rb
+++ b/app/controllers/concerns/active_storage/set_disk_blob.rb
@@ -5,7 +5,7 @@ module ActiveStorage
     extend ActiveSupport::Concern
 
     included do
-      before_action :verify_request, only: %i[set_disk_blob]
+      before_action :verify_request
       before_action :set_disk_blob, if: :not_from_bops?
     end
 


benjamineskola:
Is this happening because the callback is defined in the `included` block?

Also, couldn't we simply call `verify_request` at the beginning of `set_disk_blob`, since it's just a normal method AIUI? Do we gain anything by doing it as a callback rather than a method call? (especially when we can't use all of the features of callbacks like the `only:` argument.)

benjamineskola (approved):

EGiataganas (approved):

pixeltrix:
> Is this happening because the callback is defined in the `included` block?

No, the `:only` option never worked as `:set_disk_blob` isn't an action and now Rails 7.1 is raising an error due to the new default for the config option [`raise_on_missing_callback_actions`](https://guides.rubyonrails.org/configuring.html#config-action-controller-raise-on-missing-callback-actions)

> Also, couldn't we simply call `verify_request` at the beginning of `set_disk_blob`, since it's just a normal method AIUI? Do we gain anything by doing it as a callback rather than a method call? (especially when we can't use all of the features of callbacks like the `only:` argument.)

Unsure whether the return value of `verify_request` can be used as a `return unless verify_request` short circuit since the method is written to be part of a callback chain - if it renders or redirects and still returns a truthy value it could cause a `AbstractController::DoubleRenderError` exception to occur.

PR #1586, opened 2024-02-26
Add show page for validating ownership certificates

cdccollins:
### Description of change

Right now the journey is confusing as even if the task has been completed, the form still shows. This adds a show page to make it clearer for users

### DIFF
--- a/app/components/task_list_items/validating/ownership_certificate_component.rb
+++ b/app/components/task_list_items/validating/ownership_certificate_component.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module TaskListItems
+  module Validating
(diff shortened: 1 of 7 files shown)


pixeltrix (approved):

EGiataganas (approved):

app/components/task_list_items/validating/ownership_certificate_component.rb
```diff
@@ -0,0 +1,34 @@
...
+        @planning_application = planning_application
+        @ownership_certificate = @planning_application.ownership_certificate
+      end
+
+      private
+
+      attr_reader :planning_application, :ownership_certificate
+
+      def link_text
+        "Check ownership certificate"
```
EGiataganas, line 16: How about moving it in locales?

PR #1589, opened 2024-02-27
Add considerations

Rhian:
### Description of change

This changes the way officers add considerations when assessing against local policy. In the previous iteration, they could choose from Design, Impact and Other, which were checklist items on a local policy page and there were no separate views for each local policy area (consideration). There was also no way to delete considerations that had been added to the policy, although they could be updated.

Now, they choose from an autocomplete dropdown list and can view a table of considerations from which records can be deleted or edited.

### Story Link

https://trello.com/c/Zdo7YvWy/2509-able-to-add-as-many-considerations-as-necessary

### Screenshots

consideration_form

considerations_table

### Decisions 

Initially I had aimed to use the existing structure but that caused complications when it came to removing local policy areas from the local policy. After discussion, it was decided to use existing Rails functionality rather than extend the JavaScript on the old page.

### Known issues

Laura is doing a content review with POs so some of the language, style of buttons etc may change to bring a common style to other elements.


app/controllers/planning_applications/assessment/local_policy_areas_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+
+module PlanningApplications
+  module Assessment
+    class LocalPolicyAreasController < AuthenticationController
+      before_action :set_planning_application
+      before_action :set_local_policy, only: %i[show edit update new]
+      before_action :set_local_policy_area, only: %i[edit show]
+
+      def show
+        set_local_policy_area
```
pixeltrix, line 11: Why do we need this here if we're using the `before_action`?

app/controllers/planning_applications/assessment/local_policy_areas_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+      end
+
+      def new
+        @local_policy_area = LocalPolicyArea.new
```
pixeltrix, line 15: Shouldn't this belong to a local policy? e.g. 

``` ruby
@local_policy_area = @local_policy.local_policy_areas.new
```

app/controllers/planning_applications/assessment/local_policy_areas_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+      end
+
+      def edit
+        set_local_policy_area
```
pixeltrix, line 19: Same as above - why do we need this here if we're using the before_action?

app/controllers/planning_applications/assessment/local_policy_areas_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+      end
+
+      def destroy
+        @local_policy_area = LocalPolicyArea.find(params[:id])
```
pixeltrix, line 23: We should always scope our finds through the association otherwise you can walk through id values to find other records. Not particularly sensitive information here but it's good practice to follow.

app/controllers/planning_applications/assessment/local_policy_areas_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+            notice: I18n.t("local_policies.successfully_updated")
+        else
+          render :edit
+        end
+      end
+
+      private
+
+      def set_local_policy
+        @local_policy = @planning_application.local_policy.presence || LocalPolicy.create!(planning_application_id: @planning_application.id)
```
pixeltrix, line 57: Again, scope operations to the parent association, e.g.

``` ruby
@local_policy = @planning_application.local_policy || @planning_application.create_local_policy!
```

app/controllers/planning_applications/assessment/local_policy_areas_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+
+        if @local_policy_area.save
+          if @planning_application.local_policy.present?
+            redirect_to edit_planning_application_assessment_local_policy_path(@planning_application, @planning_application.local_policy)
+          else
+            redirect_to new_planning_application_assessment_local_policy_path(@planning_application)
+          end
+        else
+          render :new
+        end
```
pixeltrix, line 41: We should always use `respond_to` blocks in actions so that we define the formats that we accept. If we don't then someone can edit the url to include a format like `.json` and Rails will try to render it.

pixeltrix (changes requested):
Need to scope everything via the associations in the controller

app/controllers/planning_applications/assessment/local_policy_areas_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      def show
+      end
+
+      def new
+        @local_policy_area = @local_policy.local_policy_areas.new
+      end
+
+      def edit
+      end
```
pixeltrix, line 18: Sorry, wasn't clear - we should be using `respond_to` blocks in all actions

Rhian: Apologies - updated.

config/routes.rb
```diff
@@ -104,7 +104,11 @@
           end
         end
 
-        resources :local_policies, except: %i[destroy index]
+        resources :local_policies, except: %i[destroy index] do
+        end
+
+        resources :local_policy_areas, except: %i[index] do
+        end
```
pixeltrix, line 111: Don't need the empty blocks here

pixeltrix (changes requested):

app/controllers/planning_applications/assessment/local_policy_areas_controller.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+module PlanningApplications
+  module Assessment
+    class LocalPolicyAreasController < AuthenticationController
+      before_action :set_planning_application
+      before_action :set_local_policy, only: %i[create destroy edit new show update]
```
pixeltrix, line 7: I think we can remove the `:only` option - that's all the actions isn't it?

app/controllers/planning_applications/assessment/local_policy_areas_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      end
+
+      def edit
+        respond_to do |format|
+          format.html
+        end
+      end
+
+      def destroy
+        @local_policy.local_policy_areas.destroy(params[:id])
```
pixeltrix, line 30: The @local_policy_area is set using the before_action so you can just do:

``` ruby
@local_policy_area.destroy
```

app/controllers/planning_applications/assessment/local_policy_areas_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+          respond_to do |format|
+            format.html { render :new }
+          end
+        end
+      end
+
+      def update
+        if @local_policy_area.update(local_policy_params)
+          redirect_to edit_planning_application_assessment_local_policy_path(@planning_application, @planning_application.local_policy),
+            notice: I18n.t("local_policies.successfully_updated")
```
pixeltrix, line 50: I think we're using just `t` now rather than `I18n.t` - also other controllers are using relative keys, e.g. `t(".success")`

app/controllers/planning_applications/assessment/local_policy_areas_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      def set_local_policy
+        @local_policy = @planning_application.local_policy || @planning_application.create_local_policy!
+      end
+
+      def set_local_policy_area
+        @local_policy_area = @local_policy.local_policy_areas.find(params[:id])
+      end
+
+      def local_policy_params
+        params.require(:local_policy_area).permit([:area, :id, :policies, :policy, :assessment, :guidance, :local_policy_id])
```
pixeltrix, line 69: If we always scope our actions through the association then we don't need the `:local_policy_id` in the params which would allow someone to detach a record from one policy and attach it to another.

Same goes for `:id` as well - shouldn't be permitted here.

pixeltrix (changes requested):

app/controllers/planning_applications/assessment/local_policy_areas_controller.rb
Rhian: Can't believe that didn't occur to me, haha 🤦‍♀️

cdccollins (approved):

pixeltrix (approved):

PR #1593, opened 2024-02-29
Set up global admin infrastructure

benbaumann95:
### Description of change

- Add BopsConfig engine
- Scope this engine to when "config" is in the subdomain
- Wrap main application routes with local authority subdomain constraint
- Allow both LA and config subdomain to access shared devise resources
- Global admin users do not have an associated local authority
- Reuse admin user management for managing global admin users

### Story Link

https://trello.com/c/5RJfvTzj/2446-create-global-admin-infrastructure

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -71,9 +71,7 @@ class ApplicationController < ActionController::Base
   end
 
   def find_current_local_authority_from_subdomain
-    return if @current_local_authority ||= LocalAuthority.find_by(subdomain: request.subdomains.first)
-
-    render plain: "No Local Authority Found", status: :not_found
+    @current_local_authority ||= LocalAuthority.find_by(subdomain: request.subdomains.first)
   end
 
   def prevent_caching
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -2,270 +2,272 @@
 
 # require "rswag-api"
 
-require "sidekiq/web"
-
 Rails.application.routes.draw do
-  root to: "planning_applications#index"
+  constraints Constraints::DeviseSubdomain do
+    devise_for :users, controllers: {
+      sessions: "users/sessions",
+      confirmations: "confirmations"
+    }
 
-  mount BopsApi::Engine, at: "/api"
-  get "/api-docs(/index)", to: redirect("/api/docs")
-
-  mount BopsAdmin::Engine, at: "/admin"
-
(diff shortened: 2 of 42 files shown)


app/controllers/application_controller.rb
```diff
@@ -71,9 +71,12 @@ def set_back_path
   end
 
   def find_current_local_authority_from_subdomain
-    return if @current_local_authority ||= LocalAuthority.find_by(subdomain: request.subdomains.first)
+    subdomain = request.subdomains.first
 
-    render plain: "No Local Authority Found", status: :not_found
+    return if subdomain == "config"
```
pixeltrix, line 76: There's a way to avoid this palaver in the engine's routes - have a catchall route at the end and return a `x-casade` header with a 404, e.g. 

``` ruby
get "/*path", to: proc { [404, { "x-cascade" => "pass" }, %w[]] }
```

This will trigger the Rails router to stop looking for routes.

Another approach is to wrap the main application routes in a constraint that checks for valid LA subdomains (which is probably a better approach long term).

benbaumann95: Gone with the wrapping main application routes in a constraint

config/routes.rb
```diff
@@ -2,266 +2,268 @@
 
 # require "rswag-api"
 
-require "sidekiq/web"
```
cdccollins, line 5: Did you mean to get rid of this?

cdccollins (approved):

config/routes.rb
```diff
@@ -2,266 +2,268 @@
...
     end
+
+    get :healthcheck, to: proc { [200, {}, %w[OK]] }
   end
 
-  get :healthcheck, to: proc { [200, {}, %w[OK]] }
+  constraints Constraints::ConfigSubdomain do
+    mount BopsConfig::Engine => "/"
+  end
 end
```
EGiataganas, line 269: could we maybe extract the main BOPS routes to a separate file as is getting too long?

config/routes.rb
```diff
@@ -2,266 +2,268 @@
 
 # require "rswag-api"
```
EGiataganas, line 3: This is unrelated with your PR but I think it would be good to delete it as is commended thus, not needed

config/routes.rb
benbaumann95: Yeah, moving it to the engine routes instead

PR #1596, opened 2024-03-04
Allow selecting consultee and associating with constraint

benjamineskola:
### Description of change

This associates a Consultee with a PlanningApplicationConstraint and adds an extra step to the consultee workflow (see [figma](https://www.figma.com/file/AXJ7hvBaXVyJnh74Grwbml/BoPS-Prototype?type=design&node-id=8804-10481&mode=design&t=iaomaW64tY9N2fVY-4)).

### Story Link

https://trello.com/c/twXRQN1M/2194-see-constraints-when-selecting-consultees

### Screenshots

Screenshot 2024-03-06 at 12 21 15

### Known issues

- The transition to the new workflow is unfinished here. Currently the 'add consultee' form is still on the original page, which is now the second step. However it seemed easiest to break it up here and make the change in two stages.
- There are no designs in Figma for the form for actually selecting a consultee for a constraint, so there might be further refinement needed to that page.

### DIFF
--- a/app/components/task_list_items/select_consultees_component.rb
+++ b/app/components/task_list_items/select_consultees_component.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module TaskListItems
+  class SelectConsulteesComponent < TaskListItems::BaseComponent
+    def initialize(planning_application:)
+      @planning_application = planning_application
+    end
+
+    private
+
+    attr_reader :planning_application
+
+    delegate :consultation, to: :planning_application
+    delegate :consultee_emails_status, to: :consultation
+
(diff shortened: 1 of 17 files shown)


cdccollins (approved):
I know you're breaking this up into stages, so are you leaving a system spec until it works as per the designs?

benjamineskola:
> I know you're breaking this up into stages, so are you leaving a system spec until it works as per the designs?

Yeah, my thinking is that a big part of the system spec is going to involve having the add-consultee form on the same page. Trying to write a passing test with the current structure probably involves writing some code that will immediately get reverted.

PR #1597, opened 2024-03-04
Move healthcheck outside domain constraints

pixeltrix:
The AWS ECS healthcheck uses an internal hostname which fails to route and causes the cycling of tasks due to healthcheck failures.


cdccollins (approved):

pixeltrix:
Thanks 👍🏻

PR #1604, opened 2024-03-06
Show errors correctly on pre-commencement conditions

cdccollins:
### Description of change

Errors weren't showing correctly when not filling in pre-commencement conditions correctly


benjamineskola (approved):

spec/system/planning_applications/assessing/adding_pre_commencement_conditions_spec.rb
```diff
@@ -242,8 +245,13 @@
 
       click_button "Submit"
 
-      expect(page).to have_content "Enter the text of this condition"
-      expect(page).to have_content "Enter the title of this condition"
+      within(:css, "#other-conditions .condition:nth-of-type(1)") do
```
pixeltrix, line 248: No need to bother changing it but the default selector format is :css so the first argument is redundant

pixeltrix (approved):

PR #1607, opened 2024-03-06
Update planning applications endpoint

benbaumann95:
### Description of change

This introduces these changes:

**API V2**
- Authenticated index endpoint that returns planning applications (add a query service to allow page, maxresults and ids as query parameters)
- Unauthenticated index (determined) endpoint that only returns determined planning applications
- Authenticated show endpoint that takes a planning application id as a path parameter and returns a single planning application

**API V1**
- Update show endpoint to add authentication as this wasn't scope to determined planning applications
- Remove unused index endpoint (this was returning a 403 anyway due to performance reasons because of no pagination)

- Add bops-applicants API User to the relevant databases

**TODO**

- The queries can probably be made more efficient to avoid N+1 and certain indexes added to the database but that can be done as a separate PR when improving performance 

### Story Link

https://trello.com/c/PicL1gqK/2619-clarify-api-results-unauthenticated-users-should-see

### Screenshots

**Swagger**
Screenshot 2024-03-06 at 16 34 16

**Get planning applications query parameters**
Screenshot 2024-03-06 at 16 34 32


engines/bops_api/app/controllers/bops_api/v2/planning_applications_controller.rb
```diff
@@ -3,7 +3,36 @@
...
+          planning_applications_scope.determined.by_determined_at_desc
+        ).call
+
+        respond_to do |format|
+          format.json
+        end
+      end
+
+      def show
+        @planning_application = planning_applications_scope.find(params[:id])
```
benjamineskola, line 30: Should we be validating that the param is numeric? I'm not sure whether it matters in all cases or if I'm just being overcautious:
```suggestion
        @planning_application = planning_applications_scope.find(Integer(params[:id]))
```

cdccollins (approved):

benjamineskola (approved):

PR #1608, opened 2024-03-06
Feature/request time extension

Rhian:
### Description of change

This story allows officers and managers to ask the applicant to extend the expiry date of the application. By implication this also extends the target date.

### Story Link

https://trello.com/c/nNApgOyZ/2515-as-a-planning-officer-i-need-to-be-able-to-request-an-extension-of-time-so-that-that-i-have-time-to-deal-with-new-more-informati

### Screenshots

banner
extension_2
extension_3

### Decisions 

Although from a technical point of view, time extension validation requests are part of the ValidationRequest class, they are not part of the validation process as they can be sent at any point during the process and do not need to be resolved in order for the application to be validated. Certain decisions have been taken such as:

- excluding them from the display list of validation requests on the validations index page as the messaging would be wrong
-  adding a standalone response banner


app/models/planning_application.rb
```diff
@@ -704,7 +705,7 @@ def open_post_validation_requests
   end
 
   def closed_pre_validation_requests
-    validation_requests.closed - validation_requests.closed.post_validation
+    validation_requests.where.not(type: "time_extension_validation_request").closed - validation_requests.closed.post_validation
```
benjamineskola, line 708: We're using this condition in a few places (the controller too) — should we make it a scope or something perhaps?

app/presenters/concerns/validation_tasks/items_counter_presenter.rb
```diff
@@ -21,8 +21,12 @@ def items_count
 
     attr_reader :validation_requests
 
+    def validation_requests_not_time_extension
+      validation_requests.select { |request| request.type != "TimeExtensionValidationRequest" }
```
benjamineskola, line 25:
```suggestion
      validation_requests.reject { |request| request.time_extension_request? }
```

app/models/planning_application.rb
```diff
@@ -735,6 +736,30 @@ def replacement_document_validation_request
...
+  def has_only_time_extension_requests?
+    validation_requests.length == time_extension_validation_requests.length
+  end
+
+  def open_time_extension_requests
+    time_extension_validation_requests.where(state: "open")
+  end
+
+  def closed_time_extension_request
+    time_extension_validation_requests.where(state: "closed").order(:created_at).last
```
benjamineskola, line 756: Maybe we can use `closed_time_extension_requests` in the definition of `closed_time_extension_request` which also makes the relationship between them more clear (I initially missed why there were two with such similar names at first glance.)

```suggestion
    closed_time_extension_requests.order(:created_at).last
```

app/services/validation_request_update_service.rb
```diff
@@ -16,6 +16,10 @@ def call!
       validation_request.update!(validation_request_params)
       validation_request.close!
       validation_request.create_api_audit!
+      if validation_request.time_extension_request?
+        validation_request.update_planning_application!(params)
+        Rails.logger.debug "true" if validation_request_params[:approved]
```
benjamineskola, line 21: Can we remove the debug line?

```suggestion
```

app/helpers/description_change_validation_request_helper.rb
```diff
@@ -8,4 +8,9 @@ def change_rejected?(description_change_validation_request)
   def sequence_description(description_change_validation_request)
     "description##{description_change_validation_request.sequence}"
   end
+
+  def format_datetime(datetime_str)
+    datetime = DateTime.parse(datetime_str)
+    datetime.strftime("%-d %B %Y")
```
benjamineskola, line 14: It might be best to use `to_fs` here, as that makes it easier to keep the formats consistent (it uses the ones defined in `config/initializers/date_formatter.rb`).

```suggestion
    datetime.to_date.to_fs
```

app/models/planning_application.rb
benjamineskola: Although actually I can't tell — do we use `closed_time_extension_request` (singular) anywhere? I can only find the plural version.

app/views/planning_applications/validation/validation_requests/_activity_log.html.erb
```diff
@@ -0,0 +1,54 @@
...
+            <% else %>
+              <tr class="govuk-table__row">
+                <td scope="row" class="govuk-table__cell"><%= request.approved? ? "Applicant accepts" : "Request rejected by applicant" %></td>
+                <td class="govuk-table__cell"><%= request.approved? ? "Approved" : "Rejected" %></td>
+                <td class="govuk-table__cell">Applicant</td>
+                <td class="govuk-table__cell"><%= request.created_at.to_fs(:day_month_year_slashes) %></td>
+              </tr>
+            <% end %>
+            <tr class="govuk-table__row">
+              <td scope="row" class="govuk-table__cell">Request an extension until <%= request.proposed_expiry_date.strftime("%-d %B %Y") %></td>
```
benjamineskola, line 37: Same here: we should use `to_fs` where possible for consistency.

```suggestion
              <td scope="row" class="govuk-table__cell">Request an extension until <%= request.proposed_expiry_date.to_date.to_fs %></td>
```

db/migrate/20240303120759_add_proposed_expiry_date_to_validation_requests.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddProposedExpiryDateToValidationRequests < ActiveRecord::Migration[7.1]
+  def change
+    add_column :validation_requests, :proposed_expiry_date, :datetime
```
benjamineskola, line 5: Just checking: will we be using the time part of this, or will it only ever be a date? If the latter it might be better to use a `date` column, because we've had cases in the past where we need to round a datetime to the beginning or the end of the day every time it's referenced.

app/views/shared/_dates_and_assignment_header.html.erb
```diff
@@ -64,6 +64,12 @@
           <%= @planning_application.days_status_tag %>
         </span>
       <% end %>
+
+      <% if @planning_application.open_time_extension_requests.any? %>
+        <%= link_to "Extension requested", planning_application_validation_validation_request_path(@planning_application, @planning_application.time_extension_validation_requests.last), class: "govuk-link" %>
```
benjamineskola, line 69: Can we use `time_extension_request` here? That also sorts them.

```suggestion
        <%= link_to "Extension requested", planning_application_validation_validation_request_path(@planning_application, @planning_application.time_extension_request, class: "govuk-link" %>
```

db/migrate/20240303120759_add_proposed_expiry_date_to_validation_requests.rb
Rhian: Yes makes much more sense!

app/views/planning_applications/validation/validation_requests/index.html.erb
```diff
@@ -23,7 +25,7 @@
     <% if @active_validation_requests.any? %>
       <% if @planning_application.not_started? %>
         <p class="govuk-body">
-          The following requests will be sent when the application is invalidated.
+          <%= "The following requests will be sent when the application is invalidated." unless @planning_application.has_only_time_extension_requests? %>
         </p>
       <% end %>
       <% if post_validation_requests_index? %>
```
benjamineskola, line 31: It looks like this will lead to an empty `<p>` if the application is not started and it has only time extension requests. Probably not a big deal, but perhaps the if statement could be `if @planning_application.not_started? && !@planning_application.has_only_time_extension_requests?` instead?

app/helpers/description_change_validation_request_helper.rb
Rhian: I hadn't spotted that method... makes life so much easier!

app/models/time_extension_validation_request.rb
```diff
@@ -0,0 +1,56 @@
...
+  validates :cancel_reason, presence: true, if: :cancelled?
+  validates :reason, presence: true
+  validate :rejected_reason_is_present?
+  validate :allows_only_one_open_time_extension, on: :create
+  validate :proposed_expiry_is_later?
+
+  def proposed_expiry_is_later?
+    return if planning_application.nil? || proposed_expiry_date.blank?
+
+    if proposed_expiry_date < planning_application.expiry_date
```
benjamineskola, line 14: Minor point, but what happens if the proposed date is the same as the current expiry date? It's not an error in the same kind of way (because going backwards isn't allowed but staying the same is fine), but it's still sort of nonsensical to request a change where the proposed is the same as the existing.

```suggestion
    if proposed_expiry_date <= planning_application.expiry_date
```

app/views/planning_applications/validation/time_extension_validation_requests/_show.html.erb
```diff
@@ -0,0 +1,57 @@
...
+    </h1>
+    <p class="govuk-body">
+      <strong>At:</strong> <%= @planning_application.full_address  %> <br>
+      <strong>Date received:</strong> <%= @planning_application.received_at.to_date.to_fs %> <br>
+      <strong>Application number:</strong> <%= @planning_application.reference  %>
+    </p>
+
+    <div>
+      <p class="govuk-body"> <strong> Previous expiry date:</strong><br>
+        <%= @planning_application.expiry_date.strftime("%-d %B %Y")  %>
```
benjamineskola, line 22: Another of these:

```suggestion
        <%= @planning_application.expiry_date.to_fs  %>
```

benjamineskola (approved):
Just a couple more suggestions but it's all minor things, otherwise looks good to me!

app/views/planning_applications/validation/validation_requests/index.html.erb
Rhian: As if by magic, I realised that applying a scope as per your previous suggestion means I don't need that `unless` as the time extension requests are already excluded on line 25 :)

Rhian:
> Just a couple more suggestions but it's all minor things, otherwise looks good to me!

Suggestions gratefully received and implemented, thanks!

PR #1613, opened 2024-03-08
Allow officer to upload site visits anytime

cdccollins:
### Description of change

Allow planning officers to upload site visits at any point, not just if there are neighbour objections (as is the case for prior approvals)

### Story Link

https://trello.com/c/obyo7YJb/2657-able-to-add-site-visit-evidence-at-any-point-for-planning-applications#1612

### DIFF
--- a/app/models/consultation.rb
+++ b/app/models/consultation.rb
@@ -363,6 +363,10 @@ class Consultation < ApplicationRecord
     end
   end
 
+  def needs_site_visit?
+    planning_application.prior_approval? ? neighbour_responses.objection.any? : true
+  end
+
   private
 
   def unknown_placeholders(string)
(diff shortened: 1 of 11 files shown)


app/models/consultation.rb
```diff
@@ -363,6 +363,10 @@ def period_days
     end
   end
 
+  def needs_site_visit?
+    planning_application.prior_approval? ? neighbour_responses.objection.any? : true
+  end
```
benbaumann95, line 368: Is there a way we could use the `features` attribute on application type instead of checking directly. This will help when we scope out the features per application type when working on the global admin. Currently we have permitted development rights and planning conditions on the attribute

benbaumann95 (approved):

PR #1617, opened 2024-03-11
Allow officer to recommend application for committee

cdccollins:
### Description of change

Officer can submit a recommendation that the planning application go to committee to the reviewer

### Story Link

https://trello.com/c/PRVhM99P/2614-able-to-record-whether-an-application-will-be-decided-by-committee-or-not

### Screenshots

screencapture-lambeth-bops-localhost-3000-planning-applications-2216-assessment-recommendations-new-2024-03-11-17_28_51


db/schema.rb
```diff
@@ -111,6 +111,16 @@
     t.index ["user_id"], name: "ix_comments_on_user_id"
   end
 
+  create_table "committee_decisions", force: :cascade do |t|
+    t.bigint "planning_application_id"
+    t.boolean "recommend", default: false, null: false
```
Rhian, line 116: Just one very tiny comment that perhaps `recommended` would be better as it reads more like an attribute but `recommend` is more readable, so probably leave it as is.

Rhian (approved):

db/schema.rb
cdccollins: Haha yes I can argue it either way in my head

PR #1619, opened 2024-03-12
Add primary navigation component to header

benbaumann95:
### Description of change

Add primary navigation component to header.
This is still a prototype component so this is not included in govuk-frontend yet. Rather than importing the entire @x-govuk/govuk-prototype-components library, just add the css and we can switch to using the appropriate css when this is incorporated with govuk-frontend.

- https://x-govuk.github.io/govuk-prototype-components/get-started/
- https://x-govuk.github.io/govuk-prototype-components/primary-navigation/

### Story Link

https://trello.com/c/WdTO1yPr/2677-navigation-in-global-admin

### Screenshots

**ADMIN**
Screenshot 2024-03-11 at 21 13 23

**GLOBAL ADMIN**
Screenshot 2024-03-12 at 10 40 57


app/views/application/_header.html.erb
```diff
@@ -28,3 +30,18 @@
     <% end %>
   </div>
 </header>
+
+<% if respond_to?(:nav_items) %>
```
benjamineskola, line 34: I don't quite like that we use `respond_to` for this. Could we define a default that's nil or empty, and then use `if nav_items.present?` instead?

benjamineskola (approved):
Above is not really a blocker if we can't do it that way.

app/views/application/_header.html.erb
benbaumann95: Yeah I can define an empty array as we will probably want to add nav items to the main app at some point anyway

pixeltrix:
Example in a window has the blue bar across the whole width:

https://x-govuk.github.io/govuk-prototype-components/examples/primary-navigation-below-phase-banner/

image

benjamineskola:
> Example in a window has the blue bar across the whole width:

I wonder if this relates to changing the body width a little while back. It might be using the wrong width still somewhere.

I don't think that's actually a new issue in this PR, because it's happening on main too for me, but adding a new coloured element below it makes it look more obvious. I'll take a closer look.

benjamineskola:
> I don't think that's actually a new issue in this PR, because it's happening on main too for me, but adding a new coloured element below it makes it look more obvious. I'll take a closer look.

Yes, got it. In `app/views/application/_header.html.erb` the classes `govuk-header__container` and `govuk-width-container` are defined on the same div. Instead, that div should have only `govuk-header__container` and there should be a new div inside it (wrapping all its contents) with `govuk-width-container`. This will make the blue bar full-width but allow the text content to remain only the current width. 

i.e.,

```diff
-   <div class="govuk-header__container govuk-width-container">
+.  <div class="govuk-header__container">
+     <div class="govuk-width-container">
        <!-- all content unchanged -->
+     </div>
    </div>
```

benjamineskola:
Actually it's easier just top open a new PR: #1621

PR #1621, opened 2024-03-12
Fix header blue bar width

benjamineskola:
Separate govuk-header__container and govuk-width-container

This allows the blue bar to extend across the full width of the viewport, like the black bar above it.

Screenshot 2024-03-12 at 14 41 24

### DIFF
--- a/app/views/application/_header.html.erb
+++ b/app/views/application/_header.html.erb
@@ -1,33 +1,35 @@
 <header class="govuk-header" role="banner" data-module="govuk-header">
-  <div class="govuk-header__container govuk-width-container">
-    <%= link_to(
-      home_path,
-      class: "govuk-header__link govuk-header__link--homepage"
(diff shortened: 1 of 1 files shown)


cdccollins (approved):

cdccollins:
wait, is it supposed to go all the way across? Doesn't look like it does on govuk?

benjamineskola:
> wait, is it supposed to go all the way across? Doesn't look like it does on govuk?

According to this example it does: https://x-govuk.github.io/govuk-prototype-components/examples/primary-navigation-below-phase-banner/ maybe only with the navigation thing because it looks weird if it's not all the way in that case?

benbaumann95 (approved):

benbaumann95 (approved):
Merged https://github.com/unboxed/bops/pull/1619 now

PR #1622, opened 2024-03-12
Add new application type wizard

pixeltrix:
### Description of change

As a global admin
I need to create a new application type
So that I can support further types of planning application

### Story Link

https://trello.com/c/bok2DajY

### Decisions

Had to duplicate the existing `lawfulness_certificate` to support the existing and proposed variants. Main code needs to be refactored away from using the name to the new code value but is left for future PRs so that we can merge the first step of the wizard. Also removed the ability to set the work status but it remains reported in the API JSON based on the application type code.

### Screenshots

**Form Page**
image

**Summary Page**

image


db/migrate/20240312092625_add_active_to_application_types.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class AddActiveToApplicationTypes < ActiveRecord::Migration[7.1]
+  class ApplicationType < ActiveRecord::Base; end
+
+  def change
+    add_column :application_types, :active, :boolean
```
benbaumann95, line 7: Not for now but wondering if there is any chance we might want to archive or expire an application type and in that case might require an enum for status.

EGiataganas: @benbaumann95 This is what I've mentioned in stand up, this should be an enum as based on Ed's comments, there are three statuses:  `Active, Inactive, Retired`

pixeltrix: Updated PR to use a `status` column with `inactive`, `active` and `retired`

benbaumann95 (approved):

EGiataganas (approved):

config/locales/odp.yml
```diff
@@ -0,0 +1,62 @@
+en:
+  odp:
+    application_types:
```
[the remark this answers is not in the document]
pixeltrix: No, because the keys contains `.` which is not allowed in locale files

PR #1629, opened 2024-03-13
Allow only determined applications in API

Rhian:
### Description of change

Fixes the API call to display individual applications so it returns 404 for pre-determined applications.

### Story Link

https://trello.com/c/GdxDvWit/2555-swagger-retrieve-function-api-v1-planningapplications-id-states-it-returns-only-determined-planning-applications-when-it-infact


Rhian:
Closing for now.

PR #1631, opened 2024-03-14
Avoid FeeCalculation errors on missing PlanX data

benjamineskola:
- **Don't try to create fee calculations if the ODP data is missing**
- **Add an additional FeeCalculation test**

### Description of change

### Story Link

https://link-to-issue

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


app/models/fee_calculation.rb
```diff
@@ -11,20 +11,19 @@ class FeeCalculation < ApplicationRecord
 
   class << self
     def from_odp_data(odp_data)
+      return unless odp_data.present? && odp_data.is_a?(Hash)
```
Rhian, line 14: Is there ever a case where some of the fee calculation fields in the hash might be missing or malformed that we may need to put in an additional check for?

benjamineskola: I _think_ that that would probably be handled by the schema validation at API level. It's theoretically possible that they send some kind of technically-valid data that breaks the assumptions made here but at that point I'd be totally guessing about what kind of validations were necessary.

PR #1635, opened 2024-03-15
Add determination period days to application type

benbaumann95:
### Description of change

Add determination period days to application type

### Story Link

https://trello.com/c/7qaNs63M/2682-enter-determination-period-for-new-application-types

### Screenshots

Screenshot 2024-03-15 at 12 54 42

Screenshot 2024-03-15 at 12 56 47

### DIFF
--- a/engines/bops_config/app/controllers/bops_config/application_controller.rb
+++ b/engines/bops_config/app/controllers/bops_config/application_controller.rb
@@ -22,5 +22,12 @@ module BopsConfig
       session[:back_path] = request.referer if request.get?
       @back_path = session[:back_path]
     end
+
+    def application_type_id
+      param_value = params[request.path_parameters.key?(:application_type_id) ? :application_type_id : :id]
+      Integer(param_value)
+    rescue
+      raise ActionController::BadRequest, "Invalid application type id: #{param_value}"
+    end
   end
 end
--- a/engines/bops_config/spec/system/application_types_spec.rb
+++ b/engines/bops_config/spec/system/application_types_spec.rb
@@ -22,9 +22,32 @@ RSpec.describe "Application Types", type: :system, bops_config: true do
     fill_in "Suffix", with: "MAJR"
 
     click_button "Continue"
+    # Set determination period
+    expect(page).to have_selector("h1", text: "Set determination period")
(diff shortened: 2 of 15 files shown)


engines/bops_config/app/controllers/bops_config/application_controller.rb
```diff
@@ -22,5 +22,15 @@ def set_back_path
...
     end
+
+    def application_type_id
+      if request.path_parameters.key?(:application_type_id)
+        Integer(params[:application_type_id])
+      else
+        Integer(params[:id])
+      end
+    rescue
+      raise ActionController::BadRequest, "Invalid application type id: #{params[:id]}"
```
benjamineskola, line 33: In the exception we always reference `id` but in the body we sometimes reference `application_type_id`. Is this going to matter, if the exception is thrown in the first branch of the conditional, or would `id` still be meaningful here?

engines/bops_config/spec/system/application_types_spec.rb
```diff
@@ -104,7 +127,7 @@
     visit "/application_types/#{application_type.id}"
     expect(page).to have_selector("h1", text: "Review the application type")
 
-    within "dl div:nth-child(3)" do
+    within "dl div:nth-child(4)" do
```
benjamineskola, line 130: Not for this PR but it might be worth reviewing some of these tests that reference a particular child element by number — I had to alter one recently too. Perhaps we could give list entries etc an ID and then be able to use `within` on that.

benjamineskola (approved):
Just one small thing to double check.

engines/bops_config/spec/system/application_types_spec.rb
benbaumann95: Yeah that is usually my preference using the id but i've gone with being consistent with how the other tests are set up

engines/bops_config/app/controllers/bops_config/application_controller.rb
benbaumann95:
```ruby
    def application_type_id
      param_value = params[request.path_parameters.key?(:application_type_id) ? :application_type_id : :id]
      Integer(param_value)
    rescue
      raise ActionController::BadRequest, "Invalid application type id: #{param_value}"
    end
```

PR #1643, opened 2024-03-18
Upload committee decision

cdccollins:
### Description of change

Able to add what the committee decided about the planning application

### Story Link

Second half of https://trello.com/c/VB8rgwp0/2613-able-to-send-a-committee-meeting-notification


app/controllers/planning_applications/review/recommendations_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+          format.html
+        end
+      end
+
+      def create
+        @recommendation = Recommendation.new(
+          planning_application: @planning_application.planning_application,
+          status: "review_complete",
+          reviewer: Current.user
+        )
```
pixeltrix, line 31: We should construct the model through the association, e.g.

``` ruby
@recommendation = @planning_application.recommendations.new(status: "review_complete", reviewer: Current.user)
```

app/controllers/planning_applications/review/recommendations_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+
+        if @recommendation.save
+          @planning_application.update!(decision: recommendation_form_params[:decision], public_comment: recommendation_form_params[:public_comment])
```
pixeltrix, line 34: This should be handled in the model layer - either detect the update needs to be made and use a callback or add an explicit method like this:

``` ruby
def save_and_submit(params)
  transaction do
    save!
    planning_application.update!(params.slice(:decision, :public_comment))
    planning_application.submit!
  end

  true
rescue ActiveRecord::RecordNotSaved
  false
end
```

Then here you can do:

``` ruby
if @recommendation.save_and_submit(recommendation_form_params)
  redirect_to planning_application_review_tasks_path(@planning_application), notice: t(".success")
else
  render :new
end
```

(the model method may need extra classes to rescue, etc.)

app/controllers/planning_applications/review/recommendations_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+                redirect_to new_planning_application_review_recommendation_path(@planning_application)
+              else
+                redirect_to planning_application_review_tasks_path(@planning_application),
+                  notice: t(".success")
+              end
+            else
+              redirect_to planning_application_review_tasks_path(@planning_application),
+                notice: t(".success")
+            end
+          end
```
pixeltrix, line 65: Again move all the model logic to the model class and then extract the redirect path logic to a private method so you can do something like:

``` ruby
if @recommendation.save_and_review(recommendation_params)
  redirect_to after_save_and_review_location_url, notice: t(".success")
else
  render :edit
end
```

pixeltrix (changes requested):

PR #1646, opened 2024-03-19
View edit application type

EGiataganas:
### Description of change

- Use application type show page
- Improvements on application type show/index to match

### Story Link

https://trello.com/c/Cv6jDoBY


engines/bops_config/app/views/bops_config/application_types/show.html.erb
```diff
@@ -67,11 +67,13 @@
...
         <dd class="govuk-summary-list__value">
-          <%= @application_type.status.humanize %>
+          <span class="govuk-tag <%= "govuk-tag--#{tag_colour(@application_type.status)}" %>">
+            <%= @application_type.status.humanize %>
+          </span>
         </dd>
         <dd class="govuk-summary-list__actions">
           <%= link_to [:edit, @application_type, :status], class: "govuk-link" do %>
-            <%= t(".edit") %><span class="govuk-visually-hidden"> <%= t(".status") %></span>
+            <%= t(".change") %><span class="govuk-visually-hidden"> <%= t(".status") %></span>
```
pixeltrix, line 76: I used 'Edit' as that was what was in the Figma design for 'Live date for staging' which is what we've replaced with 'Status' - could be an oversight on Fede's part or may be intentional so we should check with him.

pixeltrix (approved):
Approved pending checking on Edit/Change

engines/bops_config/app/views/bops_config/application_types/show.html.erb
EGiataganas: Yeah, I've had a catch-up with Fede and made these changes based on his feedback! I did mention that on the Figma design is `edit` but in some order places `change`. He said that it might be something he forgot to update and we agreed to update it to `change` so we are consistent

PR #1647, opened 2024-03-20
Feature/confirm consultees consulted

Rhian:
### Description of change

This PR introduces a new assessment task, which is to check that all relevant consultees have been contacted. The assessor can choose to mark as checked without any validation, as it's possible that some constraints may not have consultees attached.

### Story Link

https://trello.com/c/RxWrC2IZ/2502-confirm-that-i-consulted-all-the-right-people

### Screenshots

consulted_2
confirm_consulted


db/schema.rb
```diff
@@ -647,6 +647,7 @@
     t.geography "neighbour_boundary_geojson", limit: {:srid=>4326, :type=>"geometry_collection", :geographic=>true}
     t.string "documents_status", default: "not_started", null: false
     t.datetime "in_committee_at"
+    t.boolean "consultees_checked", default: false, null: false
```
cdccollins, line 650: I'm going to be annoying... 😬 
Could you use the Review table instead? And then when the reviewer has to check consultees, we can update the review associated with the consultees? 

If that doesn't work, let me know

app/components/consultee_status_component.html.erb
```diff
@@ -0,0 +1 @@
+<%= consultee_status %>
```
cdccollins, line 1: Is there a reason you're not using the `base_component.html.erb` here?

cdccollins (changes requested):
Feel free to argue with me 😄

db/schema.rb
Rhian: Ahahaha, I think this does make a lot of sense. So treat this step as a sort of "pre-review"? Happy to do this :)

app/components/consultee_status_component.html.erb
Rhian: Because I didn't realise it existed, lol. Will update.

Rhian: ... actually, I've realised I can't now I've looked at it, as I have more statuses to account for than are in the `base_component`. So that either means :
- loading this with more statuses
- having more conditionals in the table column
- using the status tag

Any of these would work - not sure which would be best?

cdccollins: I think there are maybe 5 new ones you'd be introducing? I'd chuck them in the table, that's what I've been doing. I think (I hope) all these components will get a nice refactor soon, so easier to sellotape for now 😆

Rhian: OK cool 😎

app/controllers/planning_applications/assessment/consultees_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+      before_action :set_consultation
+
+      def index
+        respond_to do |format|
+          format.html
+        end
+      end
+
+      def check
+        @consultation.create_assessor_consultee_review
```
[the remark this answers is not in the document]
Rhian: I've just added a generic error message saying it can't be saved (for example, if something strange happens with the database). Because the assessor isn't passing any params, there isn't really any action they can take to remedy it - not sure if there is a better way to handle this?

app/components/task_list_items/assessment/consultees_consulted_component.rb
```diff
@@ -0,0 +1,35 @@
...
+      def initialize(planning_application:)
+        @planning_application = planning_application
+      end
+
+      private
+
+      attr_reader :planning_application
+
+      def link_text
+        "Check consultees consulted"
```
cdccollins, line 15: Can you put this in the .yml file, that's what we've been doing (I also always get told to do it)

app/controllers/planning_applications/assessment/consultees_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+        respond_to do |format|
+          format.html
+        end
+      end
+
+      def check
+        if @consultation.create_consultees_review
+          redirect_to planning_application_assessment_tasks_path(@planning_application)
+        else
+          flash.now(alert: "Consultees were not marked as checked for this application")
```
cdccollins, line 21: Do you also need to render :index, or does it work without that?

app/models/review.rb
```diff
@@ -59,6 +59,7 @@ class NotCreatableError < StandardError; end
   scope :neighbour_reviews, -> { review_type("neighbour") }
   scope :not_accepted, -> { where(action: "rejected").order(created_at: :asc) }
   scope :reviewer_not_accepted, -> { not_accepted.where.not(reviewed_at: nil) }
+  scope :consultees_checked, -> { where(owner_type: "Consultation").where("specific_attributes->>'review_type' = ?", "consultees_checked") }
```
cdccollins, line 62: Can you follow the pattern I've used for `neighbour_reviews`?

cdccollins (changes requested):
Just a couple of minor things!

app/controllers/planning_applications/assessment/consultees_controller.rb
Rhian: I *think* it works without, but will check! May add anyway for safety

app/views/planning_applications/assessment/consultees/index.html.erb
```diff
@@ -0,0 +1,45 @@
+<% content_for :page_title do %>
+  Confirm all consultees have been consulted - <%= t("page_title") %>
+<% end %>
+
+<% add_parent_breadcrumb_link "Home", root_path %>
+<% add_parent_breadcrumb_link "Application", planning_application_path(@planning_application) %>
+<% add_parent_breadcrumb_link "Consultation", planning_application_consultation_path(@planning_application) %>
+<% content_for :title, "Confirm all consultees have been consulted" %>
```
cdccollins, line 8: Sorry, just noticed this, we have a shared component that deals with this. 

e.g.

```
<%= render(
      partial: "shared/assessment_task_breadcrumbs",
      locals: { planning_application: @planning_application }
    ) %>

<% content_for :title, "Add #{"pre-commencement" if @condition_set.pre_commencement?} conditions" %>
```

cdccollins (approved):
One tiny thing, but otherwise 🎉

PR #1649, opened 2024-03-20
Allow entering of the legislation for an application type

pixeltrix:
### Story Link

https://trello.com/c/Bhj83JuR

### Screenshots

Choosing an existing legislation:
image

Entering an new legislation:
image


pixeltrix:
Some things to note:

1. `form.govuk_submit` can take a block for additional buttons and sets up the wrapper `<div>` tag for you
2. Don't disable labels on form elements as that makes them inaccessible (and harder to reference in specs) - use either the heading inside a `<fieldset>` or use `hidden: true` to make it visually hidden but still accessible to screen readers
3. `form.govuk_text_field` takes options for `:prefix_text` and `:suffix_text` and it will set up the correct HTML for you

pixeltrix:
Additionally the "Continue" and "Back" button behaviour probably need some further revision - I've made them so that they're at least consistent now as the dynamic `back_path` helper gets confused.

benbaumann95 (approved):

PR #1650, opened 2024-03-20
Reviewer can review neighbour notifications

cdccollins:
### Description of change

Reviewer can check what neighbour notifications were sent and the relevant dates

### Story Link

https://trello.com/c/Lg9ucw88/2662-reviewer-is-able-to-quickly-check-neighbour-notifications

### Screenshots
screencapture-lambeth-bops-localhost-3000-planning-applications-2437-review-consultation-351-neighbour-responses-2024-03-20-17_09_09

### DIFF
--- a/app/controllers/planning_applications/review/neighbour_responses_controller.rb
+++ b/app/controllers/planning_applications/review/neighbour_responses_controller.rb
@@ -11,7 +11,16 @@ module PlanningApplications
       before_action :set_consultation
       before_action :set_neighbour_review
 
-      def index
+      def show
+        respond_to do |format|
+          format.html
+        end
+      end
+
+      def edit
+        respond_to do |format|
+          format.html
+        end
       end
 
       def update
@@ -20,7 +29,7 @@ module PlanningApplications
             if @neighbour_review.update(review_params) && @consultation.update(status: consultation_status)
               redirect_to planning_application_review_tasks_path(@planning_application), notice: t(".success")
             else
-              render :index
+              render :edit
             end
           end
         end
@@ -34,7 +43,7 @@ module PlanningApplications
             if @neighbour_review.save && @consultation.update(status: consultation_status)
               redirect_to planning_application_review_tasks_path(@planning_application), notice: t(".success")
             else
-              render :index
+              render :edit
             end
           end
         end
--- a/app/controllers/planning_applications/review/neighbour_responses_controller.rb
+++ b/app/controllers/planning_applications/review/neighbour_responses_controller.rb
@@ -17,7 +17,7 @@ module PlanningApplications
       def update
         respond_to do |format|
           format.html do
-            if @neighbour_review.update(review_params)
+            if @neighbour_review.update(review_params) && @consultation.update(status: consultation_status)
               redirect_to planning_application_review_tasks_path(@planning_application), notice: t(".success")
             else
               render :index
@@ -31,7 +31,7 @@ module PlanningApplications
 
         respond_to do |format|
           format.html do
-            if @neighbour_review.save
+            if @neighbour_review.save && @consultation.update(status: consultation_status)
               redirect_to planning_application_review_tasks_path(@planning_application), notice: t(".success")
             else
               render :index
@@ -59,6 +59,14 @@ module PlanningApplications
           :to_be_reviewed
         elsif mark_as_complete?
           :complete
+        end
+      end
+
+      def consultation_status
+        if return_to_officer?
+          :to_be_reviewed
+        else
+          @consultation.status
         end
       end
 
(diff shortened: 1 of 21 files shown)


benjamineskola (approved):

app/models/consultation.rb
```diff
@@ -49,6 +49,11 @@ class AddNeighbourAddressesError < StandardError; end
     has_many :neighbour_responses
   end
 
+  has_many :reviews, as: :owner, dependent: :destroy
+  has_many :neighbour_reviews, -> { where(specific_attributes: {consultation_type: "neighbour"}) }, as: :owner, class_name: "Review", dependent: :destroy
```
pixeltrix, line 53: I'd move that `where` to a scope on review and then you can just call that in the proc here, e.g.

``` ruby
with_options as: :owner do
  has_many :reviews, dependent: :destroy
  has_many :neighbour_reviews, -> { neighbour_reviews }, class_name: "Review"
end
```

Note when you're associating to the same table multiple times that you should only have one association with a `dependent: :destroy` option otherwise Active Record will try to load and destroy the the same objects multiple times.

Depending on the expected number of reviews you might be better off using association extension methods, e.g. 

``` ruby
has_many :reviews, as: :owner, dependent: :destroy do
  def neighbours
    select(&:neighbour?)
  end

  def latest_neighbour
    neighbours.sort_ by(&:created_at).last
  end
end
```

That way there's a single set of review objects in memory and you don't run the risk of having multiple instances of a review where one is loaded via `reviews` and another loaded by `neighbour_reviews`.

app/models/consultation.rb
```diff
@@ -371,6 +379,14 @@ def needs_site_visit?
     planning_application.prior_approval? ? neighbour_responses.objection.any? : planning_application.application_type.site_visits?
   end
 
+  def neighbour_review
+    neighbour_reviews.order(:created_at).last
+  end
```
pixeltrix, line 384: This will always run an SQL query every time that it's called - depending on how you address the source association comment I made above you can either implement this as a `has_one` association that will cache the lookup or use the association extension method.

app/models/consultation.rb
```diff
@@ -371,6 +379,14 @@ def needs_site_visit?
...
+
+  def create_neighbour_review!
+    reviews.create!(specific_attributes: {consultation_type: "neighbour"}, status: "complete", assessor: Current.user)
```
pixeltrix, line 387: One reason for keeping the separate associations and not using association extensions is that you can use the `neighbour_reviews` to create the instance and it will set `specific_attributes` for you, e.g.

``` ruby
neighbour_reviews.create!(status: "complete", assessor: Current.user)
```

app/models/planning_application.rb
```diff
@@ -542,7 +542,8 @@ def updates_required?
     assessment_details_for_review.any?(&:update_required?) ||
       permitted_development_right&.update_required? ||
       policy_classes.any?(&:update_required?) ||
-      (committee_decision.present? && committee_decision.current_review.rejected?)
+      (committee_decision.present? && committee_decision.current_review.rejected?) ||
+      neighbour_review_requested?
```
pixeltrix, line 546: Something to look at in the next state of work is whether there's a better way to do this such as a `updates_required` boolean column that gets set to true in a callback.

app/models/planning_application.rb
```diff
@@ -1055,4 +1056,8 @@ def filter_documents_for_tab(documents, tab)
 
     documents.select { |document| (tags & document.tags).any? }
   end
+
+  def neighbour_review_requested?
+    consultation.present? && consultation.reviews.any? { |review| review.to_be_reviewed? }
```
pixeltrix, line 1061: You can rewrite this as

``` ruby
consultation && consultation.reviews.any?(&:to_be_reviewed?)
```

app/controllers/planning_applications/review/neighbour_responses_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+              redirect_to planning_application_review_tasks_path(@planning_application), notice: t(".success")
+            else
+              render :edit
+            end
+          end
+        end
+      end
+
+      def create
+        @neighbour_review = @consultation.reviews.new(review_params.merge(specific_attributes: {consultation_type: "neighbour"}))
```
pixeltrix, line 39: There's a store accessor for `consultation_type` - can't you just do `{consultation_type: "neighbour"}` ?

Also, can't you use `@consultation.neighbour_reviews.new` ?

app/controllers/planning_applications/review/neighbour_responses_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+        end
+      end
+
+      private
+
+      def review_params
+        params.require(:review).permit(
+          :action, :comment, :id
+        ).to_h
+          .deep_merge(
```
pixeltrix, line 58: Is the `to_h` and `deep_merge` necessary? `ActionController::Parameters` has a `merge` method.

pixeltrix (changes requested):

PR #1653, opened 2024-03-21
Add decisions to application type

EGiataganas:
### Description of change

Add decisions to application type

### Story Link

https://trello.com/c/CU1ChD2M

### DIFF
--- a/engines/bops_config/spec/system/application_types_spec.rb
+++ b/engines/bops_config/spec/system/application_types_spec.rb
@@ -18,6 +18,9 @@ RSpec.describe "Application Types", type: :system do
     create(:reporting_type, :major_industry)
     create(:reporting_type, :major_retail)
 
+    create(:decision, :full_granted)
+    create(:decision, :full_refused)
+
     visit "/application_types/new"
     expect(page).to have_selector("h1", text: "Application profile")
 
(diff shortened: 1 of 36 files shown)


engines/bops_config/app/controllers/bops_config/application_types/document_tags_controller.rb
```diff
@@ -43,7 +43,11 @@ def set_document_tags
       end
 
       def next_path
-        application_type_path(@application_type)
+        if @application_type.decision_set?
```
benbaumann95, line 46: Could we use `configured?` instead here.
And if the default is `{}` this will always return true right

engines/bops_config/spec/system/application_types_spec.rb
```diff
@@ -108,6 +108,17 @@
 
     expect(page).to have_content("Document tags successfully updated")
 
+    # Set decisions
+    expect(page).to have_selector("h1", text: "Choose decisions")
+    click_button "Continue"
+
+    expect(page).to have_content("Decisions can't be blank")
```
benbaumann95, line 115: Minor detail but do we want a more helpful error message here.
"You must choose at least one decision" etc or maybe ask Laura/Fede

benbaumann95 (commented):
Since this step is now the last one, we need to update
```ruby
before_validation on: :document_tags, unless: :configured? do
  self.configured = true
end
  ```
  
  so that we update the `configured` boolean when `:decision_set`, not `:document_tags`

benbaumann95 (commented):
Since we've added an edit system spec for all the other steps, could we please include one here for consistency. I.e `"it "allows editing of the decisions for recommendation"`

benbaumann95 (commented):
I think we also need to update the `def active_page_key` in `application_helper.rb` to include the decision sets controller page key for the navigation

PR #1654, opened 2024-03-21
Improvements to neighbour_reviews association

cdccollins:
### Description of change

Improvements suggested here https://github.com/unboxed/bops/pull/1650 after I merged too early!

### DIFF
--- a/app/models/consultation.rb
+++ b/app/models/consultation.rb
@@ -49,8 +49,11 @@ class Consultation < ApplicationRecord
     has_many :neighbour_responses
   end
 
-  has_many :reviews, as: :owner, dependent: :destroy
-  has_many :neighbour_reviews, -> { where(specific_attributes: {consultation_type: "neighbour"}) }, as: :owner, class_name: "Review", dependent: :destroy
+  with_options as: :owner do
+    has_many :reviews, dependent: :destroy
+    has_many :neighbour_reviews, -> { neighbour_reviews }, class_name: "Review" # rubocop:disable Rails/HasManyOrHasOneDependent
+    has_one :neighbour_review, class_name: "Review" # rubocop:disable Rails/HasManyOrHasOneDependent
+  end
 
   accepts_nested_attributes_for :reviews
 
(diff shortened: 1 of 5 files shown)


app/models/planning_application.rb
```diff
@@ -1058,6 +1058,6 @@ def filter_documents_for_tab(documents, tab)
   end
 
   def neighbour_review_requested?
-    consultation.present? && consultation.reviews.any? { |review| review.to_be_reviewed? }
+    consultation.present? && consultation.reviews.any?(&:to_be_reviewed?)
```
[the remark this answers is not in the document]
pixeltrix: Returning `nil` should be fine and treated as falsey though it can be contentious - there was a Rails PR that ran to 96 comments about making `request.xhr?` return true/false before being closed unmerged 😂

app/models/review.rb
```diff
@@ -57,6 +57,7 @@ class NotCreatableError < StandardError; end
   scope :enforcement, -> { where("specific_attributes->>'review_type' = ?", "enforcement") }
   scope :not_accepted, -> { where(action: "rejected").order(created_at: :asc) }
   scope :reviewer_not_accepted, -> { not_accepted.where.not(reviewed_at: nil) }
+  scope :neighbour_reviews, -> { where(specific_attributes: {consultation_type: "neighbour"}) }
```
pixeltrix, line 60: Will this work when there are multiple specific attributes?

app/models/consultation.rb
```diff
@@ -49,8 +49,11 @@ class AddNeighbourAddressesError < StandardError; end
     has_many :neighbour_responses
   end
 
-  has_many :reviews, as: :owner, dependent: :destroy
-  has_many :neighbour_reviews, -> { where(specific_attributes: {consultation_type: "neighbour"}) }, as: :owner, class_name: "Review", dependent: :destroy
+  with_options as: :owner do
+    has_many :reviews, dependent: :destroy
+    has_many :neighbour_reviews, -> { neighbour_reviews }, class_name: "Review" # rubocop:disable Rails/HasManyOrHasOneDependent
+    has_one :neighbour_review, class_name: "Review" # rubocop:disable Rails/HasManyOrHasOneDependent
```
pixeltrix, line 55: Hmm, not sure whether there's something about `dependent: :destroy` I'm missing or rubocop's rule is a bit braindead - going to have to have a look at this now for my own sanity.

pixeltrix (approved):
Just that query about the scope - if it's an issue then look at the scopes for `evidence` and `enforcement` on how to solve it.

cdccollins:
> Just that query about the scope - if it's an issue then look at the scopes for `evidence` and `enforcement` on how to solve it.

You're right it won't return reviews with multiple attributes, but then it also doesn't set the attribute correctly when I use `neighbour_reviews.new`

PR #1662, opened 2024-03-25
Small fixes for publicity review

cdccollins:
### Description of change

Fixes for publicity review
- can send a case back to an officer while consultation is happening
- don't show assessment details as started when check publicity is started

### DIFF
--- a/app/components/concerns/assessment_detailable.rb
+++ b/app/components/concerns/assessment_detailable.rb
@@ -5,7 +5,7 @@ module AssessmentDetailable
 
   included do
     def assessment_details
-      @assessment_details ||= planning_application.assessment_details_for_review.reject { |detail| detail.category == "check_publicity" }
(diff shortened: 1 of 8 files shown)


app/components/concerns/assessment_detailable.rb
```diff
@@ -5,7 +5,7 @@ module AssessmentDetailable
 
   included do
     def assessment_details
-      @assessment_details ||= planning_application.assessment_details_for_review
+      @assessment_details ||= planning_application.assessment_details_for_review.reject { |detail| detail.category == "check_publicity" }
```
pixeltrix, line 8: You can just use `reject(&:check_publicity?)` here as the `enum` creates an instance method for that category

pixeltrix (approved):
Approving as it's just that minor tweak

PR #1664, opened 2024-03-26
Check publicity feature enabled on application type

benbaumann95:


app/models/planning_application.rb
```diff
@@ -776,13 +776,11 @@ def overdue_validation_requests
   end
 
   def check_publicity?
-    return unless application_type.publicity_consultation_feature?
-
-    application_type.assessment_details.include?("check_publicity")
```
benbaumann95, line 781: I think we will be deprecating this when we fully switch over to having all these assessment details within the ApplicationTypeFeature

pixeltrix (approved):

PR #1672, opened 2024-03-28
Decision consistency

EGiataganas:
### Description of change

Make decision consistent

### DIFF
--- a/spec/system/planning_applications/review/sign_off_spec.rb
+++ b/spec/system/planning_applications/review/sign_off_spec.rb
@@ -262,7 +262,7 @@ RSpec.describe "Reviewing sign-off" do
       click_button "Save"
 
       within(".govuk-form-group--error") do
-        expect(page).to have_content("Please state the reasons why this application is, or is not lawful")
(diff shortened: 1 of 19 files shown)


spec/system/planning_applications/review/sign_off_spec.rb
```diff
@@ -262,7 +262,7 @@
       click_button "Save"
 
       within(".govuk-form-group--error") do
-        expect(page).to have_content("Please state the reasons why this application is, or is not lawful")
+        expect(page).to have_content("Please State the reasons for your recommendation.")
```
pixeltrix, line 265: `state` and not `State`

pixeltrix (approved):
Approving since there's just the typo to fix

benbaumann95 (approved):

spec/system/planning_applications/review/sign_off_spec.rb
EGiataganas: Missed it as this spec never run as is marked as `flaky`

PR #1673, opened 2024-04-02
CRUD for legislation

benbaumann95:
### Description of change

Add CRUD for legislation

- Show associated application types
- Allow deleting legislation when there are no associated application types
- Legislation title is readonly

### Story Link

https://trello.com/c/PCueJNCG/2714-as-a-global-admin-user-i-can-edit-the-legislation-description-and-link-for-an-existing-legislation-title

### Screenshots

Screenshot 2024-04-03 at 09 42 04

Screenshot 2024-04-02 at 13 25 29

### DIFF
--- a/engines/bops_config/config/routes.rb
+++ b/engines/bops_config/config/routes.rb
@@ -23,6 +23,8 @@ BopsConfig::Engine.routes.draw do
     end
   end
 
+  resources :legislation, except: %i[show]
+
   resources :users, except: %i[show destroy] do
     get :resend_invite, on: :member
   end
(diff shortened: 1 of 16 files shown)


engines/bops_config/config/routes.rb
```diff
@@ -23,6 +23,8 @@
     end
   end
 
+  resources :legislation, except: %i[show]
```
EGiataganas, line 26: should this route be plural like that `resources :legislations, except: %i[show]`?

benbaumann95: Decided to use just legislation

https://github.com/unboxed/bops/blob/main/config/initializers/inflections.rb from https://github.com/unboxed/bops/commit/ec075bf350302dd0925bd848bdf62b2cfb6fcc50

EGiataganas (approved):

PR #1675, opened 2024-04-03
Align status header in consultees table

Rhian:
### Description of change

We are using the `StatusTags::BaseComponent` in a table column and the text is aligned right as per the styling of this component. The column header was aligned left, which looked weird, so I've added an inline text-align style. I've also converted the `not_assigned` status to be used with the status tag to ensure consistent styling.

### Story Link

https://trello.com/c/j4KNLzLp/2797-layout-issue-move-table-heading-in-consultees-table

### Screenshots

status


app/views/shared/_consultees_table.html.erb
```diff
@@ -4,7 +4,7 @@
   <tr class="govuk-table__row">
     <th class="govuk-table__header">Reason/constraint</th>
     <th class="govuk-table__header">Consultees</th>
-    <th class="govuk-table__header">Status</th>
+    <th class="govuk-table__header" style="text-align:right">Status</th>
```
benbaumann95, line 7: I think this might be a class you can use
`govuk-!-text-align-right`

https://design-system.service.gov.uk/styles/font-override-classes/

benbaumann95 (approved):

benbaumann95:
Just a comment about a possible css class you can use

app/views/shared/_consultees_table.html.erb
Rhian: Yes! It worked! Thanks for spotting it :)

PR #1682, opened 2024-04-05
Track whether a constraint requires consultation

benjamineskola:
### Description of change

We need to track the difference between a constraint that doesn't need consultation, and a constraint that simply hasn't had a consultee assigned to it yet.

### Story Link

https://trello.com/c/zKNjrVNd/2709-allow-marking-constraints-as-not-requiring-a-consultee

### Screenshots

Screenshot 2024-04-08 at 11 36 08

Screenshot 2024-04-08 at 11 37 26

### DIFF
--- a/app/views/shared/_consultees_table.html.erb
+++ b/app/views/shared/_consultees_table.html.erb
@@ -24,6 +24,9 @@
         <% else %>
           <%= render StatusTags::BaseComponent.new(status: "not_assigned") %>
         <% end %>
+        <% unless constraint.consultation_required? %>
+          <%= render StatusTags::BaseComponent.new(status: "not_required") %>
+        <% end %>
       </td>
     </tr>
   <% end %>
--- a/db/migrate/20240405154811_add_consultation_required_to_planning_application_constraint.rb
+++ b/db/migrate/20240405154811_add_consultation_required_to_planning_application_constraint.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class AddConsultationRequiredToPlanningApplicationConstraint < ActiveRecord::Migration[7.1]
+  def change
+    add_column :planning_application_constraints, :consultation_required, :boolean
+
+    up_only do
+      PlanningApplicationConstraint.find_each do |constraint|
+        constraint.update!(consultation_required: true)
+      end
+
+      change_column_default :planning_application_constraints, :consultation_required, true
+      change_column_null :planning_application_constraints, :consultation_required, false
+    end
+  end
+end
--- a/app/controllers/planning_applications/consultee/assign_constraints_controller.rb
+++ b/app/controllers/planning_applications/consultee/assign_constraints_controller.rb
@@ -8,7 +8,8 @@ module PlanningApplications
 
       def create
         respond_to do |format|
-          if planning_application_constraint.update(consultee_id: consultee_id)
+          if planning_application_constraint.update(consultee_id: consultee_id,
+            consultation_required: consultation_required)
             format.html do
               redirect_to planning_application_consultees_url(@planning_application)
             end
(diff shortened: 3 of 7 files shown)


pixeltrix:
Isn't it possible we'll need multiple consultees for some constraints?

benjamineskola:
> Isn't it possible we'll need multiple consultees for some constraints?

Yes. It's a one-to-many relationship to allow for that in future. But right now the requirement is just to ensure that there is _at least_ one person being consulted.

app/views/shared/_consultees_table.html.erb
```diff
@@ -9,7 +9,7 @@
   </thead>
   <tbody class="govuk-table__body">
   <% @planning_application.planning_application_constraints.each do |constraint| %>
-    <tr class="govuk-table__row">
+    <tr class="govuk-table__row" id="constraint-<%= constraint.id %>">
```
pixeltrix, line 12: We should try to avoid injecting ERB in attributes because escaping inside HTML attributes is different from escaping HTML elements - always use a tag helper like this:

``` ruby
<%= tag.tr(id: "constraint-#{constraint.id}", class: "govuk-table__row") do %>
...
<% end %>
```

This ensures attributes are escaped properly and we don't inadvertently open an XSS hole.

app/views/shared/_consultees_table.html.erb
```diff
@@ -24,11 +24,14 @@
         <% else %>
           <%= render StatusTags::BaseComponent.new(status: "not_assigned") %>
         <% end %>
+        <% unless constraint.consultation_required %>
```
pixeltrix, line 27: Should use the query method `consultation_required?` here

app/views/shared/_consultees_table.html.erb
```diff
@@ -24,11 +24,14 @@
...
+          <%= render StatusTags::BaseComponent.new(status: "not_required") %>
+        <% end %>
       </td>
     </tr>
   <% end %>
   <% @planning_application.consultation.consultees.unassigned.find_each do |consultee| %>
-    <tr class="govuk-table__row">
+    <tr class="govuk-table__row" id="consultee-<%= consultee.id %>">
```
pixeltrix, line 34: Same comment about HTML attributes as above

db/migrate/20240405154811_add_consultation_required_to_planning_application_constraint.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddConsultationRequiredToPlanningApplicationConstraint < ActiveRecord::Migration[7.1]
+  def change
+    add_column :planning_application_constraints, :consultation_required, :boolean, null: false, default: true
```
pixeltrix, line 5: Not really a problem now but this will cause a table lock as the default is backfilled - if we were live we'd use `change_column_null` and `change_column_default` after backfilling using `Modle.find_each`.

Wonder whether it's worth adding the [strong_migrations][1] gem to avoid this scenario going forward?

[1]: https://github.com/ankane/strong_migrations

pixeltrix (changes requested):

db/migrate/20240405154811_add_consultation_required_to_planning_application_constraint.rb
benjamineskola: That's a good point — you've mentioned it a few times now and even if it's not strictly necessary yet it's probably better to start sooner than later. (Otherwise, chances are we won't get around to it until it's too late.)

I'll rewrite this migration using the pattern from the few recent migrations.

app/views/shared/_consultees_table.html.erb
benjamineskola: I actually added this in order to make the tests more specific (to pick the right row) and then ended up not using it, so I might as well just remove it.

pixeltrix (approved):

PR #1692, opened 2024-04-10
Use strong_migrations to check safety of migrations

benjamineskola:
### Description of change

Check that we're not introducing changes that might cause performance issues if applied in production (once the service becomes busier).

As recommended in #1682.

### Story Link

https://trello.com/c/3RjI1w3j/2869-add-strongmigrations-gem-to-enforce-good-database-practices

### DIFF
--- a/config/initializers/strong_migrations.rb
+++ b/config/initializers/strong_migrations.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+# Mark existing migrations as safe
+StrongMigrations.start_after = 20240410145215
+
+# Set timeouts for migrations
+# If you use PgBouncer in transaction mode, delete these lines and set timeouts on the database user
+StrongMigrations.lock_timeout = 10.seconds
+StrongMigrations.statement_timeout = 1.hour
+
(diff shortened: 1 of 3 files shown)


config/initializers/strong_migrations.rb
```diff
@@ -0,0 +1,28 @@
...
+# Add custom checks
+# StrongMigrations.add_check do |method, args|
+#   if method == :add_index && args[0].to_s == "users"
+#     stop! "No more indexes on the users table"
+#   end
+# end
+
+# Make some operations safe by default
+# See https://github.com/ankane/strong_migrations#safe-by-default
+# StrongMigrations.safe_by_default = true
```
benjamineskola, line 28: I wonder about enabling this option. It looks like it fixes a lot of the issues that we've commonly encountered.

pixeltrix: I think lets leave it off for now - there are implications around things like disabling DDL transactions that probably shouldn't be blindly added.

pixeltrix (approved):

PR #1693, opened 2024-04-10
Improve site and press notice

benbaumann95:
### Description of change

Improve site and press notice

- Remove press sent at as there was no value in having this
- Prepopulate site notice displayed at and press notice published at with today's date if there is no saved value

### Story Link

https://trello.com/c/9H9vJbp2/2801-fill-the-frequent-tasks-more-quickly
https://trello.com/c/kFvahrcO/2735-clear-instruction-of-the-required-dates

### DIFF
--- a/app/views/planning_applications/site_notices/new.html.erb
+++ b/app/views/planning_applications/site_notices/new.html.erb
@@ -38,7 +38,7 @@
       <%= form.govuk_radio_button :required, false, label: { text: "No" }, data: { action: "change->show-hide-form#handleEvent" } %>
     <% end %>
 
-    <div class="govuk-!-display-none" id="site-notice-options">
+    <%= tag.div(id: "site-notice-options", class: class_names("govuk-!-display-none": @site_notice.errors.blank?)) do %>
       <div class="grey-box" id="email-site-notice">
         <%= form.govuk_radio_buttons_fieldset(
           :method,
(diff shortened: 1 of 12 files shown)


app/views/planning_applications/site_notices/new.html.erb
```diff
@@ -38,7 +38,7 @@
       <%= form.govuk_radio_button :required, false, label: { text: "No" }, data: { action: "change->show-hide-form#handleEvent" } %>
     <% end %>
 
-    <div class="govuk-!-display-none" id="site-notice-options">
+    <div id="site-notice-options" class="<%= ' govuk-!-display-none' if @site_notice.errors.blank? %>">
```
benbaumann95, line 41: The form was previously disappearing when there was a validation error. There was a bug where you had to reload the page and choose "Yes" again

pixeltrix (changes requested):
Needs to ignore the column as well, e.g.

``` ruby
self.ignored_columns += %w[press_sent_at]
```

That makes it safe to drop the column later.

app/views/planning_applications/site_notices/new.html.erb
pixeltrix: Use a Rails helper to build this tag rather than injecting ERB into attributes, e.g.

``` erb
<%= tag.div(id: "site-notice-options", class: class_names("govuk-!-display-none": @site_notice.errors.blank?)) do %>
...
<% end %>
```

This lessens the probability of an XSS bug arising

pixeltrix (approved):

PR #1697, opened 2024-04-11
Add reset_column_information

EGiataganas:
### Description of change

Run `reset_column_information`


benbaumann95 (approved):

db/migrate/20240404113623_add_decisions_to_application_types.rb
```diff
@@ -6,6 +6,8 @@ class ApplicationType < ActiveRecord::Base; end
   def change
     add_column :application_types, :decisions, :string, array: true
 
+    ApplicationType.reset_column_information
+
```
pixeltrix, line 10: For the record, this can be inside the `up_only` block but it should be fine here as well

pixeltrix (approved):

PR #1708, opened 2024-04-16
Admin can add informatives to LA

cdccollins:
### Description of change

- Admin can add global informatives to a local authority (like contacts/consultees) 

### Story Link

https://trello.com/c/xi6OzdnR/2788-configure-add-informatives-in-council-admin


app/models/local_authority.rb
```diff
@@ -6,6 +6,7 @@ class LocalAuthority < ApplicationRecord
     has_many :planning_applications
     has_many :constraints
     has_many :contacts
+    has_many :informatives, class_name: "LocalAuthorityInformative"
```
pixeltrix, line 9: If you namespaced the model as `LocalAuthority::Informative` then you wouldn't need this `:class_name` option and the table name is still `local_authoriy_informatives`.

app/models/local_authority_informative.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class LocalAuthorityInformative < ApplicationRecord
+  belongs_to :local_authority
+
+  validates :title, :text, presence: true
+
+  class << self
+    def all_informatives(query)
+      scope = all.order(:title)
```
pixeltrix, line 10: `all` is unnecessary here - you can just use `order`

engines/bops_admin/app/controllers/bops_admin/informatives_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+          end
+        end
+      end
+    end
+
+    private
+
+    def set_informatives
+      @pagy, @informatives = pagy(current_local_authority.informatives.all_informatives(search_param), items: 10)
+    end
```
benjamineskola, line 73: If this is only being used in one action I wonder if it would be clearer just to include it in the `index` method directly, rather than as a hook? But not vital.

benjamineskola (approved):

app/models/local_authority_informative.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    def search_param(query)
+      query.to_s
+        .scan(/[-\w]{3,}/)
+        .map { |word| word.gsub(/^-/, "!") }
+        .map { |word| word.gsub(/-$/, "") }
+        .map { |word| word.gsub(/.+/, "\\0:*") }
+        .join(" & ")
+    end
```
pixeltrix, line 34: Could probably extract this to a concern inside the `bops_core` engine.

PR #1709, opened 2024-04-16
Install govuk-components gem for future use

benjamineskola:
### Description of change

This will allow us to use the GOV.UK design system components in ViewComponent form, which will be preferable in some cases.

### Story Link

<https://trello.com/c/emIN8osq/2818-review-how-we-use-components>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,7 @@ gem "devise"
 gem "devise-two-factor"
 gem "faker", require: false
 gem "faraday", "~> 2", require: false
+gem "govuk-components", "~> 5"
 gem "govuk_design_system_formbuilder"
 gem "grover"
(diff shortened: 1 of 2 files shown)


pixeltrix (approved):

pixeltrix:
There may be some breaking changes on the form builder upgrade - I remember having to use `classes` as an option on one form because we were still on 3.x so might be worthing checking the changelog.

benjamineskola:
> There may be some breaking changes on the form builder upgrade - I remember having to use `classes` as an option on one form because we were still on 3.x so might be worthing checking the changelog.

They don't actually seem to have a changelog. I'll eyeball a few of the forms locally. I presume it wouldn't cause test failures but might make some forms display weirdly.

PR #1711, opened 2024-04-17
Use ActiveSupport::Notifications for auditing

pixeltrix:
Using Active Record to inject audit records into a database table is cumbersome because we have mix them in as part of the model code. By using ActiveSupport::Notifications we can gain flexibility by making the auditing declarable and having multiple listeners, e.g. logging all the events to something like LogStash for full auditing whilst also listening on specific events for things like posting Slack notifications when a planning application status changes.

### DIFF
--- a/engines/bops_config/app/controllers/bops_config/application_controller.rb
+++ b/engines/bops_config/app/controllers/bops_config/application_controller.rb
@@ -2,6 +2,16 @@
 
 module BopsConfig
   class ApplicationController < ActionController::Base
+    include BopsCore::AuditableController
+
+    self.audit_payload = -> {
+      {
+        engine: "bops_config",
+        params: request.path_parameters,
+        user: current_user.audit_attributes
+      }
+    }
+
     default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
     before_action :authenticate_user!
--- a/engines/bops_config/app/controllers/bops_config/application_types_controller.rb
+++ b/engines/bops_config/app/controllers/bops_config/application_types_controller.rb
@@ -6,6 +6,16 @@ module BopsConfig
     before_action :set_application_types, only: %i[index]
     before_action :set_application_type, only: %i[show edit update]
 
+    audit :create, :update,
+      unless: -> { @application_type.changed? },
+      payload: -> {
+        {
+          application_type: @application_type.audit_attributes,
+          changes: @application_type.audit_changes,
+          automated: false
+        }
+      }
+
     def index
       respond_to do |format|
         format.html
--- a/engines/bops_core/app/controllers/concerns/bops_core/auditable_controller.rb
+++ b/engines/bops_core/app/controllers/concerns/bops_core/auditable_controller.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module BopsCore
+  module AuditableController
+    extend ActiveSupport::Concern
+    include Auditable
+
+    AUDITABLE_ACTIONS = {}.tap do |actions|
+      actions.merge!(
+        "create" => "created",
+        "update" => "updated",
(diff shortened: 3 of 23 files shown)


engines/bops_config/app/controllers/bops_config/application_controller.rb
```diff
@@ -2,6 +2,16 @@
...
   class ApplicationController < ActionController::Base
+    include BopsCore::AuditableController
+
+    self.audit_payload = -> {
+      {
+        engine: "bops_config",
+        params: request.path_parameters,
+        user: current_user.audit_attributes
+      }
+    }
```
pixeltrix, line 13: This is a common payload that's added to all audit events - it's executed in the context of the controller so has access to instance variables and methods.

engines/bops_config/app/controllers/bops_config/application_types_controller.rb
```diff
@@ -6,6 +6,16 @@ class ApplicationTypesController < ApplicationController
...
 
+    audit :create, :update,
+      unless: -> { @application_type.changed? },
+      payload: -> {
+        {
+          application_type: @application_type.audit_attributes,
+          changes: @application_type.audit_changes,
+          automated: false
+        }
+      }
```
pixeltrix, line 17: This is the per-audit event payload that's merged with the shared payload - it's also executed in the context of the controller.

engines/bops_core/app/controllers/concerns/bops_core/auditable_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+    extend ActiveSupport::Concern
+
+    AUDITABLE_ACTIONS = {
+      "create" => "created",
+      "update" => "updated",
+      "destroy" => "destroyed"
+    }.freeze
+
+    included do
+      class_attribute :audit_payload, instance_writer: false, default: -> { {} }
```
pixeltrix, line 14: Looking to extract this to a shared concern and add support for calling `audit` anywhere

engines/bops_core/spec/controllers/bops_core/auditable_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      post :create
+    end.to yield_with_args(
+      "event.scope.bops_audit",
+      an_instance_of(Time),
+      an_instance_of(Time),
+      a_string_matching(/^[a-z0-9]{20}$/),
+      a_hash_including(foo: "bar")
+    )
+  end
```
pixeltrix, line 27: Looking to create a matcher that wraps all this up - the arity of the block passed to subscribe affects the behaviour and in this case the RSpec `yield_with_args` matcher has a `*args` arity so the event object gets splatted to individual args.

engines/bops_core/spec/jobs/bops_core/auditable_job_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require "bops_core_helper"
+
+RSpec.describe BopsCore::AuditableJob, type: :job do
+  let(:job) do
+    Class.new(ActiveJob::Base) do
+      include BopsCore::AuditableJob
+
+      audit event: "event", scope: "scope", payload: -> { {foo: "bar"} }
```
pixeltrix, line 10: Might alter this call signature to: 

``` ruby
audit "event.scope", -> { {foo: "bar"} }
```

The convention in AS:N is to do things right to left so a typical event would be `status_updated.application.bops_audit` - the last part being added automatically

benjamineskola (approved):
I like that it ends up being a declarative thing to add an audit, rather than remembering to call a method in the right place.

EGiataganas (approved):

benbaumann95 (approved):

PR #1716, opened 2024-04-18
Improve pre-commencement conditions feature 

benbaumann95:
### Description of change

Improve pre-commencement conditions feature

- Add one condition record at a time rather than multiple forms
- Do not send request to applicant until officer has confirmed all pending requests
- Add ability to remove request before the request has been sent
- Tidy up code for conditions creating a separate controller for pre-commencement conditions
- Only show the notified requests to applicant

### Story Link

https://trello.com/c/8iYydyMp/2778-improve-add-pre-commencement-conditions-when-multiple-conditions-are-added


app/components/status_tags/base_component.rb
```diff
@@ -40,7 +41,7 @@ def colour_class
     end
 
     def task_list?
-      true
+      task_list != false
```
benjamineskola, line 44: I think this might return true if task_list is nil, which probably isn't what we want? Maybe better to check if it's true:

```suggestion
      task_list == true
```

cdccollins (approved):

app/components/status_tags/base_component.rb
benbaumann95: If it's nil or true we want this css class `app-task-list__task-tag`.
But I agree this looks a bit confusing so i'll try to rework this

PR #1723, opened 2024-04-19
Fede secret design

EGiataganas:
### Description of change

Improvements on validation task
- Make Red line boundary change pages consistent
- Change heading to medium
- Add document accordion component on description changes
- Make radios inline
- Validating document redirects back to supply documents path
- Add heading on validation documents
- Fee validation improvements
- Add heading on validation cil
- Add documents on environment impact assessment
- Change form's label to medium
- Switch it to h2
- Redesign document tags to be same as config

### Story Link

https://trello.com/c/bjjdP7Wx

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -137,7 +137,7 @@ class DocumentsController < AuthenticationController
 
   def redirect_url
     if @validate_document
-      planning_application_validation_tasks_path(@planning_application)
+      supply_documents_planning_application_path(@planning_application)
     elsif session[:return_to]
       return_to_session
     else
--- a/app/models/fee_calculation.rb
+++ b/app/models/fee_calculation.rb
@@ -53,6 +53,10 @@ class FeeCalculation < ApplicationRecord
     end
   end
 
+  def discount
+    total_fee.to_d - payable_fee.to_d
+  end
+
   private
 
   def audit_updated!
--- a/app/views/documents/form_partials/_tags.html.erb
+++ b/app/views/documents/form_partials/_tags.html.erb
@@ -8,13 +8,6 @@
       </h2>
     </legend>
 
-    <% if form.object.errors[:tags].any? %>
-      <% form.object.errors[:tags].each do |error| %>
-        <span id="status-error" class="govuk-error-message">
-          <span class="govuk-visually-hidden">Error:</span><%= error %></span>
-      <% end %>
-    <% end %>
-
     <%= govuk_tabs do |tabs| %>
       <% document_tags.tag_groups.each do |group| %>
         <% tabs.with_tab(label: t(".groups.#{group}"), id: "#{group}-tags" ) do %>
--- a/app/views/documents/form_partials/_tags.html.erb
+++ b/app/views/documents/form_partials/_tags.html.erb
@@ -1,11 +1,11 @@
 <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
 
-<div class="govuk-form-group <%= form.object.errors[:tags].any? ? 'govuk-form-group--error' : '' %>" data-controller="show-hide">
+<div class="govuk-form-group">
   <fieldset class="govuk-fieldset govuk-!-margin-top-4">
     <legend class="govuk-fieldset__legend">
-      <h3 class="govuk-heading-s">
+      <h2 class="govuk-heading-m">
         What does the document contain?
-      </h3>
+      </h2>
     </legend>
 
     <% if form.object.errors[:tags].any? %>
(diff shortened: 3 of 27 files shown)


app/views/documents/form_partials/_tags.html.erb
```diff
@@ -1,6 +1,6 @@
 <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
 
-<div class="govuk-form-group <%= form.object.errors[:tags].any? ? 'govuk-form-group--error' : '' %>" data-controller="show-hide">
+<div class="govuk-form-group <%= form.object.errors[:tags].any? ? 'govuk-form-group--error' : '' %>">
```
pixeltrix, line 3: Is this needed since we're now using `govuk_check_boxes_fieldset` inside the tab which will add the error styling

app/views/documents/form_partials/_validate_document.html.erb
```diff
@@ -6,9 +6,8 @@
 
   <%= hidden_field_tag(:validate, "yes") %>
 
-  <%= form.govuk_radio_buttons_fieldset(:validated, legend: { size: 's', text: 'Is the document valid?' }) do %>
+  <%= form.govuk_radio_buttons_fieldset(:validated, legend: { size: 's', text: 'Is the document valid?' }, class: "govuk-radios govuk-radios--inline") do %>
```
pixeltrix, line 9: Remove `govuk-radios` and use `inline: true` again

app/controllers/documents_controller.rb
```diff
@@ -137,7 +137,8 @@ def replacement_document_validation_request
 
   def redirect_url
     if @validate_document
-      planning_application_validation_tasks_path(@planning_application)
+
```
pixeltrix, line 140: Extra line here - would've thought rubocop would capture this

app/models/fee_calculation.rb
```diff
@@ -53,6 +53,10 @@ def from_planning_portal_data(pp_data)
     end
   end
 
+  def discount
+    total_fee.to_d - payable_fee.to_d
```
pixeltrix, line 57: We're converting `nil` to `0.0` by doing this - is that correct?

app/views/documents/form_partials/_tags.html.erb
```diff
@@ -15,57 +15,39 @@
       <% end %>
     <% end %>
```
pixeltrix, line 16: Are there likely to be multiple errors here? Either way we can just loop over an empty array without the `any?` check. If there are possibly multiple errors should the wrapper tag be a `<div>` instead of a `<span>` so that they're not running together.

``` erb
<% form.object.errors[:tags].each do |error| %>
  <span id="status-error" class="govuk-error-message">
    <span class="govuk-visually-hidden">Error:</span>
    <%= error %>
  </span>
<% end %>
```

app/views/planning_applications/validation/additional_document_validation_requests/_form.html.erb
```diff
@@ -14,11 +14,11 @@
       ) %>
 
       <%= form.govuk_text_field :document_request_type,
-       label: { text: 'Please specify the new document type:', size: 's' },
+       label: { size: 'm', class: "govuk-label govuk-label--m", text: 'Please specify the new document type:' },
```
pixeltrix, line 17: Doesn't the class get applied automatically here?

app/views/planning_applications/validation/additional_document_validation_requests/_form.html.erb
```diff
@@ -14,11 +14,11 @@
...
        hint: { text: 'Eg. Existing floor plans' } %>
 
       <%= form.govuk_text_area :reason,
-       label: { text: 'Please specify the reason you have requested this document?', size: 's'},
+       label: { size: 'm', class: "govuk-label govuk-label--m", text: 'Please specify the reason you have requested this document?'},
```
pixeltrix, line 21: Same here - the `govuk-label` and `govuk-label--m` should be added automatically

app/views/planning_applications/validation/description_change_validation_requests/_new.html.erb
```diff
@@ -58,7 +58,7 @@
         <% end %>
         <%= form.hidden_field(:type, value: "DescriptionChangeValidationRequest") %>
         <%= form.govuk_text_area :proposed_description,
-          label: { text: 'Suggest a new application description', class: 'govuk-label govuk-label--s'},
+          label: { size: 'm', class: "govuk-label govuk-label--m", text: 'Suggest a new application description' },
```
pixeltrix, line 61: Same here with class

pixeltrix (changes requested):

app/models/fee_calculation.rb
EGiataganas: I think so, just moved that logic from [this](https://github.com/unboxed/bops/pull/1723/files#diff-8ea26bef723ba324e88ffe97e985288c676e6e99dbce657d6361a28a3875a379L59) view to a model method.

PR #1728, opened 2024-04-22
Further content updates to planning guides

cdccollins:
### Description of change

Few extra content updates to the planning guides

### DIFF
--- a/app/views/public/planning_guides/drawings.html.erb
+++ b/app/views/public/planning_guides/drawings.html.erb
@@ -22,8 +22,7 @@
         <li>have a unique drawing reference number</li>
       </ul>
       <p class="govuk-body">
-        Your application may not be marked as valid if the drawings do not clearly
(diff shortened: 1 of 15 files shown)


app/views/public/planning_guides/drawings.html.erb
```diff
@@ -22,8 +22,7 @@
         <li>have a unique drawing reference number</li>
       </ul>
       <p class="govuk-body">
-        Your application may not be marked as valid if the drawings do not clearly
-        and accurately describe the site and the proposed changes.
+        Your application may not be valid if the drawings do not clearly and accurately describe the site and the proposed changes. Upload scans of physical documents rather than photos.
```
benjamineskola, line 25: Cam we extract this string into the translations file and reuse it in all of the places it appears?

pixeltrix (approved):
This content could probably live inside a bops_public engine eventually

PR #1731, opened 2024-04-23
Lint erb files with erb_lint

benjamineskola:
### Description of change

Ensure consistency especially of ruby embedded in erb files using erb_lint to run checks including rubocop


cdccollins (approved):
Going to assume this is all sensible!

PR #1733, opened 2024-04-23
Use address finder creating planning applications

cdccollins:
### Description of change

- When manually creating planning application, use address finder with OS Places API
- Fill in form using result and get latitude and longitude

### Story Link

https://trello.com/c/LSH77cLG/2831-spike-use-address-to-set-the-position-of-the-red-line-map-when-setting-up-cases-manually


benjamineskola (approved):

app/models/planning_application.rb
```diff
@@ -633,6 +634,16 @@ def longitude
     super || lonlat.try(:x)
   end
 
+  def update_lonlat
+    return unless longitude.present? && latitude.present?
+
+    update!(lonlat: factory.point(longitude, latitude))
+  end
```
pixeltrix, line 641: Can't we do this in a `before_create` rather than running an `UPDATE` right after the `INSERT` ?

pixeltrix (changes requested):

pixeltrix (approved):

PR #1736, opened 2024-04-23
Use govuk_inset_text component

benbaumann95:
### Description of change

Use govuk_inset_text component
- https://govuk-components.netlify.app/components/inset-text/

### DIFF
--- a/app/components/accordion_sections/pre_assessment_outcome_component.html.erb
+++ b/app/components/accordion_sections/pre_assessment_outcome_component.html.erb
@@ -31,11 +31,8 @@
     <p class="govuk-body">
       <%= t(".the_applicant_believes").html_safe %>
     </p>
-    <div class="govuk-inset-text">
-      <p class="govuk-body">
-        <%= planning_application.result_override %>
-      </p>
-    </div>
+    
+    <%= govuk_inset_text(text: planning_application.result_override) %>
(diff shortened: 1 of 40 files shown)


pixeltrix:
Not sure about these - we're calling ruby code instead of just returning a static string. I think using the helper makes sense where we're doing other dynamic stuff like:

``` erb
<div class="govuk-inset-text" id="<%= dom_id(@document) %>">
```
which would be replaced with:

``` erb
<%= govuk_inset_text(id: dom_id(@document)) do %>
```

but replacing `<div class="govuk-inset-text">` with `<%= govuk_inset_text %>` doesn't seem to offer anything visual improvement.

benbaumann95:
> Not sure about these - we're calling ruby code instead of just returning a static string. I think using the helper makes sense where we're doing other dynamic stuff like:
> 
> ```
> <div class="govuk-inset-text" id="<%= dom_id(@document) %>">
> ```
> 
> which would be replaced with:
> 
> ```
> <%= govuk_inset_text(id: dom_id(@document)) do %>
> ```
> 
> but replacing `<div class="govuk-inset-text">` with `<%= govuk_inset_text %>` doesn't seem to offer anything visual improvement.

Fair enough, I'll close for now if it's not adding value.

PR #1742, opened 2024-04-25
Make use of buttons and links consistent

cdccollins:
### Description of change

Use buttons and links consistently across BOPS
- validation
- consultation
- assessment

(Review coming in separate PR to keep this one small-ish)

### Story Link

https://trello.com/c/b1A1AKI2/2855-pattern-main-navigation-buttons

### DIFF
--- a/app/views/planning_applications/validation/cil_liability/_form.html.erb
+++ b/app/views/planning_applications/validation/cil_liability/_form.html.erb
@@ -16,5 +16,8 @@
     <% end %>
   <% end %>
 
-  <%= render(partial: "shared/submit_buttons", locals: {form:}) %>
+  <div class="govuk-button-group">
+    <%= form.submit "Save", class: "govuk-button", data: {module: "govuk-button"} %>
+    <%= link_to "Back", planning_application_validation_tasks_path(@planning_application), class: "govuk-button govuk-button--secondary" %>
+  </div>
 <% end %>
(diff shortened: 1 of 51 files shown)


app/views/planning_applications/validation/cil_liability/_form.html.erb
```diff
@@ -16,5 +16,8 @@
     <% end %>
   <% end %>
 
-  <%= render(partial: "shared/submit_buttons", locals: {form:}) %>
```
benjamineskola, line 19: It's a shame that we can't keep using the partial when we don't want the 'save and come back later' button.

I wonder if we could turn this into a view component with options for which buttons to show. (Not for this PR though.)

app/views/shared/_validation_request_show_actions.html.erb
```diff
@@ -1,27 +1,34 @@
...
     <% end %>
 
     <% if planning_application.not_started? %>
       <% if local_assigns.fetch(:include_edit, true) %>
         <%= edit_request_url(
               planning_application,
               validation_request,
-              classname: "govuk-button govuk-button--secondary"
+              classname: "govuk-body govuk-link"
             ) %>
```
benjamineskola, line 30: I might be missing something obvious here but is this actually creating a link, or just including the URL on the page?

benjamineskola (approved):

app/views/shared/_validation_request_show_actions.html.erb
cdccollins: Yeah it's kind of weird, for some reason the `link_to` is in the helper rather than here 🤷‍♀️

PR #1748, opened 2024-04-26
Rename methods for clarity

benjamineskola:
### Description of change

- Rename confusing methods that behave like `link_to` but are named like ordinary url helpers
- Use govuk_link_to helper

### DIFF
--- a/app/helpers/validation_request_helper.rb
+++ b/app/helpers/validation_request_helper.rb
@@ -3,14 +3,13 @@
 module ValidationRequestHelper
   def edit_request_link(planning_application, validation_request, classname: nil)
     govuk_link_to "Edit request",
-      send(:edit_planning_application_validation_validation_request_path, planning_application,
-        validation_request), class: classname
+      edit_planning_application_validation_validation_request_path(planning_application, validation_request),
+      class: classname
   end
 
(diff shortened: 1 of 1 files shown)


cdccollins (approved):

app/helpers/validation_request_helper.rb
```diff
@@ -1,14 +1,14 @@
...
+    govuk_link_to "Edit request",
       send(:edit_planning_application_validation_validation_request_path, planning_application,
         validation_request), class: classname
   end
 
-  def delete_confirmation_request_url(planning_application, validation_request, classname: nil)
-    link_to "Delete request",
+  def delete_confirmation_request_link(planning_application, validation_request, classname: nil)
+    govuk_link_to "Delete request",
       send(:planning_application_validation_validation_request_path,
```
pixeltrix, line 12: Just out of interest - why the `send` here? We should be able to call the url helper directly

pixeltrix (approved):

EGiataganas (approved):

app/helpers/validation_request_helper.rb
benjamineskola: It looks like we used to use string interpolation to determine which method to call, and when we stopped needing to do that the `send` was never cleaned up.

PR #1755, opened 2024-04-29
Fix html in notification banner

cdccollins:
### Description of change

Linter added a close tag when it shouldn't have


pixeltrix (changes requested):
We should use the `govuk_notification_banner` component:
https://govuk-components.netlify.app/components/notification-banner/#output-html-notification-banner-with-a-block

pixeltrix (approved):

PR #1756, opened 2024-04-29
Use a matrix test for linters to reduce boilerplate

benjamineskola:
### Description of change

The way we invoke linters has become increasingly convoluted by passing variables through from one workflow to another in order to disable all but one of the linter steps on each occurrence. In practice all we need is a single step that varies, running a different command each time, and a few shared setup steps — a matrix seems perfect for this.


benbaumann95 (approved):
:thu

EGiataganas (approved):

PR #1760, opened 2024-04-30
Record whether regulations 3 and 4 apply for reporting purposes

cdccollins:
### Description of change

Officer can record whether Regulation 3 and Regulation 4 applies to a planning application for reporting purposes

### Story Link

https://trello.com/c/05RK6NgY/2717-able-to-record-regulation-3-council-own-land-and-4-land-owned-by-the-council-applications


app/controllers/planning_applications/validation/reporting_types_controller.rb
```diff
@@ -30,7 +30,7 @@ def update
       private
 
       def reporting_type_params
-        params[:planning_application] ? params.require(:planning_application).permit(:reporting_type) : params.permit(:reporting_type)
+        params[:planning_application] ? params.require(:planning_application).permit(:reporting_type, :regulation_3, :regulation_4) : params.permit(:reporting_type, :regulation_3, :regulation_4)
```
benjamineskola, line 33: This might be more readable as an if statement.

```suggestion
        if params[:planning_application]
          params.require(:planning_application).permit(:reporting_type, :regulation_3, :regulation_4)
        else
          params.permit(:reporting_type, :regulation_3, :regulation_4)
        end
```

Or actually, since only the beginning of the chain changes:

```suggestion
        (params[:planning_application] ? params.require(:planning_application) : params).permit(:reporting_type, :regulation_3, :regulation_4)
```

app/models/planning_application.rb
```diff
@@ -1075,6 +1076,12 @@ def changed_to_prior_approval?
     application_type_changed? && application_type.name == "prior_approval"
   end
 
+  def ensure_reporting_is_correct
+    return if regulation_3
+
+    self.regulation_4 = false if !regulation_3
```
benjamineskola, line 1082: It might not be necessary to have both these lines, I think? Don't need to check whether `regulation_3` is false because we've already returned if it's true.

benjamineskola (approved):

PR #1761, opened 2024-04-30
Apply new design colour palette on status task list

EGiataganas:
### Description of change

Apply new guidelines on status task list

### Story Link

https://trello.com/c/IUfOWTmb

### Screenshots

Guidelines

### DIFF
--- a/app/components/status_tags/base_component.html.erb
+++ b/app/components/status_tags/base_component.html.erb
@@ -1,1 +1,9 @@
-<%= tag.strong(t("status_tag_component.#{status}"), class: html_classes) %>
+<% if %i[review_complete complete awaiting_response awaiting_responses checked granted valid posted supportive approved auto_approved granted_legal_agreement].include?(status.to_sym) %>
+  <div class="app-task-list__task-tag">
+    <div class="govuk-task-list__status" id="<%= t("status_tag_component.#{status}") %>">
+      <%= t("status_tag_component.#{status}") %>
+    </div>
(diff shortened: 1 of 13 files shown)


app/components/status_tags/base_component.html.erb
```diff
@@ -1 +1,9 @@
-<%= tag.strong(t("status_tag_component.#{status}"), class: html_classes) %>
+<% if %i[review_complete complete awaiting_response awaiting_responses checked granted valid posted supportive approved auto_approved granted_legal_agreement].include?(status.to_sym) %>
+  <div class="app-task-list__task-tag">
+    <div class="govuk-task-list__status" id="<%= t("status_tag_component.#{status}") %>">
+      <%= t("status_tag_component.#{status}") %>
+    </div>
+  </div>
+<% else %>
+  <%= tag.strong(t("status_tag_component.#{status}"), class: html_classes) %>
+<% end %>
```
benjamineskola, line 9: What about just using the `govuk_tag` helper here, to make use of the upstream code?

```suggestion
<%= govuk_tag(text: t("status_tag_component.#{status}"), class: html_classes) %>
```

And in fact instead of having a separate `era` file, I think it could be defined as a `call` method on the `BaseComponent` class.

EGiataganas: tried it but many specs have failed, so I've reverted back

benjamineskola (approved):

PR #1765, opened 2024-05-01
Improve reliability of new planning application dependency job

cdccollins:
### Description of change

If constraint is not successfully fetched, record the fact and stop the job retrying

### Story Link

https://trello.com/c/luYMXBVQ/2913-improve-reliability-of-new-planning-application-dependency-job

### DIFF
--- a/db/migrate/20240501134003_add_data_fetched_to_planning_application_constraint.rb
+++ b/db/migrate/20240501134003_add_data_fetched_to_planning_application_constraint.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class AddDataFetchedToPlanningApplicationConstraint < ActiveRecord::Migration[7.1]
+  def change
+    add_column :planning_application_constraints, :status, :string, default: "pending", null: false
+
+    PlanningApplicationConstraint.find_each do |constraint|
+      status = constraint.data.blank? ? "pending" : "success"
+
+      constraint.update!(status:)
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -613,6 +613,7 @@ ActiveRecord::Schema[7.1].define(version: 2024_05_03_104253) do
     t.string "identified_by", null: false
     t.bigint "consultee_id"
     t.boolean "consultation_required", default: true, null: false
+    t.string "status", default: "pending", null: false
     t.index ["constraint_id"], name: "ix_planning_application_constraints_on_constraint_id"
     t.index ["consultee_id"], name: "ix_planning_application_constraints_on_consultee_id"
     t.index ["planning_application_constraints_query_id"], name: "ix_planning_application_constraints_on_planning_application_con"
--- a/engines/bops_api/app/jobs/bops_api/planning_application_dependency_job.rb
+++ b/engines/bops_api/app/jobs/bops_api/planning_application_dependency_job.rb
@@ -67,7 +67,12 @@ module BopsApi
         next unless designation.fetch(:intersects)
 
         if (constraint = ::Constraint.for_type(designation.fetch(:value)))
-          planning_application_constraint = create_planning_application_constraint(planning_application, designation, constraint)
+          planning_application_constraint = if planning_application.planning_application_constraints.find_by(constraint_id: constraint.id)&.pending?
+            planning_application.planning_application_constraints.find_by(constraint: constraint)
+          else
+            create_planning_application_constraint(planning_application, designation, constraint)
+          end
+
           entities = designation.fetch(:entities, [])
 
           if entities.present?
@@ -151,7 +156,13 @@ module BopsApi
     end
 
     def fetch_constraint_entities(planning_application_constraint, entities)
-      planning_application_constraint.update!(data: fetch_entities(entities))
+      planning_application_constraint.update!(data: fetch_entities(entities), status: "success")
+    rescue BopsApi::Errors::EntityNotFoundError
+      planning_application_constraint.update!(data: [], status: "not_found")
+    rescue BopsApi::Errors::EntityRemovedError
+      planning_application_constraint.update!(data: [], status: "removed")
+    rescue BopsApi::Errors::EntityFetchFailedError
+      planning_application_constraint.update!(data: [], status: "failed")
     end
 
     def fetch_entities(entities)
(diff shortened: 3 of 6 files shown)


benjamineskola (approved):

engines/bops_api/app/jobs/bops_api/planning_application_dependency_job.rb
```diff
@@ -151,7 +153,27 @@ def process_immunity_details(planning_application)
...
+        status: any_error_responses?(fetched_entities) ? planning_application_status(fetched_entities) : "success"
+      )
+    end
+
+    def any_error_responses?(fetched_entities)
+      fetched_entities.empty? || fetched_entities.include?("failed") ||
+        fetched_entities.include?("removed") || fetched_entities.include?("not_found")
+    end
+
+    def planning_application_status(fetched_entities)
```
benjamineskola, line 169: I think if this had an `else "success"` at the end, then line 160 could just be `status: planning_application_status(fetched_entities)`.

benjamineskola (approved):

db/migrate/20240501134003_add_data_fetched_to_planning_application_constraint.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AddDataFetchedToPlanningApplicationConstraint < ActiveRecord::Migration[7.1]
+  def change
+    add_column :planning_application_constraints, :status, :string, default: "success", null: false
+
+    PlanningApplicationConstraint.find_each do |constraint|
+      status = constraint.data.nil? ? "failed" : "success"
```
pixeltrix, line 8: Not sure this is a valid check - I think it's an empty array by default:

https://github.com/unboxed/bops/blob/f70c5f465df1c87c213992ddca0c456dc376b947/engines/bops_api/app/jobs/bops_api/planning_application_dependency_job.rb#L55-L63

Maybe if it's populated mark it as `success` but `pending` otherwise?

db/schema.rb
```diff
@@ -613,6 +613,7 @@
     t.string "identified_by", null: false
     t.bigint "consultee_id"
     t.boolean "consultation_required", default: true, null: false
+    t.string "status", default: "success", null: false
```
pixeltrix, line 616: I think we need a default status of `pending` so that we can re-run the job and not fetch them again.

engines/bops_api/app/jobs/bops_api/planning_application_dependency_job.rb
```diff
@@ -67,6 +67,8 @@ def process_planning_designations(planning_application)
         next unless designation.fetch(:intersects)
 
         if (constraint = ::Constraint.for_type(designation.fetch(:value)))
+          next if planning_application.planning_application_constraints.find_by(constraint_id: constraint.id)
+
```
pixeltrix, line 71: With the `pending` status we can check to see if we need to run it again

engines/bops_api/app/jobs/bops_api/planning_application_dependency_job.rb
```diff
@@ -176,7 +200,11 @@ def fetch_entity(source)
 
       response.body
     rescue Faraday::ResourceNotFound
-      nil
+      "not_found"
+    rescue Faraday::ServerError, BopsApi::Errors::InvalidEntityResponseError
+      "failed"
+    rescue Faraday::ClientError => error
+      (error.response_status == 410) ? "removed" : "failed"
```
pixeltrix, line 207: Looking at the planning data server a 404 request returns a HTML page but does have the correct status. I think we should remove the `faraday.response :raise_error` line and check the status of the response. We should also rescue the generic `Faraday::Error` and re-raise with our own errors as returning these statuses means knowledge of how the constraint model works is leaking into our method to fetch entities. I suspect it would look something like this:

``` ruby
def fetch_entity(source)
  uri = normalise_url(source)

  return unless uri

  connection = Faraday.new(uri.origin) do |faraday|
    faraday.response :json, content_type: /\bjson$/
  end

  response = connection.get(uri.request_uri)
  
  case response.status
  when 200
    if response.body.is_a?(Hash)
      response.body
    else
      raise BopsApi::Errors::InvalidEntityResponseError, "Request for entity #{uri} returned a non-JSON response"
    end
  when 204
    nil
  when 404
    raise BopsApi::Errors::EntityNotFoundError, "Entity #{uri} was not found"
  when 410
    raise BopsApi::Errors::EntityRemovedError, "Entity #{uri} has been removed"
  else
    raise BopsApi::Errors::EntityFetchFailedError, "Entity #{uri} could not be fetched"
  end
rescue Faraday::Error
  raise BopsApi::Errors::EntityFetchFailedError, "Entity #{uri} could not be fetched"
end
```

You'd then update the `fetch_constraint_entities` method to update the status like this:

``` ruby
def fetch_constraint_entities(planning_application_constraint, entities)
  planning_application_constraint.update!(data: fetch_entities(entities), status: "success")
rescue BopsApi::Errors::EntityNotFoundError
  planning_application_constraint.update!(data: [], status: "not_found")
rescue BopsApi::Errors::EntityRemovedError
  planning_application_constraint.update!(data: [], status: "removed")
rescue BopsApi::Errors::EntityFetchFailedError
  planning_application_constraint.update!(data: [], status: "failed")
end
```

You'd also need to add the additional error classes and make `BopsApi::Errors::InvalidEntityResponseError` inherit from `BopsApi::Errors::EntityFetchFailedError` so the final `rescue` clause above captures the general case and the specific non-JSON response case.

engines/bops_api/app/jobs/bops_api/planning_application_dependency_job.rb
```diff
@@ -151,7 +153,29 @@ def process_immunity_details(planning_application)
...
+    def planning_application_constraint_status(fetched_entities)
+      if fetched_entities.include?("removed")
+        "removed"
+      elsif fetched_entities.include?("not_found")
+        "not_found"
+      elsif fetched_entities.include?("failed")
+        "failed"
+      else
+        "success"
+      end
```
pixeltrix, line 178: This set of changes goes away if you implement what I outline below

pixeltrix (changes requested):

PR #1766, opened 2024-05-01
Fix task list status tags

benjamineskola:
### Description of change

Removing the `task_list` parameter in #1763 caused a rendering error, but strictly speaking the classes it added aren't part of the status tag, they're part of the task list. Added them to the base TaskListItem and then additionally to any uses of StatusTag components in a hardcoded task list.

### DIFF
--- a/app/views/planning_applications/assessment/tasks/_complete_assessment.html.erb
+++ b/app/views/planning_applications/assessment/tasks/_complete_assessment.html.erb
@@ -17,11 +17,13 @@
               class: "govuk-link"
             ) %>
       </span>
+      <div class="govuk-task-list__status app-task-list__task-tag">
       <%= render(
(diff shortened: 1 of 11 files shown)


app/views/planning_applications/assessment/tasks/_complete_assessment.html.erb
```diff
@@ -17,11 +17,13 @@
               class: "govuk-link"
             ) %>
       </span>
+      <div class="govuk-task-list__status app-task-list__task-tag">
       <%= render(
             StatusTags::AssessRecommendationComponent.new(
               planning_application: @planning_application
             )
           ) %>
+      </div>
```
benbaumann95, line 26: Thought erblint would add an indent here but no biggie

benbaumann95 (approved):
Definitely a better approach 👍

PR #1767, opened 2024-05-01
Use application description as default type

benbaumann95:
### Description of change

If the mapping does not exist for translation `I18n.t(application_type.name, scope: "application_types"` then default back to the application description which is mapped in `odp.yml`

### Story Link

https://trello.com/c/XbI6XNdf/2930-squiffy-text-in-application-header

### DIFF
--- a/app/models/concerns/planning_application_decorator.rb
+++ b/app/models/concerns/planning_application_decorator.rb
@@ -32,7 +32,7 @@ module PlanningApplicationDecorator
   end
 
   def type
-    I18n.t(application_type.name, scope: "application_types")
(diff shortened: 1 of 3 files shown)


cdccollins (approved):
We also discussed in planning whether we could move this into the BOPS global admin somehow. But I don't know the ins and outs to know whether that's possible/ a good idea?

benbaumann95:
> We also discussed in planning whether we could move this into the BOPS global admin somehow. But I don't know the ins and outs to know whether that's possible/ a good idea?

Yeah I did think that too and mentioned in the card.
I guess could be added as part of the application type but since we have that big mapping in the yml file anyway may as well use this for now

PR #1769, opened 2024-05-02
Convert validation tasks to govuk-components task list

benjamineskola:
### Description of change

Use the govuk-components task list view component on the validation tasks page.

The component doesn't accept arbitrary input to the list: as far as I can tell, only its own methods can be used to add items, which means we can't use our existing TaskListItem components. However, in most cases they're simple enough that it doesn't seem to matter a great deal.

(We _can_ use a component as input to the title or status parameters, so I've done so for StatusTag subclasses; but the TaskListItem components generate the whole item, including its container, not just the left-hand content as would be needed.)

### Story Link

https://trello.com/c/QUiuHLZi/2918-use-tasklist-view-component

### Screenshots

Screenshot 2024-05-07 at 17 00 26


app/views/planning_applications/validation/tasks/_application_details.html.erb
```diff
@@ -2,76 +2,75 @@
...
-      <% else %>
-        <%= render(
-              TaskListItems::Validating::LegislationComponent.new(
-                planning_application: @planning_application
+        task_list.with_item(html_attributes: {id: "check-red-line-boundary"}) do |item|
+          task = Validation::CheckRedLineBoundaryTask.new(@planning_application)
+          item.with_title(text: task.task_list_link_text,
+            href: task.task_list_link)
+          item.with_status(text: render(StatusTags::BaseComponent.new(status: task.task_list_status)))
+        end
```
benjamineskola, line 19: I would have liked to move this block inside the Task class, so that you could pass in the task list component and have the task add itself as an item with all the relevant fields, but `render` isn't available from inside the task.

app/views/planning_applications/validation/tasks/_application_details.html.erb
```diff
@@ -2,76 +2,43 @@
...
-    <% end %>
-    <li id="development-type-for-reporting-task" class="app-task-list__item">
-      <span class="app-task-list__task-name">
-        <%= govuk_link_to "Add reporting details", edit_planning_application_validation_reporting_type_path(@planning_application) %>
-      </span>
+  <%= govuk_task_list(id_prefix: "application-details-tasks") do |task_list|
+        task_list.with_item(html_attributes: {id: "draw-red-line-boundary"}) do |item|
+          task = Validation::DrawRedLineBoundaryTask.new(@planning_application)
+          item.with_title(text: task.task_list_link_text, href: task.task_list_link)
+          item.with_status(text: render(task.task_list_status_component))
```
benjamineskola, line 9: I'm not certain about this. On the one hand if we want this to be automatically generated from a list of workflow tasks, then all the task-specific logic needs to be inside the task class; on the other hand, does that mix logic and presentation too much?

If we wanted to do that, then would be possible to move all of the calls to `StatusTags::BaseComponent` into `WorkflowTask` and then just override it in the handful of tasks that need a specific status component rather than the default.

app/views/planning_applications/validation/tasks/_application_details.html.erb
benjamineskola: I've addressed this for now by making it a partial.

PR #1776, opened 2024-05-03
Remove use of `type_with_status`

pixeltrix:
The application type description is more useful and it won't fit on the dashboard.

## Dashboard

image

## Planning application header

image


benbaumann95 (approved):

app/models/concerns/planning_application_decorator.rb
```diff
@@ -35,7 +35,7 @@ def type
     I18n.t(application_type.name, scope: "application_types", default: application_type.description)
```
benbaumann95, line 35: Can this and the relevant yml entries be removed if no longer using?

pixeltrix: Possibly - will check

PR #1778, opened 2024-05-07
Add more spacing between tasks lists

benjamineskola:
This was inconsistent and also less than the previous task lists.

Screenshot 2024-05-07 at 15 46 12


cdccollins (approved):
thank you!

PR #1789, opened 2024-05-09
Avoid collisions of name parameter on test matrix run

benjamineskola:
If more than one test group in the matrix fails, all but the first will fail to save the screenshots because they all try to upload files to the same location. Distinguish them by name of the test group instead.


pixeltrix (approved):

pixeltrix:
Do we have the same issue with logs?

benjamineskola:
> Do we have the same issue with logs?

Oh yes, good point. I hadn't noticed that one before but I'll add it now.

PR #1796, opened 2024-05-13
Create public planning applications controller with search action for API

benjamineskola:
### Description of change

Creating an unauthenticated endpoint for public search. Search is by complete or partial reference (`LIKE`), falling back to full-text search of description.

### Story Link

https://trello.com/c/wBhBhcQ1/2932-priority-1-the-ability-to-search-based-on-complete-or-partial-application-reference-referenceinfull

### Known issues

Future work:
 
1. more fields in output
2. more fields in full text search
3. refactor search methods to avoid duplication with frontend search.

### DIFF
--- a/engines/bops_api/app/controllers/bops_api/v2/public/planning_applications_controller.rb
+++ b/engines/bops_api/app/controllers/bops_api/v2/public/planning_applications_controller.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module BopsApi
+  module V2
+    module Public
+      class PlanningApplicationsController < PublicController
+        def search
+          @pagy, @planning_applications = search_service.call
+
+          respond_to do |format|
+            format.json
+          end
+        end
+
+        private
+
+        def planning_applications_scope
+          @local_authority.planning_applications.where(make_public: true)
+        end
+
+        def search_params
+          params.permit(:page, :maxresults, :q)
+        end
+
+        def search_service(scope = planning_applications_scope.by_created_at_desc)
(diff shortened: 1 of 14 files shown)


engines/bops_api/app/controllers/bops_api/v2/public/planning_applications_controller.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module BopsApi
+  module V2
+    module Public
+      class PlanningApplicationsController < PublicController
```
benjamineskola, line 6: On reflection, I wonder if we should just inherit from `ApplicationController` here and only add a `PublicController` if we find some code needs to be shared between controllers?

On the other hand, perhaps it's a nice way to indicate intent to make something public: we'll know `ApplicationController` is a mistake and everything should always be explicitly `Public` or `Authenticated`.

engines/bops_api/lib/bops_api/pagination.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
benbaumann95: Won't be hard to move there if needed

engines/bops_api/app/controllers/bops_api/v2/public/planning_applications_controller.rb
benbaumann95: Yeah it's not needed but inclined to just leave it in for now as I agree with your second point. Although I am struggling to think what would be explicit to this controller

benbaumann95 (approved):

PR #1802, opened 2024-05-15
Refactor position controllers and extract to a concern

pixeltrix:
* Extract common controller methods to core concern
* Make update payload model agnostic
* Use routing concerns to extract resource definition
* Allow options to routing concern for module scoping
* Don't put `<hr>` elements inside list element
* Use semantic `<ol>` instead of `<ul>`
* Use `respond_to` and set the route default format to JSON

### DIFF
--- a/app/views/planning_applications/assessment/informatives/index.html.erb
+++ b/app/views/planning_applications/assessment/informatives/index.html.erb
@@ -17,15 +17,15 @@
 <div class="govuk-grid-row" data-controller="informatives">
   <div class="govuk-grid-column-full">
     <% if @informative_set.informatives.select(&:persisted?).any? %>
-      <ul data-controller="sortable" class="govuk-list">
-        <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+      <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+      <ol data-controller="sortable" class="govuk-list">
         <% @informative_set.informatives.select(&:persisted?).each do |informative| %>
           <%= content_tag :li,
                 class: "sortable-list",
                 id: dom_id(informative),
                 data: {
                   model_name: informative.class.name,
-                  sortable_url: planning_application_assessment_informative_positions_path(@planning_application, informative),
+                  sortable_url: planning_application_assessment_informative_position_path(@planning_application, informative),
                   sortable_handle: true
                 } do %>
 
@@ -47,7 +47,7 @@
             <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
           <% end %>
         <% end %>
-      </ul>
+      </ol>
     <% else %>
       <p class="govuk-body"><strong>No informatives added yet</strong></p>
     <% end %>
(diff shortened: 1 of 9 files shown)


app/views/planning_applications/assessment/informatives/index.html.erb
```diff
@@ -17,15 +17,15 @@
 <div class="govuk-grid-row" data-controller="informatives">
   <div class="govuk-grid-column-full">
     <% if @informative_set.informatives.select(&:persisted?).any? %>
-      <ul data-controller="sortable" class="govuk-list">
-        <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+      <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+      <ol data-controller="sortable" class="govuk-list">
```
pixeltrix, line 21: The content model for `<ul>` and `<ol>` models only allows `<li>` and script supporting elements (i.e. `<script>` and `<template>` elements).

https://html.spec.whatwg.org/multipage/grouping-content.html#the-ol-element

config/routes.rb
```diff
@@ -19,6 +19,12 @@
   constraints Constraints::LocalAuthoritySubdomain do
     root to: "planning_applications#index"
 
+    concern :positionable do |options|
+      defaults format: "json" do
+        resource :position, {only: %i[update]}.merge(options)
```
pixeltrix, line 24: Uses `:position` and not `:positions` since it's an action on a singular item in the collection

engines/bops_core/app/controllers/concerns/bops_core/positions_controller.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module BopsCore
+  module PositionsController
+    extend ActiveSupport::Concern
+
+    included do
+      wrap_parameters false
```
pixeltrix, line 8: Ensures our `{"positions": 1}` JSON doesn't get embedded in another hash

config/routes.rb
```diff
@@ -19,6 +19,12 @@
   constraints Constraints::LocalAuthoritySubdomain do
     root to: "planning_applications#index"
 
+    concern :positionable do |options|
+      defaults format: "json" do
```
pixeltrix, line 23: Makes the default accept type to be `application/json` and not `text/html`

engines/bops_core/app/controllers/concerns/bops_core/positions_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      before_action :set_planning_application
+      before_action :set_collection
+      before_action :set_record
+    end
+
+    def update
+      respond_to do |format|
+        format.json do
```
pixeltrix, line 17: Ensure that we only ever deal with JSON responses and we'll return an "406 Not Acceptable" to things like bots and webcrawlers

engines/bops_core/app/controllers/concerns/bops_core/positions_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+          if @record.insert_at(position)
+            head :no_content
+          else
+            render json: @record.errors, status: :unprocessable_entity
```
pixeltrix, line 21: Not sure this is the best error response but don't have anything better at the moment

benjamineskola (approved):

EGiataganas (approved):

PR #1804, opened 2024-05-15
Key dates in status panel container

EGiataganas:
### Description of change

- Extract `status-panel` to its own component styling file
- Use sass ampersand
- Reuse status panel to show key dates 
- Use `govuk` classes for margin & padding bottom

### Story Link

https://trello.com/c/fa2bTFRA

### Screenshots

Screenshot 2024-05-15 at 16 18 27

### DIFF
--- a/app/views/shared/_dates_and_assignment_header.html.erb
+++ b/app/views/shared/_dates_and_assignment_header.html.erb
@@ -3,18 +3,16 @@
     <% if @planning_application.consultation&.end_date %>
       <% if display_or_publish_required?(@planning_application, :site) %>
         <div id="confirm-site-notice-warning">
-          <%= render(
-                partial: "shared/warning_text",
-                locals: {message: link_to("Confirm site notice display date", edit_planning_application_site_notice_path(@planning_application, @planning_application.site_notice))}
+          <%= govuk_warning_text(
+                text: link_to("Confirm site notice display date", edit_planning_application_site_notice_path(@planning_application, @planning_application.site_notice))
               ) %>
         </div>
       <% end %>
 
       <% if display_or_publish_required?(@planning_application, :press) %>
         <div id="confirm-press-notice-warning">
-          <%= render(
-                partial: "shared/warning_text",
-                locals: {message: govuk_link_to("Confirm press notice publication date", planning_application_press_notice_confirmation_path(@planning_application))}
+          <%= govuk_warning_text(
+                text: govuk_link_to("Confirm press notice publication date", planning_application_press_notice_confirmation_path(@planning_application))
               ) %>
         </div>
       <% end %>
--- a/app/views/shared/_dates_and_assignment_header.html.erb
+++ b/app/views/shared/_dates_and_assignment_header.html.erb
@@ -1,74 +1,23 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds" id="dates-and-assignment-details">
-    <p class="govuk-body">
-      <strong>Validation date:</strong>
-      <% if @planning_application.valid_from.present? %>
-        <%= time_tag @planning_application.valid_from %>
-      <% else %>
-        Not yet valid
+    <% if @planning_application.consultation&.end_date %>
+      <% if display_or_publish_required?(@planning_application, :site) %>
+        <div id="confirm-site-notice-warning">
+          <%= render(
+                partial: "shared/warning_text",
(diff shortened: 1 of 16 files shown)


app/views/shared/_dates_and_assignment_header.html.erb
```diff
@@ -1,74 +1,23 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds" id="dates-and-assignment-details">
-    <p class="govuk-body">
-      <strong>Validation date:</strong>
-      <% if @planning_application.valid_from.present? %>
-        <%= time_tag @planning_application.valid_from %>
-      <% else %>
-        Not yet valid
+    <% if @planning_application.consultation.present? && @planning_application.consultation.end_date.present? %>
```
benjamineskola, line 3: I think this line could be

```suggestion
    <% if @planning_application.consultation&.end_date %>
```

We don't need `present?` because it can only be set or `nil`, there is no 'not present' value other than nil. And if we use `&.` we don't need to check separately for its parent object.

EGiataganas: Good point, copied the previous behaviour without trying to improve it

app/views/shared/_dates_and_assignment_header.html.erb
```diff
@@ -1,74 +1,23 @@
...
+      <% if display_or_publish_required?(@planning_application, :site) %>
+        <div id="confirm-site-notice-warning">
+          <%= render(
+                partial: "shared/warning_text",
+                locals: {message: link_to("Confirm site notice display date", edit_planning_application_site_notice_path(@planning_application, @planning_application.site_notice))}
+              ) %>
```
benjamineskola, line 9: How about using the upstream component helper:

```suggestion
          <%= govuk_warning_text(text: link_to("Confirm site notice display date", edit_planning_application_site_notice_path(@planning_application, @planning_application.site_notice))) %>
```

The partial is just the same but more verbose :)

benjamineskola: I'll get rid of all the others in #1806 :)

EGiataganas: Thanks for sharing that, I was going to start replacing the others too 😂

app/views/shared/_dates_panel.html.erb
```diff
@@ -0,0 +1,54 @@
...
+    </div>
+  </div>
+</div>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <p class="govuk-body">
+      <ul class="govuk-list govuk-list--bullet">
+        <% if @planning_application.environment_impact_assessment_required? %>
+          <li id="view-eia-details">The expiry date has been extended to 16 weeks  <span><%= govuk_link_to "View details", planning_application_validation_environment_impact_assessment_path(@planning_application) %></span></li>
```
benjamineskola, line 41: Is the `<span>` here necessary? The `link_to` will create an `<a>` and the span ha no classes etc.

benjamineskola:
```suggestion
          <li id="view-eia-details">The expiry date has been extended to 16 weeks <%= govuk_link_to "View details", planning_application_validation_environment_impact_assessment_path(@planning_application) %></li>
```

app/views/shared/_dates_panel.html.erb
```diff
@@ -0,0 +1,54 @@
...
+        <% end %>
+        <% if @planning_application.consultation&.end_date %>
+          <% unless display_or_publish_required?(@planning_application, :site) || display_or_publish_required?(@planning_application, :press) %>
+            <li id="edit-consultation-change-end-date">
+              <% days_left = @planning_application.consultation.days_left %>
+              The consultation end date is <%= (days_left > 0) ? t("planning_applications.days_left", count: days_left) : t("planning_applications.overdue", count: days_left.abs) %> <span><%= govuk_link_to "Change end date", edit_planning_application_consultation_path(@planning_application) %></span>
```
benjamineskola, line 47: Same here:

```suggestion
              The consultation end date is <%= (days_left > 0) ? t("planning_applications.days_left", count: days_left) : t("planning_applications.overdue", count: days_left.abs) %> <%= govuk_link_to "Change end date", edit_planning_application_consultation_path(@planning_application) %>
```

benjamineskola (approved):

PR #1809, opened 2024-05-16
Rebalance test groups

benjamineskola:
We have some groups that take two or three times longer than others. Trying to shift the boundaries so the slowest are a little faster.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,25 +55,23 @@ jobs:
     strategy:
       matrix:
         specs:
-          - { group: "models", pattern: "[a-p]*_spec.rb" }
-          - { group: "models", pattern: "[q-z]*_spec.rb" }
-          - { group: "controllers", pattern: "*_spec.rb" }
+          - { group: "models", pattern: "*_spec.rb" }
           - { group: "requests", pattern: "*_spec.rb", want_pdf: true }
-          - { group: "system", pattern: "a[a-f]*_spec.rb", want_pdf: true }
-          - { group: "system", pattern: "a[g-s]*_spec.rb", want_pdf: true }
-          - { group: "system", pattern: "a[t-z]*_spec.rb" }
+          - { group: "system", pattern: "a[a-r]*_spec.rb", want_pdf: true }
+          - { group: "system", pattern: "a[s-z]*_spec.rb", want_pdf: true }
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -55,25 +55,23 @@ jobs:
...
-          - { group: "models", pattern: "[a-p]*_spec.rb" }
-          - { group: "models", pattern: "[q-z]*_spec.rb" }
-          - { group: "controllers", pattern: "*_spec.rb" }
+          - { group: "models", pattern: "*_spec.rb" }
           - { group: "requests", pattern: "*_spec.rb", want_pdf: true }
-          - { group: "system", pattern: "a[a-f]*_spec.rb", want_pdf: true }
-          - { group: "system", pattern: "a[g-s]*_spec.rb", want_pdf: true }
-          - { group: "system", pattern: "a[t-z]*_spec.rb" }
+          - { group: "system", pattern: "a[a-r]*_spec.rb", want_pdf: true }
+          - { group: "system", pattern: "a[s-z]*_spec.rb", want_pdf: true }
```
benjamineskola, line 61: The main problem here is that almost all system tests starting with `a` are either `add`/`adding` or `assess`/`assessing`/`assessment`, so breaking it down further seems pointless. The `a[t-z]*` group had only one file in it, for example.

.github/workflows/build.yml
```diff
@@ -55,25 +55,23 @@ jobs:
...
-          - { group: "system", pattern: "[s-t]*_spec.rb" }
-          - { group: "system", pattern: "[u-z]*_spec.rb" }
+          - { group: "system", pattern: "s[a-e]*_spec.rb" }
+          - { group: "system", pattern: "s[f-z]*_spec.rb" }
+          - { group: "system", pattern: "[t-z]*_spec.rb" }
           - {
               group: "other",
               pattern: "*_spec.rb",
-              directories: "{components,form_models,helpers,presenters,mailer,jobs,services,translations}",
+              directories: "{components,controllers,form_models,helpers,presenters,mailer,jobs,services,translations}",
```
benjamineskola, line 74: note that we don't actually _have_ any controller tests any more, but I wanted to ensure that if we added some again that they wouldn't be overlooked.

pixeltrix (approved):

PR #1820, opened 2024-05-21
Create suggested conditions automatically rather than prefilling form

benjamineskola:
### Description of change

In the new design the conditions will be created by default and able to be removed by the user, rather than being a prefilled form that they can optionally add.

This creates them in the controller action if none exist.

### Story Link

https://trello.com/c/oFK58Hnp/3000-design-improvements-for-add-conditions

### Screenshots

Screenshot 2024-05-21 at 16 32 16

### DIFF
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -12,6 +12,8 @@ class Condition < ApplicationRecord
   after_create :create_validation_request!, if: :pre_commencement?
   before_update :create_validation_request!, if: -> { pre_commencement? && should_create_validation_request? }
 
+  scope :sorted, -> { sort_by(&:sort_key) }
+
   def checked?
     persisted? || errors.present?
   end
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -32,6 +32,12 @@ class Condition < ApplicationRecord
     (text + "\n\nReason: #{reason}").truncate(100, separator: "Reason")
   end
 
+  class << self
+    def standard_conditions
+      I18n.t(:conditions_list).map { |k, v| Condition.new(v) }
+    end
+  end
+
   private
 
   def titles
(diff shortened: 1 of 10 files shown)


app/models/condition.rb
```diff
@@ -32,6 +32,16 @@ def truncated_comment
...
   end
 
+  class << self
+    def standard_conditions
+      I18n.t(:conditions_list).map do |k, v|
+        c = Condition.new(v)
+        c.standard = true
+        c
+      end
+    end
```
pixeltrix, line 42: Isn't `standard` set to `true` in the locale file?

app/models/planning_application.rb
```diff
@@ -671,6 +671,14 @@ def condition_set
     super || create_condition_set!
   end
 
+  def create_condition_set!
+    super
+
+    Condition.standard_conditions.map { |condition| condition.update!(condition_set: condition_set) }
+
+    condition_set
```
pixeltrix, line 679: Wouldn't it be better to do this in a callback on `ConditionSet`?

app/models/condition_set/conditions_extension.rb
```diff
@@ -7,19 +7,13 @@ def sorted
     end
 
     def standard
-      standard_conditions.map do |condition|
+      Condition.standard_conditions.map do |condition|
```
pixeltrix, line 10: Trying to work out why we don't do `select(&:standard?)` here - legacy data issue maybe?

app/models/condition.rb
benjamineskola: Oh yes, seems unnecessary. Will remove.

app/models/condition_set/conditions_extension.rb
benjamineskola: I'm actually not sure if that's used at all now.

pixeltrix (approved):
Just a couple of minor improvements

PR #1821, opened 2024-05-21
Validate example json against schema

benbaumann95:
### Description of change

Validate example json against schema

- Add camel case for response keys for consistency
- We can remove the duplicate keys once they aren't being used

### DIFF
--- a/engines/bops_api/app/views/bops_api/v2/public/_show.json.jbuilder
+++ b/engines/bops_api/app/views/bops_api/v2/public/_show.json.jbuilder
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+# json.key_format! camelize: :lower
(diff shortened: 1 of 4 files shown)


benbaumann95:
Should probably add a rule to enforce using camel case

engines/bops_api/app/views/bops_api/v2/public/_show.json.jbuilder
```diff
@@ -1,13 +1,17 @@
 # frozen_string_literal: true
 
+# json.key_format! camelize: :lower
```
benbaumann95, line 3: Just need to uncomment this line when we can drop `full_reference` and `received_at`

benjamineskola (approved):

PR #1825, opened 2024-05-23
Use job metadata to store the user

pixeltrix:
This is cleaner than mutating the job arguments.

### DIFF
--- a/app/jobs/concerns/current_user_for_job.rb
+++ b/app/jobs/concerns/current_user_for_job.rb
@@ -3,27 +3,28 @@
 module CurrentUserForJob
   extend ActiveSupport::Concern
 
-  included do
-    before_enqueue do |job|
-      next unless Current.user
+  attr_accessor :enqueueing_user
 
-      if job.arguments.last.is_a?(Hash)
-        job.arguments.last[:current_user] = Current.user
-      else
-        job.arguments << {current_user: Current.user}
+  def serialize
+    super.tap do |job_data|
+      if Current.user.present?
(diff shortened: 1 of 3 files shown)


spec/jobs/application_job_spec.rb
```diff
@@ -8,99 +8,47 @@
...
 
-      it "performs the job" do
-        perform_enqueued_jobs {
-          MyApplicationJob.perform_later("positional", keyword: "keyword")
-        }
-
-        expect(logs).to include <<~LOGS
-          INFO -- : Performed MyApplicationJob for Background User with arguments: ["positional", keyword: "keyword"]
-        LOGS
-      end
```
pixeltrix, line 103: Since we're not mutating arguments we don't need the spec variations

app/jobs/concerns/current_user_for_job.rb
```diff
@@ -3,27 +3,28 @@
...
+  attr_accessor :enqueueing_user
 
-      if job.arguments.last.is_a?(Hash)
-        job.arguments.last[:current_user] = Current.user
-      else
-        job.arguments << {current_user: Current.user}
+  def serialize
+    super.tap do |job_data|
+      if Current.user.present?
+        job_data["enqueueing_user"] = Current.user.to_gid_param
```
pixeltrix, line 11: This is a base64 encoded GlobalID uri (e.g. `gid://bops/User/108`)

benjamineskola (approved):

benbaumann95 (approved):

PR #1826, opened 2024-05-23
Validate applicant and agent email address format

benbaumann95:
### Description of change

- Validate applicant and agent email address format
- Display errors inline

### Story Link

https://trello.com/c/72sCyM2w/3012-validate-emails-on-manual-application-form

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -175,6 +175,17 @@ class PlanningApplication < ApplicationRecord
     numericality: {greater_than_or_equal_to: 0, less_than_or_equal_to: 1_000_000},
     allow_nil: true
 
+  validate do
+    errors.add(:application_type_id, :blank) unless application_type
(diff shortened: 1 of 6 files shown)


app/models/planning_application.rb
```diff
@@ -175,6 +175,17 @@ class WithdrawOrCancelError < RuntimeError; end
     numericality: {greater_than_or_equal_to: 0, less_than_or_equal_to: 1_000_000},
     allow_nil: true
 
+  validate do
+    errors.add(:application_type_id, :blank) unless application_type
+  end
+
+  with_options format: {with: URI::MailTo::EMAIL_REGEXP} do
```
pixeltrix, line 182: For the future but we probably should add a email validator that matches Notify - I think this will allow some that Notify rejects.

pixeltrix (approved):

PR #1827, opened 2024-05-24
Cancelled conditions

EGiataganas:
### Description of change

- Allow to cancel pre-commencement condition validation request, and add a timestamp record on condition
- Hide any cancelled conditions

### Story Link

https://trello.com/c/izFLUu53/2772-patterns-for-showing-when-review-tasks-have-been-completed#comment-664dd882204f3133db09353d

### DIFF
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -13,6 +13,7 @@ class Condition < ApplicationRecord
   before_update :create_validation_request!, if: -> { pre_commencement? && should_create_validation_request? }
 
   scope :sorted, -> { sort_by(&:sort_key) }
+  scope :not_cancelled, -> { where(cancelled_at: nil) }
 
   def checked?
     persisted? || errors.present?
--- a/app/models/condition_set.rb
+++ b/app/models/condition_set.rb
@@ -34,7 +34,7 @@ class ConditionSet < ApplicationRecord
   end
 
   def not_cancelled_conditions
-    conditions.reject(&:cancelled_at)
+    conditions.not_cancelled
   end
 
   def confirm_pending_requests!
--- a/app/models/condition_set.rb
+++ b/app/models/condition_set.rb
@@ -33,6 +33,10 @@ class ConditionSet < ApplicationRecord
     conditions.joins(:validation_requests).where(validation_requests: {approved: true})
   end
 
+  def not_cancelled_conditions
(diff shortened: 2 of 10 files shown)


app/models/condition_set.rb
```diff
@@ -33,6 +33,10 @@ def approved_conditions
     conditions.joins(:validation_requests).where(validation_requests: {approved: true})
   end
 
+  def not_cancelled_conditions
+    conditions.reject(&:cancelled_at)
```
benjamineskola, line 37: I'm not sure if it matters but is it better to use the active record method for this?

```suggestion
    conditions.where(cancelled_at: nil)
```

I _think_ `reject` might select all of them and then filter, instead of filtering in the database.

app/models/pre_commencement_condition_validation_request.rb
```diff
@@ -4,6 +4,7 @@ class PreCommencementConditionValidationRequest < ValidationRequest
   RESPONSE_TIME_IN_DAYS = 10
 
   validates :cancel_reason, presence: true, if: :cancelled?
+  validates :condition_cancelled_at_now, presence: true, if: :cancelled?
```
benjamineskola, line 7: Should this be a `before_save` or `before_validation` hook instead of a validation? I think a validation should only return true or false rather than making a change.

EGiataganas: I've used the same pattern as line above. However it could also be a `before_validation`.

benjamineskola: It's not the same: `cancel_reason` is a database field, not a method. The `validates :cancel_reason` just checks that it's present; it doesn't set it to anything.

app/models/condition.rb
```diff
@@ -13,6 +13,7 @@ class Condition < ApplicationRecord
   before_update :create_validation_request!, if: -> { pre_commencement? && should_create_validation_request? }
 
   scope :sorted, -> { sort_by(&:sort_key) }
+  scope :cancelled, -> { where(cancelled_at: nil) }
```
benjamineskola, line 16: Maybe this should be called `not_cancelled` or something? Otherwise it might be confusing.

benjamineskola (approved):

PR #1829, opened 2024-05-24
Allow for numbers in application type suffix

benbaumann95:
An application type suffix can include numbers such as PA1A. This change will match on these cases too.


engines/bops_api/swagger/v2/swagger_doc.yaml
```diff
@@ -16130,8 +16130,13 @@ components:
                     type: string
                   full_reference:
                     type: string
+                  fullReference:
```
benbaumann95, line 16133: Accidentally left out from before

pixeltrix (approved):

PR #1830, opened 2024-05-28
Use rack_test for system tests where possible

benjamineskola:
rack_test is noticeably faster on tests where we don't need javascript or other interactivity, so use capybara only where necessary.

In practice I think even more of these tests could be rewritten without capybara but I'm not sure. This is just a quick fix.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,35 +58,15 @@ jobs:
           - { group: "models" }
           - { group: "requests", want_pdf: true }
           - {
-              group: "system: assessing",
-              directories: "system/planning_applications/assessing",
+              group: "system: rack_test",
+              directories: "system",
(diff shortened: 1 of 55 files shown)


pixeltrix:
Is the base for this right?

benjamineskola:
@pixeltrix it is now. I spotted the time validation error when trying to get these to pass which is why I ended up rebasing this onto that branch.

benjamineskola:
It looks like this has roughly doubled the test speed: the largest groups have gone from about 4 minutes to about 2 minutes.

It might be feasible to group them based on capybara and non-capybara; on my M1 all the non-capybara system tests run in under 4 minutes total, but I'm not sure how that translates to GHA performance.

benbaumann95 (approved):

PR #1835, opened 2024-05-30
Add multiple press notices

benbaumann95:
### Description of change

- Allow for multiple press notices
- Show previous press notices 
- Add show and edit pages when confirming the publication

### Story Link

https://trello.com/c/W2W8RmGJ/2793-history-of-previous-dates-entered-press-notice

### DIFF
--- a/app/controllers/planning_applications/press_notices_controller.rb
+++ b/app/controllers/planning_applications/press_notices_controller.rb
@@ -6,7 +6,14 @@ module PlanningApplications
 
     before_action :set_planning_application
     before_action :ensure_publicity_is_permitted
-    before_action :set_press_notice
+    before_action :build_press_notice, only: [:new, :create]
+    before_action :set_press_notice, only: [:show, :update]
+
+    def new
+      respond_to do |format|
+        format.html
+      end
+    end
 
     def show
       respond_to do |format|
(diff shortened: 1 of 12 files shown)


app/controllers/planning_applications/press_notices_controller.rb
```diff
@@ -28,16 +51,18 @@ def create
...
     private
 
     def press_notice_params
       params.require(:press_notice).permit(:required, :other_reason, reasons: [])
     end
 
+    def build_press_notice
+      @press_notice = @planning_application.press_notices.new
+    end
+
```
benjamineskola, line 63: If this is being moved out of `PlanningApplication` can it be deleted there too? Or do we need two different versions of it?

spec/system/planning_applications/publicity/press_notice_spec.rb
```diff
@@ -530,6 +540,41 @@
...
+              expect(page).to have_link("View in new window")
+            end
+
+            within("td:nth-child(2)") do
+              expect(page).to have_selector(".govuk-tag", text: "Published")
+              expect(page).to have_selector("p", text: "Date requested: #{press_notice.requested_at.to_date.to_fs}")
+              expect(page).to have_selector("p", text: "Date published: #{press_notice.reload.published_at.to_date.to_fs}")
+              expect(page).to have_selector("p", text: "Comments: Edited press notice comment")
+            end
+          end
```
benjamineskola, line 576: Using `nth-child` always feels a little fragile. Maybe adding classes to the rows and/or cells might be a bit clearer? `<td class="document-links">` and `<td class="document-details">` perhaps and then we can just check for the classes in the `within` blocks.

`tr:nth-child(1)` is probably more justified in this case as it's presumably important that it really is the first row and now some other.

benjamineskola (approved):

PR #1838, opened 2024-06-03
Add make commands for bops-applicants to migrate

EGiataganas:
### Description of change

Add make commands for bops-applicants to migrate


benjamineskola:
Should this also happen as part of the deployment process?

EGiataganas:
@benjamineskola yeah, it's the Makefile which was missing this link 😄

PR #1845, opened 2024-06-05
Add valid_from and published_at dates to API

benjamineskola:
### Description of change

- **Add valid_from date to API responses**
- **Add a published_at date to be set when an application is made public**
- **Add published_at date to API responses**
- **Add make_public to ignored_columns in advance of removing it entirely**

### Story Link

https://trello.com/c/wEnTHZQf/3011-add-validatedat-and-published-to-register-date-to-api-output

### DIFF
--- a/db/migrate/20240605143118_add_published_at_to_planning_applications.rb
+++ b/db/migrate/20240605143118_add_published_at_to_planning_applications.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class AddPublishedAtToPlanningApplications < ActiveRecord::Migration[7.1]
+  def change
+    add_column :planning_applications, :published_at, :datetime
+
+    up_only do
+      PlanningApplication.find_each(make_public: true) do |application|
+        application.update(published_at: application.validated_at)
+      end
+    end
+  end
+end
--- a/engines/bops_api/app/controllers/bops_api/v2/public/planning_applications_controller.rb
+++ b/engines/bops_api/app/controllers/bops_api/v2/public/planning_applications_controller.rb
@@ -15,7 +15,7 @@ module BopsApi
         private
 
         def planning_applications_scope
-          @local_authority.planning_applications.where(make_public: true)
+          @local_authority.planning_applications.where.not(published_at: nil)
         end
 
         def search_params
--- a/engines/bops_api/app/controllers/bops_api/v2/public/planning_applications_controller.rb
+++ b/engines/bops_api/app/controllers/bops_api/v2/public/planning_applications_controller.rb
@@ -15,7 +15,7 @@ module BopsApi
         private
 
         def planning_applications_scope
-          @local_authority.planning_applications.where.not(published_at: nil)
+          @local_authority.planning_applications.published
         end
 
         def search_params
--- a/engines/bops_api/app/views/bops_api/v2/shared/_application.json.jbuilder
+++ b/engines/bops_api/app/views/bops_api/v2/shared/_application.json.jbuilder
@@ -8,5 +8,6 @@ json.application do
   json.reference planning_application.reference
   json.fullReference planning_application.reference_in_full
   json.receivedAt planning_application.received_at
+  json.validAt planning_application.validated_at
   json.status planning_application.status
 end
--- a/engines/bops_api/app/views/bops_api/v2/shared/_application.json.jbuilder
+++ b/engines/bops_api/app/views/bops_api/v2/shared/_application.json.jbuilder
@@ -9,5 +9,6 @@ json.application do
   json.fullReference planning_application.reference_in_full
   json.receivedAt planning_application.received_at
   json.validAt planning_application.validated_at
+  json.publishedAt planning_application.published_at
   json.status planning_application.status
 end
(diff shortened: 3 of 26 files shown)


engines/bops_api/app/views/bops_api/v2/public/_show.json.jbuilder
```diff
@@ -12,6 +12,8 @@ json.application do
   json.fullReference planning_application.reference_in_full
   json.received_at planning_application.received_at
   json.receivedAt planning_application.received_at
+  json.valid_from planning_application.validated_at
+  json.published_at planning_application.published_at
```
benbaumann95, line 16: Can we use camel case here, I'm enforcing it with my submission endpoint PR

benjamineskola: Sure thing — I remembered we were trying to move from one to the other but lost track of which way around we were going 🤦🏻

engines/bops_api/app/controllers/bops_api/v2/public/planning_applications_controller.rb
```diff
@@ -15,7 +15,7 @@ def search
         private
 
         def planning_applications_scope
-          @local_authority.planning_applications.where(make_public: true)
+          @local_authority.planning_applications.where.not(published_at: nil)
```
EGiataganas, line 18: how about moving it on a scope?

db/migrate/20240605143118_add_published_at_to_planning_applications.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddPublishedAtToPlanningApplications < ActiveRecord::Migration[7.1]
+  def change
+    add_column :planning_applications, :published_at, :datetime
+
+    up_only do
+      # https://en.wikipedia.org/wiki/Time_immemorial
+      PlanningApplication.where(make_public: true).update_all(published_at: Time.zone.local(1189, 7, 6, 0))
```
benbaumann95, line 9: I'm a bit wary of having this on production data.
Is there a more sensible proxy we could use or only run this for staging.
Not sure we'd have many `make_public: true` cases on prod with this tbh but maybe worth querying to see beforehand

engines/bops_api/app/views/bops_api/v2/shared/_application.json.jbuilder
```diff
@@ -8,5 +8,7 @@ json.application do
   json.reference planning_application.reference
   json.fullReference planning_application.reference_in_full
   json.receivedAt planning_application.received_at
+  json.validFrom planning_application.validated_at
```
benbaumann95, line 11: Should we have this as `validAt` to keep consistent with other datetime fields we expose?

engines/bops_api/schemas/odp/v0.6.0/applicationSubmission.json
```diff
@@ -30,6 +30,14 @@
...
           "type": "string"
         },
+        "validFrom": {
+          "format": "datetime",
+          "type": ["string", "null"]
+        },
+        "publishedAt": {
+          "format": "datetime",
+          "type": ["string", "null"]
+        },
```
benbaumann95, line 40: Not for now, and a note for me too, but next time we touch the application details let's extract all this into a reusable schema as we use it here and the search

spec/system/planning_applications/make_public_spec.rb
```diff
@@ -39,12 +39,16 @@
 
     expect(page).to have_content("Public on BOPS Public Portal: Yes")
 
+    expect(planning_application.reload.published_at).not_to be_nil
```
benbaumann95, line 42: Can't we just set this expectation to the time we travel to in this block `"2022-01-01"`

benbaumann95 (commented):
Think you need to run `make api-docs` so that the swagger gets updated

db/migrate/20240605143118_add_published_at_to_planning_applications.rb
benjamineskola: My thinking was that this needed to not be null (so that `make_public?` returns true), but also shouldn't be possible to mistake for any real date, so a date long in the past seemed to make the most sense. That way we can tell that it's public but see that we also don't know the real date that it was marked public.

benbaumann95: Yeah maybe just the validation date could potentially be better

benbaumann95: With the scope you added, we'd need to set a date on here anyway

engines/bops_api/app/views/bops_api/v2/shared/_application.json.jbuilder
benbaumann95: Even `validFromAt` would be more consistent

benbaumann95: Hmm but the more I look at that the weirder it gets though.
Think `validAt` would suffice

engines/bops_api/schemas/odp/v0.6.0/search.json
```diff
@@ -96,6 +96,14 @@
                   "format": "datetime",
                   "type": "string"
                 },
+                "validAt": {
+                  "format": "datetime",
+                  "type": "string"
```
benbaumann95, line 101: Any reason for adding `"type": ["string", "null"]` above but not here?

benbaumann95 (approved):

PR #1847, opened 2024-06-06
Extract sortable list item into component

benbaumann95:
### Description of change

Extract sortable list item into component

### Story Link

https://trello.com/c/kPQVv7zd/3022-extract-considerations-component-for-reuse

### DIFF
--- a/app/assets/stylesheets/components/_sortable.scss
+++ b/app/assets/stylesheets/components/_sortable.scss
@@ -1,10 +1,33 @@
 .sortable-list {
   cursor: move;
+
+  .nav-links {
+    display: flex;
+    padding-top: 10px;
+
+    a {
+      padding-right: 0.5rem;
+    }
+
+    a + a { 
+      border-left: 1px solid govuk-colour("mid-grey");
+      padding-left: 0.5rem;
+    }
+  }
+}
+
+.sortable-list li {
+  padding-bottom: 2rem;
+  border-bottom: 1px solid govuk-colour("mid-grey");
 }
 
 .sortable-container {
   display: flex;
   justify-content: space-between;
+
+  .sortable-list-item {
+    width: 98%;
+  }
 }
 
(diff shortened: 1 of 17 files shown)


app/assets/stylesheets/components/_sortable.scss
```diff
@@ -2,10 +2,18 @@
...
+.sortable-list li {
+  padding-bottom: 2rem;
+  border-bottom: 1px solid govuk-colour("mid-grey");
+}
+
 .sortable-container {
   display: flex;
   justify-content: space-between;
 }
+.sortable-container div {
```
benjamineskola, line 14: I'm a little cautious about using `div` here as well, if we ever add more divs to any of the list items it'll affect those too. Perhaps give it a class name and use that as the selector?

app/assets/stylesheets/components/_sortable.scss
```diff
@@ -14,10 +22,23 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: left;
+  padding-top: 10px;
+}
+.divider {
+  display: inline-block;
+  margin: 0 0.5rem;
+  color: govuk-colour("mid-grey");
+}
```
benjamineskola, line 44: These are quite generic names so I wonder if it would make sense to nest them within `.sortable-list` or something?

Actually for the divider it's probably better to use a `border` on the nav-list elements.

benjamineskola: something like

```css
.nav-links {
  a + a { /* only links that are preceded by another link i.e. not the first */
    border-left: 1px solid govuk-colour("mid-grey");
    padding-left: 2rem;
  }
}
```

benbaumann95: Will give this a go 👍

benjamineskola (approved):

PR #1848, opened 2024-06-07
Cancel heads of terms

EGiataganas:
### Description of change

- Add cancelled_at to heads_of_terms
- Hide any cancelled term
- Show cancelled button for any non pending validation request
- Change Update heads of term to Edit

### Story Link

https://trello.com/c/K16GYm1M


app/models/heads_of_terms_validation_request.rb
```diff
@@ -35,4 +37,8 @@ def audit_api_comment
   def audit_comment
     {created_at:}.to_json
   end
+
+  def term_cancelled_at_now
```
benjamineskola, line 41: Small suggestion: maybe we could call this `cancel_now!` or similar, to make it clear that it's modifying the term?

benjamineskola (approved):
One suggestion but otherwise LGTM.

PR #1852, opened 2024-06-10
Press notice improvements

benbaumann95:
### Description of change

- Press notice improvements

### Story Link

https://trello.com/c/EawUn4KH/3031-press-notices-tidy-up-and-improvements


app/views/planning_applications/press_notices/_press_notices.html.erb
```diff
@@ -8,6 +8,8 @@
 
   <tbody class="govuk-table__body">
     <% press_notices.each do |press_notice| %>
+      <% next if press_notice.documents.empty? %>
```
benjamineskola, line 11: I wonder if adding a scope to `PressNotice` would be worthwhile?

`scope :with_documents, -> { where.not(documents: nil) }` or something. Then use that in the loop rather than a `next`.

Not particularly important though

benbaumann95: Might need to do a join but yes probs better to add a scope

benbaumann95: Using a joins

PR #1853, opened 2024-06-11
Add documents endpoint

benjamineskola:
### Description of change

Add a public endpoint to return all public documents for an application.

### Story Link

https://trello.com/c/jBGKKOOM/3029-add-documents-to-new-api-endpoint

### Decisions [OPTIONAL]

Moved `find_planning_application` (which finds by reference or ID) into the parent class so that we can use it elsewhere. It isn't a strict requirement (yet?) but if we want to start moving in that direction it seems like a good idea to have any new controllers making use of that, especially since it's backwards-compatible anyway so there's no loss.

### DIFF
--- a/engines/bops_api/app/controllers/bops_api/v2/public/documents_controller.rb
+++ b/engines/bops_api/app/controllers/bops_api/v2/public/documents_controller.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module BopsApi
+  module V2
+    module Public
+      class DocumentsController < PublicController
+        def show
+          @planning_application = find_planning_application params[:planning_application_id]
+
+          respond_to do |format|
+            format.json
+          end
+        end
+
+        private
+
+        def planning_applications_scope
+          @local_authority.planning_applications.published
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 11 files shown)


engines/bops_api/app/controllers/bops_api/v2/public/documents_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+            format.json
+          end
+        end
+
+        private
+
+        def params
+          # TODO find_planning_application looks for params[:id] but we have params[:planning_application_id]
+          # this is a hack
+          {id: super[:planning_application_id]}
```
benbaumann95, line 20: Could we just define the param as a method

We do in other place

```
  def planning_application_param
    request.path_parameters.key?(:planning_application_id) ? :planning_application_id : :id
  end
  ```
  
  Then can just use the relevant key where needed

**Review by benbaumann95 — APPROVED** [review]

File: engines/bops_api/app/views/bops_api/v2/public/documents/show.json.jbuilder [hunk 2]
```diff
@@ -0,0 +1,20 @@
...
+
+json.partial! "bops_api/v2/shared/application", planning_application: @planning_application
+
+json.files @planning_application.documents.for_publication do |document|
+  json.name document.name
+  json.url main_app.api_v1_planning_application_document_url(@planning_application, document)
+  json.extract! document,
+    :created_at,
+    :tags,
+    :numbers,
```
benbaumann95, line 13: as discussed, just remove these for now 👍

PR #1864, opened 2024-06-21
Remove duplicate test

benbaumann95:


engines/bops_api/spec/requests/v2/public/documents_spec.rb
```diff
@@ -88,13 +88,6 @@
             expect(data["files"].last["type"].first["description"]).to eq("Site Notice")
           end
         end
-
-        it "validates successfully against the example documents json" do
```
benbaumann95, line 92: same as https://github.com/unboxed/bops/pull/1864/files#diff-26f7ad442c5885274e292fbc0f56e58ccc236cdc83ccec7ed0fd5ce40c0d71d2R37

benjamineskola (approved):

PR #1875, opened 2024-07-04
Set cors headers for public API

benbaumann95:
### Description of change

Allow all origins to access public API

### DIFF
--- a/engines/bops_api/app/controllers/bops_api/v2/public_controller.rb
+++ b/engines/bops_api/app/controllers/bops_api/v2/public_controller.rb
@@ -3,6 +3,17 @@
 module BopsApi
   module V2
     class PublicController < ApplicationController
+      before_action :set_cors_headers
+
+      def set_cors_headers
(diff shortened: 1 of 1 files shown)


engines/bops_api/app/controllers/bops_api/v2/public_controller.rb
```diff
@@ -3,6 +3,17 @@
 module BopsApi
   module V2
     class PublicController < ApplicationController
+      before_action :set_cors_headers
+
+      def set_cors_headers
+        response.set_header("Access-Control-Allow-Origin", "*")
+        response.set_header("Access-Control-Allow-Methods", "*")
```
benjamineskola, line 10: Would it make sense to limit this to `GET`? I don't think we have any other types for the public endpoints.

benbaumann95: Yep 👍

benjamineskola (approved):

EGiataganas (approved):

PR #1880, opened 2024-07-08
Include document tags in redacted documents

EGiataganas:
### Description of change

Include document tags in redacted documents

### Story Link

https://trello.com/c/tfSlDIvi


app/controllers/planning_applications/validation/document/redactions_controller.rb
```diff
@@ -34,7 +35,17 @@ def planning_application_params
...
+        end
+
+        def filter_tags!
+          return unless params[:planning_application][:documents_attributes]
+
+          params[:planning_application][:documents_attributes].each do |index, document_attributes|
+            if document_attributes[:tags].present? && document_attributes[:tags].is_a?(String)
+              params[:planning_application][:documents_attributes][index][:tags] = document_attributes[:tags].split
+            end
+          end
```
benjamineskola, line 48: I'm confused by the format of `documents_attributes` — is it a hash containing other hashes? And sometimes `tags` is a string and sometimes it's an array, so we have to convert the strings to arrays?

benjamineskola: Btw `present?` might be unnecessary here, because `nil.is_a?(String)` is false:
```suggestion
          params[:planning_application][:documents_attributes].each do |index, document_attributes|
            if document_attributes[:tags].is_a?(String)
              params[:planning_application][:documents_attributes][index][:tags] = document_attributes[:tags].split
            end
          end
```

app/controllers/planning_applications/validation/document/redactions_controller.rb
```diff
@@ -34,7 +35,17 @@ def planning_application_params
         end
 
         def documents_params
-          %i[file redacted publishable validated]
+          [:file, :redacted, :publishable, :validated, tags: []]
+        end
+
+        def filter_tags!
```
benbaumann95, line 41: I'm a bit confused with what this method is trying to achieve. Are we making this more complicated than necessary?

benjamineskola (approved):

app/controllers/planning_applications/validation/document/redactions_controller.rb
benjamineskola: It seems like the problem here is that the tags are passed back through a hidden field, which joins the array into a string, and therefore needs to be split back out.

It's probably worth refactoring this later though.

PR #1881, opened 2024-07-09
Filter out existing neighbours before adding

benjamineskola:
### Description of change

Neighbour records must be unique for a given consultation, but because the map input automatically creates a list of these it's very easy for existing neighbours to be included in the list to be added. Filtering out the existing addresses in the controller means that repeatedly submitting the same area, or overlapping areas, will only add the new neighbours each time, if any, rather than warning of duplicates.

### Story Link

https://trello.com/c/LZLwdKVE/3047-user-cannot-continue-adding-new-neighbour-address-in-boundary-of-neighbours-already-selected

### Decisions

This could probably have been done at a few different levels, e.g. by somehow causing Rails to use a `find_or_create` rather than `create`, but this seemed like the most straightforward.

### DIFF
--- a/app/models/consultation.rb
+++ b/app/models/consultation.rb
@@ -115,7 +115,8 @@ class Consultation < ApplicationRecord
 
   before_save :apply_deadline_extension, if: -> { deadline_extension.present? }
 
-  accepts_nested_attributes_for :consultees, :neighbours
+  accepts_nested_attributes_for :consultees
+  accepts_nested_attributes_for :neighbours, reject_if: :neighbour_exists?
 
   enum status: {
     not_started: "not_started",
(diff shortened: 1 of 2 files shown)


app/controllers/planning_applications/neighbours_controller.rb
```diff
@@ -14,7 +14,13 @@ def index
 
     def create
       begin
-        @consultation.update!(consultation_params)
+        existing_neighbour_addresses = @consultation.neighbours.map(&:address)
```
benbaumann95, line 17: Instead of loading all these records in memory (it could be 100s) can we just do a direct query.
Or i'm probably leaning towards the find_or_create_by approach

benjamineskola: My concern was that if we queried it each time in the loop that might end up being worse. But that depends a lot on usage.

If there are already a lot of neighbours associated with the consultation, loading them into memory is not great as you say, and if only a couple of addresses are being added then you end up with potentially hundreds stored to avoid only a couple of queries.

But on the other hand if there are none currently associated, and a large area is searched for, then it'll do hundreds of queries, but caching the existing addresses would have stored none in memory.

Perhaps you're right that the find-or-create way is better.

EGiataganas (approved):

PR #1888, opened 2024-07-17
Add public link to public api search get call

HonTastic2:
### Description of change

Added an extra line to also send the public link of an application if it has consultation


benbaumann95 (approved):

engines/bops_api/schemas/odp/v0.6.0/shared/definitions.json
```diff
@@ -58,6 +58,10 @@
             "endDate": {
               "type": ["string", "null"],
               "format": "date"
+            },
+            "publicUrl": {
+              "type": ["string", "null"],
+              "format": "string"
```
EGiataganas, line 64: How about you’re moving this change to latest version 0.7.0?

benbaumann95: Yeah good catch.

@HonTastic2 We've just recently updated the schema version so we'll need to rebase this PR and add this to the 0.7.0 file before you can merge

benbaumann95 (changes requested):
https://github.com/unboxed/bops/pull/1888/files#r1682783565

engines/bops_api/schemas/odp/v0.6.0/shared/definitions.json
HonTastic2: done

benbaumann95 (approved):

PR #1893, opened 2024-07-23
Tidy production dockerfile

benjamineskola:
### Description of change

Ours has grown organically for a while and there are a couple of unnecessary things (like layers that do nothing and multiple apt install steps). This version is based on the default Rails 7.1 dockerfile.

### DIFF
--- a/Dockerfile.production
+++ b/Dockerfile.production
@@ -1,143 +1,89 @@
-# Base image for all other build stages
+# syntax = docker/dockerfile:1
+
+# Make sure RUBY_VERSION matches the Ruby version in .ruby-version and Gemfile
 ARG RUBY_VERSION=3.3.4
+FROM registry.docker.com/library/ruby:$RUBY_VERSION-slim as base
 
-FROM ruby:$RUBY_VERSION-slim AS base_image
+# Rails app lives here
+WORKDIR /rails
 
-# Run security updates and install apt-utils curl and locales
-RUN bash -c "export DEBIAN_FRONTEND=noninteractive && \
-    apt-get update -qq && \
-    echo 'en_US.UTF-8 UTF-8' > /etc/locale.gen && \
-    apt-get install -y apt-utils curl locales && \
-    apt-get upgrade -y && apt-get clean && \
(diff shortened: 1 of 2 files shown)


EGiataganas (approved):

benjamineskola:
Tested and this seems to work, not sure if there's something I've overlooked though.

Dockerfile.production
```diff
@@ -1,144 +1,90 @@
...
 
-FROM ruby:$RUBY_VERSION-slim AS base_image
+# Rails app lives here
+WORKDIR /rails
+
+# Set production environment
+ENV RAILS_ENV="production" \
+    BUNDLE_DEPLOYMENT="1" \
+    BUNDLE_PATH="/usr/local/bundle" \
+    BUNDLE_WITHOUT="development"
```
pixeltrix, line 14: Should this be `BUNDLE_WITHOUT="development test"` ?

Dockerfile.production
```diff
@@ -1,144 +1,90 @@
...
 
-# Install NodeJS
+# Install JavaScript dependencies
 ARG NODE_VERSION=20.11.0
 ARG YARN_VERSION=1.22.19
 ENV PATH=/usr/local/node/bin:$PATH
-RUN bash -c "curl -sL https://github.com/nodenv/node-build/archive/master.tar.gz | tar xz -C /tmp/ && \
-    /tmp/node-build-master/bin/node-build '$NODE_VERSION' /usr/local/node && \
+RUN curl -sL https://github.com/nodenv/node-build/archive/master.tar.gz | tar xz -C /tmp/ && \
+    /tmp/node-build-master/bin/node-build "${NODE_VERSION}" /usr/local/node && \
```
pixeltrix, line 29: `RUN` by itself uses `/bin/sh` so that's something to be aware of when running a script like this that may be expecting bash or zsh. It may be that `/bin/sh` is a symlink to `/bin/bash` anyway.

benjamineskola: The default Rails Dockerfile uses just `RUN` for this rather than execing bash, so I've assumed it's safe to do so.

PR #1895, opened 2024-07-29
Dynamically calculate public holidays

benjamineskola:
### Description of change

As a precursor to work improving the handling of working days in consultation periods, dynamically load the list of public holidays. Also expose a helper for accessing these.

### Story Link

https://trello.com/c/9rNaBy0q/2498-deal-with-application-types-where-bank-holidays-are-excluded-from-the-consultation-period


app/models/consultation.rb
```diff
@@ -196,7 +196,14 @@ def end_date_from(now = Time.zone.today)
...
     # Royal Mail delivers from Monday to Saturday, excluding bank holidays.
-    default_start_date(now) + [consultee_response_period.days, period_days].max
+
+    effective_period_days = [consultee_response_period.days, period_days].max
+
+    if planning_application.application_type.include_bank_holidays?
+      Bops::Holidays.days_after_plus_holidays(from_date: default_start_date(now), count: effective_period_days)
+    else
+      default_start_date(now) + effective_period_days
+    end
```
benbaumann95, line 206: Just to confirm, when we say include bank holidays we are referring to treating these days as a normal day in the calculation right?
So it should behave like default. 
 
But when bank holidays are excluded then we factor these specific days in and the end date might be end up being a date more into the future than the default.

Not sure if I'm asking this properly so happy to have a chat 😄

benbaumann95 (commented):
Worth adding a system spec to `bops_config/spec/system/application_types_spec.rb` similar to 
```
it "allows editing of the determination period days" do
```

At the moment, is this just configurable from an existing application edit page? 
Not sure if we need to add a step in the application type wizard but I think this seems fine.

app/models/consultation.rb
benjamineskola: It's a very weird calculation. The normal situation is that the consultation is N days (default 21). That doesn't account for weekends or anything, it's just 21 calendar days.

In _some_ cases (I don't think we're sure which), then it's 21 calendar days _not counting bank holidays_. i.e,. if there's a single bank holiday within the 21 day period, the consultation actually ends 22 days later. (Weekends are still ignored for this calculation, so we can't say '21 business days' or anything like that.)

'Include bank holidays' is not necessarily a very obvious name for it but it feels hard to summarise the requirements.

benjamineskola: Or are you saying the meaning of the setting should be reversed: default (true) to be the normal behaviour of 21 days means 21 days, false means do the weird calculation?

That might make it more obvious, I'd just left it the way round it was when I'd added my own setting for this.

benbaumann95: Just on first pass, when I read include bank holidays I would assume it doesn't matter if it is a bank holiday, treat it like a normal day. But one could have the opposite interpretation too 🤔 
Agree that the naming is quite tricky with this.

Might be helpful if we can add some content for the user so we show specific guidance that if they check this checkbox then we will be calculating the consultation end date ignoring bank holidays or factoring them in (perhaps with an example). 
Then as long as we are consistent with what that flag is doing in our calculations it shouldn't matter much.

benbaumann95: This sounds like an outcome we can have from user testing

benbaumann95 (approved):

app/models/consultation.rb
benjamineskola: Yeah, I did wonder if additional hint text might be useful for this option.

PR #1905, opened 2024-08-02
Capture schema validation error for easier debugging

benbaumann95:
### Description of change

Capture schema validation error to help with debugging

### Story Link

https://trello.com/c/FYWWuQt7/3074-investigate-production-cases-not-being-pulled-through-to-staging


benjamineskola (approved):

EGiataganas:
How about adding a specific test for this and make the error more specific to schema version change?

benbaumann95:
> How about adding a specific test for this and make the error more specific to schema version change?

Sorry just saw this comment after merging.
The error could also be unrelated to schema change so should capture all but yeh can add in the message what schema version we are trying to validate with and what schema version is being sent in the metadata.

PR #1913, opened 2024-08-12
Add neighbour locations to review map

benjamineskola:
### Description of change

Expanding MapController to load more than one layer, and adding neighbour locations to be passed from Rails into stimulus in order to load them as leaflet layers.

Addresses are fetched by a background job to populate a field on the model with the lat/long as a GeoJson point.

These are then combined into three layers to display them on the map in three colours, for supportive/objection/neutral responses. These will be possible to hide and show individually or all together once we add a layer picker.

### Story Link

https://trello.com/c/XAFlIrLw/3115-overview-of-neighbours

### Screenshots

Screenshot 2024-08-12 at 17 38 50


benbaumann95 (commented):
This is looking really exciting 👏 . Glad we've finally decided to go with Leaflet.
As our map functionality becomes richer I think we should invest some time in writing specs so we can keep on top with all the changes and permutations.

db/migrate/20240812105008_add_latlong_to_neighbour.rb
```diff
@@ -0,0 +1,5 @@
+class AddLatlongToNeighbour < ActiveRecord::Migration[7.1]
+  def change
+    add_column :neighbours, :latlong, :jsonb
```
benbaumann95, line 3: Is it not worth using rgeo here for the way Leaflet wants to consume coordinates?
I think on planning application we store this as `factory.point(longitude, latitude)`

app/services/neighbour_lat_long_update_service.rb
```diff
@@ -0,0 +1,17 @@
+module NeighbourLatLongUpdateService
+  def self.call(neighbour, *args)
+    response = Apis::OsPlaces::Query.new.find_addresses(neighbour.address)
```
benbaumann95, line 3: Are we handling the case if this query fails or response returns with nothing?

app/models/neighbour.rb
```diff
@@ -18,6 +18,8 @@ class AddressValidationError < StandardError; end
 
   validate :validate_address_format
 
+  before_save :update_latlong, if: :address_changed?
```
benbaumann95, line 21: Think it would be useful to add a model spec when the address is added/updated that the `NeighbourLatLongUpdateJob` is called

PR #1929, opened 2024-08-27
Rails 7.2.2

EGiataganas:
### Description of change

Rails 7.2.1

### Story Link

https://trello.com/c/ZAd8mStM


benjamineskola:
It looks like we have a very old version of activerecord-postgis-adapter: 3.1.2 when the current is 9.0.2. I'm not sure why we have that version, which is older than BOPS itself, but I think this is the cause of the errors.

However, the latest version (9.0.2) only supports Rails 7.1, so we'll need to wait for them to release a new version.

EGiataganas:
Yes I am aware of that, there's an open [issue][1] on the activerecord-postgis-adapter gem. Last month they merged a PR but it looks like is still not fully supported. So it will need to wait

[1]:https://github.com/rgeo/activerecord-postgis-adapter/issues/402

benjamineskola:
Yep, just found the same; it looks like we could depend on the github version:

```ruby
gem "activerecord-postgis-adapter", github: "rgeo/activerecord-postgis-adapter"
```

But I suggest waiting for an actual release unless we consider this upgrade urgent.

.github/dependabot.yml
```diff
@@ -1,16 +1,21 @@
...
 updates:
-  - package-ecosystem: "bundler"
+  - package-ecosystem: bundler
     directory: "/"
     schedule:
-      interval: "daily"
-    open-pull-requests-limit: 0 # security only
-  - package-ecosystem: "docker"
+      interval: daily
+    open-pull-requests-limit: 10 # security only
```
benjamineskola, line 7: if we're changing this limit we should remove the comment, as it's no longer accurate: `0` means only security-related updates, anything `> 0` means all updates.

app/models/neighbour.rb
```diff
@@ -7,7 +7,7 @@ class AddressValidationError < StandardError; end
   has_many :neighbour_letters, dependent: :destroy
   has_many :neighbour_responses, dependent: :destroy
 
-  enum source: {manual_add: "manual_add", map_selection: "map_selection", sent_comment: "sent_comment"}
+  enum :source, {manual_add: "manual_add", map_selection: "map_selection", sent_comment: "sent_comment"}
```
pixeltrix, line 10: We can avoid the duplication here by doing this:

``` ruby
enum :source, %w[manual_add map_selection sent_comment].index_by(&:itself)
```

benjamineskola: That's useful to know, there were a few cases like this I think.

pixeltrix: Actually, this is better:

``` ruby
enum :source, %i[manual_add map_selection sent_comment].index_with(&:to_s)
```

That gives symbol keys and string values.

benjamineskola: Naturally we have one enum where the keys use snake case and the values use kebab case, and another where the keys are lowercase and the values uppercase. I'm not sure those are worth trying to deduplicate.

benbaumann95 (approved):

PR #1940, opened 2024-08-30
Adjust application types with schema v0.7.0

EGiataganas:
### Description of change

Adjust application types with schema v0.7.0

### Story Link

https://trello.com/c/cxm54Hh4


benbaumann95 (approved):

config/locales/odp.yml
```diff
@@ -53,8 +53,8 @@ en:
       - ['pa.part17', 'Prior Approval - Coal mining']
       - ['pa.part17.classB', 'Prior Approval - Other developments ancillary to mining operations']
       - ['pa.part17.classC', 'Prior Approval - Developments for maintenance or safety']
-      - ['pa.part17.classG', 'Prior Approval - Coal mining development by the Coal Authority for maintenance or safety']
-      - ['pa.part18.classA', 'Prior Approval - Specific Acts of Parliament or Local Orders']
+      - ['pa.part17.classG', 'Prior Approval - Coal mining development by the Coal Authority for maintence or safety']
```
benjamineskola, line 56: Has this misspelling been introduced by them or by us?

benjamineskola (changes requested):

config/locales/odp.yml
benjamineskola:
```suggestion
      - ['pa.part17.classG', 'Prior Approval - Coal mining development by the Coal Authority for maintenance or safety']
```

config/locales/odp.yml
```diff
@@ -24,7 +24,7 @@ en:
       - ['approval', 'Planning approval']
       - ['approval.reservedMatters', 'Approval of reserved matters']
       - ['complianceConfirmation', 'Written confirmation of compliance with a planning condition']
-      - ['environmentalImpact', 'Environmental Impact Decision']
+      - ['environmnentalImpact', 'Environmental Impact Decision']
```
benjamineskola, line 27:
```suggestion
      - ['environmentalImpact', 'Environmental Impact Decision']
```

config/locales/odp.yml
EGiataganas: I've basically copy and paste, I saw them as well

config/locales/odp.yml
EGiataganas: I've basically copy and paste, I saw them as well

config/locales/odp.yml
benjamineskola: It looks like they've corrected their spelling errors in `main`. I think we should adopt the corrected spellings; it's not going to make any sense to temporarily have a misspelling in our code.

Particularly in this case where it's only in the description, and not in the identifier itself.

EGiataganas: yes, that's why I am very keen to improve the process for things like that. so we can avoid human errors

EGiataganas:
@benjamineskola I've made the suggested changes which are introduced in PlanX main branch and not on `v.0.7.0`

PR #1946, opened 2024-09-09
Add policy classes in assessment area

benbaumann95:
### Description of change

- Use dynamic policies created in global admin

### Story Link

https://trello.com/c/LX8kHPTj/3169-add-new-assessment-area-using-dynamic-policy-data

### Decisions [OPTIONAL]

The existing functionality for assess against legislation is very indexed to the set of hardcoded policy references yml file and the previous schema set up.

It is going to be more trouble than it's worth to update the existing functionality so this is the first step to build in parallel and do a switch over once it's complete. This will just exist in staging for now

### DIFF
--- a/app/controllers/planning_applications/assessment/policy_areas/parts_controller.rb
+++ b/app/controllers/planning_applications/assessment/policy_areas/parts_controller.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module PlanningApplications
+  module Assessment
+    module PolicyAreas
+      class PartsController < BaseController
+        before_action :ensure_can_assess_planning_application
+        before_action :find_policy_parts
+
+        def index
+          respond_to do |format|
+            format.html
+          end
+        end
+
+        private
+
+        def find_policy_parts
+          @policy_parts = PolicySchedule.schedule_2.policy_parts
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/planning_applications/assessment/policy_areas/policy_classes_controller.rb
+++ b/app/controllers/planning_applications/assessment/policy_areas/policy_classes_controller.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module PlanningApplications
+  module Assessment
+    module PolicyAreas
+      class PolicyClassesController < BaseController
+        before_action :ensure_can_assess_planning_application
+        before_action :find_policy_parts
+        before_action :find_part, only: %i[new create]
+
+        def new
+          respond_to do |format|
+            format.html
+          end
+        end
+
+        def create
+          class_ids = params[:new_policy_classes].compact_blank
+
+          @part.new_policy_classes.where(id: class_ids).find_each do |policy_class|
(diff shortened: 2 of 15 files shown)


app/controllers/planning_applications/assessment/policy_areas/parts_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+        end
+
+        private
+
+        def ensure_can_assess_planning_application
+          render plain: "forbidden", status: :forbidden and return unless @planning_application.can_assess?
+        end
+
+        def find_policy_parts
+          @policy_parts = PolicySchedule.find_by_number(2).policy_parts
```
benjamineskola, line 23: What is the hardcoded 2 here? Is it something we could define a constant for to make it clearer?

app/controllers/planning_applications/assessment/policy_areas/parts_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+            format.html
+          end
+        end
+
+        private
+
+        def ensure_can_assess_planning_application
+          render plain: "forbidden", status: :forbidden and return unless @planning_application.can_assess?
+        end
+
```
benjamineskola, line 21: I think this can be removed as it's duplicating the base controller now

```suggestion
```

app/views/planning_applications/assessment/tasks/index.html.erb
```diff
@@ -32,6 +32,12 @@
 
       <%= render "assess_against_legislation" %>
 
+      <% unless Bops.env.production? %>
+        <% unless @planning_application.application_type.name == "planning_permission" %>
```
benjamineskola, line 36: How come planning permission specifically is excluded? Is this something we can check application type features for instead?

app/controllers/planning_applications/assessment/policy_areas/policy_classes_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+            @planning_application.planning_application_policy_classes.find_or_create_by!(new_policy_class_id: policy_class.id)
+          end
+
+          redirect_to planning_application_assessment_tasks_path(@planning_application), notice: t(".success")
+        end
+
+        private
+
+        def find_policy_parts
+          @policy_parts = PolicySchedule.find_by_number(2).policy_parts
```
benjamineskola, line 30: Same as above

app/controllers/planning_applications/assessment/policy_areas/parts_controller.rb
benbaumann95: good spot!

app/views/planning_applications/assessment/tasks/index.html.erb
benbaumann95: Just a copy + paste of the previous logic. We don't currently capture this in application type features so we'd need to add something to the choose features page. I think this can be done in a following PR

benjamineskola (approved):

PR #1949, opened 2024-09-12
Assess policy sections

benbaumann95:
### Description of change

**Assess policy sections**
- Mark as complies, does not comply and to be determined
- Ability to save and mark as complete or save and come back later

### Story Link

https://trello.com/c/DrGoMWyy/3170-view-and-respond-to-policy-sections-when-assessing-a-policy-class

### DIFF
--- a/app/models/new_policy_class.rb
+++ b/app/models/new_policy_class.rb
@@ -3,6 +3,7 @@
 class NewPolicyClass < ApplicationRecord
   belongs_to :policy_part
   has_many :policy_sections, -> { order(:section) }, dependent: :restrict_with_error
+  has_many :planning_application_policy_sections, through: :policy_sections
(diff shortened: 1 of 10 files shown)


app/models/new_policy_class.rb
```diff
@@ -22,4 +23,10 @@ def menu
       pluck(:id, :section, :name)
     end
   end
+
+  def all_policies_are_determined
+    return if planning_application_policy_sections.none?(&:to_be_determined?)
+
+    errors.add(:base, :policies_to_be_determined)
+  end
```
benjamineskola, line 31: Is this used? There's a similar method in the `Review` model, but this version of it doesn't seem to get added as a validator unless I'm missing something.

benbaumann95: No, not used. Good spot 👍

benjamineskola (approved):

PR #1955, opened 2024-09-18
Allow revocation of API keys

benjamineskola:
### Description of change

- **Add revoked_at column to api_users**
- **Limit api users to unrevoked by default**

### Story Link

https://trello.com/c/VNmws12n/3191-add-revokedat-column-to-apiuser

### Decisions [OPTIONAL]

I've used a default scope here, with caution, because we'll almost always want only unrevoked users. I've added scopes for showing revoked users and all users, just for convenience and as a reminder that they exist.

### DIFF
--- a/app/models/api_user.rb
+++ b/app/models/api_user.rb
@@ -19,4 +19,8 @@ class ApiUser < ApplicationRecord
       active.find_by(token: token)
     end
   end
+
+  def revoke!
+    touch(:revoked_at)
+  end
 end
--- a/app/models/api_user.rb
+++ b/app/models/api_user.rb
@@ -8,12 +8,15 @@ class ApiUser < ApplicationRecord
 
   has_secure_token :token, length: 36
 
+  scope :active, -> { where(revoked_at: nil) }
+  scope :revoked, -> { where.not(revoked_at: nil) }
+
   validates :name, presence: true, uniqueness: {scope: :local_authority_id}
   validates :file_downloader, store_model: {allow_blank: true}
 
   class << self
     def authenticate(token)
-      find_by(token: token)
+      active.find_by(token: token)
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/models/api_user.rb
```diff
@@ -8,6 +8,10 @@ class ApiUser < ApplicationRecord
 
   has_secure_token :token, length: 36
 
+  default_scope { where(revoked_at: nil) }
```
benbaumann95, line 11: Shall we avoid using default scope given the documents?

app/models/api_user.rb
```diff
@@ -8,6 +8,10 @@ class ApiUser < ApplicationRecord
...
+  scope :with_revoked, -> { unscope(where: :revoked_at) }
```
benbaumann95, line 12: How about just an `active` and `revoked` scope?

scope :active, -> { where(revoked_at: nil) }
scope :revoked, -> { where.not(revoked_at: nil) }

Then can update the authenticate class method with `active.find_by(token: token)`

app/models/api_user.rb
benjamineskola: For future reference: adding the default scope is the pattern followed by gems like [`paranoia`](https://github.com/rubysherpas/paranoia) and [`soft_deletion`](https://github.com/grosser/soft_deletion) which was what I followed here. (Actually via [this medium article](https://medium.com/@cathmgarcia/soft-deletion-in-ruby-on-rails-a1d65d0172ab).)

I think it's probably safer when applied to a deleted/revoked state than for a fuzzier concept of what records to access, as we have in `document` — I think for documents it really is just that some views want to exclude some types and others don't, so default_scope is too blunt an instrument for it. We really would _almost never_ want to access a revoked key.

However, again, this is in practice probably one single method call in the authentication handler, so we're not gaining anything from the scopes anyway.

PR #1957, opened 2024-09-18
Add assessment against legislation as an application type feature

benjamineskola:
### Description of change

Replacing some conditionals based on application type with an application type feature associated with the appropriate types.

At the same time enabling separate conditionals for two features that were previously considered mutually exclusive.

### Story Link

https://trello.com/c/tqG7H2Tw/3173-add-assess-against-legislation-as-choose-a-feature-option-in-global-admin

https://trello.com/c/KL5uvleQ/3164-able-to-assess-against-policies-for-prior-approval

### DIFF
--- a/engines/bops_config/app/views/bops_config/application_types/features/edit.html.erb
+++ b/engines/bops_config/app/views/bops_config/application_types/features/edit.html.erb
@@ -21,6 +21,8 @@
 
       <%= form.fields_for :features, @application_type.features do |ff| %>
         <%= ff.govuk_fieldset legend: {text: t("application_type_features.legends.check_application_details"), size: "s"} do %>
+          <%= ff.govuk_check_box :assess_against_policies, 1, 0, multiple: false, label: {text: t("application_type_features.labels.assess_against_policies")} %>
+          <%= ff.govuk_check_box :considerations, 1, 0, multiple: false, label: {text: t("application_type_features.labels.considerations")} %>
           <%= ff.govuk_check_box :informatives, 1, 0, multiple: false, label: {text: t("application_type_features.labels.informatives")} %>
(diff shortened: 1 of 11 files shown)


benbaumann95 (approved):
Looks great.
Mind if we just add a tiny system spec (in `bops/engines/bops_config/spec/system/application_types_spec.rb`) where we check in the application type summary if we select one option, for example: 

```
expect(page).not_to have_selector("li", text: "Assess policies and guidance")
expect(page).to have_selector("li", text: "Assess considerations")
```

engines/bops_config/app/views/bops_config/application_types/features/edit.html.erb
```diff
@@ -21,6 +21,8 @@
 
       <%= form.fields_for :features, @application_type.features do |ff| %>
         <%= ff.govuk_fieldset legend: {text: t("application_type_features.legends.check_application_details"), size: "s"} do %>
```
benjamineskola, line 23: I wonder if this list might be better generated by iterating over an array of symbols, if we create a method on `ApplicationType` that returns all the available features.

engines/bops_config/app/views/bops_config/application_types/show.html.erb
```diff
@@ -59,9 +59,15 @@
             row.with_value { %>
               <p class="govuk-body"><strong>Application details</strong></p>
               <ul class="govuk-list govuk-list--bullet">
```
benjamineskola, line 61: This one too.

engines/bops_config/app/views/bops_config/application_types/features/edit.html.erb
benbaumann95: Yeah it's the same boilerplate for each option so iterating makes sense here

PR #1960, opened 2024-09-19
Skip flaky drag and drop informatives test

benbaumann95:
### Description of change

This is now failing more frequently than not so skip for now until we can reassess this test set up.


spec/system/planning_applications/assessing/adding_informatives_spec.rb
```diff
@@ -173,7 +173,7 @@
       let!(:informative_two) { create(:informative, informative_set:, title: "Title 2", text: "Text 2", position: 2) }
       let!(:informative_three) { create(:informative, informative_set:, title: "Title 3", text: "Text 3", position: 3) }
 
-      it "I can drag and drop to sort the informatives", :capybara do
+      it "I can drag and drop to sort the informatives", :capybara, pending: "flaky" do
```
benjamineskola, line 176:
```suggestion
      it "I can drag and drop to sort the informatives", :capybara, skip: "flaky" do
```

`pending` will fail if the test passes (it assumes when it passes you've implemented the thing so need to remove the `pending`). We need `skip` for tests that only fail some of the time.

benbaumann95: Woops, thanks

benjamineskola (approved):

PR #1977, opened 2024-10-04
Drop past applications

EGiataganas:
### Description of change

- Drop past_applications from assessment_details
- Drop any record related to past_applications category
- Drop additional_information from assessment_details

### Story Link

https://trello.com/c/I5dxeLSh


benjamineskola (approved):
Nice! Looks like this simplifies some things and removes special cases.

PR #1994, opened 2024-10-17
Add ability to search via address

benbaumann95:
### Description of change

Enable search by address on API and planning applications dashboard
- Add virtual address_search column to combine address fields and add index to this column to improve search performance
- Add this to the API and dashboard search

### Story Link

https://trello.com/c/ix7VwSXL/2944-the-ability-to-search-based-on-address
https://trello.com/c/Cd9TnjmO/2948-search-applications-by-address-on-the-application-dashboard

### DIFF
--- a/app/services/planning_application_search.rb
+++ b/app/services/planning_application_search.rb
@@ -108,13 +108,20 @@ class PlanningApplicationSearch
   end
 
   def records_matching_query
-    records_matching_reference.presence || records_matching_address.presence || records_matching_description
+    records_matching_reference.presence || records_matching_address_search.presence || records_matching_description
   end
 
   def records_matching_reference
     current_planning_applications.where(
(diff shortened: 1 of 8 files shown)


app/services/planning_application_search.rb
```diff
@@ -145,6 +163,10 @@ def query_terms
     @query_terms ||= query.split.join(" | ")
   end
 
+  def address_query_terms
+    @address_query_terms ||= query.split.map { |term| term }.join(" & ")
+  end
+
```
benjamineskola, line 169: I feel like this could be a local variable inside the `records_matching_address` method — it doesn't seem like it's used anywhere else, or used more than once per query — unless I'm missing something?

benjamineskola: Also I don't think `map { |term| term }` does anything, does it?

PR #1995, opened 2024-10-18
Track history of neighbour letters

benjamineskola:
### Description of change

Tracking a history of batches of letters, so that groups of letters sent at the same time as each other can be seen, including the recipients and the text.

### Story Link

https://trello.com/c/8eLmtFtA/3232-store-history-of-neighbour-letters

### Known issues [OPTIONAL]

This currently only applies to letters sent as part of a consultation, and conversely we can't yet send email updates as part of the consultation (only committee decisions). There's [a card for refactoring to make that possible](https://trello.com/c/PRmqbLxM/3247-refactor-lettersendingservice-to-send-emails-as-needed).

### DIFF
--- a/app/services/letter_sending_service.rb
+++ b/app/services/letter_sending_service.rb
@@ -5,13 +5,14 @@ require "notifications/client"
 class LetterSendingService
   attr_reader :neighbour, :consultation, :letter_content, :resend_reason
 
-  def initialize(neighbour, letter_content, letter_type:, resend_reason: nil)
+  def initialize(neighbour, letter_content, letter_type:, resend_reason: nil, batch: nil)
     @local_authority = neighbour.consultation.planning_application.local_authority
     @neighbour = neighbour
     @consultation = neighbour.consultation
     @letter_content = letter_content
     @resend_reason = resend_reason
     @letter_type = letter_type
+    @batch = batch
   end
 
   def deliver!
@@ -22,6 +23,10 @@ class LetterSendingService
     ActiveRecord::Base.transaction do
       letter_record.save!
       consultation.start_deadline if consultation_letter?
+    end
+
+    if @batch
+      @batch.neighbour_letters << letter_record
     end
 
     if resend_reason.present?
--- a/app/services/letter_sending_service.rb
+++ b/app/services/letter_sending_service.rb
@@ -3,19 +3,17 @@
 require "notifications/client"
 
 class LetterSendingService
-  attr_reader :neighbour, :consultation, :letter_content, :resend_reason
+  attr_reader :consultation, :letter_content, :resend_reason
 
-  def initialize(neighbour, letter_content, letter_type:, resend_reason: nil, batch: nil)
-    @local_authority = neighbour.consultation.planning_application.local_authority
-    @neighbour = neighbour
-    @consultation = neighbour.consultation
+  def initialize(letter_content, consultation:, letter_type:, resend_reason: nil)
+    @consultation = consultation
+    @local_authority = consultation.planning_application.local_authority
     @letter_content = letter_content
     @resend_reason = resend_reason
     @letter_type = letter_type
-    @batch = batch
   end
 
-  def deliver!
+  def deliver!(neighbour)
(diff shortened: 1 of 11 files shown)


app/services/letter_sending_service.rb
```diff
@@ -3,27 +3,31 @@
...
-    @neighbour = neighbour
-    @consultation = neighbour.consultation
+  def initialize(letter_content, letter_type:, resend_reason: nil)
     @letter_content = letter_content
     @resend_reason = resend_reason
     @letter_type = letter_type
   end
 
-  def deliver!
+  def deliver!(neighbour)
```
benbaumann95, line 14: We have quite a lot of updates / db transactions happening in this method. Is it worth putting everything in the ActiveRecord::Base.transaction in case we need to rollback if one step fails?

app/services/letter_sending_service.rb
```diff
@@ -46,6 +50,14 @@ def deliver!
     neighbour.touch :last_letter_sent_at
   end
 
+  def deliver_batch!(neighbours)
+    @batch = neighbours.first.consultation.neighbour_letter_batches.new(text: letter_content)
+
+    neighbours.each do |neighbour|
+      deliver!(neighbour)
+    end
+  end
```
benbaumann95, line 59: Can we add a rescue block here to continue the batch processing even if a neighbour fails and send this error to AppSignal

```ruby
rescue => e
    # Log and continue processing the rest of the batch
    Appsignal.send_error(e)
    next
  end
```

app/services/letter_sending_service.rb
benbaumann95: Think we can also add more specific error handling i.e `ActiveRecord::RecordInvalid` to this method

app/services/letter_sending_service.rb
benbaumann95: Might be `report_error` now actually

app/services/letter_sending_service.rb
benbaumann95:
```ruby
def deliver!(neighbour)

  @local_authority = neighbour.consultation.planning_application.local_authority
  @consultation = neighbour.consultation
  letter_record = NeighbourLetter.new(neighbour:, text: letter_content, resend_reason:)

    begin
      letter_record.save!
      consultation.start_deadline if consultation_letter?

      @batch.neighbour_letters << letter_record if @batch

      if resend_reason.present?
        letter_content.prepend "# Application updated\nThis application has been updated. Reason: #{resend_reason}\n\n"
      end

      personalisation = { message: letter_content, heading: }.merge(neighbour.format_address_lines)

      response = client.send_letter(
        template_id: @local_authority.letter_template_id,
        personalisation:
      )

      update_letter!(letter_record, response)
      neighbour.touch :last_letter_sent_at

    rescue ActiveRecord::RecordInvalid => e
      Appsignal.report_error(e)
    rescue Notifications::Client::RequestError => e
      letter_record.update!(status: "rejected", failure_reason: e.message)
      Appsignal.report_error(e)
    rescue StandardError => e
      # General fallback
      letter_record.update!(status: "rejected", failure_reason: e.message)
      Appsignal.report_error(e)
    end
  end
end

def deliver_batch!(neighbours)
  @batch = neighbours.first.consultation.neighbour_letter_batches.new(text: letter_content)
  neighbours.each do |neighbour|
    deliver!(neighbour)
  rescue => e
    Appsignal.report_error(e)
    next
  end
end

```

benbaumann95: Had a go at writing that method. Still seems like it's doing too much and perhaps we need to rollback in the rescue blocks if we don't want the partial updates going ahead?

benjamineskola: Wouldn't moving everything inside the transaction actually backfire slightly? It would mean, for example, that if a letter failed to send, it wouldn't get added to the batch, because that would get rolled back.

I'm also not sure it would be necessary to handle the errors in `deliver_batch!`, because every exception is caught inside `deliver!`.

benjamineskola: You're right though that I'm not completely sure about the error handling.

benbaumann95 (approved):

PR #1999, opened 2024-10-23
Add local authority management of API tokens

pixeltrix:
### Description of change

As more LPAs sign up to BOPS manually concierging API tokens becomes more of a roadblock and especially so when dealing with revocations. To address this add self-service generation and revocation of tokens to the LPA admin portal.

### Story Link

https://trello.com/c/zNruHDTe
https://trello.com/c/10i6BC3A

### Screenshots

southwark bops localhost_3000_admin_tokens (1)

southwark bops localhost_3000_admin_tokens (2)

southwark bops localhost_3000_admin_tokens_new

southwark bops localhost_3000_admin_tokens

### Known issues

We should hash the token in the database to protect against mass extraction like happened with Heroku and their cache of GitHub oauth tokens. However it seemed like gold-plating to do it in this PR.

### DIFF
--- a/app/models/api_user.rb
+++ b/app/models/api_user.rb
@@ -22,7 +22,7 @@ class ApiUser < ApplicationRecord
 
   class << self
     def authenticate(token)
-      active.find_by(token: token)
+      active.find_by(token: token).tap { |user| user.try(:touch, :last_used_at) }
     end
 
     def generate_unique_secure_token(length: 36)
--- a/app/models/api_user.rb
+++ b/app/models/api_user.rb
@@ -4,9 +4,12 @@ require "base64"
 require "zlib"
 
 class ApiUser < ApplicationRecord
+  include StoreModel::NestedAttributes
+
   TOKEN_FORMAT = /\Abops_[a-zA-Z0-9]{36}[-_a-zA-Z0-9]{6}\z/
 
   attribute :file_downloader, FileDownloaders.to_type
+  alias_attribute :value, :token
 
   belongs_to :local_authority
   has_many :audits, dependent: :nullify
@@ -15,10 +18,13 @@ class ApiUser < ApplicationRecord
 
   scope :active, -> { where(revoked_at: nil) }
   scope :revoked, -> { where.not(revoked_at: nil) }
+  scope :by_name, -> { reorder(:name) }
 
-  validates :name, presence: true, uniqueness: {scope: :local_authority_id}
-  validates :token, format: {with: TOKEN_FORMAT}
-  validates :file_downloader, store_model: {allow_blank: true}
+  validates :name, presence: true, uniqueness: {scope: :local_authority_id, conditions: -> { where(revoked_at: nil) }}
+  validates :token, format: {with: TOKEN_FORMAT}, on: :create
+  validates :file_downloader, store_model: true
+
+  accepts_nested_attributes_for :file_downloader
 
   class << self
     def authenticate(token)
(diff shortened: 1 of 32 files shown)


app/models/api_user.rb
```diff
@@ -1,7 +1,15 @@
...
+require "base64"
+require "zlib"
+
 class ApiUser < ApplicationRecord
+  include StoreModel::NestedAttributes
+
+  TOKEN_FORMAT = /\Abops_[a-zA-Z0-9]{36}[-_a-zA-Z0-9]{6}\z/
+
   attribute :file_downloader, FileDownloaders.to_type
+  alias_attribute :value, :token
```
pixeltrix, line 12: Ideally we'd rename this model as `ApiToken` as it's not really a user and to that aim I've aliased the `token` to `value` so we don't have `token.token` where it's used.

app/models/api_user.rb
```diff
@@ -10,17 +18,43 @@ class ApiUser < ApplicationRecord
...
+  accepts_nested_attributes_for :file_downloader
 
   class << self
     def authenticate(token)
-      active.find_by(token: token)
+      active.find_by(token: token).tap { |user| user.try(:touch, :last_used_at) }
+    end
+
+    def generate_unique_secure_token(length: 36)
+      super.then { |token| "bops_#{token}#{checksum(token)}" }
```
pixeltrix, line 35: GitHub suggests adding the checksum so you can validate it without making a request but unless I'm missing something an attacker could generate the valid checksum anyway. Better option maybe to create a signed secret value that we can change the key to and invalidate all tokens if there were to be a mass leak.

engines/bops_admin/app/views/bops_admin/tokens/_file_downloader.html.erb
```diff
@@ -0,0 +1,20 @@
+<% if file_downloader.errors.any? %>
+  <a name="token-file-downloader-field-error"></a>
+<% end %>
```
pixeltrix, line 3: The automatic link generation for error messages doesn't seem to work for StoredModel nested fields and didn't want to go down a rabbit hole.

engines/bops_admin/app/controllers/bops_admin/tokens_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+    def edit
+      respond_to do |format|
+        format.html
+      end
+    end
+
+    def create
+      respond_to do |format|
+        if @token.save
+          format.html { render :show }
```
pixeltrix, line 30: Normally you'd redirect here but we render so we can show the token a single time

app/models/api_user.rb
benjamineskola: Yeah I think you're right, it would in theory allow us to check that the _format_ was valid but that's hardly worth the saving — unless I'm missing something. (It's not like ISBNs or credit card numbers where checking for misentered numbers might be useful.)

benbaumann95 (commented):
Did we want to remove the token displayed in `/admin/profile` ?
We are displaying `<%= current_local_authority.default_api_token %>`

benbaumann95 (approved):

app/models/api_user.rb
```diff
@@ -10,17 +18,43 @@ class ApiUser < ApplicationRecord
...
+    end
+
+    def valid_token?(token)
+      TOKEN_FORMAT.match?(token) && checksum(token[5..40]) == token[41..46]
+    end
```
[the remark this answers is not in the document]
pixeltrix: It will be used in the future, once we're happy we've updated everyone's tokens

EGiataganas (approved):

benjamineskola (approved):

PR #2001, opened 2024-10-23
Track and highlight notify config errors

benjamineskola:


EGiataganas (approved):

app/services/letter_sending_service.rb
```diff
@@ -16,6 +16,8 @@ def initialize(letter_content, consultation:, letter_type:, resend_reason: nil)
   def deliver!(neighbour)
     return if resend_reason.nil? && NeighbourLetter.find_by(neighbour:).present? && consultation_letter?
 
+    return if @local_authority.notify_error_status.present?
```
benbaumann95, line 19: What happens when the user tries to send a letter if this conditional is met - do they get any feedback?
Perhaps we need to alert the user that we are aborting this request in the case of this notify error and give them an action they can take

benjamineskola: Good point. I think I was expecting that it would show up in the list of unsent letters (same as any Notify error would do) but on reflection I'd want to check that it actually did so, at least.

benjamineskola: OK, took a second look and moved the check around, so now those letters will be saved and show up in the list but won't send. Perhaps adding the error message there, where the sending statuses show, might work?

benjamineskola: One thing to investigate will be how/whether these messages are resent. I think currently we just keep on trying for a certain number of times but I'm not sure whether that will be sufficient for this situation, when the configuration might get fixed after the retries have expired.

benbaumann95 (approved):

PR #2006, opened 2024-10-28
Report error in planning application dependency job to Appsignal

benbaumann95:
### Description of change

- Applications may get caught in a pending state due to an error in the PlanningApplicationDependencyJob.
Add retry and report error to Appsignal to help with debugging and proactively monitoring the situation

### DIFF
--- a/engines/bops_api/app/jobs/bops_api/planning_application_dependency_job.rb
+++ b/engines/bops_api/app/jobs/bops_api/planning_application_dependency_job.rb
@@ -8,6 +8,14 @@ module BopsApi
     queue_as :submissions
     discard_on ActiveJob::DeserializationError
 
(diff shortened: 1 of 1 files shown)


benjamineskola (approved):

benjamineskola (commented):
Just checking: should the Faraday error handler come first?

If Faraday::TimeoutError inherits from StandardError will that cause it to be handled by that one instead? Actually the result will be the same in that case, but then do we need both?

(I don't know the behaviour of retry_on so maybe this way is fine.)

benbaumann95:
> Just checking: should the Faraday error handler come first?
> 
> If Faraday::TimeoutError inherits from StandardError will that cause it to be handled by that one instead? Actually the result will be the same in that case, but then do we need both?
> 
> (I don't know the behaviour of retry_on so maybe this way is fine.)

You might be right but will just merge and monitor the behaviour. Can remove if unnecessary

PR #2008, opened 2024-10-29
Map neighbour's response correctly

EGiataganas:
### Description of change

Remove unnecessary =


benbaumann95 (approved):

benjamineskola (changes requested):
This will break.

PR #2010, opened 2024-10-31
Bops accordion

EGiataganas:
### Description of change

BOPS-Accordion redesign

### Story Link

https://trello.com/c/zZjmHi7i

### DIFF
--- a/app/assets/stylesheets/components/_accordion.scss
+++ b/app/assets/stylesheets/components/_accordion.scss
@@ -1,0 +1,279 @@
+.bops-accordion {
+  @include govuk-responsive-margin(6, "bottom");
+}
+
+.govuk-frontend-supported {
+  .govuk-accordion__section-toggle {
+    display: none;
+  }
+
+  .govuk-accordion__section-heading-text {
+    margin-bottom: 0px;
+  }
+}
+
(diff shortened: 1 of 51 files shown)


app/assets/stylesheets/components/_accordion.scss
benjamineskola, line 1: I'm concerned that a lot of this will clash with the existing `govuk-accordion` classes. 

I wonder if it might be best to rename all of the classes to `bops-accordion`, _or_ figure out if we can use only those classes which deviate from the existing `govuk-accordion` component.

app/components/planning_applications/search_component.html.erb
benjamineskola, line 1: Could we avoid having the HTML structure duplicated everywhere we use the component, please? We've defined a component class, after all.

benjamineskola (changes requested):

app/assets/stylesheets/components/_accordion.scss
benjamineskola: It looks like you've reverted the change to scope this within the `bops-accordion` selector — but if we can't make this work with proper scoping it would be good to ensure that this at least doesn't break existing default accordions.

PR #2027, opened 2024-11-08
Don't exclude pre-review applications for reviewers

benjamineskola:
If assigned to a user it's probably for a reason. This filter makes it difficult for a reviewer to see all the cases assigned to them, as it will default to only certain states if filters aren't set. (It might work if they select every status in the filter options (untested) but if they select none it defaults.)

This brings the reviewer behaviour in line with the assessor behaviour: an assessor is always (I think) shown any application assigned to them in the 'my applications' tab.


benbaumann95 (approved):
Shall we just confirm this change with an officer before merging

PR #2034, opened 2024-11-14
Allow v0.7.1 ODP submissions with an alternative schema url

pixeltrix:
What's being sent doesn't match what's on GitHub.


benjamineskola (approved):

pixeltrix:
OSL will be fixing PlanX to use the correct schema URL

PR #2039, opened 2024-11-18
Review assessment summaries

benbaumann95:
### Description of change

Review assessment summaries
- Move all assessment details into accordion with individual forms
- Allow reviewer to accept or return with comments
- When reviewer returns, validate that a comment is entered before saving as complete
- If returned, assessor will have to submit a new assessment and this back and forth interaction can continue indefinitely. Show previous summaries /comments within the task

### Story Link

https://trello.com/c/CyfSt5Ef/12-redesign-of-review-assessment-summaries

### Screenshots

Screenshot 2024-11-18 at 15 08 43
--
Screenshot 2024-11-18 at 15 09 10
--
Screenshot 2024-11-18 at 15 08 33


app/views/planning_applications/review/assessment_details/_form.html.erb
```diff
@@ -1,94 +1,116 @@
...
-            StatusTags::Reviewing::AssessmentDetailComponent.new(
-              planning_application: planning_application,
-              assessment_detail: form.object.send(assessment_detail)
-            )
-          ) %>
-      <% if assessment_detail == "site_description" %>
-        <p class="govuk-body govuk-!-margin-top-2"><%= govuk_link_to "View site on Google Maps", map_link(@planning_application.full_address), new_tab: true %></p>
-      <% end %>
-      <% if assessment_detail == "consultation_summary" && @consultation&.consultees&.any? %>
+<%= bops_task_accordion do |accordion| %>
```
EGiataganas, line 1: Could we leave this as just being a form and move the `bops_task_accordion` in the `app/views/planning_applications/review/tasks/_review_assessment.html.erb`?

benjamineskola: I think if this isn't _only_ a single form we should avoid using the `_form.html.erb` conventional name for it, since `render "form"` won't behave in the same way as the Rails default arrangement.

app/views/planning_applications/assessment/assessment_details/neighbour_summary/new.html.erb
```diff
@@ -69,7 +69,7 @@
               :"#{tag}",
               label: {text: "Summary of #{tag.to_s.humanize.downcase} comments", size: "m"},
               rows: 10,
-              value: (@assessment_detail.entry[/(?<=#{tag.to_s.humanize}:)(.*)/] unless @assessment_detail.entry.nil?)
+              value: @rejected_assessment_detail&.entry&.[](/(?<=#{tag.to_s.humanize}:)(.*)/)
```
benjamineskola, line 72: this line isn't super obvious to me 

first off, is it intentional that we've gone from `[/regex/]` to `[](/regex/)`? I tested briefly and this doesn't seem to work but I might be missing something. adding the `&` seems also to be invalid syntax.

secondly, i'm guessing this is meant to extract the part of the entry that follows the tagged name and a colon. but i'm not a fan of interpolating a variable into the regex. would it actually work to just use `entry&.split("#{tag.to_s.humanize}:")[1]` for example?

finally maybe extracting this line into a variable would help readability

app/views/planning_applications/review/tasks/index.html.erb
```diff
@@ -36,20 +36,26 @@
 <% if @planning_application.application_type.consultation? %>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
-        <%= render "review_consultation" %>
+      <%= render "planning_applications/review/tasks/review_consultation" %>
```
benjamineskola, line 39: I wonder if we can make some of these `render` calls less longwinded by moving the partials into the right place for it to detect them without a full path?

benbaumann95: If we are building out separate controllers for these forms, I think when rendering the page again due to an error we'll need the full context of the partial.

app/views/planning_applications/review/assessment_details/_form.html.erb
benbaumann95: Yeah I'll move all this into "review_assessment_summaries"

benbaumann95: Moved the accordion logic to there now which makes more sense 👍

benjamineskola (approved):

PR #2040, opened 2024-11-18
Move CIL review to new layout

benjamineskola:
### Description of change

Move CIL liability review in the new review workflow layout.

### Story Link

https://trello.com/c/41duAr9g/19-redesign-of-review-for-check-community-infrastructure-levy-cil

### Screenshots

Screenshot 2024-11-19 at 15 12 31


app/views/planning_applications/review/tasks/_review_assessment.html.erb
```diff
@@ -1,17 +1,19 @@
...
-          StatusTags::BaseComponent.new(
-            status: (@planning_application.audits.review_cil_liability.any? ? :complete : :not_started)
-          )
-        ) %>
-    </div>
-  </li>
-</ul>
+<%= bops_task_accordion(id: "review-assessment") do |accordion| %>
+  <% accordion.with_heading(text: "Review assessment") %>
+  <% accordion.with_section(id: "check-cil", expanded: false) do |section| %>
```
benbaumann95, line 3: Can we check for errors to determine whether this section is expanded or not. What happens if a comment is blank when submitting this form?

EGiataganas: it should also have an error if someone clicks`Save and mark as complete` without selecting Yes or No, right?

benjamineskola: @EGiataganas I'm not sure. There's currently a test that ensures the assessor can leave it blank, and the reviewer can continue without changing it. So I presume this is intended behaviour.

EGiataganas: Ok if that's the case, it can be changed at a later stage

app/views/planning_applications/review/cil_liability/_form.html.erb
```diff
@@ -3,6 +3,8 @@
       class: "govuk-!-margin-top-7" do |form| %>
   <%= form.govuk_error_summary %>
 
+  <%= render "planning_applications/review/cil_liability/audit_trail" %>
+
```
EGiataganas, line 7: you could change line 1 to be

```ruby
<%= form_with model: planning_application,
      url: planning_application_review_cil_liability_url(planning_application),
````
as you pass the reference of planning_application

benjamineskola: I've removed the unnecessary local assignment instead, since this can simply use the instance variable app/views/planning_applications/review/tasks/_review_assessment.html.erb

PR #2043, opened 2024-11-19
Small fixes to neighbour letter history

benjamineskola:
Update accordion to new version and bring some small things inline with the design

https://trello.com/c/mqUkdGnL/5-view-a-copy-of-neighbour-letters

Screenshot 2024-11-19 at 17 08 37

### DIFF
--- a/app/views/planning_applications/neighbour_letter_batches/index.html.erb
+++ b/app/views/planning_applications/neighbour_letter_batches/index.html.erb
@@ -28,7 +28,14 @@
           Date sent: <%= batch.created_at.to_date.to_fs %>
         <% end %>
         <% section.with_footer do %>
-          <%= govuk_link_to "Download letter", planning_application_consultation_neighbour_letter_batch_path(@planning_application, batch, format: "pdf") %>
+          <div class="display-flex" style="justify-content: space-between">
+            <p class="govuk-body">
+              Response period: <%= @planning_application.consultation.consultee_response_period %> days
+            </p>
+            <p class="govuk-body">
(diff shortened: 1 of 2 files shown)


app/views/planning_applications/neighbour_letter_batches/index.html.erb
```diff
@@ -14,30 +14,46 @@
...
-        <%= govuk_table do |table|
-              table.with_head do |head|
-                head.with_row do |row|
-                  row.with_cell(text: "Address")
-                  row.with_cell(text: "Sent by")
+  <%= bops_task_accordion(id: "neighbour-letter-batches") do |accordion| %>
+    <% accordion.with_heading(text: "Neighbour letters") %>
+    <% @planning_application.consultation.neighbour_letter_batches.each_with_index do |batch, i| %>
+      <% accordion.with_section do |section| %>
+        <% section.with_heading(text: "Neighbour letter #{i + 1}") %>
```
pixeltrix, line 26: I wonder if this is something we could add to the accordion component - a running counter of section number, e.g.

``` erb
<% section.with_heading(text: "Neighbour letter %{counter}") %>
```

Obviously not as part of this PR, but it's likely others will need something like it.

pixeltrix (approved):

app/views/planning_applications/neighbour_letter_batches/index.html.erb
benjamineskola: In this case it would better be implemented as a property of the letter because it's also used in the CSV.

PR #2045, opened 2024-11-21
Update recommendation to committee review

benbaumann95:
### Description of change

Update recommendation to committee review

### Story Link

https://trello.com/c/LGcQv3Ot/22-redesign-of-recommendation-to-committee

### Screenshots

Screenshot 2024-11-21 at 13 14 16


app/components/reviewer_comment_component.html.erb
```diff
@@ -3,7 +3,7 @@
     <p class="govuk-body govuk-!-font-weight-bold govuk-!-margin-bottom-0">
       Reviewer comment
     </p>
-    <p class="govuk-hint govuk-!-font-size-16 govuk-!-margin-top-1">Sent on <%= comment.created_at.to_fs %> by <%= comment.user.name %></p>
+    <p class="govuk-hint govuk-!-font-size-16 govuk-!-margin-top-1">Sent on <%= (comment.try(:reviewed_at) || comment.created_at).to_fs %> by <%= comment.user.name %></p>
```
benjamineskola, line 6: Is `try` necessary here? `comment.reviewed_at` should be fine, I think — or do we have multiple comment models and not all of them have `reviewed_at`?

benjamineskola: Would it be better to have this logic in the model? e.g, have a `sent_at` method that's an alias for `reviewed_at` in types that have it, and for `created_at` in types that don't? Or even just alias `reviewed_at` to `created_at` for the types that don't. That way we can just call a single method in the view.

benbaumann95: Thing is, it's not always the comment model. Various things are being passed to this component including the comment model and others with a comment attribute

benbaumann95: I'll add a method on the component

benjamineskola: Took a quick look and I think it'd be ImmunityDetail and AssessmentDetail that we'd need to handle (as well as Review) — but you're right, moving it to the component is the simplest.

benbaumann95: Yep, only that component needs to know about this so added
```
  def sent_at
    (comment.try(:reviewed_at) || comment.created_at).to_fs
  end
  ```

benjamineskola (approved):

PR #2047, opened 2024-11-22
Move heads of terms review into new layout

benjamineskola:
### Description of change

Move heads of terms review into new review layout.

### Story Link

https://trello.com/c/Ae4hmaT1/26-redesign-of-review-heads-of-terms

### Screenshots


app/views/planning_applications/review/heads_of_terms/_form.html.erb
```diff
@@ -9,44 +8,23 @@
...
-            <% end %>
-          <% end %>
-        <% end %>
-
         <%= review_form.govuk_radio_button(
               :action, :rejected,
               checked: current_review.rejected?,
-              label: {text: "Return to officer with comment"},
-              disabled: !editable
+              label: {text: "Return to officer with comment"}
```
EGiataganas, line 17: new wording is `Return to officer`

EGiataganas (approved):

benbaumann95 (approved):

app/views/planning_applications/review/tasks/_review_conditions.html.erb
```diff
@@ -1,28 +1,56 @@
 <%= bops_task_accordion(id: "review-conditions-section") do |accordion| %>
   <% accordion.with_heading(text: "Review conditions") %>
+  <% if @planning_application.review_heads_of_terms? %>
+    <% accordion.with_section(id: "review-heads-of-terms", expanded: false) do |section| %>
```
benbaumann95, line 4: Should this be `expanded: false` ?

benbaumann95: Think we want to add this conditionally if an error is present or at least not hardcode this value

benbaumann95 (changes requested):

benbaumann95 (approved):

app/controllers/planning_applications/review/heads_of_terms_controller.rb
```diff
@@ -3,26 +3,15 @@
...
-      end
-
       def update
         respond_to do |format|
           format.html do
             if heads_of_term.update(review_params)
               redirect_to planning_application_review_tasks_path(@planning_application),
                 notice: I18n.t("review.heads_of_terms.update.success")
             else
-              render :show
```
EGiataganas, line 25: could we apply same pattern?

change the 

```ruby 
def heads_of_term
  @planning_application.heads_of_term
end
```
to 

```ruby 
def set_heads_of_term
  @heads_of_term = @planning_application.heads_of_term
end
```

happy to do it as additional PR if you want

so this would be
```ruby
flash.now[:alert] = @heads_of_term.errors.messages.values.flatten.join(", ")
```

benjamineskola: I don't really like the pattern of using a method to set an instance variable and then using the instance variable. I think it's preferable to use a method call where possible.

It makes sense for variables that will be used in a view, because instance variables are needed for that, methods aren't accessible from the view. But for cases like this I don't think it makes sense.

PR #2059, opened 2024-12-02
Review legislation

benbaumann95:
### Description of change

Redesign review legislation

### Story Link

https://trello.com/c/oLEcNiDf/183-redesign-review-against-legislation

### Screenshots

Screenshot 2024-12-03 at 15 05 32
Screenshot 2024-12-03 at 15 05 38


config/routes.rb
```diff
@@ -290,7 +290,7 @@
 
           resources :immunity_enforcements, only: :update
 
-          resource :considerations, only: %i[show edit update] do
+          resource :considerations, only: %i[edit update] do
```
benjamineskola, line 293: Is there a controller method that can also be deleted as a result of this?

benjamineskola (approved):

config/routes.rb
benbaumann95: Already removed in a previous PR so this should have been captured then

PR #2066, opened 2024-12-05
Add pre app type

EGiataganas:
### Description of change

Add pre app type

### Story Link

https://trello.com/c/2eTr9eIe


config/locales/odp.yml
```diff
@@ -96,6 +96,7 @@ en:
       - ['pa.part9.classD', 'Prior Approval - Development of toll facilities']
       - ['pip', 'Permission in Principle - Consent for the principle of a project with less than 1,000 square metres floor area on a site of less than 1 hectare']
       - ['pip.technicalDetails', 'Permission in Principle - Approval of technical details']
+      - ['preApp', 'Pre App']
```
pixeltrix, line 99: Maybe the description should be something like:

```
- ['preApp', 'Pre-application Advice']
```

EGiataganas: How about changing `preApp` to `pre.app` so it match with the above overall format?

benjamineskola (approved):

PR #2069, opened 2024-12-10
Add and edit preapplication services

benjamineskola:
### Description of change

Add preapplication services to an application and allow editing them

### Story Link

https://trello.com/c/zqIX6axa/198-able-to-see-any-additional-services-required-to-be-provided-during-pre-app

### Screenshots

Screenshot 2024-12-10 at 14 37 28

Screenshot 2024-12-10 at 14 37 36

### DIFF
--- a/app/models/preapplication_service.rb
+++ b/app/models/preapplication_service.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class PreapplicationService < AdditionalService
+  TYPES = %i[
+    written_advice
+    meeting
+    site_visit
+  ].freeze
+end
--- a/app/views/planning_applications/additional_services/edit.html.erb
+++ b/app/views/planning_applications/additional_services/edit.html.erb
@@ -1,0 +1,34 @@
+<% content_for :page_title do %>
+  Pre-application services - <%= t("page_title") %>
+<% end %>
+
+<% add_parent_breadcrumb_link "Home", root_path %>
+<% add_parent_breadcrumb_link "Application", planning_application_path(@planning_application) %>
+
+<% content_for :title, "Pre-application services" %>
+
+<%= render(
+      partial: "shared/proposal_header",
+      locals: {heading: "Edit pre-application services"}
+    ) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @planning_application,
+          url: planning_application_additional_services_url(@planning_application) do |form| %>
+      <%= form.govuk_error_summary %>
+
+      <%= form.govuk_check_boxes_fieldset :additional_services,
+            legend: {text: "Select required services"} do %>
+            <% PreapplicationService::TYPES.each do |type| %>
+              <%= form.govuk_check_box :additional_services, type, label: {text: type.to_s.humanize}, checked: @planning_application.additional_services.map(&:name).include?(type) %>
+            <% end %>
+      <% end %>
+
+      <div class="govuk-button-group">
+        <%= form.submit "Save", class: "govuk-button", data: {module: "govuk-button"} %>
+        <%= govuk_button_link_to "Back", planning_application_path(@planning_application), secondary: true %>
+      </div>
+    <% end %>
+  </div>
+</div>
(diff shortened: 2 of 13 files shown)


app/models/preapplication_service.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class PreapplicationService < ApplicationRecord
```
pixeltrix, line 3: Wonder whether this should be more generic like `AdditionalService` so that it's reusable

app/models/preapplication_service.rb
```diff
@@ -0,0 +1,15 @@
...
+  belongs_to :planning_application
+
+  TYPES = %i[
+    written_advice
+    1-2-1_meeting
+    site_visit
+  ].freeze
```
pixeltrix, line 10: Snake or kebab case - no mixed 🙈

pixeltrix: Also this seems a hostage to fortune to define these in code and not via a configurable list per local authority but MVP, yada.

app/controllers/planning_applications/preapplication_services_controller.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module PlanningApplications
+  class PreapplicationServicesController < AuthenticationController
```
pixeltrix, line 4: If we change the name to `AdditionalService` for the model then we should change this as well

pixeltrix (changes requested):

app/models/preapplication_service.rb
benjamineskola: Perhaps we can create an `AdditionalService` and turn `PreapplicationService` into a subclass of it if it turns out we need them in future?

app/models/preapplication_service.rb
benjamineskola: It's not kebab case, it's snake case for "1-2-1 meeting". Though I also hate it. I'll find a better name.

app/models/preapplication_service.rb
pixeltrix: It's just weird as a data model to have an association that's specific to a subset of that model - if we use STI and had `class PreApplication < PlanningApplication` then I'd be less bothered if it was defined in the subclass.

Leave it if you want but what will happen if we need it for other services in the future is we'll end up with non-preapplication services in that table but the model will still be called `PreapplicationService` because MVP trumps everything - either that or we'll add another model. 😭

app/models/preapplication_service.rb
pixeltrix: > I'll find a better name.

Meeting?

app/models/preapplication_service.rb
benjamineskola: Oh yeah, I forgot that Rails wants you to design everything up front.

pixeltrix: Lets just leave it as `PreapplicationService` and if it needs changing I'll make sure it happens

pixeltrix: Ignore the last comment - I see you've changed it now

app/views/planning_applications/additional_services/edit.html.erb
```diff
@@ -0,0 +1,34 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @planning_application,
+          url: planning_application_additional_services_url(@planning_application) do |form| %>
+      <%= form.govuk_error_summary %>
+
+      <%= form.govuk_check_boxes_fieldset :additional_services,
+            legend: {text: "Select required services"} do %>
+            <% PreapplicationService::TYPES.each do |type| %>
```
benjamineskola, line 23: For now, we only have one subtype of AdditionalService so I'm going to leave this as it is, but when we have multiple types we'll need more logic here to determine which options to show. (No idea in advance if the multiple types will be mutually exclusive or whether showing options from more than one type will make sense, so not going to try to predict.)

pixeltrix (approved):

PR #2070, opened 2024-12-10
Add session ID to check fee page

benbaumann95:
### Description of change

Add session ID to check fee page
- This enables better fee reconciliation between PlanX and BOPS

### Story Link

https://trello.com/c/bPwMdB10/68-add-planx-session-id-to-bops-application

### Screenshots

Screenshot 2024-12-10 at 13 36 56


benjamineskola (approved):
My only small concern is whether 'session ID' will be clear enough in this context. But we can always adjust later if it turns out not to be.

PR #2072, opened 2024-12-11
Filter by pre-application type

benbaumann95:
### Description of change

Filter by pre-application type

### Story Link

https://trello.com/c/j1KiM0Bt/235-show-pre-applications-in-dashboard-and-specify-application-type

### DIFF
--- a/app/models/application_type.rb
+++ b/app/models/application_type.rb
@@ -6,7 +6,7 @@ class ApplicationType < ApplicationRecord
 
   self.audit_attributes = %w[code]
 
-  NAME_ORDER = %w[prior_approval planning_permission lawfulness_certificate other].freeze
+  NAME_ORDER = %w[prior_approval planning_permission lawfulness_certificate pre_application other].freeze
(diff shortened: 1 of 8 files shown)


app/models/application_type.rb
```diff
@@ -115,6 +115,8 @@ class ApplicationType < ApplicationRecord
         "lawfulness_certificate"
       when /\App\./
         "planning_permission"
+      when "preApp"
```
pixeltrix, line 118: Might want to future-proof this, e.g.

``` ruby
when /\ApreApp(?:\z|\.)/
```

spec/factories/application_type.rb
```diff
@@ -556,6 +556,7 @@
     end
 
     trait :pre_application do
+      name { "pre_application" }
```
pixeltrix, line 559: Doesn't the `before_validation` callback set this?

pixeltrix (approved):
Just a couple of optional suggestions

EGiataganas (approved):

PR #2078, opened 2024-12-16
Pre app determination period days

EGiataganas:
### Description of change

- Add application_type_overrides to local_authority
- Allow local admin to set determination period day for pre app

### Story Link

https://trello.com/c/Zpi72FoV

### Screenshots

Screenshot 2024-12-16 at 09 41 18
Screenshot 2024-12-16 at 09 43 01


app/models/planning_application.rb
```diff
@@ -1029,7 +1029,11 @@ def set_key_dates
   end
 
   def application_type_determination_period
-    application_type.determination_period_days || DAYS_TO_EXPIRE
+    if pre_application?
+      local_authority.application_type_overrides.determination_period_days
+    else
+      application_type.determination_period_days || DAYS_TO_EXPIRE
+    end
```
benjamineskola, line 1036: What happens here if the LA hasn't configured a determination period? Can we have it all back to the default if one isn't set?

EGiataganas: Good spot, I had this logic but looks like it might be change when I was rebasing.

app/models/planning_application.rb
```diff
@@ -1029,7 +1029,11 @@ def set_key_dates
   end
 
   def application_type_determination_period
-    application_type.determination_period_days || DAYS_TO_EXPIRE
+    if pre_application?
+      local_authority.application_type_overrides.determination_period_days || DAYS_TO_EXPIRE
+    else
+      application_type.determination_period_days || DAYS_TO_EXPIRE
+    end
```
benbaumann95, line 1036: Could we have a method on the application type instead?

I think we don't need to check whether it's pre-app or not. Just check if the override value has been set and if so, use this value but if not use the default value

benjamineskola: Yeah, e.g.:

```suggestion
      local_authority.application_type_overrides.determination_period_days ||
        application_type.determination_period_days ||
        DAYS_TO_EXPIRE
```

benbaumann95: Just need to consider when the application type overrides is something like

```
[
  {"code" => "preApp", "determination_period_days" => 30}, 
  {"code" => "HAPP", "determination_period_days" => 21}
]
```

benbaumann95: Speaking of, don't we want to store these overrides as an array of objects?
Unless i'm missing something

benbaumann95 (approved):

PR #2083, opened 2024-12-17
Hide check legislative requirements link

benbaumann95:
### Story Link

https://trello.com/c/ZeG5Nn6Q/240-show-only-tasks-relevant-to-pre-app

### DIFF
--- a/app/workflows/validation/check_legislation_description_task.rb
+++ b/app/workflows/validation/check_legislation_description_task.rb
@@ -3,11 +3,7 @@
 module Validation
   class CheckLegislationDescriptionTask < WorkflowTask
     def task_list_link_text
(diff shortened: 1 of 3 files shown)


app/workflows/validation/check_legislation_description_task.rb
```diff
@@ -21,7 +17,8 @@ def task_list_status
     end
 
     def render?
-      planning_application.application_type.legislation_description
+      application_type = planning_application.application_type
+      application_type.legislative_requirements? && application_type.legislation_description
```
benjamineskola, line 21: Should this have a `.present?`? Not sure if it's really likely ever to be non-nil but empty, admittedly.

PR #2087, opened 2024-12-19
Improvements

EGiataganas:
### Description of change

- Create a helper method to check `lawfulness_certificate`
- Create a helper method to check `planning_permission`


benjamineskola:
Are these actually cases where it'd be better to turn something into a feature rather than depend on the type name?

PR #2088, opened 2024-12-20
Restrict access to uploads

pixeltrix:
### Description of change

Use CloudFront signed cookies to prevent premature access to planning application documents.

### Story Link

https://trello.com/c/1VP9wgYz


app/models/document.rb
```diff
@@ -209,12 +226,11 @@ class NotArchiveableError < StandardError; end
   validate :numbered
   validate :created_date_is_in_the_past
 
-  default_scope -> { no_owner.or(not_excluded_owners) }
-
```
pixeltrix, line 213: This needed to be removed so the uploads engine could find all documents for checking the blob key

app/models/document.rb
```diff
@@ -247,7 +263,33 @@ class NotArchiveableError < StandardError; end
...
+      ]
+
+      left_joins(*blob_associations)
+        .where(file_blob: {key:})
+        .or(where(file_variant_blobs: {key:}))
+        .or(where(file_preview_blob: {key:}))
+        .or(where(file_preview_variant_blobs: {key:}))
+        .distinct.sole
+    end
+
```
pixeltrix, line 292: This huge block is so we can find a document through a local authority and prevent someone using access in one local authority to access a document in another authority.

.env.example
```diff
@@ -6,3 +7,4 @@ GROUPED_PROPOSAL_DETAILS_FEATURE=enabled
 GOOGLE_TAG_MANAGER_ID=GTM-XXXXXXXX
 UPLOADS_HOSTNAME=uploads.bops.localhost
 UPLOADS_BASE_URL=http://uploads.bops.localhost:3000
+USE_SIGNED_URLS=false
```
pixeltrix, line 10: Need to deploy this to staging so we can test it but don't want to break the app so this controls whether it uses the current urls or the new urls.

app/models/document.rb
pixeltrix: Doing an `EXPLAIN` on the query shows it's not great so continuing to look at other options.

app/models/document.rb
```diff
@@ -225,8 +241,8 @@ class NotArchiveableError < StandardError; end
   )
   scope :publishable, -> { where(publishable: true) }
 
-  scope :for_publication, -> { active.publishable }
-  scope :for_display, -> { active.referenced_in_decision_notice }
+  scope :for_publication, -> { publishable }
+  scope :for_display, -> { referenced_in_decision_notice }
```
benjamineskola, line 245: Doesn't this change the behaviour? If there's a document where `publishable: true` and an archived date set, then previously `for_publication` would have excluded it but now would include it. But it depends on whether that could actually happen, of course.

pixeltrix: Yes, that's why where these scopes are used I've prefixed the call with `active` - doing it inside the scope affects how the `OR` is constructed elsewhere.

benjamineskola (approved):

EGiataganas (approved):

PR #2099, opened 2025-01-01
Import PAAPI data when the planning application is submitted

pixeltrix:
The current merged table is too confusing.

### DIFF
--- a/app/models/site_history.rb
+++ b/app/models/site_history.rb
@@ -1,7 +1,35 @@
 # frozen_string_literal: true
 
 class SiteHistory < ApplicationRecord
+  include DateValidateable
+
   belongs_to :planning_application
 
-  validates :date, :application_number, :description, :decision, presence: true
+  alias_attribute :reference, :application_number
+
+  normalizes :decision, with: ->(decision) { normalize_decision(decision) }
+
+  with_options presence: true do
+    validates :reference, :description, :decision
+    validates :date, date: {before: :current}
+  end
+
+  class << self
+    private
+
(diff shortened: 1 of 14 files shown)


app/models/site_history.rb
```diff
@@ -1,7 +1,33 @@
...
 
 class SiteHistory < ApplicationRecord
+  include DateValidateable
+
   belongs_to :planning_application
 
-  validates :date, :application_number, :description, :decision, presence: true
+  alias_attribute :reference, :application_number
+
+  normalizes :decision, with: ->(decision) { normalize_decision(decision) }
```
pixeltrix, line 10: Not sure about this - might need to do a distinct query on the PAAPI data to see the full extent of the variations

pixeltrix:
Also took the time to rework the form/table to use GOV.UK components and form builder features and share the form between the index and edit actions.

southwark bops localhost_3000_planning_applications_24-00481-HAPP_assessment_site_histories

app/models/site_history.rb
```diff
@@ -1,7 +1,35 @@
 # frozen_string_literal: true
 
 class SiteHistory < ApplicationRecord
+  include DateValidateable
+
   belongs_to :planning_application
 
-  validates :date, :application_number, :description, :decision, presence: true
+  alias_attribute :reference, :application_number
```
pixeltrix, line 8: I aliased this because elsewhere in the app we use reference but have the label as "Application number" - not sure what the design guidance is for this?

EGiataganas (approved):

PR #2100, opened 2025-01-02
Reviewe bug fix for permitted development rights and assessment against legislation

EGiataganas:
### Description of change

Fix reviewer bugs
- Show permitted development rights when their is one
- Show assessment against legislation when their is one

### Story Link

https://trello.com/c/RL2TdiOD

### DIFF
--- a/app/views/planning_applications/review/tasks/_review_conditions.html.erb
+++ b/app/views/planning_applications/review/tasks/_review_conditions.html.erb
@@ -75,7 +75,7 @@
     <% end %>
   <% end %>
 
-  <% if @planning_application.check_permitted_development_rights? %>
+  <% if @planning_application.check_permitted_development_rights? &&  @planning_application.permitted_development_right_in_progress? %>
     <% @permitted_development_right ||= @planning_application.permitted_development_right %>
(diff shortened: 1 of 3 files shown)


app/views/planning_applications/review/tasks/_review_conditions.html.erb
```diff
@@ -75,7 +75,7 @@
     <% end %>
   <% end %>
 
-  <% if @planning_application.check_permitted_development_rights? %>
+  <% if @planning_application.check_permitted_development_rights? &&  @planning_application.permitted_development_right_in_progress? %>
```
pixeltrix, line 78: This isn't a suggestion for change but more of an observation about the app in general. We have so much business logic tied up in view conditionals that it seems like we should extract that some how - probably worth an innovation day on patterns that others have used to address this since it's a common problem.

pixeltrix (approved):

app/views/planning_applications/review/tasks/_review_conditions.html.erb
EGiataganas: Yes this is something I've discussed previously with Ben B. We shouldn't be doing all this logic here. This is something that should be happened dynamic

PR #2104, opened 2025-01-03
Raise error when neighbor comments sent to invalid application type

benjamineskola:
### Description of change

If neighbour comments are sent to an application of a type that doesn't support them, the creation fails with a `NoMethodError`. Instead we should explicitly reject them and return an intentional error.

### Story Link

https://trello.com/c/NKZ1z9tn/60-neighbour-responses-to-an-application-without-a-consultation-are-handled-badly

### DIFF
--- a/app/controllers/api/v1/neighbour_responses_controller.rb
+++ b/app/controllers/api/v1/neighbour_responses_controller.rb
@@ -9,6 +9,10 @@ module Api
       skip_before_action :authenticate_api_user!
 
       def create
+        unless @planning_application.application_type.consultation_steps.include? "neighbour"
+          raise NeighbourResponseCreationService::CreateError, "This application type cannot accept neighbour responses"
+        end
+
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/neighbour_responses_controller.rb
```diff
@@ -9,6 +9,10 @@ class NeighbourResponsesController < Api::V1::ApplicationController
       skip_before_action :authenticate_api_user!
 
       def create
+        unless @planning_application.application_type.consultation_steps.include? "neighbour"
```
EGiataganas, line 12: how about moving it on its own method?

benjamineskola: I'm not sure that there would be much benefit, when it's quite short and only used once.

EGiataganas (approved):

PR #2112, opened 2025-01-13
Don't allow local admins to create global admins

EGiataganas:
### Description of change

Don't allow local admins to create global admins

### Story Link

https://trello.com/c/YUAg3Ixr

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -4,6 +4,8 @@ class User < ApplicationRecord
   include BopsCore::AuditableModel
   include Discard::Model
 
+  GLOBAL_ROLES = %w[global_administrator].freeze
+
   self.audit_attributes = %w[id name role]
   self.discard_column = :deactivated_at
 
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -47,6 +47,10 @@ def menu(scope = User.all)
     def by_name
       order(name: :asc)
     end
+
+    def roles_without_global_administrator
+      roles.except("global_administrator")
+    end
```
pixeltrix, line 53: There might possibly be multiple roles within the BopsConfig engine eventually so lets leave some space like this:

``` ruby
GLOBAL_ROLES = %w[global_administrator].freeze

def local_roles
  roles.keys - global_roles
end

def global_roles
  GLOBAL_ROLES
end
```

engines/bops_admin/app/views/bops_admin/users/_form.html.erb
```diff
@@ -13,7 +13,7 @@
     <% unless form.object == current_user %>
       <%= form.govuk_collection_select(
             :role,
-            User.roles.keys,
+            User.roles_without_global_administrator.keys,
```
pixeltrix, line 16: With the above change this becomes:

``` erb
<%= form.govuk_collection_select(:role, User.local_roles, :itself, :capitalize) %>
```

app/models/user.rb
pixeltrix: Could also add a validation to ensure that someone doesn't hack a form:

``` ruby
validates :role, inclusion: {in: :local_roles}, if: -> { local_authority.present? } 
```

pixeltrix (changes requested):

app/models/user.rb
```diff
@@ -106,6 +117,14 @@ def valid_persistence_token?(token)
...
   end
 
+  def local_roles
+    self.class.local_roles
+  end
+
+  def global_roles
+    self.class.global_roles
+  end
+
```
pixeltrix, line 127: This could be:

``` ruby
delegate :local_roles, :global_roles, to: :class
```

See the documentation for [`delegate`](https://api.rubyonrails.org/classes/Module.html#method-i-delegate)

pixeltrix (approved):
Just a suggestion for improvement but it's fine as it is

PR #2116, opened 2025-01-15
Consultee magic link auth using sgid

benbaumann95:
### Description of change

Use global signed ID for magic link auth
- Handle valid and invalid sgid in magic link
- Handle expired resource so that end users can request a new magic link if previous link has expired

### Story Link

https://trello.com/c/8VoNhP0c/320-authentication-for-consultees-with-magic-link

### DIFF
--- a/engines/bops_core/app/controllers/concerns/bops_core/magic_link_authenticatable.rb
+++ b/engines/bops_core/app/controllers/concerns/bops_core/magic_link_authenticatable.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module BopsCore
+  module MagicLinkAuthenticatable
+    extend ActiveSupport::Concern
+
+    included do
+      rescue_from ActionController::ParameterMissing do |exception|
+        render plain: "Not Found", status: :not_found
+      end
+    end
+
+    def authenticate_with_sgid!
+      resource = sgid_authentication_service.locate_resource
+
+      handle_expired_or_invalid_sgid if resource.nil?
+    end
+
+    private
+
+    def sgid_authentication_service
+      @sgid_authentication_service ||= SgidAuthenticationService.new(sgid)
+    end
+
+    def sgid
+      params.require(:sgid)
+    end
+
+    def handle_expired_or_invalid_sgid
+      if sgid_authentication_service.expired_resource
+        render plain: "Magic link expired", status: :unprocessable_entity
+      else
+        render plain: "Forbidden", status: :forbidden
+      end
+    end
+  end
+end
--- a/engines/bops_core/app/services/bops_core/sgid_authentication_service.rb
+++ b/engines/bops_core/app/services/bops_core/sgid_authentication_service.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module BopsCore
+  class SgidAuthenticationService
+    attr_reader :sgid, :purpose
+
+    def initialize(sgid, purpose: "magic_link")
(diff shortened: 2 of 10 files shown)


engines/bops_core/app/controllers/concerns/bops_core/magic_link_authenticatable.rb
```diff
@@ -0,0 +1,31 @@
...
+    SGID_FOR = "magic_link"
+
+    included do
+      rescue_from ActionController::ParameterMissing do |exception|
+        render plain: "Not Found", status: :not_found
+      end
+    end
+
+    def authenticate_with_sgid!
+      resource = GlobalID::Locator.locate_signed(sgid, for: SGID_FOR)
```
benjamineskola, line 16: I don't think we really need the constant for this.

```suggestion
    included do
      rescue_from ActionController::ParameterMissing do |exception|
        render plain: "Not Found", status: :not_found
      end
    end

    def authenticate_with_sgid!
      resource = GlobalID::Locator.locate_signed(sgid, for: "magic_link")
```

benjamineskola: Yeah, I saw there were a few other places in other files that used the literal string that could in theory be changed to use the constant but in practice I think it's probably not worth even that.

benbaumann95: Now that I've added a check if the sgid has expired with base64 decoding, need to check against the specified sgid for so leaving this as a constant

engines/bops_core/app/controllers/concerns/bops_core/magic_link_authenticatable.rb
```diff
@@ -0,0 +1,61 @@
...
+      expired_resource = sgid_expired_resource
+
+      if expired_resource
+        render plain: "Magic link expired", status: :unprocessable_entity
+      else
+        redirect_failed
+      end
+    end
+
+    def parse_global_id_from_sgid(sgid)
```
benbaumann95, line 45: If link is expired, we want to check that the magic link was previously valid and then extract the consultee resource so we can query with the id and grab the email address to resend the magic link

engines/bops_core/app/controllers/concerns/bops_core/magic_link_authenticatable.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    def sgid
+      params.require(:sgid)
+    end
+
+    def handle_expired_or_invalid_sgid
+      expired_resource = sgid_authentication_service.expired_resource
+
+      if expired_resource
```
benjamineskola, line 32: since this doesn't seem to get used again does it need to be assigned to a variable?

```suggestion
      if sgid_authentication_service.expired_resource
```

engines/bops_core/app/services/bops_core/sgid_authentication_service.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    def expired_resource
+      gid = parse_global_id
+      return nil unless gid
+
+      gid.model_class.find_by(id: gid.model_id)
+    rescue ActiveRecord::RecordNotFound
+      nil
+    end
```
benjamineskola, line 23: Do we ever use the return value of this, or is it effectively just a boolean? calling it `resource_expired?` or something and making it true or false might be nicer if we only ever care that the resource exists.

PR #2118, opened 2025-01-15
Update seeds.rb file to create global admin and confirmed users

rebeccaoneill:
### Description of change

Update seeds file to create Global Admin account and confirm other accounts on creation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,12 @@ git submodule update --init --recursive
 
 This links the main repo to the [BOPS applicants project](https://github.com/unboxed/bops-applicants/).
 
+If it's your first time setting up the project or you are changing the docker file first run:
+
+```sh
+docker compose build 
+```
+
 Then build and launch the images:
 
 ```sh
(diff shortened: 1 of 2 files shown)


db/seeds.rb
```diff
@@ -26,6 +26,18 @@
   end
 end
 
+User.find_or_create_by!(role: "global_administrator") do |user|
+  user.role = "global_administrator"
+  user.password = PasswordGenerator.call
+  user.otp_required_for_login = false
+  user.name = "#{Faker::Name.first_name} #{Faker::Name.last_name}"
+  user.email = "test@globaladmin.com"
```
benjamineskola, line 34: It might be best to use a domain we control — globaladmin.com is a real site and so test@globaladmin.com might actually be deliverable.

```suggestion
  user.email = "globaladmin@example.com"
```

benjamineskola (approved):
Apart from the suggested change this looks good to me!

README.md
```diff
@@ -20,6 +20,12 @@ git submodule update --init --recursive
 
 This links the main repo to the [BOPS applicants project](https://github.com/unboxed/bops-applicants/).
 
+If it's your first time setting up the project or you are changing the docker file first run:
+
+```sh
+docker compose build
```
EGiataganas, line 26: We have centralised a lot of docker commands under a [Makefile](https://github.com/unboxed/bops/blob/main/Makefile)

You can use [make build](https://github.com/unboxed/bops/blob/main/Makefile#L5), which will also update the submodule of BOPS-Applicants

Feel free to include more `make` commands.

PR #2121, opened 2025-01-17
Update ODP schema to 0.7.2

pixeltrix:
### Story Link

https://trello.com/c/PYNZ29Tn


config/locales/en.yml
```diff
@@ -605,6 +605,8 @@ en:
   constraint_type_codes:
     article4: Article 4 direction area
     article4_caz: Central activities zones
+    articleFour: Article 4 direction area
+    articleFour_caz: Central activities zones
```
EGiataganas, line 609: Do we also want to run any migration to change it on the [database](https://github.com/unboxed/bops/blob/main/db/schema.rb#L227)?

benjamineskola (approved):

config/locales/en.yml
pixeltrix: Not yet - it's not in production yet so we'd get new requests with `article4` in them.

EGiataganas (approved):

PR #2133, opened 2025-01-27
Documents list for consultees

benjamineskola:
### Description of change

Show a list of documents on the consultee landing page.

### Story Link

https://trello.com/c/kd2916G2/364-as-a-consultee-i-can-view-documents-so-that-i-can-check-related-documents-quickly

### Screenshots
Screenshot 2025-01-28 at 12 27 17

### DIFF
--- a/engines/bops_consultees/app/controllers/bops_consultees/documents_controller.rb
+++ b/engines/bops_consultees/app/controllers/bops_consultees/documents_controller.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module BopsConsultees
+  class DocumentsController < ApplicationController
+    before_action :authenticate_with_sgid!, only: :show
+    before_action :set_planning_application, only: %i[show]
+
+    def show
+      document = @planning_application.documents.find(params[:id])
+      redirect_to main_app.uploaded_file_url(document.blob), allow_other_host: true
+    end
+
+    private
+
+    def set_planning_application
+      @planning_application = planning_applications_scope.find_by!(reference:)
+    rescue ActiveRecord::RecordNotFound
+      render_not_found
+    end
+
+    def planning_applications_scope
+      @current_local_authority.planning_applications.active
+    end
+
+    def reference
+      params[:planning_application_reference]
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


engines/bops_consultees/app/controllers/bops_consultees/documents_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
```
benbaumann95, line 1: Is this controller needed?
Are we not using the planning applications controller

engines/bops_consultees/config/routes.rb
```diff
@@ -6,5 +6,7 @@
   resource :dashboard, only: %i[show]
   resources :planning_applications, param: :reference, only: %i[show] do
     post :resend_link, on: :member
+
+    resources :documents, only: %i[show]
```
benbaumann95, line 10: Not needed if we are using the planning applications controller

engines/bops_consultees/app/controllers/bops_consultees/documents_controller.rb
benjamineskola: I think we've used this pattern elsewhere, yeah: a `Document` controller with a `show` method rather than adding a `document` method to the `PlanningApplication` controller.

benbaumann95: Perhaps I'm missing something but aren't you just rendering the "documents_list" within the planning application show?

benbaumann95: Ah I think i've misunderstood. This is just to view a single document

benjamineskola: Ah yeah, no, this is for the download links.

There's an existing one in the api-v1 controller but that feels like the wrong thing to use, particularly because it's limited to public files and this one presumably needs to access some unpublished files — but that reminds me that the scope is something that needs confirming.

benbaumann95 (approved):

PR #2137, opened 2025-01-29
Consultee response

EGiataganas:
### Description of change

Consultees submit their comments

### Story Link

https://trello.com/c/8BneHnuB

### Screenshots

submition
show

### DIFF
--- a/app/models/consultee/response.rb
+++ b/app/models/consultee/response.rb
@@ -13,8 +13,8 @@ class Consultee
     attr_readonly :response
 
     enum :summary_tag, {
-      amendments_needed: "amendments_needed",
       approved: "approved",
+      amendments_needed: "amendments_needed",
       objected: "objected"
     }, scopes: false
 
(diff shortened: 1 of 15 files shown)


app/models/consultee/response.rb
```diff
@@ -13,8 +13,8 @@ class Response < ApplicationRecord
     attr_readonly :response
 
     enum :summary_tag, {
-      amendments_needed: "amendments_needed",
       approved: "approved",
+      amendments_needed: "amendments_needed",
       objected: "objected"
     }, scopes: false
```
benjamineskola, line 19: How about:

```ruby
    enum :summary_tag, %i[amendments_needed approved objected].index_with(&:to_s), scopes: false
```

engines/bops_consultees/app/helpers/bops_consultees/application_helper.rb
```diff
@@ -12,5 +12,24 @@ def nav_items
...
+        link_text,
+        url_for_document(document),
+        new_tab:,
+        **args
+      )
+    end
+
+    def url_for_document(document)
+      if document.published?
+        main_app.api_v1_planning_application_document_url(document.planning_application, document)
```
benjamineskola, line 29: We now have a url endpoint in the consultees engine that might be more appropriate?

```suggestion
        planning_application_document_url(document.planning_application, document)
```

(actually, this should work whether published or not, so we may not need the conditional.)

benjamineskola (approved):

PR #2139, opened 2025-01-30
Add notify template ID and ensure 1 min before resending magic link

benbaumann95:
### Description of change

- Add missing notify template ID and use view_mail
- Add magic_link_last_sent_at to consultees to ensure we wait at least 1 min before a magic link can be resent

### Story Link

https://trello.com/c/aZkozPGV/342-as-a-consultee-i-need-to-receive-email-with-a-secured-magic-link-to-pre-applications-so-that-i-can-respond-to-the-pre-applicatio


app/models/consultee.rb
```diff
@@ -102,6 +102,10 @@ def application_link
     end
   end
 
+  def can_resend_magic_link?
+    magic_link_last_sent_at.nil? || magic_link_last_sent_at <= 1.minute.ago
```
benjamineskola, line 106: This logic is confusing. I think it's correct but I had to double-take: here `<=` actually means _before_ one minute ago, which is correct, but it looks like it's saying _less than_ one minute ago, which is the opposite. Don't know if there's any way around that.

benjamineskola (approved):
Small concern but I'm not sure if there's any actual change that can be made (except maybe a comment?)

app/models/consultee.rb
EGiataganas: How about splitting the method in more methods?

```ruby 
def can_resend_magic_link?
    magic_link_last_sent_at_nil? || magic_link_last_sent_at_less_than_1_min?
end

def magic_link_last_sent_at_nil?
    magic_link_last_sent_at.nil?
end

def magic_link_last_sent_at_less_than_1_min?
    magic_link_last_sent_at <= 1.minute.ago
end
```

PR #2142, opened 2025-01-31
Sitemap corrects bug

EGiataganas:
### Description of change

- Their was a bug where the label for `site_map_correct` was the value of `documents_consistent`.
- Also added a validation where is setting `site_map_correct_comment` to `nil` when the user has selected `site_map_correct` to Yes

### Story Link

https://trello.com/c/1lvMCnq6

### DIFF
--- a/app/models/consistency_checklist.rb
+++ b/app/models/consistency_checklist.rb
@@ -48,6 +48,8 @@ class ConsistencyChecklist < ApplicationRecord
     end
   end
 
+  before_validation :reset_site_map_correct_comment_if_needed
+
   private
 
   CHECKS.each do |check|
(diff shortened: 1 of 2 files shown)


app/models/consistency_checklist.rb
```diff
@@ -73,4 +75,8 @@ class ConsistencyChecklist < ApplicationRecord
       )
     end
   end
+
+  def reset_site_map_correct_comment_if_needed
+    self.site_map_correct_comment = nil if site_map_correct == "yes"
```
benjamineskola, line 80: Should `site_map_correct` really be storing a string? Can't we have that as a boolean?

I realise this isn't being introduced by this PR and it's an existing issue, but it doesn't seem ideal.

benjamineskola (approved):
Approving because the only issue already existed, not a new bug.

app/models/consistency_checklist.rb
EGiataganas: Yes, unfortunately is not boolean and for this reason I did this.

PR #2149, opened 2025-02-03
Store application's previous references

benbaumann95:
### Description of change

Store application's previous references

- When changing application type for a planning application, we update the reference number to include the new application type suffix. We want to store previous versions of the reference number for transparency

### Story Link

https://trello.com/c/dFLkk4Jw/373-changing-application-type-changes-reference-number-and-the-public-status

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -1133,16 +1133,18 @@ class PlanningApplication < ApplicationRecord
     old_application_type = ApplicationType.find(changes["application_type_id"].first)
     old_reference = reference
 
-    set_application_number
-    set_reference
+    transaction do
+      set_reference
+      previous_references << old_reference
 
-    audit!(
-      activity_type: "updated",
-      activity_information: "Application type",
-      audit_comment:
-        "Application type changed from: #{old_application_type.full_name} / Changed to: #{application_type.full_name},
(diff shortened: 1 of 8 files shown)


app/models/planning_application.rb
```diff
@@ -1133,16 +1133,19 @@ def audit_update_application_type!
...
-    set_application_number
-    set_reference
-
-    audit!(
-      activity_type: "updated",
-      activity_information: "Application type",
-      audit_comment:
-        "Application type changed from: #{old_application_type.full_name} / Changed to: #{application_type.full_name},
+    transaction do
+      set_application_number
```
benjamineskola, line 1137: I think if we call this method only when the application number is nil, we'll also gain the behaviour of reusing the existing application number when the type changes. I know that wasn't a vital requirement, though.

benbaumann95: It's only triggered with `before_update :audit_update_application_type!, if: :application_type_id_changed?`
so perhaps not needed at all?

benjamineskola: Yeah, I think it should only ever need to be set once anyway. If a new application gets a number then that should remain valid forever.

benjamineskola (approved):

PR #2153, opened 2025-02-05
Use a custom serialize_from_session handler

pixeltrix:
We need to scope restoring the user from the session to the current local authority otherwise you can reuse session cookies across subdomains.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -12,10 +12,9 @@ class User < ApplicationRecord
   enum :role, {assessor: 0, reviewer: 1, administrator: 2, global_administrator: 3}
   enum :otp_delivery_method, {sms: 0, email: 1}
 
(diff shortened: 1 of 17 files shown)


app/models/user.rb
```diff
@@ -60,17 +59,18 @@ def local_roles
...
-    else
-      local_authority = LocalAuthority.find_by!(subdomain:)
-      local_authority.users.kept.find_first_by_auth_conditions(email:)
+      if subdomain == "config"
+        local_authority_id = nil
+      else
+        local_authority_id = LocalAuthority.find_by!(subdomain:).id
+      end
+
+      kept.find_first_by_auth_conditions({email:}, {local_authority_id:})
```
pixeltrix, line 73: This needs to be two hashes otherwise Devise will drop the `nil` value from the lookup.

EGiataganas (approved):

PR #2155, opened 2025-02-05
Only display list of application types of the same type when updating

benbaumann95:
### Description of change

Only display list of application types of the same type when updating

Allowing an update to any application type runs the risk of the settings being widely different between the application types. For example, some application types do not support having a consultation so if we switch from an application type that has this to one without, we will run into issues. 
This selective list only displays the application types with the same name/type. This should limit any issues with big variances in the application type settings.

### Story Link

https://trello.com/c/dFLkk4Jw/373-changing-application-type-changes-reference-number-and-the-public-status

### DIFF
--- a/app/models/application_type.rb
+++ b/app/models/application_type.rb
@@ -183,7 +183,8 @@ class ApplicationType < ApplicationRecord
       end
     end
 
-    def menu(scope = by_name)
+    def menu(scope = by_name, type: nil)
+      scope = scope.where(name: type) if type
       scope.active.order(code: :asc).map do |application_type|
         [application_type.description, application_type.id]
       end
--- a/app/views/planning_applications/_form.html.erb
+++ b/app/views/planning_applications/_form.html.erb
@@ -24,7 +24,7 @@
         <% end %>
       </p>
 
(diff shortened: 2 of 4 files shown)


app/models/application_type.rb
```diff
@@ -183,7 +183,8 @@ def code_menu
       end
     end
 
-    def menu(scope = by_name)
+    def menu(scope = by_name, type: nil)
```
benjamineskola, line 186: The mixing of `type` and `name` doesn't seem ideal, but I don't think it would make it clearer to call the parameter `type`, because then it'll be too easy to think it's referring to application_type when it's actually referring to application_type.type. Maybe that attribute of application_type should have been called category or similar, but it's too late now and not worth fixing.

benjamineskola: Actually I've just realised I got it backwards, and the model has a `name` parameter not a `type`. WOuld making the parameter to this method be `name` instead simplify things?

```suggestion
    def menu(scope = by_name, name: nil)
```

benbaumann95: I think I tried initially but name might be reserved as it gives `"ApplicationType"` for this class method

PR #2156, opened 2025-02-05
Remove legacy attr_encrypted columns to 2fa gem

pixeltrix:
The devise-two-factor gem now uses Active Record's built-in encryption.


benbaumann95 (commented):
Should we remove the method `def legacy_otp_secret` in the user model?

pixeltrix:
> Should we remove the method `def legacy_otp_secret` in the user model?

Yes, well spotted

EGiataganas (approved):

PR #2158, opened 2025-02-10
Remove executable bit from initdb-postgis.sh

pixeltrix:
The script is meant to be included and not executed. Also ensure $POSTGRES_DB is set.


EGiataganas (approved):

benjamineskola:
I'm not clear on why this would fix anything — is something deciding whether to source or execute it depending on the executable bit being set?

PR #2159, opened 2025-02-10
Allow upload & edit documents on determined applications

EGiataganas:
### Description of change

Allow upload & edit documents on determined applications

### Story Link

https://trello.com/c/sdSbb1tK

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -122,7 +122,7 @@ class DocumentsController < AuthenticationController
   end
 
   def ensure_document_edits_unlocked
-    render plain: "forbidden", status: :forbidden and return unless @planning_application.can_validate?
+    render plain: "forbidden", status: :forbidden and return unless @planning_application.can_edit_documents?
   end
 
   def validate_document?
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -381,6 +381,10 @@ class PlanningApplication < ApplicationRecord
     consultation&.consultees_checked?
   end
 
+  def can_edit_documents?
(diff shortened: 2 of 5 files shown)


app/controllers/documents_controller.rb
```diff
@@ -122,7 +122,7 @@ def document_id
   end
 
   def ensure_document_edits_unlocked
-    render plain: "forbidden", status: :forbidden and return unless @planning_application.can_validate?
+    render plain: "forbidden", status: :forbidden and return unless @planning_application.can_validate? || @planning_application.publish_complete?
```
benjamineskola, line 125: What about a new method in PlanningApplication that represents this state, since we use it twice?

It might be clearer that way — especially as `publish_complete?` is an alias for `determined?`, but `can_validate?` includes `!closed_or_cancelled?` and `closed_or_cancelled?` includes `determined?`. This is all correct but might get confusing because of the number of different states we have.

EGiataganas: I wanted to use another method, but when I tried tweaking the existing one or adding a new one, something did not work as it should, as we have many `ifs`.

benjamineskola: What I mean is
 
```suggestion
    render plain: "forbidden", status: :forbidden and return unless @planning_application.can_edit_documents?
```

And then on PlanningApplication:

```ruby
def can_edit_documents?
  can_validate? || publish_complete?
end
```

benjamineskola (approved):
I was going to suggest some further changes to the PlanningApplication state boolean methods but it gets complicated and I'm not sure all of them make sense — probably best to write a tech debt ticket to review them separately sometime.

PR #2164, opened 2025-02-12
Extract some logic about policies to concerns

EGiataganas:
### Description of change

Extract some logic about policies to concerns


app/models/concerns/planning_application_policies.rb
```diff
@@ -0,0 +1,117 @@
+# frozen_string_literal: true
+
+module PlanningApplicationPolicies
```
benjamineskola, line 3: This is a minor issue I know, but I'm not sure I like the name `PlanningApplicationPolicies` since we already have the name 'policy' for something else.

Maybe some of this could go in `PlanningApplicationStatus` since that's where the base states are all defined — we can then have all the 'compound' states there too.

PR #2168, opened 2025-02-13
Public register link

EGiataganas:
### Description of change

Allow local authorities admin to set link for public register.

### Story Link

https://trello.com/c/wUP4cRIF

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -85,6 +85,12 @@ class LocalAuthority < ApplicationRecord
     end
   end
 
+  def public_register_base_url
+    return super if super.present? && Bops.env.production?
+
+    "https://#{subdomain}.#{Rails.configuration.applicants_base_url}"
+  end
+
   def inactive?
     !active
   end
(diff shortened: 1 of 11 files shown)


app/models/local_authority.rb
```diff
@@ -85,6 +85,12 @@ def applicants_url
     end
   end
 
+  def public_register_url
```
benjamineskola, line 88: Would it make sense to call this `public_register_base_url` perhaps, to make the difference with `Consultation#public_register_url` clearer?

db/migrate/20250213111518_add_public_register_url_to_local_authority.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPublicRegisterUrlToLocalAuthority < ActiveRecord::Migration[7.2]
+  def change
+    add_column :local_authorities, :public_register_url, :string
```
benjamineskola, line 5: Same here, etc.

benjamineskola (approved):
Suggested rename is not vital.

app/models/local_authority.rb
EGiataganas: good suggestion and it makes sense!

benjamineskola (approved):

PR #2175, opened 2025-02-18
Remove dynamic category list for requirements

pixeltrix:
Just use a static based on the document tags

https://trello.com/c/Cayx5Cro


benjamineskola (approved):
Is this because they're defined in the schema or something?

pixeltrix:
> Is this because they're defined in the schema or something?

No, it was decided that they should be standardised across all LPAs and limited to a small set matching the document tag groups plus one for other requirements but they wanted 'Drawings' instead of 'Plans', hence #2173.

benjamineskola:
Standardising it seems good. Very slight concern that it makes it more difficult to update in future, but it seems like that would be rare anyway and not _very_ difficult to change if it did happen. 👍🏻

PR #2177, opened 2025-02-19
Move application state methods into concern

benjamineskola:


app/models/planning_application.rb
```diff
@@ -360,14 +312,6 @@ def pending_review?
     recommendations.pending_review.any?
   end
 
-  def pending_recommendation?
-    may_assess? && !pending_review?
```
benjamineskola, line 364: There's no such method as `may_assess?` as far as I can tell and never has been — I assume it should be `can_assess?`. But also, as far as I can tell, `pending_recommendation?` has never been used.

pixeltrix (approved):

app/models/concerns/planning_application_status.rb
```diff
@@ -149,5 +149,61 @@ module PlanningApplicationStatus
 
       after_all_transitions :timestamp_status_change # FIXME: https://github.com/aasm/aasm#timestamps
     end
+
+    def recommendable?
+      true unless closed_or_cancelled? || invalidated? || not_started?
```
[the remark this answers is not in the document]
benjamineskola: Yeah, I wasn't really sure why some of them are written as they are. I thought doing this might be a first step towards a more general tidyup of some of the states.

My bigger concern was really that some of them jump through multiple layers of method calls before getting back to the actual status attribute, but I'm not immediately sure if any of them can easily be improved in that regard.

app/models/planning_application.rb
```diff
@@ -272,10 +272,6 @@ def application_number
     self[:application_number].to_s.rjust(5, "0")
   end
 
-  def reviewer_disagrees_with_assessor?
-    to_be_reviewed?
-  end
```
benjamineskola, line 277: This one hasn't been used in four years either, and if it still was I'd probably just replace it with `to_be_reviewed?` anyway.

PR #2186, opened 2025-02-28
Able to add and check requirements in pre-app

rebeccaoneill:
### Description of change

Able to add requirements to a pre-app.

### Story Link

https://trello.com/c/p04eUIiw/447-add-requirements

### Screenshots

image
image


app/controllers/planning_applications/assessment/requirements_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      def create
+        @categories = LocalAuthority::Requirement.categories
+        @requirements.where(id: params[:requirement_ids]).find_each do |requirement|
+          application_requirement = @planning_application.requirements.new(
+            description: requirement.description,
+            guidelines: requirement.guidelines,
+            url: requirement.url,
+            category: requirement.category
+          )
+          application_requirement.save
```
benjamineskola, line 25: It might be better to use the `!` version here, because that'll throw an exception if it fails (the version without `!` just returns true or false, but we're not checking the return value).

benjamineskola:
```suggestion
          application_requirement.save!
```

benbaumann95 (changes requested):
Great effort and really well done on this!
Think there are just a few unintentional schema changes added to this PR and some inconsistent formatting in the views where 4 spaces are used instead of 2. Not sure why our rubocop isn't picking this up but should address this.

benjamineskola:
> Not sure why our rubocop isn't picking this up but should address this.

iirc only the parts of erb files within erb tags get checked by rubocop, so indentation of the html parts is left for humans to sort out.

(i haven't yet found an html formatter that'll play nicely with erb, either.)

db/schema.rb
```diff
@@ -10,7 +10,8 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2025_02_28_161501) do
+
+ActiveRecord::Schema[7.2].define(version: 2025_03_03_152123) do
```
benbaumann95, line 14: Think it's added another line break here, can we move this back up to line 13 please

benbaumann95 (approved):

PR #2189, opened 2025-03-05
Add proxy endpoint to proxy requests to the OS API

benbaumann95:
### Story Link

https://trello.com/c/VCSgKtFj/399-api-key-disclosure

### DIFF
--- a/app/controllers/map_proxy_controller.rb
+++ b/app/controllers/map_proxy_controller.rb
@@ -1,71 +1,26 @@
 # frozen_string_literal: true
 
-require "faraday"
-
 class MapProxyController < ApplicationController
   skip_before_action :verify_authenticity_token
 
   def proxy
-    faraday_response = proxy_to_ordnance_survey(build_os_url)
-    configure_response_headers
-    send_response(faraday_response)
-  end
+    response = client.proxy
 
-  private
-
-  def build_os_url
-    os_path = request.fullpath.sub("/map_proxy", "")
-    "https://api.os.uk#{os_path}"
-  end
-
-  def proxy_to_ordnance_survey(url)
-    conn = Faraday.new(url: url) do |faraday|
-      faraday.response :raise_error
-      faraday.adapter Faraday.default_adapter
-    end
-
-    conn.get do |req|
-      req.params = request.query_parameters
-      req.headers["key"] = Rails.configuration.os_vector_tiles_api_key
-      req.headers["Accept"] = "application/octet-stream"
-    end
-  end
-
-  def configure_response_headers
-    response.headers["Cross-Origin-Resource-Policy"] = "cross-origin"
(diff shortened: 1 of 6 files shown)


app/controllers/map_proxy_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+    configure_response_headers
+    send_response(faraday_response)
+  end
+
+  private
+
+  def build_os_url
+    os_path = request.fullpath.sub("/map_proxy", "")
+    "https://api.os.uk#{os_path}"
+  end
```
benjamineskola, line 19: Maybe `build_os_url` could just be `os_url` since it returns the value.

```suggestion
  def proxy
    faraday_response = proxy_to_ordnance_survey(os_url)
    configure_response_headers
    send_response(faraday_response)
  end

  private

  def os_url
    os_path = request.fullpath.sub("/map_proxy", "")
    "https://api.os.uk#{os_path}"
  end
```

app/controllers/map_proxy_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+  end
+
+  def origin
+    request.headers["Origin"]
+  end
+
+  def request_subdomain
+    request.host.split(".").first.downcase
+  end
```
benjamineskola, line 57: I wonder if all of this is actually necessary.

Won't we already have the local authority in `current_local_authority`, which in turn is set by the request parameters in the routing config?

So (untested) we might only need something more like this:

```suggestion
  def cors_allowed?
    return true if origin.nil? # Same-origin requests are allowed
    return true if origin == current_local_authority.applicants_url
    
    Rails.logger.warn "CORS denied for origin: #{origin}"
    false
  end

  def origin
    request.headers["Origin"]
  end
```

benjamineskola (approved):

app/controllers/map_proxy_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class MapProxyController < ApplicationController
+  skip_before_action :verify_authenticity_token
+
+  def proxy
+    send_response(client.proxy)
```
benjamineskola, line 7: I feel like at this point we could just inline `send_response` rather than having a separate method.

app/views/shared/_location_map.html.erb
```diff
@@ -1,6 +1,6 @@
 <my-map
   style="width: 100%; height: 600px;"
-  osVectorTilesApiKey='<%= Rails.configuration.os_vector_tiles_api_key %>'
+  osProxyEndpoint="<%= os_proxy_endpoint %>"
```
benjamineskola, line 3: Can we use the route helper for this instead? If we have `as: 'os_proxy'` in the route definition I think we can refer to it as `os_proxy_url`.

benjamineskola (approved):

PR #2195, opened 2025-03-07
Application type local authority

EGiataganas:
### Description of change

Migrate determination_period_days from JSONb to a table

### Story Link

https://trello.com/c/VAX78YOi


db/migrate/20250307153923_migrate_application_type_with_local_authority_id_data.rb
benjamineskola, line 1: Just checking this is correct: for every existing application type, it'll create a copy of the application type for each local authority?
 
 ```
before:

application_types
| code  | local_authority_id | ... |
| Type1 | null               | ... |
| Type2 | null               | ... |
...

after:
| code  | local_authority_id | ... |
| Type1 | null               | ... |
| Type1 | 1                  | ... |
| Type1 | 2                  | ... |
| Type2 | null               | ... |
| Type2 | 1                  | ... |
| Type2 | 2                  | ... |
...
```

What happens if the migration is rolled back and then re-applied?

I think because it's in a `change` method without an `up_only` block, it'll _also_ try to do this on rollback; and then when it's re-applied it'll try to do so again. And it'll include the versions that were created the first time around, the ones with `local_authority_id` set, and it'll fail (which is good because if it succeeded the table would grow exponentially).

File: db/migrate/20250313154742_migrate_determination_period_days_from_application_type_overrides_to_application_types.rb [hunk 1]
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class MigrateDeterminationPeriodDaysFromApplicationTypeOverridesToApplicationTypes < ActiveRecord::Migration[7.2]
+  class LocalAuthority < ActiveRecord::Base; end
+  class ApplicationType < ActiveRecord::Base; end
+
+  def change
```
benjamineskola, line 7: Once again, should this have an `up_only` block?

engines/bops_admin/app/controllers/bops_admin/application_types_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    private
+
+    def set_application_types
+      @application_types = current_local_authority.application_types.by_name
+    end
+
+    def set_application_type
+      @application_type = current_local_authority.application_types.find(application_type_id)
+    end
```
benjamineskola, line 28: If these hooks are each just being used once I think we could just inline them.
```suggestion
    before_action :set_application_type, only: %i[show]

    def index
      @application_types = current_local_authority.application_types.by_name

      respond_to do |format|
        format.html
      end
    end

    def show
      @application_type = current_local_authority.application_types.find(application_type_id)

      respond_to do |format|
        format.html
      end
    end

    private
```

benjamineskola (changes requested):
Just want to double-check those migrations but otherwise LGTM

db/migrate/20250307153923_migrate_application_type_with_local_authority_id_data.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class MigrateApplicationTypeWithLocalAuthorityIdData < ActiveRecord::Migration[7.2]
+  class ApplicationType < ActiveRecord::Base; end
+  class LocalAuthority < ActiveRecord::Base; end
+
+  def change
+    up_only do
+      ApplicationType.all.find_each do |application_type|
```
benjamineskola, line 9:
```suggestion
      ApplicationType.where(local_authority_id: nil).find_each do |application_type|
```

To account for the case where this has been rolled back and reapplied, or if application types have been created manually.

EGiataganas: As we are doing it in same PR it should be any case of this. Although happy to take this into account

benjamineskola: Yeah, I'm mainly thinking about for dev instances where it's not really vital but might be annoying to fix :)

PR #2196, opened 2025-03-07
Add appeals to BOPS endpoints in PostsubmissionSchema format (DSNPI-1008)

apricot13:
Hey!

We will slowly be submitting PR's to convert the public endpoints to the ODP schema for the DPR, this one is for the appeals section. 

### Description of change

This PR makes a change to the `v2 public search` and `v2 public show` endpoints to return appeals in the structure we've put in the PostsubmissionApplication schema and the DPR. 

### Decisions 

- I have purposely put files in `engines/bops_api/app/views/bops_api/v2/shared/` and not public as the intention is in the future to add a new `PublishedSubmission` schema which is a subset of the `Postsubmission` and this would require moving and renaming all sorts of files down the line when currently everything we're doing applies at this top level. 
- I have created a new helper `PostsubmissionApplicationSchemaHelper` as a starting point for Post submission specific formatting
- I have placed the appeals data inside of a data object in the returned json all on its lonesome for now so that we can keep adding to it as we migrate to the new schema.

### Known issues 

I haven't updated the `swagger_doc.yml` as I was having trouble validating it in the swagger editor - hoping to get some tips on this from this PR!

I have also kept the tests fairly minimal, I have some [example's documented](https://github.com/theopensystemslab/digital-planning-data-schemas/tree/main/types/schemas/postSubmissionApplication#test-cases-in-the-examples) in the ODP schema repo which might be good to build tests against as we build up the new schema but I felt adding all that was a little beyond the scope of this PR - something to discuss in our next call?

### DIFF
--- a/engines/bops_api/app/helpers/bops_api/postsubmission_application_schema_helper.rb
+++ b/engines/bops_api/app/helpers/bops_api/postsubmission_application_schema_helper.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module BopsApi
+  module PostsubmissionApplicationSchemaHelper
+    # Formats a given date in 'YYYY-MM-DD' format.
+    # Use this when formatting dates stored as datetimes in the database to date strings until they can be converted
+    #
+    # @param date [DateTime, Time, nil] the date to be formatted
+    # @return [String, nil] the formatted date string or nil if the date is nil
+    #
+    # Note: This method converts the given date to a Date object, which removes the time component,
+    # and then converts it to a string in 'YYYY-MM-DD' format. This method does not account for time zones
+    # as it only deals with the date part so data could be lost if the time zone is important.
+    def format_postsubmission_date(date)
+      date&.to_date&.to_s
+    end
+
+    # Formats a given date as UTC and returns it in ISO 8601 format.
+    #
+    # @param date [DateTime, Time, nil] the date to be formatted
+    # @return [String, nil] the formatted date string or nil if the date is nil
+    def format_postsubmission_datetime(date)
+      date&.utc&.iso8601
+    end
+  end
+end
--- a/engines/bops_api/app/views/bops_api/v2/public/search.json.jbuilder
+++ b/engines/bops_api/app/views/bops_api/v2/public/search.json.jbuilder
@@ -3,5 +3,6 @@
 json.partial! "bops_api/v2/shared/metadata"
 
 json.data @planning_applications do |planning_application|
-  json.partial! "bops_api/v2/public/shared/show", planning_application:
+  json.partial! "bops_api/v2/public/shared/show", planning_application: planning_application
+  json.partial! "bops_api/v2/shared/postsubmissionApplication/postsubmissionApplication", planning_application: planning_application
 end
--- a/engines/bops_api/app/views/bops_api/v2/shared/postsubmissionApplication/_appeal.json.jbuilder
+++ b/engines/bops_api/app/views/bops_api/v2/shared/postsubmissionApplication/_appeal.json.jbuilder
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+# This contains the structure for the data.appeals section of the Postsubmission Application schema
+if planning_application.appeal
+  json.appeal do
+    json.reason planning_application.appeal.reason.camelize(:lower) if planning_application.appeal.reason
+
+    json.lodgedDate format_postsubmission_date(planning_application.appeal.lodged_at) if planning_application.appeal.lodged_at
+    json.validatedDate format_postsubmission_date(planning_application.appeal.validated_at) if planning_application.appeal.validated_at
+    json.startedDate format_postsubmission_date(planning_application.appeal.started_at) if planning_application.appeal.started_at
+
+    json.decisionDate format_postsubmission_date(planning_application.appeal.determined_at) if planning_application.appeal.determined_at
+    json.decision planning_application.appeal.decision if planning_application.appeal.decision
+
+    if planning_application.appeal.documents
+      json.files planning_application.appeal.documents do |document|
+        json.partial! "bops_api/v2/shared/document", planning_application: planning_application, document:
+      end
+    end
+  end
+end
--- a/engines/bops_api/spec/helpers/postsubmission_application_schema_helper_spec.rb
+++ b/engines/bops_api/spec/helpers/postsubmission_application_schema_helper_spec.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe BopsApi::PostsubmissionApplicationSchemaHelper, type: :helper do
+  include BopsApi::PostsubmissionApplicationSchemaHelper
+
+  describe "#format_postsubmission_date" do
+    it "formats a DateTime object in YYYY-MM-DD format" do
+      date = DateTime.new(2025, 3, 6, 12, 0, 0)
+      expect(format_postsubmission_date(date)).to eq("2025-03-06")
+    end
+
+    it "formats a Date object in YYYY-MM-DD format" do
+      date = Date.new(2025, 3, 6)
+      expect(format_postsubmission_date(date)).to eq("2025-03-06")
+    end
+
+    it "returns nil if the date is nil" do
+      expect(format_postsubmission_date(nil)).to be_nil
+    end
+  end
+
+  describe "#format_postsubmission_datetime" do
+    it "formats a DateTime object to UTC and returns it in ISO 8601 format" do
+      date = DateTime.new(2025, 3, 6, 12, 0, 0)
+      expect(format_postsubmission_datetime(date)).to eq("2025-03-06T12:00:00Z")
+    end
+
+    it "returns nil if the date is nil" do
+      expect(format_postsubmission_datetime(nil)).to be_nil
+    end
+  end
+end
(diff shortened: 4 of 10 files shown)


engines/bops_api/app/helpers/bops_api/postsubmission_application_schema_helper.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
```
benjamineskola, line 1: For what it's worth, we usually convert datetimes to `YYYY-MM-DD` format with `foo.to_date.to_s`.

(Normally we'd display dates as `YYYY-MM-DD` and datetimes as full ISO8601, but we might still have a few datetimes that should probably be stored as dates and so we truncate those — and of course if there's an external requirement for a particular format.)

I'd be really suspicious if we need to display a date in full ISO8601 format though, I wouldn't want to make it look like we have more accuracy in the data than we actually do. (But I don't think the `format_postsubmission_datetime` method is actually being used?)

engines/bops_api/app/views/bops_api/v2/public/search.json.jbuilder
```diff
@@ -3,5 +3,6 @@
 json.partial! "bops_api/v2/shared/metadata"
 
 json.data @planning_applications do |planning_application|
-  json.partial! "bops_api/v2/public/shared/show", planning_application:
+  json.partial! "bops_api/v2/public/shared/show", planning_application: planning_application
+  json.partial! "bops_api/v2/shared/postsubmissionApplication/postsubmissionApplication", planning_application: planning_application
```
benjamineskola, line 7: I think you can leave the variable name out here and the variable with the same name as the key will be used:
```suggestion
  json.partial! "bops_api/v2/public/shared/show", planning_application:
  json.partial! "bops_api/v2/shared/postsubmissionApplication/postsubmissionApplication", planning_application:  
```

engines/bops_api/app/helpers/bops_api/postsubmission_application_schema_helper.rb
apricot13: Ah ok so would you suggest I switch to using `foo.to_date.to_s` in this helper? 

I still need to do a final walkthrough of the date/datetime setup for the postSubmission - for now I've followed the same setup as we had before but I still feel for some aspects a full date time would be useful for context and auditing even if its never shown in the back/frontend. 

(and yeh `format_postsubmission_datetime` was just me adding in both types of dates the postsubmission supports its not actually used yet!)

engines/bops_api/app/views/bops_api/v2/public/search.json.jbuilder
apricot13: I've made that change! its so weird because it gave me no end of arguments the other day when I left it out and now its behaving! 🤷‍♀️

engines/bops_api/app/helpers/bops_api/postsubmission_application_schema_helper.rb
benjamineskola: Yeah, there are definitely places where we store a datetime even if in practice we never actually show it: I think it's good for recording when an event actually occurred specifically. (So I think it normally ends up being datetimes for the past, dates for the future: like deadlines and so on.)

My only concern about using a datetime when we store only a date is that it might be misleading, for cases like deadlines, to include `00:00:00` when in practice it's probably actually going to be the start or end of the working day or something like that. But obviously there might be other situations I haven't thought of where it does make sense to do so.

engines/bops_api/app/views/bops_api/v2/public/search.json.jbuilder
apricot13: OK now the error has come back! any idea what would be causing this? I've put the code back to how it was for now to be able to use the endpoint (I've rebuild and rerun the container and it still errors)

```Encountered a syntax error while rendering template: check # frozen_string_literal: true

json.partial! "bops_api/v2/shared/metadata"

json.data @planning_applications do |planning_application|
  json.partial! "bops_api/v2/public/shared/show", planning_application: 
  json.partial! "bops_api/v2/shared/postsubmissionApplication/postsubmissionApplication", planning_application:   
end
```


This error only shows when `engines/bops_api/app/views/bops_api/v2/public/search.json.jbuilder` is :


```
# frozen_string_literal: true

json.partial! "bops_api/v2/shared/metadata"

json.data @planning_applications do |planning_application|
  json.partial! "bops_api/v2/public/shared/show", planning_application:
  json.partial! "bops_api/v2/shared/postsubmissionApplication/postsubmissionApplication", planning_application:
end
```

engines/bops_api/spec/helpers/postsubmission_application_schema_helper_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
```
benjamineskola, line 1: These tests all fail because they refer to methods that don't exist: should it be `format_postsubmission_date` and `format_postsubmission_datetime` instead?

benjamineskola (changes requested):
There are a few places where Rubocop is showing up formatting errors: mainly whitespace/newlines and quote marks — could you run `rubocop --autocorrect` and update the PR with the changes, please?

engines/bops_api/app/views/bops_api/v2/shared/postsubmissionApplication/_appeal.json.jbuilder
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
```
benjamineskola, line 1: We've found in the past that it's easier to define a schema for things if we have values that are always present, but potentially nil, rather than being omitted if they're nil.

So in this case we'd probably leave off the if-statements in this file and just let it return the nils.

But you're the API consumer (and probably more familiar with jsonschema than me) so would be open to your perspective on that!

apricot13: Yeh, I see what you mean.  I've mostly followed the convention from the `PrototypeApplication` and since the schema is generated from TypeScript, the general recommendation is to use `field?: string` rather than `field: string | null`. It's also **much** easier to work with `field?: string` when dealing directly with the types. 😅. It may also be, and I'd need to check this, that the library that converts TypeScript to Json-Schema (`ts-json-schema-generator`) is opinionated on this?

In the examples I have purposely left out fields that are not relevant or applicable to the application - [Appeal determined application](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/examples/postSubmissionApplication/planningPermission/fullHouseholder-05-appeal-03-appeal-determined.json) vs [Council determined application](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/examples/postSubmissionApplication/planningPermission/fullHouseholder-04-assessment-01-council-determined.json). Data not being present in a result usually means that is is optional and/or not applicable? which is definitely true in the case of appeal data though it may not be the case for some of the other fields later on! If you can no longer submit an appeal after 6 months then `appeal: null` being present would seem odd and could imply you can still appeal?

engines/bops_api/spec/helpers/postsubmission_application_schema_helper_spec.rb
apricot13: I never thought I'd miss husky running the tests every five minutes! Fixed now hopefully!

engines/bops_api/app/helpers/bops_api/postsubmission_application_schema_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module BopsApi
+  module PostsubmissionApplicationSchemaHelper
+    # Formats a given date as UTC and returns it in 'YYYY-MM-DD' format.
+    #
+    # @param date [DateTime, Time, nil] the date to be formatted
+    # @return [String, nil] the formatted date string or nil if the date is nil
+    def format_postsubmission_date(date)
+      date&.utc&.strftime("%Y-%m-%d")
```
benjamineskola, line 10: Ah, Andrew has suggested that this could be
```suggestion
      date&.utc&.to_date.iso8601
```

Which I think might make the intent even clearer?

(I'm actually not sure if `utc` is even needed when we're only dealing with dates, since the DST change should never change the date — it'd be different if we had to deal with times outside the UK.)

apricot13: I think I've been getting ahead of myself here tbh, I've changed the functionality and description of this helper to reflect its role as go between for dates stored as datetimes that should be dates until they get converted some day! 

Theres an issue [here](https://github.com/theopensystemslab/digital-planning-data-schemas/issues/295) to standardise the dates as UTC in the API endpoints and use the at/date naming convention too. Even though this standard is currently only used in the UK, it's pretty standard to use UTC in APIs for consistency. We currently get all sorts of different date formats back from BOPS/PlanX so it would be good to agree on one format.

benjamineskola: Oh yeah, I'm not arguing against it being UTC as standard, I just mean that in this case it'll always be UTC even without calling the `.utc` method (because we're removing the time part, only showing the date).

But it also doesn't really hurt to have it in there for clarity, of course.

benjamineskola (approved):
We can't merge this just yet because of CI issues but once we've got this sorted out we should be able to rebase it onto `main` (to run the workflows properly) and then merge.

PR #2199, opened 2025-03-11
Able to edit and remove requirements from a planning application

rebeccaoneill:
### Description of change

Able to edit requirements url and guidelines and remove requirement from a planning application.

### Story Link

https://trello.com/c/2nZ1eBji/448-check-and-edit-requirements

### Screenshots

image

### DIFF
--- a/app/components/status_tags/requirements_component.rb
+++ b/app/components/status_tags/requirements_component.rb
@@ -2,17 +2,22 @@
 
 module StatusTags
   class RequirementsComponent < StatusTags::BaseComponent
-    def initialize(requirements:)
-      @requirements = requirements
+    def initialize(planning_application:, requirements:)
+      @planning_application = planning_application
+      @requirements = @planning_application.requirements
       super(status:)
     end
 
     private
 
-    attr_reader :requirements
+    attr_reader :planning_application_requirement
 
     def status
-      :not_started
+      if @requirements.any?
+        :complete
+      else
+        :not_started
+      end
     end
   end
 end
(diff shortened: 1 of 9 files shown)


app/components/status_tags/requirements_component.rb
```diff
@@ -2,17 +2,22 @@
 
 module StatusTags
   class RequirementsComponent < StatusTags::BaseComponent
-    def initialize(requirements:)
-      @requirements = requirements
+    def initialize(planning_application:, requirements:)
+      @planning_application = planning_application
+      @requirements = @planning_application.requirements
```
benjamineskola, line 7: It looks like the requirements parameter is never actually used any more, so probably safe to remove

app/components/status_tags/requirements_component.rb
```diff
@@ -2,17 +2,22 @@
...
     end
 
     private
 
-    attr_reader :requirements
+    attr_reader :planning_application_requirement
 
     def status
-      :not_started
+      if @planning_application.requirements.any?
```
benjamineskola, line 16: Shorter:

```suggestion
      if @requirements.any?
```

app/controllers/planning_applications/assessment/requirements_controller.rb
```diff
@@ -31,11 +32,46 @@ def create
...
       end
 
+      def edit
+        respond_to do |format|
+          format.html
+        end
+      end
+
+      def update
+        @requirement = @planning_application.requirements.find(params[:id])
```
benjamineskola, line 42: You could remove this line and add `update` to the set_requirement hook above, since I think the code is the same.

benjamineskola (approved):
A few suggestions but none of them is actually vital as far as I can see.

PR #2201, opened 2025-03-12
Do not use secrets in test run

benjamineskola:
This allows them to be used by PRs from forks.


.env.test
```diff
@@ -2,7 +2,8 @@ DOMAIN="example.com"
 STAGING_API_URL="bops-staging.services"
 STAGING_API_BEARER="fjisdfjsdiofjdsoi"
 OS_VECTOR_TILES_API_KEY="testtest"
-NOTIFY_LETTER_API_KEY='testtest'
+NOTIFY_API_KEY="test_key_for_ci-d64cc892-8848-4663-9544-487551484e9b-d7bc72d1-8e1f-4d5a-a703-de5beecaf4a8"
+NOTIFY_LETTER_API_KEY="test_key_for_ci-d64cc892-8848-4663-9544-487551484e9b-d7bc72d1-8e1f-4d5a-a703-de5beecaf4a8"
```
pixeltrix, line 6: Are these real test keys?

benjamineskola: Yes. (Have assumed that because it's specifically for testing and therefore doesn't have permission to do anything, it's less of a risk. Can revert and revoke if that's not the case.)

Problem is that we can't keep it in GH Secrets (if we want to allow PRs from forks) and we can't use a fake key because some of the code is actually instantiating Notify, which throws an error if just given a non-uuid like `testtest`.

We'd have to track down all of those places and mock Notify (which arguably we should have been doing all along.)

pixeltrix: You can just smoosh together `test` and two random uuids like we do here and it should work as we're not calling Notify:

https://github.com/SeneddCymru/e-petitions/blob/main/.env.test#L18

benjamineskola: Yes we are:

```
     ArgumentError:
       "02A0EEC7-EBAC-4F10-8BE4-06A60018B27A" is not a valid uuid
       This error is probably caused by initializing the Notifications client with an invalid API key.
       You can generate a new API key by logging into Notify and visiting the 'API integration' page:
```

pixeltrix: We should not be calling Notify in our tests

benjamineskola: Looking into that now.

pixeltrix (approved):

PR #2203, opened 2025-03-13
Add consultee responses to planning considerations and advice

benbaumann95:
### Description of change

Add consultee responses to planning considerations and advice

### Story Link

https://trello.com/c/W7vt08EA/289-add-planning-considerations-and-advice-view-consultee-response

### Screenshots

Screenshot 2025-03-13 at 11 39 54
Screenshot 2025-03-13 at 11 39 48


app/components/consultee_response_detail_component.html.erb
benjamineskola, line 1: Is it possible to use the [gov.uk details viewcomponent](https://govuk-components.netlify.app/components/details/) at least for lines 2–5 and so avoid having to define all the css classes for it?

benjamineskola (approved):

app/components/consultee_response_detail_component.html.erb
benbaumann95: Thanks for the review but may want to check the next commit as I've extracted this into a shared component now

engines/bops_core/app/components/bops_core/status_detail_component.rb
```diff
@@ -0,0 +1,68 @@
...
+      super(classes:, html_attributes:)
+    end
+
+    def before_render
+      fail_missing_attributes
+    end
+
+    def call
+      tag.div(**html_attributes) do
+        safe_join([details_block, status_block].compact_blank)
```
benjamineskola, line 23: Not sure the `compact_blank` is necessary here because wouldn't it be an error for either the body or the status not to be present?

engines/bops_core/app/components/bops_core/status_detail_component.rb
```diff
@@ -0,0 +1,68 @@
...
+      ["govuk-details-wrapper"]
+    end
+
+    def default_attributes
+      {id: id, class: component_classes}
+    end
+
+    def details_block
+      tag.details(class: "govuk-details", open: open) do
+        safe_join([summary_tag, body_tag].compact)
```
benjamineskola, line 50: Same here.

engines/bops_core/app/components/bops_core/status_detail_component.rb
```diff
@@ -0,0 +1,68 @@
...
+      @id = id
+      @open = open
+      super(classes:, html_attributes:)
+    end
+
+    def before_render
+      fail_missing_attributes
+    end
+
+    def call
```
benjamineskola, line 21: I think having this as a method rather than a separate template file is a definite improvement in this case 👍🏻

engines/bops_core/app/components/bops_core/status_detail_component.rb
benbaumann95: Added the validation after and forgot to take this out, thanks

PR #2210, opened 2025-03-17
Add new consideration

benbaumann95:
### Description of change

### Story Link

https://trello.com/c/cAgFnMIB/468-add-a-new-consideration

### Screenshots
Screenshot 2025-03-17 at 18 27 17

### DIFF
--- a/app/controllers/planning_applications/assessment/consideration_guidances_controller.rb
+++ b/app/controllers/planning_applications/assessment/consideration_guidances_controller.rb
@@ -3,12 +3,72 @@
 module PlanningApplications
   module Assessment
     class ConsiderationGuidancesController < BaseController
-      def index
-        @consultee_responses = @planning_application.consultation.consultee_responses
+      before_action :set_consideration_set
+      before_action :set_considerations
+      before_action :build_consideration, only: [:index, :create]
+      before_action :set_consultee_responses, only: :index
+      before_action :set_consideration, only: :destroy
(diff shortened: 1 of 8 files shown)


app/controllers/planning_applications/assessment/consideration_guidances_controller.rb
```diff
@@ -3,13 +3,73 @@
 module PlanningApplications
   module Assessment
     class ConsiderationGuidancesController < BaseController
-      def index
-        @consultee_responses = @planning_application.consultation.consultee_responses
+      before_action :set_consideration_set
+      before_action :set_considerations
+      before_action :build_consideration, only: [:index, :create]
+      before_action :set_consultee_responses, only: :index
+      before_action :set_consideration, only: :destroy
```
benjamineskola, line 10: Not quite following how these are being used. Looks like `@consideration` is used in `create` but set as `only: :destroy`; `set_considerations` and `set_consideration_set` don't have any constraints but I don't know if they're actually used in all of them.

(if any are only being used in one action I'd probably suggest it's worth just inlining them.)

benbaumann95: Using `build_consideration` for the create action.
Future proofing here as next two cards will be requiring more controller actions which will need the consideration being set

benjamineskola (approved):

PR #2212, opened 2025-03-18
Add disclaimer to local admin

benjamineskola:
### Description of change

Pre-application reports need a disclaimer, so we need to be able to store and edit that in order to (later) display it.

### Story Link

https://trello.com/c/anY0pEYT/444-add-disclaimer

### Known issues [OPTIONAL]

This can be configured for any application type but might only ever be relevant for preapps, so we may want to limit it on the show/edit form. Do we want an application type feature for example?

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -61,10 +61,4 @@ module ApplicationHelper
   def nav_items
     []
   end
-
-  def markdown(text)
-    return if text.blank?
-
-    CommonMarker.render_html(text).html_safe
-  end
 end
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -88,6 +88,7 @@ class PlanningApplication < ApplicationRecord
     delegate :appeals?
     delegate :assess_against_policies?
     delegate :consultation?
+    delegate :disclaimer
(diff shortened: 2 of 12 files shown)


db/migrate/20250318164438_add_preapp_disclaimer_to_local_authority.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPreappDisclaimerToLocalAuthority < ActiveRecord::Migration[7.2]
+  def change
+    add_column :local_authorities, :preapp_disclaimer, :string
```
pixeltrix, line 5: Not a big fan of this field name - wondering whether we could take advantage of @EGiataganas work on LA application type overrides and add `disclaimer` as an optional field to the application types table?

pixeltrix:
Regarding the translations it might have been easier to use aliases, e.g.

``` yaml
show: &profile
  name: "Name"
edit:
  *profile
```

pixeltrix:
> Good idea — except the i18n-normalize rake task breaks those

Well that seems bad 😭

benbaumann95 (approved):

EGiataganas (approved):

PR #2215, opened 2025-03-19
Copy reporting_type column to reporting_type_code

benjamineskola:
### Description of change

`PlanningApplication#reporting_type` was created as a string which needs to match a code in the ReportingTypes table. To turn it into a real reference we need to move the old column out of the way and the first step for doing that is to copy it to a new column and make sure everything references that instead.

### Story Link

https://trello.com/c/GTvuVy6x/144-make-reportingtype-a-direct-association-rather-than-a-string


db/migrate/20250319163537_rename_planning_application_reporting_type.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class RenamePlanningApplicationReportingType < ActiveRecord::Migration[7.2]
+  def change
+    add_column :planning_applications, :reporting_type_code, :string
+
+    up_only do
+      PlanningApplication.all.find_each do |planning_application|
```
benbaumann95, line 8: Sql might be better here

```
execute <<~SQL
      UPDATE planning_applications
      SET reporting_type_code = reporting_type;
SQL
```

benbaumann95 (approved):
Just a note about the migration

db/migrate/20250319163537_rename_planning_application_reporting_type.rb
benjamineskola: Good call!

PR #2220, opened 2025-03-21
Recommended application type

EGiataganas:
### Description of change

Able to select application type for pre apps.

### Story Link

https://trello.com/c/a1UEB0ZM

### Screenshots

Screenshot 2025-03-21 at 15 58 26

### DIFF
--- a/app/components/task_list_items/assessment/recommended_application_type_component.rb
+++ b/app/components/task_list_items/assessment/recommended_application_type_component.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module TaskListItems
+  module Assessment
+    class RecommendedApplicationTypeComponent < TaskListItems::BaseComponent
+      def initialize(planning_application:)
+        @planning_application = planning_application
+      end
+
+      private
+
+      attr_reader :planning_application
+
+      def link_text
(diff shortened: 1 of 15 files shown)


app/components/task_list_items/assessment/recommended_application_type_component.rb
```diff
@@ -0,0 +1,35 @@
...
+          planning_application_assessment_recommended_application_type_path(planning_application)
+        else
+          edit_planning_application_assessment_recommended_application_type_path(planning_application)
+        end
+      end
+
+      def status_tag_component
+        StatusTags::BaseComponent.new(status:)
+      end
+
```
benjamineskola, line 29: I believe this is unnecessary, since it's [inherited from the superclass](https://github.com/unboxed/bops/blob/main/app/components/task_list_items/base_component.rb#L11).

```suggestion
```

benjamineskola (approved):

app/components/task_list_items/assessment/recommended_application_type_component.rb
EGiataganas: That's true! I had status there at some point but then moved it to a different method.

PR #2228, opened 2025-03-24
Show a link to document validation request where appropriate

benjamineskola:
### Description of change

Currently we're always showing a link to request a replacement even if one is already in the process of being requested. This also makes it hard to cancel requests after the validation stage (when the validation request view isn't so easily accessible). Linking to the management page for an existing request when one exists seems reasonable.

### Story Link

https://trello.com/c/hSnZOHGB/471-be-able-to-cancel-validation-requests-at-any-stage-where-appropriate

### Screenshots

Screenshot 2025-03-24 at 15 14 05

Screenshot 2025-03-24 at 15 04 31

### DIFF
--- a/app/views/documents/_active_documents_table.html.erb
+++ b/app/views/documents/_active_documents_table.html.erb
@@ -27,7 +27,11 @@
                 <p class="govuk-body govuk-!-margin-right-2 govuk-!-text-align-right">
                   <%= govuk_link_to "Edit", edit_planning_application_document_path(@planning_application, document) %><br>
                   <%= govuk_link_to "Archive", planning_application_document_archive_path(document_id: document.id) %><br>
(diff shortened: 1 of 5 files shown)


benbaumann95 (commented):
Did we want to enable cancelling at any point for other validation requests too that can occur post validation?
Or was this only applicable to documents

benjamineskola:
> Did we want to enable cancelling at any point for other validation requests too that can occur post validation? Or was this only applicable to documents

I could be wrong but it didn't seem to me like we're _preventing_ them from being cancelled, it's just not very straightforward to get to the validation request view.

(This also seemed to make sense anyway, because requesting another replacement while there's already an active replacement request doesn't seem like it would be useful.)

benbaumann95 (approved):
Might be worth adding a quick spec but looks fine

PR #2229, opened 2025-03-25
Add rich text editing to the pre-app considerations

pixeltrix:
### Description of change

As a case officer I need to enter advice in rich text so that I can structure my advice properly

### Story Link

https://trello.com/c/NwxiILpR

### Screenshots

image


app/assets/stylesheets/_main.scss
```diff
@@ -402,6 +403,16 @@ body {
...
+.govuk-form-group--error {
+  .autocomplete__input {
+    border-color: $govuk-error-colour;
+
+    &:focus {
+      border-color: $govuk-input-border-colour;
+    }
+  }
+}
+
```
pixeltrix, line 415: This makes sure that autocomplete inputs have a red border as class on the input is lost

app/components/task_list_items/assessment/considerations_component.rb
```diff
@@ -3,8 +3,10 @@
 module TaskListItems
   module Assessment
     class ConsiderationsComponent < TaskListItems::BaseComponent
-      def initialize(planning_application:)
+      def initialize(planning_application:, link_text: nil, link_path: nil)
```
pixeltrix, line 6: I'm allow the component to override the link and text so we can reuse the existing component - is the intention to replace the original considerations design with the new one?

app/controllers/planning_applications/assessment/consideration_guidances_controller.rb
```diff
@@ -76,6 +77,10 @@ def set_considerations
         @considerations = @consideration_set.considerations.select(&:persisted?)
       end
 
+      def set_review
+        @review = @consideration_set.current_review
+      end
```
pixeltrix, line 82: I'm setting the review here but the form submission goes to the original considerations controller - something to think about if we switch to the pre-app design for considerations on full applications.

app/javascript/controllers/address_fill_controller.js
```diff
@@ -10,7 +10,7 @@ export default class extends Controller {
       name: this.data.get("name"),
       source: (query, populateResults) => {
         const results = []
-        ajax({
+        Rails.ajax({
```
pixeltrix, line 13: I think something changed so that `ajax` is no longer exported so needed to update these references

app/views/active_storage/blobs/_blob.html.erb
```diff
@@ -0,0 +1,14 @@
...
+
+  <figcaption class="attachment__caption">
+    <% if caption = blob.try(:caption) %>
+      <%= caption %>
+    <% else %>
+      <span class="attachment__name"><%= blob.filename %></span>
+      <span class="attachment__size"><%= number_to_human_size blob.byte_size %></span>
+    <% end %>
+  </figcaption>
+</figure>
```
pixeltrix, line 14: Probably will need to revisit this when we get inline images working but just left the default for now

app/views/documents/form_partials/_upload.html.erb
```diff
@@ -1,14 +1,12 @@
 <% if @document.new_record? %>
   <%= form.govuk_file_field(
         :file,
-        label: {text: t(".upload_a_file")},
-        direct_upload: true
+        label: {text: t(".upload_a_file")}
```
pixeltrix, line 4: We must have used direct uploads in the past but Rails code protected us from ourselves

config/routes.rb
```diff
@@ -390,4 +390,12 @@
...
 
   mount BopsUploads::Engine, at: "/", as: :bops_uploads
+
+  direct :rails_direct_uploads do
+    ""
+  end
+
+  direct :rails_service_blob do
+    ""
+  end
```
pixeltrix, line 400: Just placeholders as Action Text blows up without these methods

benbaumann95 (approved):
🎉

EGiataganas (approved):

PR #2230, opened 2025-03-25
Permit OpenLayers stylesheet from CDN

benjamineskola:
### Description of change

Permit openlayers stylesheet to be loaded from a cdn after incorrectly blocking too enthusiastically in #2197.

### Story Link

<https://trello.com/c/j3XbGmJA/562-dark-buttons-and-misaligned-times-new-roman-text-on-the-map>

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -13,6 +13,7 @@ google_analytics_hostnames = %w[
   region1.google-analytics.com
   region1.analytics.google.com
 ].freeze
+openlayers_map_url = "https://cdn.skypack.dev/ol@%5E6.6.1/ol.css"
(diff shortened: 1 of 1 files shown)


pixeltrix:
Actually, I think we might be able to just `@import` the CSS and ignore the warning - we're already fetching the `ol` package anyway as a dependency

benjamineskola:
I did try this and it didn't appear to load — I can give it a second look, though.

Also I'm not a huge fan of having warnings that we just expect to ignore but there's not much we could do about that other than getting it changed upstream I suppose.

pixeltrix:
Ahh, the custom element prevents it from working - it isolates it from the parent document

pixeltrix (approved):
Okay, lets go with this and I'll make a PR to allow the url to be overridden in the map component.

PR #2242, opened 2025-04-01
Create new public endpoint for public and specialist comments (DSNPI-896)

apricot13:
### Description of change

This PR adds the following new endpoints documented in the [ODP API Implementation](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/IMPLEMENTATION.md)

* `/api/v2/public/planning_applications/{reference}/comments/public`
* `/api/v2/public/planning_applications/{reference}/comments/specialist`

In this PR the endpoints each return:

* `pagination` in the [ODP Pagination format](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/IMPLEMENTATION.md#offset-pagination) [Pagination type](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/types/schemas/postSubmissionApplication/implementation/Pagination.ts)
* `summary` in either the `PublicCommentSummary` or `SpecialistCommentSummary` structure from [CommentSummary type](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/types/schemas/postSubmissionApplication/data/CommentSummary.ts) (see known issues!)
* `comments` currently in the same format as the current `/show` and `/search` endpoints in `consultation.public_comments` the exception being an `id` has now been added (see known issues!)

Each endpoint supports the following query params for more details see the corresponding [ODP API Implementation](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/IMPLEMENTATION.md) section:

* `page`
* `resultsPerPage`
* `query`
* `sortBy` 
* `orderBy`

### Decisions [OPTIONAL]

The use of `public` and `specialist` over `neighbour` and `consultee` - `specialist` and `public` as terms are more understandable to the public from our research

Created new services to handle all the Postsubmission things including PostsubmissionPagination so we don't mess up any other endpoints with the new structure!

### Known issues [OPTIONAL]

1. Comments are not in the new [ODP comment structure](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/types/schemas/postSubmissionApplication/data/Comment.ts) yet but it has been outlined in the view files, this is because there are still some unknowns about the format we need to work out!
2. Technically this would come under the `@next` schema version as its `Postsubmission` but we didn't want to stray too far from the norm for this
3. Not all endpoint filters are implemented yet only the simpler ones to be easier to review/get the tests right
4. A lot of the code is duplicated right now but this will change in the new schema work
5. We're mapping approved, objected, amendments_needed to supportive, objection, neutral and setting neutral as a default to ensure it validates against the schema - this is one of the unknowns we have about the specialist comment schema 
6. I'm not 100% sure that the code for consultee responses count is correct!

```rb
          @total_responses = @consultee_responses.count
          @total_consulted = @consultation.consultees.count
```

### DIFF
--- a/engines/bops_api/app/services/bops_api/comments_public_service.rb
+++ b/engines/bops_api/app/services/bops_api/comments_public_service.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+module BopsApi
+  class CommentsPublicService
+    def initialize(scope, params)
+      @scope = scope
+      @params = params
+    end
+
+    attr_reader :scope, :params
+
+    def call
+      paginate(
+        sort_results(
+          filter_by_query(scope)
+        )
+      )
+    end
+
+    private
+
+    def filter_by_query(scope)
+      if params[:query].present?
+        scope.where("redacted_response ILIKE ?", "%#{params[:query]}%")
+      else
+        scope
+      end
+    end
+
+    def sort_results(scope)
+       # Define allowed fields and their default sort orders
+      allowed_sort_fields = {
+        "receivedAt" => {column: "received_at", default_order: "desc"},
+        "id" => {column: "neighbour_responses.id", default_order: "asc"}
+      }
+
+      # Define allowed orderBy values
+      allowed_order_values = %w[asc desc]
+
+      # Validate sortBy if it is explicitly set
+      if params[:sortBy].present?
+        sort_by = params[:sortBy]&.camelize(:lower)
+        unless allowed_sort_fields.key?(sort_by)
+          raise ArgumentError, "Invalid sortBy field: #{params[:sortBy]}. Allowed fields are: #{allowed_sort_fields.keys.join(", ")}"
+        end
+      else
+        sort_by = "receivedAt" # Default sortBy
+      end
+
+      # Validate orderBy if it is explicitly set
+      if params[:orderBy].present?
+        order_by = params[:orderBy]
+        unless allowed_order_values.include?(order_by)
+          raise ArgumentError, "Invalid orderBy value: #{params[:orderBy]}. Allowed values are: #{allowed_order_values.join(", ")}"
+        end
+      else
+        order_by = allowed_sort_fields[sort_by][:default_order] # Default orderBy
+      end
+
+      # Apply sorting to the scope
+      sort_field = allowed_sort_fields[sort_by]
+      scope.order("#{sort_field[:column]} #{order_by}")
+    end
+
+    def paginate(scope)
+      BopsApi::PostsubmissionPagination.new(scope: scope, params: params).call
+    end
+  end
+end
--- a/engines/bops_api/app/services/bops_api/comments_public_service.rb
+++ b/engines/bops_api/app/services/bops_api/comments_public_service.rb
@@ -59,7 +59,7 @@ module BopsApi
 
       # Apply sorting to the scope
       sort_field = allowed_sort_fields[sort_by]
-      scope.order("#{sort_field[:column]} #{order_by}")
+      scope.reorder("#{sort_field[:column]} #{order_by}")
     end
 
     def paginate(scope)
--- a/engines/bops_api/app/controllers/bops_api/v2/public/comments_public_controller.rb
+++ b/engines/bops_api/app/controllers/bops_api/v2/public/comments_public_controller.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module BopsApi
+  module V2
+    module Public
+      class CommentsPublicController < PublicController
+        def index
+          @planning_application = find_planning_application params[:planning_application_id]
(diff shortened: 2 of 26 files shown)


engines/bops_api/app/services/bops_api/comments_public_service.rb
benjamineskola, line 1: The two services CommentsPublicService and CommentsSpecialistService look very similar (though I haven't gone through line by line). Could we either just use one service for both of them, or at least share the common functionality between them?

apricot13: I've uhmed and ahhed soo much about this! 😅😭 any thoughts are very welcome!!

**Pros for having them separate:**

* In the next few weeks we'll be converting these endpoints to use the new schema format and having to unpick everything wouldn't make much sense? (if the timescale were greater we should definitely combine things)
* Even though right now they're the same in the new schema they are going to be very different: [SpecialistComment](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/types/schemas/postSubmissionApplication/data/Comment.ts#L84) vs [PublicComment](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/types/schemas/postSubmissionApplication/data/Comment.ts#L21C18-L21C31) 
* In the future the `query` fields could be much broader for both
* `sortBy`, `orderBy` for specialist will _definitely_ include different options as well
* There will also be a PR for the documents and applications as well which will follow the standard we setup - sortBy, orderBy & query will be in those as well

**Cons for having them separate:**

* It's not DRY - Files are literally copy paste 😅 We don't like damp code!



* at least `page`, `resultsPerPage` are ok because they're split out into `PostsubmissionPagination`!

benjamineskola: I think the starting point should be to have the code separate, and then split it out when it turns out to be necessary. At the very least, I'd be inclined to make a base class (e.g. `BaseCommentsService`) which at the very least could have `initialize` and `call` in it, and maybe `filter_by_query`.

(It might also work to have one of them inherit from the other, but if as you say they'll end up being quite different that makes less sense.)

`sort_results` obviously gets more complicated, but that could then be defined in the two subclasses. _Or_ it might also work to define `allowed_sort_fields` in the subclasses, and then have `sort_results` in the base class operate on that dynamically. But that of course depends on how different they end up needing to be.

benjamineskola: I'm also a little concerned that the naming could cause confusion. In the API we use `Public` to mean publicly-accessible, whereas I think here you mean _from_ the public, i.e. what we'd usually call `Neighbour` (although they're not necessarily literal neighbours). Then what you call `Specialist` is what we normally call `Consultee`, I think?

apricot13: I've updated the code to use a common base class! Feels much cleaner now!

I know what you mean about the naming, the concept of public and specialist is more to do with the postsubmission schema than the bops internals so would it make more sense to group things under postsubmission or public since that will eventually be all postsubmission themed? I've been trying to keep the postsubmission stuff separate from the rest of bops.

apricot13: The schema specifies public and specialist as the names so the endpoint url's would need to remain as they are in the [implementation](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/IMPLEMENTATION.md#comments) 

I've made a small change to move the services under the `postsubmission` module to make it clear theres a separation from the rest of the bops application logic, tbh I should have done that in the first place as I'd already done that with the views and the helpers as well! 

But that still leaves the controller's which in rails usually match the endpoint they correspond to? We could update the routes to point to `postsubmission_comments_public` instead of `comments_public` maybe? but eventually all the public endpoints will be postsubmission flavoured? 🤔

The reason behind the change in naming convention in the schema is to make the different types of comments more understandable and inclusive to more people across the planning process.  In our user research people consistently didn't know what "consultee" meant. Neighbours doesn’t always mean actual "neighbours" and neighbours are technically consultee’s because they’ve been consulted so the term consultee is confusing.

benjamineskola: Yes, the user-facing terminology doesn't need to change. (Changing a third-party schema would be out of scope for this PR even if I didn't like it.)

I just think it's worth keeping BOPS internal identifiers consistent — things that affect only developers, that is. (Our internal terminology tends to match what makes sense to planning officers, which isn't necessarily what makes sense to the general public as you say.)

That does mean we'll need to explicitly specify the route as it won't be able to infer it from the controller, but I think that's the least confusing option.

(I'm not sure it's ideal to have 'public' twice in the URL meaning two different things, either. 🤔 but there doesn't seem to be much way around that.)

benjamineskola (approved):

PR #2248, opened 2025-04-02
Add pre-application details to report

benbaumann95:
### Description of change

Add pre-application details to report

- Allow editing and returning back to report of individual items

### Story Link

https://trello.com/c/S0EOTvIB/443-pre-app-preview-and-submit-final-report-your-application-details

### Screenshots

Screenshot 2025-04-02 at 15 51 05

### DIFF
--- a/app/assets/stylesheets/_main.scss
+++ b/app/assets/stylesheets/_main.scss
@@ -296,7 +296,7 @@ $govuk-page-width: 1100px;
   }
 }
 
-.proposal-details-sub-heading {
+.flex-between {
   display: flex;
   justify-content: space-between;
 }
--- a/app/controllers/concerns/return_to_report.rb
+++ b/app/controllers/concerns/return_to_report.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module ReturnToReport
+  extend ActiveSupport::Concern
+
+  private
+
+  def store_return_to_report_path
+    return unless params[:return_to] == "report"
+    return unless @planning_application.pre_application?
+
+    session[:return_to_report_paths] ||= {}
+    session[:return_to_report_paths][@planning_application.reference] = true
+  end
+
+  def report_path_or(default_path)
+    if session.dig(:return_to_report_paths, @planning_application.reference)
+      session[:return_to_report_paths].delete(@planning_application.reference)
+      planning_application_review_report_path(@planning_application)
+    else
+      default_path
+    end
+  end
+end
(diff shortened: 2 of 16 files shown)


app/assets/stylesheets/_main.scss
```diff
@@ -372,6 +372,9 @@ $govuk-page-width: 1100px;
 .display-flex {
   display: flex;
 }
+.justify-between {
+  justify-content: space-between;
+}
```
benjamineskola, line 377: Not the end of the world but I'm a little suspicious of turning this into a utility class. Because `justify-content` doesn't make sense on its own it'd always have to be used alongside a class that does `display: flex` too. Would it be worth extracting out a single class that does both, instead, and giving it a name specific to the usecase?

(I see `.proposal-details-sub-heading` does exactly the same thing, and maybe for the same reason: to put a heading on the same line as a status tag.)

benjamineskola: Yeah, I can't think of a particularly good name for it (hardest problem in computer science) but it definitely seems like something we're likely to reuse.

(When I checked where it could be used/combined, I noticed `app/views/planning_applications/neighbour_letter_batches/index.html.erb` also needs it, and even worse I just used `style="justify-content: space-between"` there, no idea what I was thinking 😩.)

benbaumann95: Will clean that up too

app/views/planning_applications/review/reports/show.html.erb
```diff
@@ -25,30 +25,106 @@
...
-      # ]},
-      ## sections with headings
-      # {title: "first", links: [
-      #  ["foo", "#"]
-      # ]},
-      # {title: "second", links: [
-      #  ["bar", "#"]
-      # ]}
+      {links: [["Pre-application outcome", "#pre-application-outcome"]]},
+      {links: [["Your pre-application details", "#pre-application-details"]]}
```
benjamineskola, line 32: This will create two lists with one item each which might be weird for screenreaders. It might be better to combine them into a single list:

```suggestion
      {links: [
        ["Pre-application outcome", "#pre-application-outcome"],
        ["Your pre-application details", "#pre-application-details"]
      ]}
```

benjamineskola (approved):

EGiataganas (approved):

app/controllers/concerns/return_to_report.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def report_path_or(default_path)
+    if session.dig(:return_to_report_paths, @planning_application.reference)
+      session[:return_to_report_paths].delete(@planning_application.reference)
+      planning_application_review_report_path(@planning_application)
+    else
+      default_path
+    end
+  end
+end
```
EGiataganas, line 24: this might be good to move it on BOPS-Core in a feature PR

PR #2254, opened 2025-04-07
Add next steps and disclaimer

benbaumann95:
### Description of change

Add next steps and disclaimer

### Story Link

https://trello.com/c/ReGzOnaT/550-pre-app-preview-and-submit-final-report-next-steps-and-disclaimer


engines/bops_reports/app/views/bops_reports/planning_applications/show.html.erb
```diff
@@ -133,4 +139,39 @@
...
+    </p>
+  <% end %>
+</section>
+
+<hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+
+<section id="disclaimer">
+  <h3 class="govuk-heading-s"><%= t(".disclaimer_heading") %></h3>
+  <%= govuk_warning_text do %>
+    <%= @planning_application.disclaimer || t(".disclaimer_content") %>
```
benjamineskola, line 171: Would it be worth setting the default disclaimer at the local authority admin level instead? Probably not a big deal but it might not be very clear to an admin that leaving the field blank there will lead to a default being set (that they can't actually see).

benjamineskola (approved):

engines/bops_reports/app/views/bops_reports/planning_applications/show.html.erb
benbaumann95: Yeah fair point but inclined to wait and see if this is a problem

PR #2263, opened 2025-04-09
Add considerations, advice, requirements

benjamineskola:
### Description of change

Adding sections to preapp report: guidance, considerations etc.

### Story Link

https://trello.com/c/Ar2TQi8u/549-pre-app-preview-and-submit-final-report-advice

#### Considerations

Screenshot 2025-04-10 at 08 03 49

#### Requirements

Screenshot 2025-04-09 at 11 19 32

### DIFF
--- a/app/assets/stylesheets/_main.scss
+++ b/app/assets/stylesheets/_main.scss
@@ -471,7 +471,7 @@ body {
 
 .comment-component {
   border-left: 10px solid $govuk-link-colour;
-  background-color: #F2F9FF;
+  background-color: #f2f9ff;
 }
 
 .govuk-details--borderless {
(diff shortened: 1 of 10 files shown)


benbaumann95 (commented):
Can we add some specs please 🙏 
We have the `spec/system/pre_application_report_spec.rb` which you could add to

rebeccaoneill:
Probably one for design as I've seen it matches figma but seems weird that the Planning considerations and advice table has 'complies' but the description of the acceptability says 'supported'?

benjamineskola:
@rebeccaoneill 

Good point. I think this is just in the database as 'complies' but it sounds like we might need to map it to 'supported', etc, at least for displaying it in the report page.

EGiataganas (approved):

benbaumann95 (approved):

PR #2274, opened 2025-04-15
Add application submission url to local admin

rebeccaoneill:
### Description of change

Add url to local admin for applicants to submit application for use in pre-app report.


engines/bops_reports/app/views/bops_reports/planning_applications/show.html.erb
```diff
@@ -460,7 +460,7 @@
 <section id="next-steps">
   <h3 class="govuk-heading-s"><%= t(".next_steps_heading") %></h3>
 
-  <% if @current_local_authority.submission_guidance_url %>
+  <% if @current_local_authority.submission_url %>
```
benjamineskola, line 463: I wonder if it would make sense to check that they're both set, and warn if not? As it is, it might now be possible to send the report with no guidance URL set.

benjamineskola (approved):

PR #2286, opened 2025-04-24
Ensure a bops-applicants token is created for every new local authority

benjamineskola:
### Description of change

bops-applicants sends an authentication key with its requests: the same key for every local authority (because b-a has no concept of a local authority). However, on the bops side, every key is associated with a single local authority, which means that every local authority needs a copy of the b-a key to be created for it. This was not being done automatically when a local authority was created, which led to requests from bops-applicants failing for that local authority.

### Story Link

<https://trello.com/c/8CIOsMX6/611-api-bearer-token-not-scoped-per-local-authority>

### Known issues

Ideally we would not have copies of the same key. But as bops-applicants stores no per-authority state, currently we can't do that (and adding that would lead to additional complexity when onboarding a local authority).

A one-to-many relation (one key, many authorities) might be nicer — _if_ this was a situation we wanted to maintain in the long run, which we don't. Keys should not be reused between authorities, so we shouldn't implement that feature; instead we should merge bops-applicants into bops as an engine, so it can share state properly. But in the short term, this PR will ensure that b-a requests work.


app/services/local_authority_creation_service.rb
```diff
@@ -49,4 +50,10 @@ def create_administrator_user
       user.otp_required_for_login = false
     end
   end
+
+  def assign_bops_applicants_token
+    ApiUser.find_or_create_by!(local_authority:, name: "bops-applicants") do |user|
+      user.token = ApiUser.where(name: "bops-applicants").first.token
+    end
+  end
```
pixeltrix, line 58: This only covers running the task - the web page doesn't call this service

engines/bops_config/app/controllers/bops_config/local_authorities_controller.rb
```diff
@@ -65,7 +65,9 @@ def set_local_authority
     end
 
     def build_local_authority
-      @local_authority = LocalAuthority.new
+      token = ApiUser.where(name: "bops-applicants").first.token
```
benbaumann95, line 68: Worth actually checking in the databases if the name is set like this for the specific token

benjamineskola: In theory the name could be anything, but then in theory the token could be anything too; this seems like the safest bet.

engines/bops_config/app/controllers/bops_config/local_authorities_controller.rb
```diff
@@ -65,7 +65,9 @@ def set_local_authority
...
+      api_user = ApiUser.new(name: "bops-applicants", token:)
+      @local_authority = LocalAuthority.new(api_users: [api_user])
```
benbaumann95, line 70: Think this logic could probably sit in a before_create in the model

engines/bops_config/app/controllers/bops_config/local_authorities_controller.rb
benbaumann95: Yep just meant we might have other naming on staging and prod for this bops applicants token we want to associate

pixeltrix (approved):

engines/bops_config/app/controllers/bops_config/local_authorities_controller.rb
```diff
@@ -65,7 +65,10 @@ def set_local_authority
     end
 
     def build_local_authority
-      @local_authority = LocalAuthority.new
+      token = ApiUser.find_by!(name: "bops-applicants").token
+      api_user = ApiUser.new(name: "bops-applicants", token:)
+
+      @local_authority = LocalAuthority.new(api_users: [api_user])
```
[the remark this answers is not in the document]
benjamineskola: Ah, I knew `find_or_create_by` took a block but I don't think I'd seen it used for `new`.

PR #2293, opened 2025-04-29
Add support for ODP schema v0.7.4

pixeltrix:
### Story Link

https://trello.com/c/bN5A4dHL


benjamineskola (approved):
There's no way to usefully comment on 74k lines of JSON so let's just go for it I guess. 🤷🏻

PR #2296, opened 2025-04-30
Link application type with requirements

rebeccaoneill:
### Description of change

Create a link between application type and local authority requirements, to be configured at local admin level. When a suggested application type is selected from pre-app the requirements would then pre-populate based on this application type.

This PR deals only with the configuration step in local admin, there is a separate ticket for the assessment step.

### Story Link

https://trello.com/c/5oIjFbPA/421-create-a-link-between-application-types-and-requirements-so-that-the-requirement-list-can-be-pre-populated

### Screenshots
image
image


engines/bops_admin/app/controllers/bops_admin/application_type_requirements_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    def set_application_type_requirements
+      @application_type_requirements = ApplicationTypeRequirement.includes(:local_authority_requirement).where(
+        local_authority_requirement: {local_authority_id: current_local_authority.id},
+        application_type_id: application_type_id
+      )
+    end
+
+    def set_application_type_requirement
+      @application_type_requirement = ApplicationTypeRequirement.where(
```
benjamineskola, line 82: This might need to be `find_by`, not `where`, if it needs a single result and not a collection.

engines/bops_admin/app/controllers/bops_admin/application_types_controller.rb
```diff
@@ -12,6 +13,8 @@ def index
     end
 
     def show
+      @application_type_requirements = ApplicationTypeRequirement.where(application_type_id: params[:id])
```
benjamineskola, line 16: I think this isn't necessary because `set_application_type_requirements` will do it, except for the mismatched variable name.

spec/models/application_type_requirement_spec.rb
benjamineskola, line 1: No need to add this file

benjamineskola (approved):
Looks good — just a few small things

EGiataganas (approved):

PR #2297, opened 2025-05-01
Fix policy section comment bug

benjamineskola:
### Description of change

Previously, the comment lookup traversed the multiple steps of references incorrectly, so that for the planning application policy class in question, it was looking up the global policy class, and then from there all the planning application policy sections for that policy class. This was then used to prepopulate the form field for a particular comment, but instead of being the most recent comment on the given planning-application-policy-section, it would be the most recent comment on the instance of a section for any other planning application, potentially at a different local authority.

This PR corrects the traversal so it looks only for related policy classes/sections and the comments on them.

(The way the comments are looked up causes some less critical but still counter-intuitive behaviour, which was initially reported as a bug and will be addressed separately — actually, fixing the initial bug first seems like it might have masked this one, either way I couldn't get my head around the initial one until this was out of the way.)

### Story Link

https://trello.com/c/ghVF20VP/625-comment-updates-not-persisting-on-policy-references

### DIFF
--- a/app/models/planning_application_policy_class.rb
+++ b/app/models/planning_application_policy_class.rb
@@ -12,13 +12,14 @@ class PlanningApplicationPolicyClass < ApplicationRecord
     validates :reporting_types, presence: true
   end
 
-  delegate :section, :name, :url, :description, to: :policy_class
-  delegate :planning_application_policy_sections, to: :planning_application
+  after_create :initialize_planning_application_policy_sections
 
-  def policy_sections
-    planning_application_policy_sections
-      .includes(:comments, :policy_section)
-      .where(policy_section: {policy_class_id: policy_class_id})
+  delegate :section, :name, :url, :description, to: :policy_class
+
+  def planning_application_policy_sections
+    planning_application.planning_application_policy_sections
+      .includes(:policy_section)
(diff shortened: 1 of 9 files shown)


app/models/planning_application_policy_class.rb
```diff
@@ -12,13 +12,14 @@ class PlanningApplicationPolicyClass < ApplicationRecord
     validates :reporting_types, presence: true
   end
 
+  after_create :initialize_planning_application_policy_sections
+
   delegate :section, :name, :url, :description, to: :policy_class
-  delegate :planning_application_policy_sections, to: :planning_application
```
benjamineskola, line 16: Can't delegate this because the policy sections will be for all policy classes on this application, not just this class.

app/models/planning_application_policy_class.rb
```diff
@@ -12,13 +12,14 @@ class PlanningApplicationPolicyClass < ApplicationRecord
...
 
-  def policy_sections
```
benjamineskola, line 18: Trying to avoid having methods called `policy_section` etc that actually deal with `PlanningApplicationPolicySection`s.

PR #2298, opened 2025-05-02
DSNPI-1251 update PostSubmissionPagination to return totalResults & totalItems (if present)

apricot13:
[DSNPI-1251](https://tpximpact.atlassian.net/browse/DSNPI-1251)

### Description of change

This PR goes along with this change to the ODP schema [here](https://github.com/theopensystemslab/digital-planning-data-schemas/pull/346)

It updates the Post-submission pagination view to return `totalResults` and `totalItems` (if `totalItems` is set in the controller)

* `totalResults` is the total number of results for this request 
* `totalItems` is the total number of items regardless of filtering. 

The `totalItems` field is optional since it could be a bit of an overhead for some endpoints to provide, but in the case of comments we're already calculating `totalItems` for the commentSummary.

### DIFF
--- a/engines/bops_api/app/views/bops_api/v2/shared/postsubmissionApplication/_pagination.json.jbuilder
+++ b/engines/bops_api/app/views/bops_api/v2/shared/postsubmissionApplication/_pagination.json.jbuilder
@@ -6,7 +6,7 @@ json.pagination do
   json.totalPages @pagy.pages
   json.totalResults @pagy.count
   if @total_responses.present?
-    json.totalItems @total_responses
+    json.totalAvailableItems @total_responses
   end
 end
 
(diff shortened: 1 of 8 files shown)


benjamineskola:
I'm not sure the names are very clear: `totalResults` and `totalItems` seem basically interchangeable to me? What about something like `totalResultsUnfiltered`?

(I'm guessing this is for cases where you're showing e.g. only positive comments and want to be able to show how many comments there are in total.)

apricot13:
@benjamineskola  have made a change to the spec and this PR. 

I think if I said `totalAvailableItems` to you out of the context of the rest of the fields you would think "ah, thats not related to my request it must be related to ✨all the things✨" - I hope! then `totalResults` will be fairly standard 🤞

benjamineskola:
Yeah that seems better — adding `Available` to the field name makes it clear that it's not interchangable.

apricot13:
@benjamineskola is this ok to be merged now?

benjamineskola (approved):

PR #2314, opened 2025-05-16
Add timestamps and request-IDs to error pages

benjamineskola:
### Description of change

This should help us identify an error more specifically in our logs without revealing the specific error, when given only a screenshot of the generic error page.

Screen Shot 2025-05-22 at 11 33 10


config/application.rb
```diff
@@ -64,6 +64,8 @@ class Application < Rails::Application
     # Ensure the correct user scope is used for authentication
     config.middleware.use BopsCore::Middleware::User, global_subdomains: %w[config]
 
+    config.exceptions_app = routes
```
pixeltrix, line 67: This is not a good idea - we've just had an exception and now we're sending the request back to the rack app that generated the exception. We should build a custom exceptions app if we really want to do this.

benjamineskola:
Will be superseded by #2318

PR #2316, opened 2025-05-19
Submissions endpoint

benbaumann95:
### Description of change

Create endpoint to receive submissions 

- Implement an authenticated API endpoint that accepts POST requests from Planning Portal for submission events. Use the BOPS submissions engine to handle this functionality. This feature will be extended to receive any submission case or application from any third party.
- Add swagger docs for the submissions endpoint

### Story Link

https://trello.com/c/G2Y8fRGf/56-create-webhook-endpoint-to-receive-submissions-and-withdrawals

Screenshot 2025-05-20 at 11 28 19


engines/bops_submissions/config/initializers/rswag_ui.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Rswag::Ui.configure do |c|
+  # c.openapi_endpoint "/submissions/api-docs/v1/swagger_doc.yaml", "Submission API V1 Docs"
+  c.openapi_endpoint "/api/docs/v2/swagger_doc.yaml", "Submission API V2 Docs"
+
+  # Rswag::UI doesn't provide a helper for this so we need to set it directly on config hash
+  c.config_object[:"urls.primaryName"] = "Submission API Docs"
```
pixeltrix, line 8: We don't want to default to the submission api do we?

engines/bops_submissions/config/initializers/rswag_ui.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Rswag::Ui.configure do |c|
+  # c.openapi_endpoint "/submissions/api-docs/v1/swagger_doc.yaml", "Submission API V1 Docs"
+  c.openapi_endpoint "/api/docs/v2/swagger_doc.yaml", "Submission API V2 Docs"
```
pixeltrix, line 5: You have to remember that the RSwag configuration is a singleton so if they're both loaded then they'll conflict

pixeltrix (commented):
Not sure about using that zip file as fixture - is it sample data or a real application? Also .DS_Store files?

benbaumann95:
> Not sure about using that zip file as fixture - is it sample data or a real application? Also .DS_Store files?

It's sample data and oops just shoved everything up in the first draft but removed now 👍

engines/bops_submissions/app/services/bops_submissions/creation_service.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    attr_reader :params, :headers, :local_authority
+
+    def call
+      submission = local_authority.submissions.create!(
+        request_headers: filtered_request_headers,
+        request_body: permitted_request_params
+      )
+
+      submission.update!(external_uuid: SecureRandom.uuid_v7)
```
benjamineskola, line 30: What is this UUID for?

benbaumann95: Agreed with PP that they will receive this as a success message and possible identifier for future interaction with that resource

benbaumann95: Also don't want to be giving back a database-generated ID

benjamineskola: Agreed we shouldn't be sending back the database-generated integer primary keys.

benbaumann95: Hm I guess we could make it the primary key

app/models/submission.rb
```diff
@@ -29,4 +31,8 @@ class Submission < ApplicationRecord
       transitions from: :started, to: :completed
     end
   end
+
+  def application_reference
+    request_body["applicationRef"]
```
pixeltrix, line 36: Is this a PP specific reference? The submission model should be neutral and any service specific stuff in a class that represents the submission using a consistent interface.

benbaumann95: Yes currently it is but we can extract / modify it when adding PlanX submissions

PR #2317, opened 2025-05-19
Mark consultation as starting immediately but retain existing end date calculation

benjamineskola:
### Description of change

We want the consultation start date, as presented in the API etc, to be the actual date on which the planning application is made public; however, we need the end date to remain calculated, as currently, from the next business day. This means the old start date gets retained internally for the end date calculation, but the new start date is just immediate.

### Story Link

https://trello.com/c/3lHvT7rJ/557-consultation-start-date-doesnt-align-with-dpr-statuses

### DIFF
--- a/spec/system/planning_applications/publicity/press_notice_spec.rb
+++ b/spec/system/planning_applications/publicity/press_notice_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe "Press notice" do
   let!(:assessor) { create(:user, :assessor, local_authority:) }
 
   let!(:planning_application) do
-    create(:planning_application, :prior_approval, local_authority:)
+    create(:planning_application, :prior_approval, local_authority:, postcode: "")
   end
 
   before do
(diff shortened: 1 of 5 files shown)


spec/system/planning_applications/publicity/press_notice_spec.rb
```diff
@@ -7,7 +7,7 @@
   let!(:assessor) { create(:user, :assessor, local_authority:) }
 
   let!(:planning_application) do
-    create(:planning_application, :prior_approval, local_authority:)
+    create(:planning_application, :prior_approval, local_authority:, postcode: "")
```
benjamineskola, line 10: I'm not entirely sure why this was necessary, but changing the `travel_to`s in this spec caused webmock to start complaining about there not being any stubs for the mapit api. A blank postcode means no requests will be made to the mapit api so avoids the issue.

pixeltrix: That sounds like the behaviour I had on Petitions with that Chrome bug - requests were still happening after the end of the spec and webmock had been reset.

EGiataganas (approved):

PR #2324, opened 2025-05-22
View submission table in local admin

benbaumann95:
### Description of change

- List submission with some metadata and add show page for viewing request body and headers in more detail

Screenshot 2025-05-20 at 17 32 26
Screenshot 2025-05-20 at 17 31 55

### DIFF
--- a/engines/bops_admin/app/controllers/bops_admin/submissions_controller.rb
+++ b/engines/bops_admin/app/controllers/bops_admin/submissions_controller.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+module BopsAdmin
+  class SubmissionsController < ApplicationController
+    before_action :set_submissions, only: %i[index]
+
+    rescue_from Pagy::OverflowError do
+      redirect_to submissions_path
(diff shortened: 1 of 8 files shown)


engines/bops_admin/app/controllers/bops_admin/submissions_controller.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module BopsAdmin
+  class SubmissionsController < ApplicationController
+    before_action :set_submissions, only: %i[index]
+
+    rescue_from Pagy::OverflowError do
+      redirect_to submissions_path
+    end
```
benjamineskola, line 9: Probably fine to do it this way, but there's also an optional extra in pagy itself to configure the overflow behaviour: https://ddnexus.github.io/pagy/docs/extras/overflow/ (although it looks like redirecting to the first page isn't actually an option in that case).

benjamineskola (approved):

EGiataganas (approved):

PR #2325, opened 2025-05-22
Add permissions to ApiUser

benjamineskola:
### Description of change

Add permissions to ApiUser model:

- `planning_application:read`
- `planning_application:submit`
- `comment:read`
- `comment:submit`

We can add different types as necessary for enforcements etc.

I don't think there's really much need for a `...:all` type of permission, as we don't have enough types of permission and we probably don't have any keys that need to do both (or if they do they can just have the multiple kinds of permission).

### Story Link

<https://trello.com/c/g5wll1n9/472-add-scopes-to-api-keys>

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


engines/bops_core/app/controllers/concerns/bops_core/application_controller.rb
```diff
@@ -64,12 +64,26 @@ def current_api_user
...
+      elsif params[:controller].include?("validation_request")
+        ":update"
+      else
+        ":submit"
+      end
+
+      raise NotImplementedError, "No permissions have been defined for this authenticated method for #{params[:controller]}:#{request.method}" if required_api_key_scope.nil?
+
+      required_api_key_scope == :any || key.permissions.include?(required_api_key_scope + scope_type)
+    end
```
benjamineskola, line 85: I'm not sure if there's a better approach for this.

I had initially thought of just defining a permission for each action somehow (so mapping `CommentsController#create` to `comment:submit` and so on), but because the access control is done before the action, I don't know if that's possible. So instead this is defining the first half of the scope on the controller level, which should be fine as long as we never want some actions that need `planning_application:...` scope and some that need `comment:...` scope in the same controller; then the second half of the scope comes from whether it's a GET request (read-only) or anything else (needs write permissions).

engines/bops_core/app/controllers/concerns/bops_core/application_controller.rb
```diff
@@ -64,12 +64,24 @@ def current_api_user
     end
 
     def authenticate_api_user!
```
benbaumann95, line 66: Feel like this logic could be broken down into a concern so the controller remains focused on handling HTTP requests while the concern handles finding the ApiUser and enforcing permissions. But perhaps a wider observation than just here.

benbaumann95 (approved):

PR #2328, opened 2025-05-29
Mapping planning portal submission

rebeccaoneill:
### Description of change

Created parsers and creation service to be able to create a new `PlanningApplication` record from a PlanningPortal submission.

The different field names have been documented here: https://docs.google.com/spreadsheets/d/1iSJdMLSQ3_u4n_Z4hEbXz66v3lr6BjJRLoHJ9x2D0LE/edit?usp=sharing

### Story Link

https://trello.com/c/iJTTrSp2/58-build-planning-portal-parser-to-map-data

### Known issues

1. Planning portal submit coordinates in Easting & northing format, this needs converting to degrees coordinates. There is a separate piece of work looking at this which will be merged later. At present the `Planning Application` record can be created but we'd be unable to render the map.
2. Planning portal do not currently send `ApplicationType` as part of the submission. For purpose of passing validation I have assigned all `PlanningApplications` as Full Householder. This will need to be updated at a later stage.
3. The geojson format is slightly different from the planx submission, would need more testing to understand downstream impact.


engines/bops_submissions/spec/services/application/creation_service_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      travel_to("2023-12-13") do
+        example.run
+      end
+    end
+
+    context "when successfully calling the service with params" do
+      let(:planning_application) { PlanningApplication.last }
+
+      context "when application type is planning permission full householder" do
+        let(:params) { json_fixture("v2/valid_planning_portal_planning_permission.json").with_indifferent_access }
```
rebeccaoneill, line 26: this is working locally but not when pushed to GitHub, unsure why?

benbaumann95 (approved):

PR #2329, opened 2025-05-30
Implement submission processor job and zip extraction service

benbaumann95:
### Description of change

**Implement submission processing job and zip extraction flow**

This PR introduces the SubmissionProcessorJob and ZipExtractionService to handle processing of Planning Portal application ZIPs

**This enables us to:**
- Automatically kick off background processing once a submission is created
- Fetch and unpack ZIPs, falling back from Zip::InputStream to Zip::File when necessary (e.g. Bit 3 header flag)
- Extract and attach documents to the submission, storing metadata and handling failures per file
- Store the Application.json contents in application_payload (as json_file)
- Extract additional files and store their names in the other_files field

**Admin UI updates:**
Enhanced the index and show pages to include:
- Status, timestamps, and external UUID
- Request headers and body
- Application payload and other files
- Documents table, if files have been extracted

### Story Link

https://trello.com/c/2yHDjAAx/57-implement-submission-processor-job

### Screenshots

Screenshot 2025-05-30 at 15 15 32
Screenshot 2025-05-30 at 15 16 03


engines/bops_admin/app/helpers/bops_admin/application_helper.rb
```diff
@@ -71,5 +71,16 @@ def nav_items
         {text: "Settings", href: profile_path, active: active_page_key?("settings")}
       ]
     end
+
+    def link_to_document(link_text, document, **args)
```
benjamineskola, line 75: Would it make sense to move this into BopsCore and reference it from there? This'll be the fourth version of this helper, I think.

engines/bops_submissions/app/services/bops_submissions/zip_extraction_service.rb
```diff
@@ -0,0 +1,124 @@
...
+    def initialize(submission:)
+      @submission = submission
+    end
+
+    attr_reader :submission
+
+    def call
+      submission.document_link_urls.each do |url|
+        process_with_zip_input_stream(url)
+      rescue Zip::Error => e
```
benjamineskola, line 24: Do we know why Zip::InputStream sometimes fails when Zip::File works? Is it simpler always to rely on Zip::File or are there other downsides to that?

benjamineskola (approved):

engines/bops_submissions/app/services/bops_submissions/zip_extraction_service.rb
benjamineskola: Oh never mind, I see it's in the second commit message 👍🏻

benjamineskola: It's a little annoying that these two `extract_*` methods are kind of similar and yet not similar enough.

It'd be ideal if we were able to do something like `def extract(url, extractor_class)` and call it like `extract(url, Zip::InputStream)` or `extract(url, Zip::File)`, but it looks like the behaviour of the two classes is different in a few places. Oh well, can't be helped.

benbaumann95: Yeah I was experimenting abstracting and using a yield type approach but think it's just different enough that it makes sense to be separated

pixeltrix: Streaming will be more efficient but there may be some zip archives where you have to load it all to decompress

engines/bops_admin/app/helpers/bops_admin/application_helper.rb
benbaumann95: Yeah probably should, I'll do it now as a separate PR

PR #2339, opened 2025-06-09
Fix edge case where consultation is present but not started

pixeltrix:


benjamineskola (approved):
Took me a moment to realise why the title referred to consultations that aren't started, but the code checked the end date. But of course every started consultation has an end date.

PR #2342, opened 2025-06-10
Automatically apply changes from validation requests

pixeltrix:
If a request is approved automatically apply the changes.

https://trello.com/c/RnXntStR

### DIFF
--- a/app/models/time_extension_validation_request.rb
+++ b/app/models/time_extension_validation_request.rb
@@ -62,7 +62,7 @@ class TimeExtensionValidationRequest < ValidationRequest
     super&.to_date
   end
 
-  def update_planning_application!(params)
+  def update_planning_application!(*)
     self.previous_expiry_date = planning_application.expiry_date
     save!(validate: false)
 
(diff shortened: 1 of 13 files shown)


app/models/time_extension_validation_request.rb
```diff
@@ -62,7 +62,7 @@ def previous_expiry_date
     super&.to_date
   end
 
-  def update_planning_application!(params)
+  def update_planning_application!(*)
```
pixeltrix, line 65: Changed to `*` to make it clear they aren't used here

app/models/validation_request.rb
```diff
@@ -260,9 +260,8 @@ def time_extension_request?
     type == "TimeExtensionValidationRequest"
   end
 
-  def update_planning_application!(params)
-    # Specific types of validation request use this method, which is overwritten in those models.
-    #  A couple don't hence the empty method
+  def update_planning_application!(*)
+    # Subclasses must implement any necessary updates
```
pixeltrix, line 264: The only place that is passing params here is a legacy v1 API call for ownership certificate updates - which should be just bops-applicants so once the merge is final we can remove a lot of complexity.

PR #2343, opened 2025-06-11
Use consultee response form for all application types

pixeltrix:
### Description of change

Consultee has access to a response form on all application types

### Story Link

https://trello.com/c/fumWzEdx


app/models/consultee.rb
```diff
@@ -95,11 +95,7 @@ def last_response
   end
 
   def application_link
-    if consultation.planning_application.pre_application?
-      routes.consultees_magic_link
-    else
-      consultation.application_link
-    end
+    routes.consultees_magic_link
```
benjamineskola, line 98: `application_link` is only ever used in one place (in the consultation model) and doesn't really depend on the consultee at all, so perhaps we can just get rid of it and call `consultees_magic_link` directly? Not vital though.

benjamineskola (approved):

EGiataganas (approved):

app/models/consultee.rb
benjamineskola: Oh right, it didn't occur to me that `routes` was a method on the consultee. That makes sense (and that's where it gets the sgid). So yeah, any improvement to this is unrelated to this PR.

PR #2344, opened 2025-06-11
Add some docs about our rails engines

benbaumann95:
Add documentation around our Rails engines

### DIFF
--- a/docs/FAQs.md
+++ b/docs/FAQs.md
@@ -35,9 +35,7 @@ BOPS is integrated with:
 
 ### What does the architecture look like?
 
-Here is a very high-level view of how BOPS works:
-
-![Architecture](images/architecture.png)
+- [Docs](docs/architecture.md)
 
 ### Where can I find out more about terms that are used in the planning process?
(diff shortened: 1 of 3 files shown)


docs/FAQs.md
```diff
@@ -35,9 +35,7 @@ BOPS is integrated with:
 
 ### What does the architecture look like?
 
-Here is a very high-level view of how BOPS works:
-
-![Architecture](images/architecture.png)
```
benbaumann95, line 40: This is outdated now, might need to revisit at some point

docs/architecture.md
```diff
@@ -0,0 +1,25 @@
...
+work together to serve the back-office planning system as well as the
+public and authenticated API.
+
+## Engines and their roles
+
+| Engine               | Purpose                                                                            |
+| -------------------- | ---------------------------------------------------------------------------------- |
+| **bops_core**        | Shared helpers, routing helpers and base configuration used by the other engines.  |
+| **bops_admin**       | Admin interface for managing users, application types, consultees and policy data per local authority. |
+| **bops_applicants**  | Applicant interface for responding to officer requests.                            |
```
pixeltrix, line 14: It's also neighbour comments but maybe we want to split that out

PR #2346, opened 2025-06-11
Transform redline coordinates from PP submission to lat long

rebeccaoneill:
### Description of change

Submissions from planning portal have coordinates in easting and northing, the submitted redline boundary coordinates need to be transformed to latitude and longitude to be readable by the map component. 

### Story Link

https://trello.com/c/6wy6dVrb/205-red-line-boundary-coordinates-transfromation


engines/bops_submissions/spec/services/application/parsers/proposal_parser_spec.rb
```diff
@@ -16,11 +16,10 @@
       }
 
       it "returns a correctly formatted proposal hash" do
-        expected_geojson = params.dig("polygon", "features", 0, "geometry").to_json
-
+        parser = described_class.new(params, local_authority:)
         expect(parse_proposal).to eq(
           description: "\nDH Test Description",
-          boundary_geojson: expected_geojson
+          boundary_geojson: parser.send(:build_boundary_feature)
```
benjamineskola, line 22: This sort of just runs the `build_boundary_feature` method twice (indirectly) and checks that it returns the same thing both times.

I wonder if it might be more useful with the `hash_including` method, something like:

```suggestion
          boundary_geojson: hash_including("boundary" => hash_including(
              "coordinates" => ...
            )
          )
```

benbaumann95: Yeah this spec would never fail. We should check the actual output from this conversion so probably want to check the actual coordinates are transformed.

PR #2355, opened 2025-06-18
Redirect to review documents after adding reference

rebeccaoneill:
### Description of change

Currently when reviewing documents there is a link back to the edit document page in oder to add a reference and change the publish documents information. When you save this page it re-routes you back to the documents index page whereas it should return to the document review page, this PR corrects that.

### Story Link

https://trello.com/c/nqYT2pS7/767-able-to-navigate-back-from-the-add-drawing-reference-journey


app/controllers/documents_controller.rb
```diff
@@ -149,6 +150,8 @@ def redirect_url
       supply_documents_planning_application_path(@planning_application)
     elsif session[:return_to]
       return_to_session
+    elsif request.referer&.include?("route=review")
```
Rhian, line 153: This could maybe be a case statement?

PR #2369, opened 2025-06-25
Replace my/all planning applications button with tabs

benjamineskola:
### Description of change

- Remove the 'my applications' and 'all applications' buttons from the dashboard.
- Remove unassigned applications from the default tab , so that it contains only 'my'
- Add an 'all' tab and an 'unassigned' tab
- Remove unused code from PlanningApplicationSearch (so that it no longer does any 'smart' filtering by user or selects different application types based on the user role)
- Add a 'closed' option to the filters, to enable viewing them in 'all cases' (off by default)

### Story Link

https://trello.com/c/ikiMZDQP/570-differentiate-between-new-cases-and-assigned-cases-on-dashboard

### Screenshots

Screenshot 2025-06-27 at 16 17 58

### DIFF
--- a/spec/system/filter_spec.rb
+++ b/spec/system/filter_spec.rb
@@ -395,7 +395,7 @@ RSpec.describe "filtering planning applications", type: :system, capybara: true 
 
         within(selected_govuk_tab) do
           expect(page).to have_content(closed_planning_application.reference)
-          # expect(page).to have_content(other_closed_planning_application.reference)
+          expect(page).not_to have_content(other_closed_planning_application.reference)
           expect(page).not_to have_content(not_started_planning_application.reference)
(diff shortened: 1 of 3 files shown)


spec/system/filter_spec.rb
```diff
@@ -396,7 +395,7 @@
 
         within(selected_govuk_tab) do
           expect(page).to have_content(closed_planning_application.reference)
-          expect(page).to have_content(other_closed_planning_application.reference)
+          # expect(page).to have_content(other_closed_planning_application.reference)
```
benjamineskola, line 398: This and another assertion later are commented out because they're closed and belonging to another user, and waiting to confirm what (if anything) we want to do about that.

PR #2376, opened 2025-06-26
Documents consistency

EGiataganas:
### Description of change

- Keep consistent pattern when representing documents
- Add drawing number on the API

### Story Link

https://trello.com/c/4eqDEclW

### DIFF
--- a/app/components/evidence_groups/documents_component.html.erb
+++ b/app/components/evidence_groups/documents_component.html.erb
@@ -5,14 +5,16 @@
       <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
           <p class="govuk-body govuk-!-margin-bottom-1">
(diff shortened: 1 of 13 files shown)


app/components/evidence_groups/documents_component.html.erb
benjamineskola, line 1: Would it be worth extracting this out into a partial since it's repeated three times? Perhaps not if we're going to be redesigning this all anyway.

benjamineskola (approved):

app/components/evidence_groups/documents_component.html.erb
EGiataganas: Well I was wondering whether I should extract it on a component or not. But as we will do the DMS redesign I was like is not worth it. Plus what's common is the sequence we want to represent it.

For example on the evidence we show different information compare to others. Also in the applicant we only show the drawing number.

PR #2378, opened 2025-06-27
Remove publish to bops portal for pre-apps

rebeccaoneill:
### Description of change

PreApps are not required to be published, this PR removes the question to ask if the application should be published as well as hiding the item in the dates and assignment header.

### Story Link

https://trello.com/c/sStA0KFC/698-skip-publish-application-on-bops-public-portal-for-pre-apps

### Screenshots

image

### DIFF
--- a/app/views/planning_applications/confirm_validation.html.erb
+++ b/app/views/planning_applications/confirm_validation.html.erb
@@ -23,9 +23,13 @@
       <div class="govuk-hint" id="planning-application-validated-at-hint">
         Will be marked as valid from <%= @planning_application.valid_from_date.to_date.to_fs %>
       </div>
-      <%= form.govuk_radio_buttons_fieldset :make_public, legend: {text: "Publish application on BOPS Public Portal?"} do %>
-        <%= form.govuk_radio_button :make_public, true, label: {text: "Yes"} %>
-        <%= form.govuk_radio_button :make_public, false, label: {text: "No"} %>
+      <% if @planning_application.publishable? %>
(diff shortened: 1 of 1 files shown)


app/views/shared/_dates_and_assignment_header.html.erb
```diff
@@ -36,7 +36,7 @@
       </p>
     <% end %>
 
-    <% if @planning_application.publishable? %>
+    <% if @planning_application.publishable? && !@planning_application.pre_application? %>
```
benjamineskola, line 39: I thought for a moment that it would be good to just modify `publishable?` so that it returns false for preapps, but it looks like that's just delegated to the application type feature. That makes me wonder if the application type is configured wrongly, then?

app/views/planning_applications/confirm_validation.html.erb
```diff
@@ -23,9 +23,13 @@
       <div class="govuk-hint" id="planning-application-validated-at-hint">
         Will be marked as valid from <%= @planning_application.valid_from_date.to_date.to_fs %>
       </div>
-      <%= form.govuk_radio_buttons_fieldset :make_public, legend: {text: "Publish application on BOPS Public Portal?"} do %>
-        <%= form.govuk_radio_button :make_public, true, label: {text: "Yes"} %>
-        <%= form.govuk_radio_button :make_public, false, label: {text: "No"} %>
+      <% if @planning_application.pre_application? %>
```
benjamineskola, line 26: Might make sense for this conditional to be the same as the one in the other file.

benjamineskola (approved):
Probably fine as it is, but it might be possible to make this less dependent on checking specifically for preapps.

PR #2384, opened 2025-07-01
Create CaseRecord

benjamineskola:
### Description of change

- **Create CaseRecord model**
- **Create Enforcement model**
- **Add Caseable delegated type**

### Story Link

https://trello.com/c/3EtDzqy1/761-create-enforcement-classes-and-tables
<https://link-to-issue>

### DIFF
--- a/app/models/case_record.rb
+++ b/app/models/case_record.rb
@@ -3,6 +3,8 @@
 require "securerandom"
 
 class CaseRecord < ApplicationRecord
+  delegated_type :caseable, types: %w[Enforcement], dependent: :destroy
+
   belongs_to :local_authority
 
   after_initialize :generate_uuid
--- a/app/models/concerns/caseable.rb
+++ b/app/models/concerns/caseable.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+module Caseable
+  extend ActiveSupport::Concern
+
+  included do
+    has_one :case_record, as: :caseable, touch: true, dependent: :destroy
+    delegate :local_authority, to: :case_record
+  end
+end
--- a/app/models/enforcement.rb
+++ b/app/models/enforcement.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
 
 class Enforcement < ApplicationRecord
+  include Caseable
 end


app/models/case_record.rb
```diff
@@ -0,0 +1,17 @@
...
+
+class CaseRecord < ApplicationRecord
+  delegated_type :caseable, types: %w[Enforcement], dependent: :destroy
+
+  before_create :generate_uuid_v7
+
+  private
+
+  def generate_uuid_v7
+    return if self.class.attribute_types["id"].type != :uuid
```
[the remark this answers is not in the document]
benjamineskola: The blog post I got it from put it in ApplicationRecord so was accounting for columns of different types.

db/schema.rb
```diff
@@ -175,6 +175,13 @@
     t.index ["user_id"], name: "index_audits_on_user_id"
   end
 
+  create_table "case_records", id: :uuid, default: nil, force: :cascade do |t|
+    t.string "caseable_type"
+    t.bigint "caseable_id"
```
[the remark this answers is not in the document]
benjamineskola: I left `caseable_id` as a bigint because it needs to support `PlanningApplication` which has integer ids.

app/models/case_record.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require "securerandom"
+
+class CaseRecord < ApplicationRecord
+  delegated_type :caseable, types: %w[Enforcement], dependent: :destroy
+
+  after_initialize :generate_uuid_v7
```
pixeltrix, line 8: I'd leave the `_v7` off as that's an implementation detail of the private method

app/models/case_record.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require "securerandom"
+
+class CaseRecord < ApplicationRecord
+  delegated_type :caseable, types: %w[Enforcement], dependent: :destroy
+
```
pixeltrix, line 7: Needs the `belongs_to :local_authority`

app/models/concerns/caseable.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Caseable
+  extend ActiveSupport::Concern
+
+  included do
+    has_one :case_record, as: :caseable, touch: true, dependent: :destroy
```
pixeltrix, line 7: Add a delegation inside the `included` block:

``` ruby
delegate :local_authority, to: :case_record`
```

db/migrate/20250701092007_create_case_record.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateCaseRecord < ActiveRecord::Migration[7.2]
+  def change
+    create_table :case_records, id: false do |t|
+      t.uuid :id, primary_key: true
+      t.references :local_authority, null: false
```
pixeltrix, line 7: Need to add index/foreign key options

db/migrate/20250701092007_create_case_record.rb
```diff
@@ -0,0 +1,14 @@
...
+      t.string :caseable_type, null: false
+      t.bigint :caseable_id, null: false
```
pixeltrix, line 9: I think we want an index on these two columns so doing `enforcement.case_record` doesn't result in a table scan

pixeltrix:
Do we also need?

``` ruby
class LocalAuthority < ApplicationRecord
  has_many :case_records
end
```

So that we can scope the query in the controller.

PR #2388, opened 2025-07-02
Mark heads of terms complete when saving

benjamineskola:
Fixes broken test


pixeltrix (approved):
We should really refactor all of these different patterns to a single style of an index controller that updates the review on the collection (ConditionSet, InformativeSet, etc.) and an items controller that deals with creating, editing and updating the individual items of the collection. However that's for another day.

PR #2403, opened 2025-07-08
Create v2 post endpoint

gabimartini:
### Description of change

We Implemented the /api/v2/planning_application/{reference}/comments/public POST endpoint to accept comment submissions in the existing DprCommentSubmission format


engines/bops_api/app/controllers/bops_api/v2/planning_applications/neighbour_responses_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+        end
+
+        def required_api_key_scope = "comment"
+
+        private
+
+        def build_neighbour_response
+          @consultation.neighbour_responses.build(response_attributes).tap do |response|
+            response.neighbour = find_or_create_neighbour
+          end
```
pixeltrix, line 40: No need for the `tap` - just pass the neighbour in the attributes hash:

``` ruby
def response_attributes
  params
    .permit(:name, :email, :response, :summary_tag, tags: [])
    .merge(received_at: Time.zone.new, neighbour: find_or_create_neighbour)
end
```

engines/bops_api/app/controllers/bops_api/v2/planning_applications/neighbour_responses_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+          response_params.except(:address, :planning_application_id).merge!(
+            received_at: Time.zone.now,
+            consultation_id: @consultation.id
+          )
+        end
+
+        def response_params
+          params.permit(
+            :address, :name, :email, :response, :summary_tag,
+            :redacted_response, tags: []
```
pixeltrix, line 62: Should `redacted_response` be here? Isn't that done by a planning officer?

engines/bops_api/app/controllers/bops_api/v2/planning_applications/neighbour_responses_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+
+          render json: {
+            data: nil,
+            status: {
+              code: 200,
+              message: "OK",
+              detail: "Neighbour response created successfully"
+            }
+          }, status: :ok
+        rescue ActiveRecord::RecordInvalid
```
pixeltrix, line 22: We handle `ActiveRecord::RecordInvalid` in `BopsApi::ErrorHandler` - is there a reason why you're rescuing it here?

engines/bops_api/app/controllers/bops_api/v2/planning_applications/neighbour_responses_controller.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+module BopsApi
+  module V2
+    module PlanningApplications
+      class NeighbourResponsesController < AuthenticatedController
+        def create
+          @planning_application = find_planning_application(params[:planning_application_id])
+          @consultation = @planning_application.consultation
```
pixeltrix, line 9: It's generally idomatic Rails controller style to move these to before actions, e.g.

``` ruby
before_action :set_consultation
before_action :build_neighbour_response

def create
  @neighbour_response.save!

  respond_to do |format|
    format.json
  end
end

private

def set_planning_application
  @planning_application = find_planning_application(params[:planning_application_id])
end

def set_consultation
  @consultation = @planning_application.consultation
end

def build_neighbour_response
  @neighbour_response = @consultation.neighbour_responses.new(response_attributes)
end
```

engines/bops_api/app/controllers/bops_api/v2/planning_applications/neighbour_responses_controller.rb
apricot13: ah this one might be my fault I may have said to add it in too soon! 

We will need to make some sort of Postsubmission response handler so we can implement the [ApiResponse structure](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/IMPLEMENTATION.md#api-response-structure) for the ODP public endpoints. 

For now we can use your [previous comment](https://github.com/unboxed/bops/pull/2403/files#r2195003597) and let `BopsApi::ErrorHandler` handle it for now. 

I'll make a ticket in our backlog for the handler, @pixeltrix do you have any recommendations for how best to approach it?

bops-applicants/yarn.lock
benjamineskola, line 1: This file shouldn't be created/added, and we don't use the `bops-applicants` submodule/subdirectory any more.

apricot13:
Just a heads-up that this PR is in draft while Gabi’s away. I’ll be picking it up and working on the fixes for now.

apricot13:
I've made some updates to this PR and it's ready for re-review @benjamineskola @pixeltrix 

* The duplicate service test has been removed along with some other small files.
* Tests have been updated accordingly.
* The implementation now follows the v1 endpoint more closely.

One thing to flag: there's a potential issue with the neighbour response creation service when used by this endpoint on the DPR. If someone submits a response via the DPR from the same address as an existing neighbour, it will allow them to post a comment that appears to come from that neighbour. This could be a spoofing risk worth looking into.

benjamineskola:
> If someone submits a response via the DPR from the same address as an existing neighbour, it will allow them to post a comment that appears to come from that neighbour. This could be a spoofing risk worth looking into.

I don't think it should ever be assumed that two comments with the same address are from the same person: even without malicious intent, it's perfectly normal for a single address to have more than one resident. I don't think submitting the same address as a previous commenter gives any special access to the previous commenter's data, and if the comments are being displayed to the public they should be presented as if they are from the same user.

PR #2405, opened 2025-07-09
Create enforcement case via submissions API

benbaumann95:
### Description of change

1. New enforcement model associations & fields

- Introduced ProposalDetails, UPRN, boundary_geojson, lonlat, and application_type_id to enforcements table via migrations.
- Added proper foreign‐key constraints, and updated schema to support geographic types. 

2. Associations

- Linked CaseRecord to Submission.
- Ensured Enforcement ties to ApplicationType through enforced foreign key. 

3. API plumbing – V2 submissions endpoint

- Updated the /v2/submissions#create controller and its schema validation to accept enforcement payloads when data is present . 
- Submission processor job now routes enforcement submissions to Enforcement::CreationService, whereas planning submissions use the existing application flow. 

4. Enforcement creation logic

- Implemented Enforcement::CreationService to map submission payloads into Enforcement cases via the Submissions API. 

5. Admin UI visibility & tests

- Updated the admin “show” view to surface enforcement submission payload (headers/body).
Enhanced system specs to verify enforcement submissions appear in the admin UI and that payload details are shown.

### Story Link

https://trello.com/c/8TnuLfnX/867-create-enforcements-from-api

Screenshot 2025-07-11 at 11 25 47

### DIFF
--- a/app/models/enforcement.rb
+++ b/app/models/enforcement.rb
@@ -3,8 +3,6 @@
 class Enforcement < ApplicationRecord
   include Caseable
 
-  belongs_to :application_type
-
   composed_of :address,
     mapping: {
       address_1: :line_1,
--- a/app/models/enforcement.rb
+++ b/app/models/enforcement.rb
@@ -2,6 +2,9 @@
 
 class Enforcement < ApplicationRecord
   include Caseable
+
+  belongs_to :application_type
+
   composed_of :address,
     mapping: {
       address_1: :line_1,
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -5,6 +5,7 @@ class Submission < ApplicationRecord
 
   belongs_to :local_authority
   has_one :planning_application, dependent: :nullify
+  has_one :case_record, dependent: :nullify
   has_many :documents, dependent: :destroy
 
   validates :external_uuid, uniqueness: true, allow_nil: true
@@ -38,15 +39,22 @@ class Submission < ApplicationRecord
   store_accessor :application_payload, :json_file, :other_files
 
   def application_reference
-    request_body["applicationRef"]
+    @application_reference ||= request_body["applicationRef"] || request_body["metadata"]["id"]
   end
 
   def document_link_urls
-    request_body.fetch("documentLinks", []).pluck("documentLink")
+    @document_link_urls ||= request_body.fetch("documentLinks", []).pluck("documentLink")
   end
 
   def source
-    # TODO: this should be dynamic based on the source of the submission
-    "Planning Portal"
+    @source ||= request_body.dig("metadata", "source") || "Planning Portal"
+  end
+
+  def planning_portal?
+    source == "Planning Portal"
+  end
+
+  def planx?
+    source == "PlanX"
   end
 end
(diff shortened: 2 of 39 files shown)


app/models/enforcement.rb
```diff
@@ -11,6 +14,10 @@ class Enforcement < ApplicationRecord
       postcode: :postcode
     }
 
+  alias_attribute :address_1, :address_line_1
+  alias_attribute :address_2, :address_line_2
+  alias_attribute :description, :breach_description
```
benjamineskola, line 19: I'd like to avoid introducing compatibility aliases if we can help it, but maybe it's worth it temporarily to get this done quickly?

(If we migrate the PlanningApplication to use the address model, then we can refer to `address.line_1` etc on both PlanningApplication and Enforcement, perhaps.)

benbaumann95: Happy to do that as a separate PR and then revert/refactor this change

app/models/enforcement.rb
```diff
@@ -2,6 +2,9 @@
 
 class Enforcement < ApplicationRecord
   include Caseable
+
+  belongs_to :application_type
```
benjamineskola, line 6: I'm not sure about this. I realise RAB is representing it as a 'type' of planning application but that feels sort of like a workaround for them having to use the ODP submission schema, and I don't know if we gain anything by implementing the same workaround. Particularly when the 'type' is always going to be `breach`, which we can tell because it's stored in a different table; and any behaviour that would differ based on application type features won't be shared with enforcements anyway.

Is it going to be possible to handle this in the submission service without recording the type they send?

app/models/submission.rb
```diff
@@ -5,6 +5,7 @@ class Submission < ApplicationRecord
 
   belongs_to :local_authority
   has_one :planning_application, dependent: :nullify
+  has_one :case_record, dependent: :nullify
```
benjamineskola, line 8: This makes me think we should aim to migrate PlanningApplication to CaseRecord soon, but that's going to be fiddly. Started testing it out in #2416.

app/models/submission.rb
```diff
@@ -38,15 +39,22 @@ class Submission < ApplicationRecord
...
-    "Planning Portal"
+    @source ||= request_body.dig("metadata", "source") || "Planning Portal"
+  end
+
+  def planning_portal?
+    source == "Planning Portal"
+  end
+
+  def planx?
+    source == "PlanX"
```
benjamineskola, line 58: I wondered about using `ActiveSupport::StringEnquirer` here but it might be a little too fiddly to be worthwhile.

```ruby
def source
  @source ||= ActiveSupport::StringInquirer.new((request_body.dig("metadata", "source") || "Planning Portal").delete(" ").underscore)
end

delegate :planning_portal?, :plan_x, to: :source
```

engines/bops_submissions/app/services/bops_submissions/enforcement/creation_service.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+
+      def build_enforcement
+        local_authority.enforcements.new(enforcement_params)
+      end
+
+      def enforcement_params
+        {}.tap do |enforcement_params|
+          enforcement_params.merge!(parsed_data)
+        end
```
benjamineskola, line 41: I don't think I quite follow what this does — are we just merging a hash into an empty hash? Couldn't this just be `parsed_data` alone? Or are we predicting that the initial empty hash will need to have some other values in it in future?

benjamineskola (approved):

PR #2410, opened 2025-07-11
Hide conditions on decision notice if decision is to refuse

benjamineskola:
### Description of change

Conditions are not relevant in this case (these are probably the default ones).

### Story Link

https://trello.com/c/6C7Yvwka/885-conditions-appear-on-refusal-decision-notice


benjamineskola:
Perhaps in the `after` block of the state machine transition: `condition_set = nil; save!`

pixeltrix:
> Perhaps in the `after` block of the state machine transition: `condition_set = nil; save!`

Maybe just the conditions in the condition set

app/models/concerns/planning_application_status.rb
```diff
@@ -100,6 +100,8 @@ module PlanningApplicationStatus
         transitions from: %i[awaiting_determination in_committee], to: :determined
 
         after do
+          condition_set.update!(conditions: []) if refused?
```
pixeltrix, line 103: There's pre-commencement conditions and possibly informatives?

pixeltrix (approved):

PR #2413, opened 2025-07-11
Assessment task list redesign

EGiataganas:
### Description of change

Hide and check order of tasks on the LDC-E assessment checklist

### Story Link

https://trello.com/c/KgFK79ve

### DIFF
--- a/app/views/planning_applications/assessment/tasks/index.html.erb
+++ b/app/views/planning_applications/assessment/tasks/index.html.erb
@@ -29,9 +29,15 @@
             }
           ) %>
 
-      <%= render "additional_services" %>
+      <% if @planning_application.pre_application? %>
+        <%= render "additional_services" %>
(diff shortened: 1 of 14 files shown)


app/views/planning_applications/assessment/tasks/index.html.erb
```diff
@@ -29,10 +29,16 @@
...
           ) %>
 
-      <%= render "additional_services" %>
+      <% if @planning_application.pre_application? %>
+        <%= render "additional_services" %>
+      <% end %>
 
       <%= render "assessment_information" %>
 
+      <% if @planning_application.possibly_immune?  && !@planning_application.pre_application? %>
```
benjamineskola, line 38: Not very important but I wonder if we should make `possibly_immune?` return false for pre-applications, instead of remembering to use both checks?

PR #2434, opened 2025-07-24
Further split up test runs

benbaumann95:


benjamineskola (approved):
Minor concern about splitting up alphabetically like this is that our system tests (in particular) aren't spread equally throughout the alphabet: there's a lot bunched under certain letters, particularly because of inconsistencies in how we've named them. When I looked a while back, it was hard to find any combination at all that would split reasonably.

The ideal would be something that splits them up automatically, but I never managed to get `parallel_rspec` to work correctly in CI. (Ruby's minitest has this configured by default but we can't switch everything to minitest just for the sake of that.)

I'd also really love to know what caused the test runs to get 2x slower over the last six months, because grouping them all was previously (more or less) fine.

Another speedup might be to remove `want_pdf` on the groups that don't need it, since that installation takes a couple of minutes each time.

PR #2436, opened 2025-07-24
Add tasks schema and setup

benbaumann95:
### Description of change

Add tasks schema and setup for enforcements workflow


app/models/task.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class Task < ApplicationRecord
+  belongs_to :case_record
+  belongs_to :parent, polymorphic: true
+
+  has_many :tasks, as: :parent, class_name: "Task", dependent: :destroy, autosave: true
```
pixeltrix, line 7: No need for the `class_name` option

app/models/task.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  validates :slug, presence: true, uniqueness: {scope: :case_record_id}
+  validates :name, presence: true, uniqueness: {scope: [:parent_type, :parent_id], message: "must be unique within the same group"}
```
pixeltrix, line 10: I'd avoid the uniqueness validation and just rely on the unique index to blow up - these tasks aren't being created in a form so there's no need for user feedback error messages and it avoids extra queries.

app/models/task.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  before_save :set_slug
+
+  enum status: {
+    not_started: "not_started",
+    in_progress: "in_progress",
+    completed: "completed",
+    optional: "optional"
+  }
```
pixeltrix, line 19: I don't think `optional` is a status - it's a separate flag

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+  end
+
+  def update
+    if @form.update(task_params)
+      redirect_to @form.redirect_url, notice: "Task updated successfully"
+    else
+      render template_for(:edit)
+    end
+  end
```
benjamineskola, line 27: Should this be within a `respond_to {|format| format.html}` block?

app/forms/tasks/base_form.rb
```diff
@@ -0,0 +1,29 @@
...
+    include ActiveModel::Model
+    include Rails.application.routes.url_helpers
+
+    attr_reader :task, :case_record, :enforcement
+    delegate :parent, to: :task
+
+    def initialize(task)
+      @task = task
+      @case_record = @task.case_record
+      @enforcement = @case_record.caseable
```
benjamineskola, line 14: Safe for the time being but can't hurt to do the same check here that we do in the controller:

```suggestion
      @enforcement = @case_record.caseable if case_record.enforcement?
```

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    respond_to do |format|
+      format.html { render template_for(:edit) }
+    end
+  end
+
+  def update
+    respond_to do |format|
+      format.html do
+        if @form.update(task_params)
+          redirect_to @form.redirect_url, notice: "Task updated successfully"
```
pixeltrix, line 25: I assume at some point we'll want to customise this flash message but not a blocker to merging

app/views/enforcements/show.html.erb
```diff
@@ -12,6 +12,12 @@
...
-    end %>
+<%= govuk_task_list(id_prefix: "enforcement") do |task_list| %>
+  <% @enforcement.case_record.tasks.each do |task| %>
+    <%= task_list.with_item(
+          title: task.name,
+          href: task_path(case_type: "enforcement", case_id: @enforcement.case_record.id, slug: task.slug),
+          status: task_status_tag(task)
+        ) %>
+  <% end %>
+<% end %>
```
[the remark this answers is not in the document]
benjamineskola: No, ignore me, I'm thinking of a different page (which does have it).

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+  def set_case_record
+    @case_record = @current_local_authority.case_records.find(params[:case_id])
+  end
+
+  def set_caseable
+    @enforcement = @case_record.caseable if @case_record.enforcement?
+  end
+
+  def find_task
+    slug_segments = params[:slug].to_s.split("/")
```
pixeltrix, line 44: Not a blocker but probably want to move this the model as it's something that should be opaque to the controller

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    @task = @case_record.find_task_by_path!(*slug_segments)
+  rescue ActiveRecord::RecordNotFound
+    raise ActionController::RoutingError, "Task not found"
```
pixeltrix, line 47: No need to rescue and re-raise as `ActiveRecord::RecordNotFound` will generate a 404 error

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    @form = klass.new(@task)
+  end
+
+  def task_params
+    @form.permitted_fields(params)
+  end
+
+  def template_for(action)
+    path = "tasks/#{@task.full_slug}/#{action}"
+    lookup_context.exists?(path) ? path : "tasks/generic/#{action}"
```
pixeltrix, line 62: Again another suggestion for improvement later - we should use the `render_in` method to make the form act like a component:

https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-objects

app/forms/tasks.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  FORM_HANDLERS = {
+    "check-breach-report" => BaseForm,
+    "investigate-and-decide" => BaseForm,
+    "review-recommendation" => BaseForm,
+    "serve-notice-and-monitor-compliance" => BaseForm,
+    "process-an-appeal" => BaseForm,
+
+    "check-report-details" => CheckReportDetailsForm
+  }.freeze
```
pixeltrix, line 18: This feels like it should be defined in the workflow yaml with a method override on the task to return the default BaseForm. Also we need to handle the nesting of tasks as this is just a flat list.

app/forms/tasks/base_form.rb
pixeltrix: Feels like this should be a `before_action` and not littered through the code.

app/services/task_loader.rb
```diff
@@ -0,0 +1,36 @@
...
+  def load!
+    unless File.exist?(path)
+      raise ArgumentError, "Tasks workflow file not found: #{workflow_key}"
+    end
+
+    workflow = YAML.load_file(path)
+    raise ArgumentError, "Workflow not found: #{workflow_key}" if workflow.blank?
+
+    build_tasks_for(case_record, workflow)
+  end
```
pixeltrix, line 21: An optimization for later is to preload all the workflows and create loader instances to improve performance

config/routes/bops.rb
```diff
@@ -351,12 +351,12 @@
...
 
-  resources :enforcements, only: %i[index show] do
-    scope module: :enforcements do
-      resource :report, only: %i[show update] do
-        resource :check_details, only: :show
-      end
-    end
+  resources :enforcements, only: %i[index show]
+
+  scope "/:case_type/:case_id", constraints: {case_type: /enforcement|planning_application/} do
```
pixeltrix, line 356: We don't need the case type in the url as the case id is unique - we could just use `/cases`.

app/forms/tasks/base_form.rb
pixeltrix: I mean a ` before_action` in the tasks controller as that's where it's initially loaded.

PR #2442, opened 2025-07-30
Refactor tasks code to be more flexible

pixeltrix:


app/services/task_loader.rb
```diff
@@ -40,11 +40,12 @@ def load!
...
 
   def build_tasks_for(parent, nodes)
-    Array(nodes).each do |node|
+    Array(nodes).each_with_index do |node, index|
       task = parent.tasks.build(
         name: node["name"],
-        position: node["position"],
-        status: node["status"] || :not_started
+        slug: node["slug"],
+        status: node["status"],
```
benjamineskola, line 47: Can these few lines actually just be `**node`?

pixeltrix: Probably but I want to come back and do a bigger refactor here

benjamineskola (approved):

PR #2449, opened 2025-08-01
Able to edit site description

rebeccaoneill:
### Description of change

### Story Link

https://link-to-issue

### Screenshots

### Decisions [OPTIONAL]

### Known issues [OPTIONAL]

1. issue1
2. issue2

e.g.


app/controllers/tasks_controller.rb
```diff
@@ -22,9 +22,9 @@ def update
     respond_to do |format|
       format.html do
         if @form.update(task_params)
-          redirect_to @form.redirect_url, notice: "Task updated successfully"
+          redirect_to @form.redirect_url, notice: t(".#{@task.slug.tr("-", "_")}.success")
         else
-          render template_for(:edit)
+          render template_for(:edit), notice: t(".#{@task.slug.tr("-", "_")}.failure")
```
benjamineskola, line 27: Might be worth extracting the `task.slug.tr ...` into a method — or otherwise, can we use the hyphenated version in the locales file?

app/forms/tasks/check_description_form.rb
```diff
@@ -0,0 +1,26 @@
...
+      @enforcement = case_record.caseable
+    end
+
+    def permitted_fields(params)
+      params.require(:enforcement).permit(:description)
+    end
+
+    def update(params)
+      enforcement.update!(params)
+      task.update!(status: "completed")
```
benjamineskola, line 19: Should this be in a transaction like the other?

(Or, I wonder if we can put something in the BaseForm class so that the update method is always in a transaction?)

rebeccaoneill: Good catch, we'll handle this in a separate PR

benjamineskola (approved):

PR #2451, opened 2025-08-03
Add checked status to documents

benbaumann95:
### Description of change

Add checked status to documents

### Story Link

https://trello.com/c/f0R0pyD5/753-review-documents

### DIFF
--- a/app/components/status_tags/base_component.rb
+++ b/app/components/status_tags/base_component.rb
@@ -25,7 +25,7 @@ module StatusTags
 
     def colour
       case status.to_sym
-      when :approved, :auto_approved, :supportive, :complies
+      when :approved, :auto_approved, :supportive, :complies, :valid
         "green"
       when :not_started, :new, :review_not_started, :not_consulted, :none, :no_response
         "blue"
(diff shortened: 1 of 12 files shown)


app/components/status_tags/base_component.rb
```diff
@@ -12,6 +12,8 @@ def render?
     end
 
     def call
+      return link_text if status == :checked
```
benjamineskola, line 15: I'm a bit concerned about putting such a general condition in the base class. If we want a special case behaviour that applies to documents only then would it make sense to have a small subclass `StatusTags::DocumentComponent` where `call` is just this line then `super`?

Alternatively, we could possibly define the colour for `:checked` to be transparent, so that we still just call the `govuk_tag` method unconditionally but it doesn't add a background.

benbaumann95: Can't seem to give a transparent colour to the tag component. Will default to blue. Perhaps best to not even call this component then if we are having to render text instead of the component

benbaumann95: I've removed the dependency on being inside .govuk-task-list, .app-task-list__item, or .bops-task-accordion__section-status, as we want this to be consistent behaviour.

benjamineskola (approved):

PR #2490, opened 2025-08-14
Fix constraints toggle

benbaumann95:
### Description of change

Fix constraints toggle.

### Story Link

https://trello.com/c/IkV8UZnq/818-constraints-layers-on-map-in-review-are-not-able-to-be-switched-on-off

### DIFF
--- a/app/javascript/controllers/map_controller.js
+++ b/app/javascript/controllers/map_controller.js
@@ -253,7 +253,7 @@ export default class extends Controller {
       }
     }
 
-    const heading = this.element.querySelector("h3")
+    const heading = this.element.querySelector("#neighbour-heading")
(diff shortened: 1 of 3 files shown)


app/javascript/controllers/map_controller.js
```diff
@@ -252,9 +252,6 @@ export default class extends Controller {
         el.classList.add("govuk-!-display-none")
       }
     }
-
-    const heading = this.element.querySelector("h3")
-    heading.innerHTML = latlng === null ? "All neighbours" : "Neighbours"
```
benjamineskola, line 257: Does this need to be removed?

benbaumann95: Design feedback - shows up in the constraints list too

PR #2496, opened 2025-08-20
Set case record id in migration

benjamineskola:
Because #2495 override the CaseRecord class, this meant the after_initialize hook was no longer setting the id column, which causes validation to fail. Set the ID explicitly here so it's never null.


Rhian (approved):
LGTM!

PR #2499, opened 2025-08-27
Disable all tasks on a case when it has been closed

EGiataganas:
### Description of change

Disable all tasks on a case when it has been closed

### Story Link

https://trello.com/c/NYDdl5Fv


engines/bops_enforcements/app/controllers/bops_enforcements/tasks_controller.rb
```diff
@@ -6,6 +6,7 @@ class TasksController < ApplicationController
 
     before_action :set_enforcement
     before_action :build_form, only: %i[edit update]
+    before_action :ensure_case_is_not_closed, only: %i[show edit update]
```
benbaumann95, line 9: In this case every task is just redirecting back to the enforcement home page when it's closed - is that the expected behaviour or do we want a view only option?

PR #2510, opened 2025-08-28
Add fixes to start enforcement investigation

benbaumann95:


app/models/task.rb
```diff
@@ -34,8 +34,7 @@ def task_for(slug)
   end
 
   def title
-    I18n.t("bops_enforcements.tasks.title.#{slug}")
-  rescue I18n::MissingTranslationData
-    name
+    key = "bops_enforcements.tasks.title.#{slug}"
+    I18n.exists?(key) ? I18n.t(key) : name
```
pixeltrix, line 38: Why not use `:default` option?

``` ruby
I18n.t("bops_enforcements.tasks.title.#{slug}", default: name)
```

pixeltrix (approved):
Approving so you can merge but using `:default` change should be made

EGiataganas (approved):

PR #2516, opened 2025-09-02
Do not show consultee or neighbour response in documents list

benbaumann95:
### Description of change

Documents uploaded with neighbour and consultee responses are showing in the documents list

### Story Link

https://trello.com/c/pvozzZyV/965-documents-uploaded-with-neighbour-and-consultee-responses-are-showing-in-the-documents-list

### DIFF
--- a/app/controllers/planning_applications/neighbour_responses_controller.rb
+++ b/app/controllers/planning_applications/neighbour_responses_controller.rb
@@ -67,7 +67,7 @@ module PlanningApplications
 
     def create_files(response)
       neighbour_response_params[:files].compact_blank.each do |file|
-        @planning_application.documents.create!(file:, neighbour_response: response)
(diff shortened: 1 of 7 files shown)


benjamineskola (approved):
I like that we're moving in the direction of using the `owner` relation consistently rather than having so many different optional relationships.

Do we also want to migrate existing `documents.neighbour_response_id` -> `documents.owner_id` and then remove the `neighbour_response_id` column`?

benbaumann95:
> I like that we're moving in the direction of using the `owner` relation consistently rather than having so many different optional relationships.
> 
> Do we also want to migrate existing `documents.neighbour_response_id` -> `documents.owner_id` and then remove the `neighbour_response_id` column`?

Yes, will do this next

PR #2527, opened 2025-09-04
Add check if the proposal is development

EGiataganas:
### Description of change

Add check if the proposal is development.

### Story Link

https://trello.com/c/mujrakYV

### Screenshots

Screenshot 2025-09-05 at 15 09 29
Screenshot 2025-09-05 at 15 07 45
Screenshot 2025-09-05 at 15 07 18
Screenshot 2025-09-05 at 13 47 10

### DIFF
--- a/app/models/planning_application.rb
+++ b/app/models/planning_application.rb
@@ -129,6 +129,8 @@ class PlanningApplication < ApplicationRecord
     delegate :session_id
   end
 
+  validates :section_55_development, inclusion: [true, false], on: :require_section_55_development
+
   delegate :lodged?, :validated?, :started?, :determined?, :display_status, to: :appeal, allow_nil: true, prefix: true
 
   delegate :code, to: :reporting_type, prefix: true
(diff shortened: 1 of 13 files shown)


app/models/planning_application.rb
```diff
@@ -129,6 +129,10 @@ class WithdrawOrCancelError < RuntimeError; end
     delegate :session_id
   end
 
+  with_options on: :require_development_type do
+    validates :development_type, inclusion: [true, false]
+  end
+
```
benjamineskola, line 135: Since there's only one using this option can't we just do this as a one-liner?

```suggestion
  validates :development_type, inclusion: [true, false], on: :require_development_type
    
```

app/models/planning_application.rb
```diff
@@ -598,7 +602,7 @@ def assessment_submitted?
   end
 
   def no_policy_classes_after_assessment?
-    assessment_complete? && planning_application_policy_classes.none?
+    assessment_complete? && planning_application_policy_classes.none? && true
```
benjamineskola, line 605: What does the `&& true` do here?

benjamineskola (commented):
Just a couple of suggestions.

I'm not a huge fan of the name `development_type` because to me it sounds like that should be a string, not just true/false. But I don't have a better suggestion and it's not very important.

benjamineskola (approved):

PR #2538, opened 2025-09-10
Store a schema for submissions and set it based on the request

benjamineskola:
### Description of change

Rather than inferring the schema on-demand and defaulting to planning portal, add a `schema` column to submissions. Requests from planning portal must set `schema=planning-portal` in the parameters. Will default to ODP schema.

This is distinct from `source`, which is more specific (e.g., `schema: 'odp'` and `source: 'PlanX'`), because in principle ODP submissions could come from other sources (including, in future, maybe Planning Portal too?).

### Story Link

https://trello.com/c/p3zQECqx/1173-infer-type-of-submission-via-api-key

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -51,7 +51,7 @@ class Submission < ApplicationRecord
   end
 
   def source
-    @source ||= request_body.dig("metadata", "source") || "Planning Portal"
+    @source ||= odp? ? request_body.dig("metadata", "source") : "Planning Portal"
   end
 
   def request_body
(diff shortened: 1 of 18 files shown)


app/models/submission.rb
```diff
@@ -65,4 +63,12 @@ def planning_portal?
   def planx?
     source == "PlanX"
   end
+
+  private
+
+  def set_source
+    if (source = request_body.dig("metadata", "source"))
+      self.source = source
+    end
```
pixeltrix, line 72: Digging into the request body runs the risk of triggering an exception - is there a reason why PlanX can't include a query parameter in their request?

Rhian (approved):
LG

PR #2540, opened 2025-09-10
Feature/add notify preview

Rhian:
### Description of change

Add links to Notify settings in Admin and Global Admin to allow admins to send test emails and SMS and preview letters in order to help debug.

I have extracted as much logic as I can into concerns in `bops_core`. This works for the models but only some of the controller methods.

### Story Link

https://trello.com/c/xlxTXBxK/1193-add-feature-to-send-test-emails-sms-letters-to-govuk-notify

### Screenshots

notify

email


pixeltrix:
Looking at the screen shots can you wrap the fields in a two-thirds grid container so they're not massively wide and can you also use a textarea for the body, thanks.

config/routes/bops.rb
```diff
@@ -29,6 +29,16 @@
...
+  scope path: "/admin", module: "bops_admin", as: "bops_admin" do
+    resource :notify, only: [:show, :update], controller: "notify"
+
+    get "notify/send_test", to: "notify#send_test_new", as: :notify_send_test
+    post "notify/send_test", to: "notify#send_test_create"
+
+    get "notify/preview_letter", to: "notify#preview_letter", as: :notify_preview_letter
+    post "notify/preview_letter", to: "notify#generate_preview"
+  end
+
```
benjamineskola, line 41: I don't understand rails routing very well, but can this go in `engines/bops_admin/config/routes.rb` instead?

Rhian: > I don't understand rails routing very well, but can this go in `engines/bops_admin/config/routes.rb` instead?

You're right.. looking at `engines/bops_admin/config/routes.rb`, I can see there is already a `notify` resource in there. I'll move it

engines/bops_admin/config/routes.rb
```diff
@@ -6,7 +6,14 @@
...
   resource :accessibility, only: %i[edit update], controller: "accessibility"
-  resource :notify, only: %i[edit update], controller: "notify"
+
+  resource :notify, only: %i[show edit update], controller: "notify" do
+    get :send_test, action: :send_test_new
+    post :send_test, action: :send_test_create
+
+    get :preview_letter, action: :preview_letter
+    post :preview_letter, action: :generate_preview
+  end
```
pixeltrix, line 16: Ideally I'd have preferred more restful routes like this:

``` ruby
resource :notify, only: %i[edit update], controller: "notify" do
  resource :email, only: %i[new create]
  resource :sms, only: %i[new create]
  resource :letter, only: %i[new create]
end
```

This allows you to create individual controllers for each type of notification and makes it easier for others to read the code in future.

pixeltrix (approved):
A more restful approach would be better but it's fine as it is given the fact it's purely for global/local admins to troubleshoot during account setup.

PR #2548, opened 2025-09-15
Setup task workflow for pre-apps

rebeccaoneill:
### Description of change

Add task setup for pre-apps, including base forms, generic views and yml file.

### Story Link

https://trello.com/c/NIPm7GVa/1221-set-up-pre-apps-to-use-task-model-framework

### Screenshots

image

### Decisions

- At present the top level tasks have been added only to test the functionality and views.
- The proposed url is `/preapps/cases/reference/` this means we have had to handle the case record allocation slightly different to Enforcements where the UUID is being used in the routes
- Simple tests have been added to check the tasks are being created
- Further PRs will be needed to migrate each of the individual pre-app sections

### DIFF
--- a/engines/bops_preapps/app/controllers/bops_preapps/tasks_controller.rb
+++ b/engines/bops_preapps/app/controllers/bops_preapps/tasks_controller.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module BopsPreapps
+  class TasksController < ApplicationController
+    include BopsCore::TasksController
+
+    before_action :set_planning_application, only: %i[show]
+    before_action :build_form, only: %i[edit update]
+    # before_action :ensure_case_is_not_closed, only: %i[show edit update]
+
+    private
+
+    def set_planning_application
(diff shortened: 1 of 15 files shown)


engines/bops_preapps/app/controllers/bops_preapps/tasks_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module BopsPreapps
+  class TasksController < ApplicationController
+    include BopsCore::TasksController
+
+    before_action :set_planning_application, only: %i[show]
+    before_action :build_form, only: %i[edit update]
+    # before_action :ensure_case_is_not_closed, only: %i[show edit update]
```
EGiataganas, line 9: is there any reason of commenting this out?

rebeccaoneill: I missed that, will remove in a later PR

PR #2550, opened 2025-09-15
Prevent publication of closed/withdrawn/returned applications

benjamineskola:
### Description of change

A quick change missed over from #2545 for states which shouldn't be publishable.

In the process I've had to fix some tests which had a validation date set even when the status suggested they weren't validated; I've also taken the opportunity to clean up some duplicated factories that should have been traits on the main factory.

### Story Link

https://trello.com/c/vOQfM9PO/1267-application-which-is-not-valid-has-been-published-this-should-be-prevented


EGiataganas (approved):

Rhian (approved):
LG

PR #2576, opened 2025-09-26
Consultee constraints

benbaumann95:
### Description of change

Add multiple consultees to one constraint

### Story Link

https://trello.com/c/dBe7WUVL/1288-officer-can-add-multiple-consultees-to-one-constraint

### Screenshots

Screenshot 2025-09-26 at 16 23 59

### DIFF
--- a/app/models/consultee.rb
+++ b/app/models/consultee.rb
@@ -62,7 +62,7 @@ class Consultee < ApplicationRecord
     end
   end
 
-  scope :unassigned, lambda {
+  scope :unassigned, -> {
     where.missing(:planning_application_constraint_consultees)
   }
 
--- a/app/views/planning_applications/consultees/index.html.erb
+++ b/app/views/planning_applications/consultees/index.html.erb
@@ -1,15 +1,15 @@
 <% content_for :page_title do %>
-  Select and add consultees - <%= t("page_title") %>
+  Add and assign consultees - <%= t("page_title") %>
 <% end %>
 
 <% add_parent_breadcrumb_link "Home", root_path %>
 <% add_parent_breadcrumb_link "Application", planning_application_path(@planning_application) %>
 <% add_parent_breadcrumb_link "Consultation", planning_application_consultation_path(@planning_application) %>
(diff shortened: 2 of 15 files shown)


app/models/consultee.rb
```diff
@@ -61,7 +62,9 @@ def default_scope
     end
   end
 
-  scope :unassigned, -> { where.not(id: PlanningApplicationConstraint.pluck(:consultee_id)) } # rubocop:disable Rails/PluckInWhere
+  scope :unassigned, lambda {
```
benjamineskola, line 65: Might be preferable to use the arrow syntax consistently:

```suggestion
  scope :unassigned, -> {
```

app/views/planning_applications/consultees/index.html.erb
```diff
@@ -23,15 +23,56 @@
...
+                data: form_data_attributes,
+                class: "consultee-selection__form" do |constraint_form| %>
+            <%= constraint_form.hidden_field :constraint, value: constraint.id %>
+            <div class="govuk-checkboxes__item">
+              <% checkbox_options = {
+                   class: "govuk-checkboxes__input",
+                   id: dom_id(constraint, :consultation_required)
+                 }
+                 checkbox_options[:data] = {action: "change->auto-submit#submit"} if manageable_consultees
+                 checkbox_options[:disabled] = true unless manageable_consultees %>
```
benjamineskola, line 49: Not vital, but I think element attributes can be simplified, because they're left out if the value is nil. Which means this should work:

```suggestion
              <% checkbox_options = {
                   class: "govuk-checkboxes__input",
                   id: dom_id(constraint, :consultation_required),
                   data: manageable_consultees ? {action: "change->auto-submit#submit"} : nil,
                   disabled: !manageable_consultees
                 } %>
```

(or in fact could go inline in the `check_box_tag` call.)

app/views/shared/_consultees_table.html.erb
benjamineskola, line 1: Might be worth using the `govuk_table` helper for this view, which should reduce some of the boilerplate classes.

PR #2586, opened 2025-10-01
Use after_commit to mark consultee as responded only after response is persisted

benbaumann95:


app/components/consultee_summary_component.rb
benjamineskola, line 1: Is this change related?

benjamineskola: It's a bit of a nitpick but it might be good if we rename the methods to `_at` instead of `_on` if they're returning a time instead of just a date. (I think we're mostly consistent with that elsewhere.)

app/jobs/update_consultee_email_status_job.rb
```diff
@@ -30,11 +30,20 @@ def perform(consultee_email)
...
+          last_resp_at = consultee.last_response_at
+
+          attrs = {
+            email_delivered_at: consultee.email_delivered_at || current_time,
+            last_email_delivered_at: current_time
+          }
+
+          if last_resp_at.nil? || last_resp_at < current_time
+            attrs[:status] = :awaiting_response
+          end
```
benjamineskola, line 43: Can last_response_at ever be in the future?

benbaumann95: That's the race condition I believe. I think Tom managed to add a response quick enough that it did seem to cause to revert back to awaiting_response here.

benjamineskola: I'm trying to understand how that could happen. It's possible that if we were taking the date of an email, for example, and trusting that, it would be possible to set the email's sent date to the future and we'd need to account for that.

But in this case it sounds like the dates are both being set from our system and so I don't see how it can be possible for a response to arrive in the future.

(i.e., `last_resp_at == current_time` at the instant the response is created; so `last_resp_at` should by definition be equal to or greater than `current_time`.)

benjamineskola (approved):

PR #2591, opened 2025-10-06
Use a LPA's GOV.UK Notify account where possible

pixeltrix:
We still have to use Southwark's for the BOPS Config emails/SMS messages.

### DIFF
--- a/db/migrate/20251127173911_validate_add_enable_notify_to_local_authorities.rb
+++ b/db/migrate/20251127173911_validate_add_enable_notify_to_local_authorities.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class ValidateAddEnableNotifyToLocalAuthorities < ActiveRecord::Migration[8.0]
+  def up
+    validate_check_constraint :local_authorities, name: "local_authorities_enable_notify_null"
+    change_column_null :local_authorities, :enable_notify, false
+    remove_check_constraint :local_authorities, name: "local_authorities_enable_notify_null"
+  end
+
+  def down
+    add_check_constraint :local_authorities, "enable_notify IS NOT NULL", name: "local_authorities_enable_notify_null", validate: false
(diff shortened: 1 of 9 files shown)


pixeltrix:
> Mainly for my own information: in a few places there's a delegation to local_authority with `allow_nil: true`, but I don't think most of it will work if local_authority is nil. Is `allow_nil` necessary or am I misunderstanding how `delegate` works?

It's so the fall back will work - for example on global administrators there's no local authority so it will return nil and the `||` will fall back to the default api key and template id. The places where it isn't `allow_nil: true` should blow up if there's no local authority because it means something has gone wrong.

db/migrate/20251127173911_validate_add_enable_notify_to_local_authorities.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class ValidateAddEnableNotifyToLocalAuthorities < ActiveRecord::Migration[8.0]
+  def up
+    validate_check_constraint :local_authorities, name: "local_authorities_enable_notify_null"
+    change_column_null :local_authorities, :enable_notify, false
+    remove_check_constraint :local_authorities, name: "local_authorities_enable_notify_null"
+  end
+
+  def down
+    add_check_constraint :local_authorities, "enable_notify IS NOT NULL", name: "local_authorities_enable_notify_null", validate: false
```
benjamineskola: The check constraint is removed in the `up` and added back in the `down` — I understand that, because it's being replaced by the `change_column_null`.

But in the other migration, the check constraint is only being added in the up. Doesn't that mean that if they were both rolled back, the check constraint wouldn't be removed (and then applying it again would fail)?

pixeltrix: At first I thought it would but the column goes away and PostgreSQL removes the check constraint on that column like an index on the column is removed.

PR #2593, opened 2025-10-07
Suggest consultees based on assigned constraints

benbaumann95:
### Description of change

Suggest consultees based on assigned constraints

### Story Link

https://trello.com/c/JH82JIQ6/1286-bops-suggests-consultees-based-on-constraints-that-are-present-in-application


benbaumann95:
@codex review

app/models/planning_application_constraint.rb
```diff
@@ -63,8 +64,39 @@ def planning_data_dataset
...
+      email_address: email,
+      name: contact.name,
+      origin: contact.origin,
+      role: contact.role,
+      organisation: contact.organisation
+    }
+
+    consultation.consultees.find_or_create_by!(attrs) do |consultee|
+      consultee.assign_attributes(attrs)
+    end
```
benjamineskola, line 97: I don't think you need to assign the attributes in the block if they're also passing them in to `find_or_create_by`. The block is only needed if you're setting attributes that aren't used to find the object by.

benbaumann95:
@codex review

app/models/consultee_constraint.rb
```diff
@@ -0,0 +1,15 @@
...
+  belongs_to :constraint
+
+  validates :constraint_id, uniqueness: {scope: :consultee_id}
+  validate :consultee_has_consultee_category
+
+  def consultee_has_consultee_category
+    return if consultee&.consultee?
+
+    errors.add(:consultee, :invalid)
+  end
```
benjamineskola, line 14: Took me a moment to understand this validation until I realised that the `consultee` relationship isn't actually to the `consultee` model. I get why, but it feels confusing.

Perhaps if the validation was called something like `contact_is_consultee` or `contact_category_is_consultee` or something?

app/models/planning_application_constraint.rb
```diff
@@ -63,8 +64,37 @@ def planning_data_dataset
...
+    additions = consultee_ids - existing_consultee_ids
+
+    additions.each do |consultee_id|
+      planning_application_constraint_consultees.find_or_create_by!(consultee_id: consultee_id)
+    end
+  end
+
   private
 
+  def find_or_create_consultee!(contact)
```
benjamineskola, line 84: I sort of feel like this whole method could just be inlined into the `map` block above where it's used., since it's a one-time thing.

PR #2602, opened 2025-10-08
Use main app route to display consultees in report

benbaumann95:


benjamineskola:
If this was able to break without us noticing, does it imply that we have a missing test?

PR #3003, opened 2026-04-13
Handle polygon_geojson read/write for JSONB compatibility

benbaumann95:
### Description of change

The geojson value was passed as a raw string to super, causing double-serialization in the JSONB column. Also guard against undefined EPSG:3857 data in the JS to prevent invalid submissions.

### Story Link

https://trello.com/c/3tk4NN0F/1909-neighbour-geojson-boundary-is-stored-as-a-string-in-the-jsonb-field-and-causes-errors-when-trying-to-re-read

### DIFF
--- a/app/models/consultation.rb
+++ b/app/models/consultation.rb
@@ -316,17 +316,8 @@ class Consultation < ApplicationRecord
     end_date >= Time.zone.today
   end
 
-  def polygon_geojson
-    value = super
-    value.is_a?(String) ? JSON.parse(value) : value
-  end
-
-  def polygon_geojson_before_type_cast
-    polygon_geojson.to_json
-  end
-
   def polygon_geojson=(value)
-    if value.is_a?(String) && value.present?
+    if value.present?
       parsed = JSON.parse(value)
       self.polygon_search = geometry_collection(value) if parsed.present?
       super(parsed)
--- a/app/models/consultation.rb
+++ b/app/models/consultation.rb
@@ -316,9 +316,23 @@ class Consultation < ApplicationRecord
     end_date >= Time.zone.today
   end
 
+  def polygon_geojson
+    value = super
+    value.is_a?(String) ? JSON.parse(value) : value
+  end
+
+  def polygon_geojson_before_type_cast
+    polygon_geojson.to_json
+  end
+
   def polygon_geojson=(value)
-    self.polygon_search = geometry_collection(value) if value.present? && JSON.parse(value).present?
-    super
+    if value.is_a?(String) && value.present?
+      parsed = JSON.parse(value)
+      self.polygon_search = geometry_collection(value) if parsed.present?
+      super(parsed)
+    else
+      super
+    end
   end
 
   def polygon_fill_colour
--- a/db/migrate/20260416090501_fix_double_serialized_polygon_geojson.rb
+++ b/db/migrate/20260416090501_fix_double_serialized_polygon_geojson.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+class FixDoubleSerializedPolygonGeojson < ActiveRecord::Migration[8.0]
+  def change
+    up_only do
+      Consultation.find_each do |consultation|
+        polygon_geojson = consultation.read_attribute(:polygon_geojson)
+        next if polygon_geojson.blank?
+
+        if polygon_geojson.is_a?(String)
+          begin
+            parsed_json = JSON.parse(polygon_geojson)
+            consultation.update_column(:polygon_geojson, parsed_json)
+          rescue JSON::ParserError
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/views/shared/_location_map.html.erb
+++ b/app/views/shared/_location_map.html.erb
@@ -72,7 +72,10 @@
     const map = document.querySelector("my-map");
 
     map.addEventListener("geojsonChange", ({ detail: geoJSON }) => {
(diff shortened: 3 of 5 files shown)


app/models/consultation.rb
```diff
@@ -316,9 +316,23 @@ def publicity_active?
     end_date >= Time.zone.today
   end
 
+  def polygon_geojson
```
benbaumann95: Maybe better to do a data migration for legacy data stored as strings?

benjamineskola: For another recent change we did just wrap it like this (I forget which — one of @pixeltrix's)

But in this case, because it's the format change and there are performance benefits, I do wonder if properly migrating might be worthwhile.

benbaumann95: Hm maybe can just leave for now and then I might do a batch change to update all these relevant fields

app/models/consultation.rb
```diff
@@ -316,9 +316,18 @@ def publicity_active?
...
   def polygon_geojson=(value)
-    self.polygon_search = geometry_collection(value) if value.present? && JSON.parse(value).present?
-    super
+    if value.is_a?(String) && value.present?
+      parsed = JSON.parse(value)
+      self.polygon_search = geometry_collection(value) if parsed.present?
+      super(parsed)
+    else
+      super
+    end
```
benjamineskola: Just double-checking this logic:

If the value being set is _not_ a string (so it's already been deserialised from JSON, I guess?), is it correct that we then don't need to worry about setting `polygon_search`? Will `polygon_search` already have been set in that case / do we need it?

Or, taking a step back even further: if we're handling this all correctly, will we _ever_ be assigning a string here, or should it always be getting an object? I don't know if we can enforce that so maybe it's better to be safe than sorry.

benbaumann95: I think the value will always be a string. Every code path comes from form params via a hidden textarea populated by JSON.stringify() in _location_map.html.erb:77. The else branch only handles nil/blank, where setting polygon_search isn't needed.

No code path assigns a hash directly.
There might be a bigger refactor to move the string parsing to the controller/form layer so the setter only receives hashes.

My vote is let's just get this fix in though.

benjamineskola: If it'll always be a string, then I don't think we need to check whether it's a string. We can just check `value.present?`.

(And I wonder if the `else` branch is actually necessary in that case, because it'll already presumably be blank, right? Or is it possible to overwrite an existing one with a blank one?)

benbaumann95: how about

```
  def polygon_geojson=(value)
    return super unless value.present?

    parsed = JSON.parse(value)
    self.polygon_search = geometry_collection(value) if parsed.present?
    super(parsed)
  end
  ```

File: app/models/consultation.rb [hunk 3]
```diff
@@ -316,9 +316,16 @@ def publicity_active?
     end_date >= Time.zone.today
   end
 
+  def polygon_geojson_before_type_cast
+    polygon_geojson.to_json
+  end
```
benjamineskola: What is this used for?

db/migrate/20260416090501_fix_double_serialized_polygon_geojson.rb
```diff
@@ -0,0 +1,21 @@
...
+      Consultation.find_each do |consultation|
+        polygon_geojson = consultation.read_attribute(:polygon_geojson)
+        next if polygon_geojson.blank?
+
+        if polygon_geojson.is_a?(String)
+          begin
+            parsed_json = JSON.parse(polygon_geojson)
+            consultation.update_column(:polygon_geojson, parsed_json)
+          rescue JSON::ParserError
+            consultation.update_column(:polygon_geojson, nil)
```
benjamineskola: I'm a little concerned that this would end up just deleting data in the case of an error. It might be worth doing a trial run and confirming that all existing entries are able to be parsed, and then committing this code without the nil assignment.

app/models/consultation.rb
benbaumann95: oh yep can remove now

db/migrate/20260416090501_fix_double_serialized_polygon_geojson.rb
benbaumann95: Copied over from a previous migration but yep better to just leave empty.

db/migrate/20260416090501_fix_double_serialized_polygon_geojson.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class FixDoubleSerializedPolygonGeojson < ActiveRecord::Migration[8.0]
+  def change
+    up_only do
+      Consultation.find_each do |consultation|
+        polygon_geojson = consultation.read_attribute(:polygon_geojson)
```
benjamineskola, line 7: I know there are a bunch of different attributes-related methods that I can't really keep track of and have to look up every time. Do we specifically need to use read_attribute here or would `consultation.polygon_geojson` work as well?

Fine either way, just wondering.

benjamineskola (approved):

db/migrate/20260416090501_fix_double_serialized_polygon_geojson.rb
benbaumann95: Yeah `consultation.polygon_geojson` would work fine here now that we've removed the getter override. `read_attribute` is just a bit more defensive since it bypasses any model methods but wouldn't matter here. But no harm.

PR #3004, opened 2026-04-13
Add view neighbour response task

rebeccaoneill:
### Description of change

Add view neighbour responses task.

### Story Link

https://trello.com/c/uFSWHkAT/1861-consultation-view-neighbour-responses

### Screenshots

image

### Decisions 

- Removed the redacted comment field from the edit page as there is a specific flow for redaction
- Once generated the comment is read only so this field is rendered as a `<p>` on edit page
- Kept the redact and publish content in the old pages with conditional sidebar and redirect. The task already had a

### DIFF
--- a/app/controllers/planning_applications/redact_neighbour_responses_controller.rb
+++ b/app/controllers/planning_applications/redact_neighbour_responses_controller.rb
@@ -5,6 +5,7 @@ module PlanningApplications
     before_action :set_planning_application
     before_action :set_consultation
     before_action :set_neighbour_response
+    before_action :show_sidebar
 
     def edit
       respond_to do |format|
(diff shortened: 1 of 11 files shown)


app/controllers/planning_applications/redact_neighbour_responses_controller.rb
```diff
@@ -45,7 +46,19 @@ def set_neighbour_response
...
+      else
+        planning_application_consultation_neighbour_responses_path(@planning_application)
+      end
+    end
+
+    def show_sidebar
+      @show_sidebar = if use_new_sidebar_layout?(@planning_application)
+        if @planning_application.case_record.find_task_by_slug_path("consultees-neighbours-and-publicity/neighbours/view-neighbour-responses")
+          @planning_application.case_record.tasks.find_by(section: "Consultation")
+        end
```
benjamineskola: Do we need this extra inner conditional? I think most of the others just do the `use_new_sidebar_layout?` check.

```suggestion
        @planning_application.case_record.tasks.find_by(section: "Consultation")
```

rebeccaoneill: Good question, I was getting a failing spec before which I applied a fix for (was failing on `find_task_by_slug_path!`) in the `neighbour_responses_path` method for an application type that doesn't have the task. I removed the `!` but then added an additional guard here which I see wasn't needed. Will revert.

benjamineskola (approved):

PR #3020, opened 2026-04-21
Fixes for resending neighbour letters

benjamineskola:
### Description of change

Removing the superfluous save buttons from the send-letters task, and fixing a pre-existing bug where the resend reason would be duplicated across multiple resent letters.

### Story Link

https://trello.com/c/unmtwwgk/1860-consultation-send-letters-to-neighbours

### DIFF
--- a/app/services/letter_sending_service.rb
+++ b/app/services/letter_sending_service.rb
@@ -13,29 +13,28 @@ class LetterSendingService
     @letter_type = letter_type
   end
 
-  def deliver!(neighbour)
+  def deliver!(neighbour, text: letter_content, batch: nil)
     return if resend_reason.nil? && NeighbourLetter.find_by(neighbour:).present? && consultation_letter?
 
-    letter_record = NeighbourLetter.new(neighbour:, text: letter_content, resend_reason:)
+    if resend_reason.present?
+      text = "# Application updated\nThis application has been updated. Reason: #{resend_reason}\n\n" + text
+    end
+
+    letter_record = NeighbourLetter.new(neighbour:, text:, resend_reason:)
 
     ActiveRecord::Base.transaction do
       letter_record.save!
       consultation.start_deadline if consultation_letter?
     end
 
-    if @batch
-      @batch.neighbour_letters << letter_record
+    if batch
+      batch.neighbour_letters << letter_record
+      batch.update!(text:)
     end
-
-    if resend_reason.present?
-      @letter_content.prepend "# Application updated\nThis application has been updated. Reason: #{resend_reason}\n\n"
-    end
-
-    @batch&.update!(text: letter_content)
 
     return if @local_authority.notify_error_status.present?
 
-    personalisation = {message: letter_content, heading:}
+    personalisation = {message: text, heading:}
     personalisation.merge! neighbour.format_address_lines
 
     begin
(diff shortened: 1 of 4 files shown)


app/services/letter_sending_service.rb
```diff
@@ -13,29 +13,28 @@ def initialize(letter_content, consultation:, letter_type:, resend_reason: nil)
     @letter_type = letter_type
   end
 
-  def deliver!(neighbour)
+  def deliver!(neighbour, text: letter_content, batch: nil)
     return if resend_reason.nil? && NeighbourLetter.find_by(neighbour:).present? && consultation_letter?
 
-    letter_record = NeighbourLetter.new(neighbour:, text: letter_content, resend_reason:)
+    if resend_reason.present?
+      text.prepend "# Application updated\nThis application has been updated. Reason: #{resend_reason}\n\n"
```
benbaumann95: Perhaps I'm struggling to follow but my thinking is:

In a batch of 3, neighbour 2 still gets the prefix twice, neighbour 3 three times. I think we could swap to a non mutating assignment 

`text = "# Application updated\nThis application has been updated. Reason: #{resend_reason}\n\n#{text}"`

benbaumann95 (commented):
Might be worth adding a test too to cover this fix?

app/services/letter_sending_service.rb
benjamineskola: It's not the mutation that was the issue, it was the instance variable. A non-mutating assignment would still have been a problem. But now, every invocation will now have a fresh copy of the text.

benbaumann95: When `deliver_batch!` calls `deliver!(neighbour, text: letter_content, batch:)`, text points to the same String object as `@letter_content` - it's not a fresh copy. `text.prepend` then mutates that shared object in place, so the next iteration's `letter_content` has already been modified.

Maybe a test will cover this or can verify with a script in the console.
Perhaps worth another look @pixeltrix in case I'm mistaken

benjamineskola: I've only tested this in the console and it appeared to be fixed then, but writing a test now to confirm anyway.

(I had switched away from `prepend` to assignment and then back, it's possible I missed this.)

Either way I think it's better to avoid using instance variables like this, so I'll leave that change in place too.

spec/system/planning_applications/consulting/send_letters_to_neighbours_spec.rb
```diff
@@ -329,12 +335,15 @@
       fill_in("Resend reason",
         with: "Previous letter mistakenly listed applicant's address as Buckingham Palace.")
 
-      expect_any_instance_of(Notifications::Client).to receive(:send_letter).with(template_id: anything,
-        personalisation: hash_including(message: match_regex(/# Application updated\nThis application has been updated. Reason: Previous letter mistakenly listed applicant's address as Buckingham Palace.\n\n# The Town and Country Planning \(General Permitted Development\) \(England\) Order 2015 Part 1, Class A/))).and_call_original
+      expect_any_instance_of(Notifications::Client).to receive(:send_letter).twice.with(template_id: anything,
+        personalisation: hash_including(message: match_regex(/^# Application updated\nThis application has been updated. Reason: Previous letter mistakenly listed applicant's address as Buckingham Palace.\n\n# The Town and Country Planning \(General Permitted Development\) \(England\) Order 2015 Part 1, Class A/))).and_call_original
```
benjamineskola: Note adding `\A` here (and below): the previous regex would have matched without error because the buggy behaviour repeated the string: `"#Application updated\nReason\n\n#Application updated\nReason\n\n"`.

benbaumann95 (approved):

PR #3022, opened 2026-04-21
Add schema validations

benbaumann95:
### Description of change

Add schema validations.
Schema validations will flag when new fields are added to our endpoints. 

### Story Link

https://trello.com/c/25IWaqZV/1931-add-schema-validations

### DIFF
--- a/engines/bops_api/app/views/bops_api/v2/shared/_application.json.jbuilder
+++ b/engines/bops_api/app/views/bops_api/v2/shared/_application.json.jbuilder
@@ -17,7 +17,6 @@ json.application do
   json.status planning_application.appeal_display_status || planning_application.status
   json.session_id planning_application.session_id
   json.alternative_reference planning_application.alternative_reference
-  json.secret_test_field "this should be caught"
 
   if (consultation = planning_application.consultation)
     json.consultation do
--- a/engines/bops_api/app/views/bops_api/v2/shared/_application.json.jbuilder
+++ b/engines/bops_api/app/views/bops_api/v2/shared/_application.json.jbuilder
@@ -17,6 +17,7 @@ json.application do
   json.status planning_application.appeal_display_status || planning_application.status
   json.session_id planning_application.session_id
   json.alternative_reference planning_application.alternative_reference
+  json.secret_test_field "this should be caught"
 
   if (consultation = planning_application.consultation)
     json.consultation do
--- a/engines/bops_api/schemas/odp/v0.7.7/public/show.json
+++ b/engines/bops_api/schemas/odp/v0.7.7/public/show.json
@@ -221,7 +221,7 @@
                     }
                   },
                   "createdAt": {
-                    "format": "datetime",
+                    "format": "date-time",
                     "type": ["string", "null"]
                   },
                   "applicantDescription": {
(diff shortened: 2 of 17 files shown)


engines/bops_api/app/views/bops_api/v2/shared/_application.json.jbuilder
```diff
@@ -17,6 +17,7 @@ json.application do
   json.status planning_application.appeal_display_status || planning_application.status
   json.session_id planning_application.session_id
   json.alternative_reference planning_application.alternative_reference
+  json.secret_test_field "this should be caught"
```
pixeltrix: Obviously this is going to be removed and wondering whether we should add a test that purposefully stubs an invalid response to a request and sets an expectation on failing.

engines/bops_api/schemas/odp/v0.7.7/public/show.json
```diff
@@ -0,0 +1,253 @@
...
+                        },
+                        "description": {
+                          "type": "string"
+                        }
+                      },
+                      "required": ["value", "description"]
+                    }
+                  },
+                  "createdAt": {
+                    "format": "datetime",
```
pixeltrix: This should be `date-time` and it should be added to all timestamps

engines/bops_api/schemas/odp/v0.7.7/public/show.json
```diff
@@ -0,0 +1,253 @@
...
+                  },
+                  "applicantDescription": {
+                    "type": ["string", "null"]
+                  },
+                  "metadata": {
+                    "type": "object",
+                    "additionalProperties": false,
+                    "properties": {
+                      "byteSize": {
+                        "type": "number"
```
pixeltrix: This should be `integer`

engines/bops_api/spec/swagger_helper.rb
```diff
@@ -110,6 +116,10 @@
 
           CommentsSpecialistResponse: comments_specialist,
 
+          Show: show,
+
+          PublicShow: public_show,
+
```
pixeltrix, line 123: These names appear in the Swagger docs and they're a bit cryptic if you're not a Rails developer - can we have something more descriptive like `PlanningApplication` and `PublicPlanningApplication` ?

engines/bops_api/schemas/odp/v0.7.7/public/show.json
pixeltrix: List of valid format strings:
https://www.learnjsonschema.com/2020-12/format-annotation/format/

engines/bops_api/app/views/bops_api/v2/shared/_application.json.jbuilder
benbaumann95: Add a spec `it "rejects responses with undeclared additional properties"`

benjamineskola:
Rather than specifying `additionalProperties: false` everywhere I think the RSwag `no_additional_properties` option might be useful — we disabled it in #2556 but might be worth revisiting. (It's possible it has too many unintended consequences though.)

benbaumann95:
> Rather than specifying `additionalProperties: false` everywhere I think the RSwag `no_additional_properties` option might be useful — we disabled it in #2556 but might be worth revisiting. (It's possible it has too many unintended consequences though.)

Oh yep, think that'll work too.

pixeltrix (approved):
The spec failure is one that Rebecca has already fixed - rebasing against main should clear it.

PR #3027, opened 2026-04-23
Update check publicity urls and warning

rebeccaoneill:
### Description of change

Update link urls which had erroneously included wrong parent slug. Conditionally update warnings in dates and assignment header to include new task links.

### Story Link

https://trello.com/c/tl49DBcw/1935-update-links-in-check-publicity-task-to-new-consultation-tasks


benjamineskola (approved):
I feel like we should find a better approach than hardcoding a list of slugs in the CaseRecord class but that's a more general problem, not something to fix here.

PR #3028, opened 2026-04-23
Fix document review form

benjamineskola:
### Description of change

The return value of the method is used as the overall success status of the form submission; but because this ends with a conditional it returns false if the `validated` field is `true`.

### Story Link

https://trello.com/c/5pZEXCzQ/1926-when-reviewing-supplied-documents-there-is-an-error-message-shown-after-user-saved-the-task

### DIFF
--- a/app/forms/tasks/review_documents_form.rb
+++ b/app/forms/tasks/review_documents_form.rb
@@ -14,7 +14,7 @@ module Tasks
 
     with_options on: :save_document do
       validates :validated, inclusion: {in: [true, false, "true", "false"], message: "Select whether the document is valid"}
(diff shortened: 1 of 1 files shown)


app/forms/tasks/review_documents_form.rb
```diff
@@ -63,17 +59,19 @@ def save_document
         referenced_in_decision_notice: referenced_in_decision_notice,
         available_to_consultees: available_to_consultees,
         validated: validated,
-        invalidated_document_reason: document_invalid? ? invalidated_document_reason : nil,
+        invalidated_document_reason: invalidated_document_reason.presence,
```
benjamineskola, line 62: Validations mean it's not possible for this to be set if the document is being marked as valid, so no need to check again.

pixeltrix (approved):

PR #3037, opened 2026-04-28
Refactor task_path logic

benjamineskola:
Refactor repeated logic around `task_path`: e.g., we always need an extra parameter (though we always know it'll be the parent application), and in several places need to do the same `pre_application?` check to pick the right helper.

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -81,6 +81,10 @@ class Task < ApplicationRecord
     tasks.first&.first_child
   end
 
+  def url(**args)
+    @url ||= case_record.url_helpers.task_path(case_record.caseable, self, **args)
+  end
+
   private
 
   def raise_not_saved(transition)
--- a/app/components/planning_applications/assessment/assessment_report_component.html.erb
+++ b/app/components/planning_applications/assessment/assessment_report_component.html.erb
@@ -43,7 +43,7 @@
   <% if show_edit_links && current_user.assessor? %>
     <p class="govuk-body-s">
       <% if task.present? %>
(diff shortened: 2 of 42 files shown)


app/models/task.rb
```diff
@@ -81,6 +81,17 @@ def first_child
...
+  def url(**args)
+    @url ||=
+      if case_record.caseable_type == "Enforcement"
+        BopsEnforcements::Engine.routes.url_helpers.task_path(case_record.caseable, slug: self, **args)
+      elsif case_record.caseable.pre_application?
+        BopsPreapps::Engine.routes.url_helpers.task_path(case_record.caseable, slug: self, **args)
+      else
+        Rails.application.routes.url_helpers.task_path(case_record.caseable, slug: self, **args)
+      end
+  end
```
pixeltrix, line 86: Feels like we could implement this on the delegated types and then call it on the case record, e.g.

``` ruby
class CaseRecord < ApplicationRecord
  delegated_type :caseable, types: %w[Enforcement PlanningApplication], dependent: :destroy

  delegate :url_helpers, to: :caseable
end

class Enforcement < ApplicationRecord
  def url_helpers
    BopsEnforcements::Engine.routes.url_helpers  
  end
end

class PlanningApplication < ApplicationRecord
  def url_helpers
    if pre_application?
      BopsPreapps::Engine.routes.url_helpers
    else
      Rails.application.routes.url_helpers
    end
  end
end

def url(**args)
  @url ||= case_record.url_helpers.task_path(case_record.caseable, slug: self, **args)
end
```

benjamineskola: This also makes it easier to use the logic in places where `task_path` is being called with a string slug instead of a task object.

pixeltrix (approved):

PR #3038, opened 2026-04-28
Update design for adding and assigning consultees

pixeltrix:
## Story Link

https://trello.com/c/xjLIsBAY

### Main screen:

southwark bops localhost_3000_planning_applications_26-00732-HAPP_consultees-neighbours-and-publicity_consultees_add-and-assign-consultees

### Consultees for a constraint (consultation required):

southwark bops localhost_3000_planning_applications_26-00732-HAPP_consultees-neighbours-and-publicity_consultees_add-and-assign-consultees_2131_edit

### Consultees for a constraint (consultation not required):
 southwark bops localhost_3000_planning_applications_26-00732-HAPP_consultees-neighbours-and-publicity_consultees_add-and-assign-consultees_2131_edit (1)

### Other consultees:

southwark bops localhost_3000_planning_applications_26-00732-HAPP_consultees-neighbours-and-publicity_consultees_add-and-assign-consultees (1)


app/views/tasks/consultees-neighbours-and-publicity/consultees/add-and-assign-consultees/edit.html.erb
benjamineskola, line 1: If this file didn't exist, it _should_ load the file with the same name from `bops_core`; that would need the bops_core version to have `consultees-neighbours-and-publicity/` in the path, though. Same goes for the other templates like this, I think.

edit: just noticed the path is different for preapp so we'd need the explicit `render` for at least one of them, never mind.

benjamineskola (approved):

app/views/tasks/consultees-neighbours-and-publicity/consultees/add-and-assign-consultees/edit.html.erb
pixeltrix: This doesn't work for this task as the slug path is different between preapps and planning applications so I create a shared partial to load

PR #3039, opened 2026-04-28
Pause API tokens to hold submissions for manual review

benbaumann95:
### Description of change

Pause API tokens to hold submissions for manual review

### Story Link

https://trello.com/c/B9ELI8wV/1724-enable-selective-addition-of-planning-portal-submissions

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -4,6 +4,7 @@ class Submission < ApplicationRecord
   include AASM
 
   belongs_to :local_authority
+  belongs_to :api_user, optional: true
   has_one :case_record, dependent: :nullify
   has_many :documents, dependent: :destroy
 
@@ -21,14 +22,18 @@ class Submission < ApplicationRecord
 
   scope :by_created_at_desc, -> { order(created_at: :desc) }
 
-  aasm column: :status, timestamps: true do
+  aasm column: :status, timestamps: true, whiny_transitions: false do
     state :submitted, initial: true
     state :started
     state :failed
     state :completed
 
     event :start do
-      transitions from: [:failed, :submitted], to: :started
+      transitions from: [:failed, :submitted], to: :started, unless: :api_user_paused?
+
+      after do
+        BopsSubmissions::SubmissionProcessorJob.perform_later(self)
+      end
     end
 
     event :fail do
@@ -64,5 +69,9 @@ class Submission < ApplicationRecord
 
   def odp?
     schema == "odp"
+  end
+
+  def api_user_paused?
+    api_user&.paused?
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_04_21_135314) do
+ActiveRecord::Schema[8.1].define(version: 2026_04_30_154433) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gin"
   enable_extension "pg_catalog.plpgsql"
@@ -72,6 +72,7 @@ ActiveRecord::Schema[8.1].define(version: 2026_04_21_135314) do
     t.datetime "last_used_at"
     t.bigint "local_authority_id"
     t.string "name", null: false
+    t.boolean "paused", default: false, null: false
     t.string "permissions", array: true
     t.string "product_id"
     t.datetime "revoked_at"
@@ -1265,6 +1266,7 @@ ActiveRecord::Schema[8.1].define(version: 2026_04_21_135314) do
   end
 
   create_table "submissions", force: :cascade do |t|
+    t.bigint "api_user_id"
     t.jsonb "application_payload", default: {}, null: false
     t.datetime "completed_at"
     t.datetime "created_at", null: false
@@ -1278,6 +1280,7 @@ ActiveRecord::Schema[8.1].define(version: 2026_04_21_135314) do
     t.datetime "started_at"
     t.string "status", default: "submitted", null: false
     t.datetime "updated_at", null: false
+    t.index ["api_user_id"], name: "ix_submissions_on_api_user_id"
     t.index ["external_uuid"], name: "ix_submissions_on_external_uuid", unique: true
     t.index ["local_authority_id"], name: "ix_submissions_on_local_authority_id"
     t.index ["status"], name: "ix_submissions_on_status"
--- a/engines/bops_admin/app/controllers/bops_admin/submissions_controller.rb
+++ b/engines/bops_admin/app/controllers/bops_admin/submissions_controller.rb
@@ -5,7 +5,7 @@ module BopsAdmin
     self.page_key = "submissions"
 
     before_action :set_submissions, only: %i[index]
-    before_action :set_submission, only: %i[show]
+    before_action :set_submission, only: %i[show update]
     before_action :set_planning_application, only: %i[show]
 
     rescue_from Pagy::OverflowError do
@@ -21,6 +21,14 @@ module BopsAdmin
     def show
       respond_to do |format|
         format.html
+      end
+    end
+
+    def update
+      if @submission.start!
+        redirect_to submission_path(@submission), notice: t(".success")
+      else
+        redirect_to submission_path(@submission), alert: t(".not_eligible")
       end
     end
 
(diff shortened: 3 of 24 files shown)


db/migrate/20260427122615_add_planning_portal_manual_review_enabled_to_local_authorities.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPlanningPortalManualReviewEnabledToLocalAuthorities < ActiveRecord::Migration[8.0]
+  def change
+    add_column :local_authorities, :planning_portal_manual_review_enabled, :boolean, default: false, null: false
```
benbaumann95: I'm open to better names for this..

engines/bops_admin/app/views/bops_admin/submissions/show.html.erb
```diff
@@ -116,12 +124,24 @@
...
+    <% elsif @submission.planning_portal? && @submission.started? %>
+      <p><%= t("bops_admin.submissions.processing.body") %></p>
     <% else %>
       <p>There is no case associated with this submission.</p>
     <% end %>
 
-    <div class="govuk-button-group">
-      <%= back_link %>
-    </div>
+    <% if @submission.planning_portal? && @submission.submitted? %>
```
benjamineskola: Should this also have the `planning_portal_manual_review_enabled?` check? There will be no point showing this to most users, since this is only intended for use when trialling BOPS and not expected to be used permanently.

engines/bops_admin/app/views/bops_admin/submissions/show.html.erb
benjamineskola, line 1: Not vital but I think some of these translations can use the shorter form, e.g., `.process_submission.button` — `bops_admin.submissions` should be implicit. (I think it's only in view components that we need to use the full path.)

engines/bops_admin/app/views/bops_admin/submissions/show.html.erb
benbaumann95: Yep missed this one out.

engines/bops_admin/app/views/bops_admin/submissions/show.html.erb
```diff
@@ -116,12 +124,24 @@
         and case id:
         <%= @submission.case_record.id + "." %>
       </p>
+    <% elsif show_submit_action? %>
+      <p><%= t(".awaiting_review.no_application_yet") %></p>
+    <% elsif @submission.planning_portal? && @submission.started? %>
```
benjamineskola: Doesn't this apply regardless of source?

```suggestion
    <% elsif @submission.started? %>
```

benjamineskola (approved):
Meant to approve: above comment isn't a blocker.

engines/bops_submissions/app/services/bops_submissions/creation_service.rb
```diff
@@ -13,19 +13,21 @@ class CreationService
       Referer
     ].freeze
 
-    def initialize(params:, headers:, local_authority:, schema: nil)
+    def initialize(params:, headers:, local_authority:, api_user: nil, schema: nil)
```
benjamineskola: Might not be significant but I don't think we want a default api_user, since I think it'll break if it's nil. Hard to be sure but I think the DocumentService requires a non-nil user, as it tries to read the `file_downloader` attribute.

This also implies that the PlanxCreationService (which calls the DocumentService) shouldn't default to nil either.

(Maybe insignificant since in practice we're only calling this in one place, and the api user is being set correctly there.)

db/schema.rb
```diff
@@ -1265,6 +1266,7 @@
   end
 
   create_table "submissions", force: :cascade do |t|
+    t.bigint "api_user_id"
```
pixeltrix, line 1269: I was wondering whether we needed a data migration but it's only really relevant for processing after the request so I think it's fine to not do that.

engines/bops_admin/app/views/bops_admin/tokens/_table.html.erb
```diff
@@ -42,6 +45,15 @@
...
           </td>
+          <% unless revoked %>
+            <td class="govuk-table__cell">
+              <% if token.paused? %>
+                <%= govuk_tag(text: "Paused", colour: "yellow") %>
+              <% else %>
+                <%= govuk_tag(text: "Active", colour: "green") %>
+              <% end %>
+            </td>
+          <% end %>
```
benjamineskola: This is immediately followed by `if revoked`, so could this actually go into the `else` block there instead of being an `unless` beforehand?

engines/bops_admin/app/controllers/bops_admin/submissions_controller.rb
```diff
@@ -24,6 +24,16 @@ def show
...
 
+    def submit
+      if @submission.may_start?
+        @submission.start!
+        BopsSubmissions::SubmissionProcessorJob.perform_later(@submission)
+        redirect_to submission_path(@submission), notice: t(".success")
+      else
+        redirect_to submission_path(@submission), alert: t(".not_eligible")
+      end
+    end
```
pixeltrix, line 33: Rather than using a custom action we could just use `update` and use the AASM callbacks to enqueue the job, e.g.

``` ruby
class Submission < ApplicationRecord
  aasm column: :status, timestamps: true, whiny_transitions: false do
    state :submitted, initial: true
    state :started
    state :failed
    state :completed

    event :start do
      transitions from: [:failed, :submitted], to: :started

      after do
        BopsSubmissions::SubmissionProcessorJob.perform_later(self)
      end
    end
  end
end

module BopsAdmin
  class SubmissionsController < ApplicationController
    def update
      if @submission.start!
        redirect_to submission_path(@submission), notice: t(".success")
      else
        redirect_to submission_path(@submission), alert: t(".not_eligible")
      end
    end
  end
end
```

This will obviously need changes elsewhere in `SubmissionProcessorJob` where we're breaking the encapsulation of the state by interrogating it with `may_start?` and then calling `start!` - we should be just sending events/messages to the object following the [Tell, Don't Ask](https://martinfowler.com/bliki/TellDontAsk.html) principle

pixeltrix (approved):
Approving since it's good enough but I think we could make it better - however that would lead to refactoring elsewhere so I'll leave it up to you to make the call.

db/schema.rb
benjamineskola: I wondered whether this should be `null: false`, because as mentioned elsewhere it seems like one is actually required for some parts of the application processing. But that would mean backfilling old submissions, and I'm not sure it's worth it.

engines/bops_admin/app/controllers/bops_admin/submissions_controller.rb
benbaumann95: This feels better, I'll give it a go.

engines/bops_submissions/app/controllers/bops_submissions/v2/submissions_controller.rb
```diff
@@ -8,7 +8,7 @@ class SubmissionsController < AuthenticatedController
 
       def create
         @submission = creation_service.call
-        SubmissionProcessorJob.perform_later(@submission.id, @current_api_user)
+        @submission.start!
```
pixeltrix, line 11: We're not checking the response here

benbaumann95: This is deliberate. If the token's paused the AASM guard returns false and leaves the submission in submitted - we don't want to expose that internal state to Planning Portal. Held submissions show up for the LPA admin instead.

app/models/submission.rb
```diff
@@ -21,14 +22,18 @@ class Submission < ApplicationRecord
...
-  aasm column: :status, timestamps: true do
+  aasm column: :status, timestamps: true, whiny_transitions: false do
     state :submitted, initial: true
     state :started
     state :failed
     state :completed
 
     event :start do
-      transitions from: [:failed, :submitted], to: :started
+      transitions from: [:failed, :submitted], to: :started, unless: :api_user_paused?
```
benjamineskola, line 32: Doesn't this guard mean `start!` can _never_ be used when the ApiUser is paused, even when manually submitting it?

pixeltrix: Hmm, possibly - I assume this is here to prevent the submission being started in the API request but feels like this in the controller would be better:

``` ruby
@submission.start! unless current_api_user.paused?
```

And then remove the guard clause

PR #3040, opened 2026-04-28
Able to add signature to decision notice

rebeccaoneill:
### Description of change

Add optional field to local admin to upload a signature image for decision notice, which is a requirement for some LAs.
The approach also handles removing previously uploaded signatures, and clearing the form before submission.

I have not done any work to remove background of the signature would suggest the user does so on the file itself.

### Story Link

https://trello.com/c/owVMkHND/1905-feedback-support-image-of-hand-drawn-signatures-on-decision-notices-issued-from-bops

### Screenshots

image
image (2)
image

### DIFF
--- a/app/views/planning_applications/_decision_notice.html.erb
+++ b/app/views/planning_applications/_decision_notice.html.erb
@@ -194,7 +194,10 @@
   </h3>
   <p class="govuk-!-margin-bottom-0">
     <%= planning_application.local_authority.signatory_name %><br>
-    <%= planning_application.local_authority.signatory_job_title %>
+    <%= planning_application.local_authority.signatory_job_title %><br>
+    <% if  planning_application.local_authority.signature.present? %>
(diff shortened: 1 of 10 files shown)


app/views/planning_applications/_decision_notice.html.erb
```diff
@@ -194,7 +194,10 @@
   </h3>
   <p class="govuk-!-margin-bottom-0">
     <%= planning_application.local_authority.signatory_name %><br>
-    <%= planning_application.local_authority.signatory_job_title %>
+    <%= planning_application.local_authority.signatory_job_title %><br>
+    <% if  planning_application.local_authority.signature.present? %>
```
benjamineskola, line 198: Tiny nitpick: this `<br>` would end up being visible even if there's no signature.

```suggestion
    <%= planning_application.local_authority.signatory_job_title %>
    <% if  planning_application.local_authority.signature.present? %>
    <br>
```

benjamineskola (approved):

PR #3043, opened 2026-04-29
Keep sidebar visible through consultee response sub-flows

benbaumann95:
### Description of change

Keep sidebar visible through consultee response sub-flows.

Mirror the bops_preapps pattern

### Story Link

https://trello.com/c/Qac2YiSN/1856-consultation-view-consultee-responses


app/components/planning_applications/consultation/response_component.rb
benjamineskola, line 1: Would it be easier to use the `govuk_tag` helper rather than constructing it from a span ourselves?

benbaumann95: Might as well, I'll need to add this for pre-apps too.

benjamineskola (approved):

PR #3051, opened 2026-05-05
Fix manual submit blocked by paused token

benbaumann95:


pixeltrix (approved):

engines/bops_admin/spec/system/submissions_spec.rb
```diff
@@ -293,13 +293,26 @@
     end
 
     context "when the LPA has a paused API token" do
-      before { create(:api_user, :planning_portal, local_authority:, paused: true) }
+      let!(:paused_api_user) { create(:api_user, :planning_portal, local_authority:, paused: true) }
```
benjamineskola, line 296: Nitpick but this doesn't need a `!`, since the variable is referenced in the test.

benjamineskola (approved):

PR #3052, opened 2026-05-05
Use action_required state for assessment tasks returned from review

benjamineskola:
### Description of change

When the assessment tasks are returned from review, we should update their Task record state to `action_required` so that this is flagged in the UI.

Also ensuring Task status uses AASM for state.

### Story Link

https://trello.com/c/VJiOYzNs/1963-implement-status-symbols-for-each-task-in-the-sidebar-of-assessment-to-show-if-they-need-review

### DIFF
--- a/app/controllers/planning_applications/review/assessment_details_controller.rb
+++ b/app/controllers/planning_applications/review/assessment_details_controller.rb
@@ -4,9 +4,12 @@ module PlanningApplications
   module Review
     class AssessmentDetailsController < BaseController
       before_action :set_assessment_detail
+      before_action :set_task, only: :update
 
       def update
         if @assessment_detail.update(review_assessment_details_params)
+          @task.action_required! if @task && return_to_officer?
+
           redirect_to(
             planning_application_review_tasks_path(@planning_application, anchor: "#{@assessment_detail.category}_section"),
             notice: t(".success", category: @assessment_detail.category.humanize.downcase)
@@ -23,6 +26,10 @@ module PlanningApplications
         @assessment_detail = @planning_application.assessment_details.find(assessment_detail_id)
       end
 
+      def set_task
+        @task = @assessment_detail.task
+      end
+
       def assessment_detail_id
         Integer(params[:id])
       end
@@ -33,6 +40,10 @@ module PlanningApplications
           :entry,
           comment_attributes: [:text]
         ).merge(review_status: :complete)
+      end
+
+      def return_to_officer?
+        params.dig(:assessment_detail, :reviewer_verdict) == "rejected"
       end
     end
   end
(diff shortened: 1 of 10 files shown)


app/models/task.rb
```diff
@@ -10,33 +12,30 @@ class Task < ApplicationRecord
...
+  aasm column: :status, enum: true, whiny_persistence: true do
+    state :not_started, initial: true
+    state :in_progress, :completed, :cannot_start_yet, :action_required
 
-  def start
-    completed? || update(status: :in_progress)
-  end
+    event :start do
+      transitions from: :completed, to: :completed
+      transitions from: %i[not_started in_progress], to: :in_progress
```
benbaumann95, line 21: Do we need to add `action_required` here?

app/controllers/planning_applications/review/assessment_details_controller.rb
```diff
@@ -23,6 +26,29 @@ def set_assessment_detail
         @assessment_detail = @planning_application.assessment_details.find(assessment_detail_id)
       end
 
+      def set_task
```
benbaumann95, line 29: This isn't amazing either but could be better? Still think there ought to be a better way for this but anyway..

```
CATEGORY_TO_SLUG = {
  "summary_of_work"      => "check-and-assess/assessment-summaries/summary-of-works",
  "site_description"     => "check-and-assess/assessment-summaries/site-description",
  "additional_evidence"  => "check-and-assess/assessment-summaries/other-considerations",
  "consultation_summary" => "check-and-assess/assessment-summaries/summary-of-consultation",
  "neighbour_summary"    => "check-and-assess/assessment-summaries/summary-of-neighbour-responses",
  "amenity"              => "check-and-assess/assessment-summaries/amenity"
}.freeze

def set_task
  slug = CATEGORY_TO_SLUG[@assessment_detail.category] or return
  @task = @planning_application.case_record.find_task_by_slug_path(slug)
end
```

app/models/task.rb
benjamineskola: I sort of assumed not: once it's `action_required`, I don't think we ever need it to move back to `in_progress`, it only moves forwards again to `completed`.

```mermaid
flowchart TD
    not_started --> in_progress
    in_progress --> complete
    complete --> action_required
    action_required --> complete
```

app/controllers/planning_applications/review/assessment_details_controller.rb
benjamineskola: There isn't really any way of doing this that isn't horrible, IMO, because the fundamental problem is that we need to keep a big list of tasks somewhere and however you arrange that it's bad.

pixeltrix: Is there a difference between planning applications/preapps on these slugs? Wondering whether it's better to move this logic into the `AssessmentDetail` model

pixeltrix (approved):

PR #3054, opened 2026-05-05
Use SiteLocationWGS84.geojson for boundary data in PP submissions

pixeltrix:
Planning Portal is now submitting both OSGB36 and WGS84 spatial references.

https://trello.com/c/tPgxGVEY


engines/bops_submissions/app/services/bops_submissions/parsers/proposal_parser.rb
```diff
@@ -24,33 +24,16 @@ def parse_planning_portal
 
         {
           description: description,
-          boundary_geojson: build_boundary_feature
+          boundary_geojson: params.dig("polygon")
```
pixeltrix, line 27: PP is submitting the boundary wrapped in a `FeatureCollection` and I'm loathe to go digging in there as it's possible that a boundary could have more than one feature.

engines/bops_submissions/app/services/bops_submissions/parsers/proposal_parser.rb
```diff
@@ -24,33 +24,16 @@ def parse_planning_portal
...
         }
       end
 
       def parse_planx
         {
           description: params[:description],
-          boundary_geojson: params.dig("boundary", "site")&.to_json
+          boundary_geojson: params.dig("boundary", "site")
```
pixeltrix, line 34: I removed this as I think we should be returning a hash and not a string here

engines/bops_submissions/app/services/bops_submissions/zip_extraction_service.rb
```diff
@@ -11,6 +11,7 @@ class ZipExtractionService
 
     ATTACHABLE_EXTENSIONS = %w[.pdf .docx .doc .jpg .jpeg .png].freeze
     APPLICATION_FILE = "Application.json"
+    SITE_LOCATION_FILE = "SiteLocationWGS84.geojson"
```
pixeltrix, line 14: I did originally look to pull in both files but there seemed little point other than to bloat the database

engines/bops_submissions/app/services/bops_submissions/zip_extraction_service.rb
```diff
@@ -131,15 +132,17 @@ def process_entry(name, io)
...
+        update_json_file(io, :json_file)
+      elsif name == SITE_LOCATION_FILE
+        update_json_file(io, :site_location_file)
       else
         store_other_file(name)
       end
     end
 
-    def update_json_file(io)
+    def update_json_file(io, attribute)
```
pixeltrix, line 143: Probably could rename this method to `update_submission` or even get rid of it and inline the `update!` calls

pixeltrix (commented):
Not really happy about the messy nature of this but no time to refactor everything.

benjamineskola (approved):

PR #3055, opened 2026-05-06
Fix policy review list

benjamineskola:
### Description of change

Review page was hiding all policies, because the review status hadn't been set by the assessment task.

### Story Link

https://trello.com/c/MGO9h4Ey/1929-improve-assess-against-legislation-and-review-area-page


pixeltrix (approved):
Do we need a regression test?

PR #3073, opened 2026-05-13
Resubmit failed submission

benjamineskola:
### Description of change

Show the resubmit button for failed as well as paused submissions.

### Story Link

https://trello.com/c/ukP3BNwf/1916-able-to-re-submit-if-bops-fails-to-create-the-application-and-its-our-fault

### DIFF
--- a/app/models/assessment_detail.rb
+++ b/app/models/assessment_detail.rb
@@ -107,7 +107,7 @@ class AssessmentDetail < ApplicationRecord
   end
 
   def task
-    planning_application.case_record.find_task_by_slug_path(task_slug)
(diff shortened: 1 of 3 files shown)


app/models/assessment_detail.rb
rebeccaoneill, line 1: is this supposed to be in this PR?

rebeccaoneill (approved):
Left one comment about a stray file included

app/models/assessment_detail.rb
benjamineskola: That was just a quick fix that I was going to include in whichever PR came next since I didn't expect to have another task-related PR yet. In practice I'll merge it in #3072 (it's the same commit, branched from there).

PR #3084, opened 2026-05-19
Fix cancel replacement document request link

benbaumann95:
### Description of change

Fix cancel replacement document request link for pre-applications

### Story Link

https://trello.com/c/v0XiCaIM/1987-unable-to-cancel-validation-request-for-pre-app-in-review-validation-requests-task

### DIFF
--- a/app/helpers/validation_request_helper.rb
+++ b/app/helpers/validation_request_helper.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module ValidationRequestHelper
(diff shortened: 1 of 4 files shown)


benjamineskola (approved):
For assessment details, we have the equivalent code in the model itself. Because the validation requests are all individual subclasses they might be even simpler than assessment details: a `task_slug` method (or public class variable) on each class can just be a simple string method.

Not really necessary for this PR though since it was introduced previously.

PR #3087, opened 2026-05-20
Use local authority scope for planning applications

benjamineskola:
### Description of change

Always use `current_local_authority.planning_applications` when looking up a reference, not `PlanningApplications`.

In particular when using `find_by(reference:)`, because references are not globally unique, and `find_by` will return the first if there is more than one (rather than treating it as an error).

### DIFF
--- a/app/services/planning_applications_creation.rb
+++ b/app/services/planning_applications_creation.rb
@@ -65,7 +65,10 @@ class PlanningApplicationsCreation
   attr_reader(*ATTRIBUTES)
 
   def importer
-    pa = PlanningApplication.find_or_initialize_by(reference: reference)
+    pa = PlanningApplication.find_or_initialize_by(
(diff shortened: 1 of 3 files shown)


pixeltrix (approved):
The service one is a bit sketchy but I think it's okay - not sure how it was working before

benjamineskola:
> The service one is a bit sketchy but I think it's okay - not sure how it was working before

Yeah, I wasn't sure if this one was actually necessary to change. I think it's not actually normally used — the API endpoints have different creation services. If I'm reading correctly it's only for the site history import, which isn't in use.

It might be preferable to rewrite the code so it's also more obviously correct (i.e., treat any reference to query methods on a literal constant `PlanningApplication` as a code smell and only ever go through the LocalAuthority relationships) but this was the minimal change to fix it and it may actually be preferable to remove that code altogether and unify on the submissions API.

PR #3088, opened 2026-05-21
Rubocop check for global queries

benjamineskola:
### Description of change

Warn about 'global' queries for planning applications, i.e., `PlanningApplication.find_by` and similar, rather than `local_authority.planning_applications.find_by`.

Detect the kind of error fixed in #3087.

A couple of details: firstly, there are clearly some false positives. For example I haven't tried to distinguish when the code is manually scoped to the local authority, e.g., `.where(local_authority: local_authority)`. It's easier just to make sure we always use the scope and then consider `PlanningApplication.where` to be a code smell. (This also avoids potential issues with the `kept`/`discarded` scopes being ignored.)

Second, I haven't included an autocorrect. Writing one was more complicated than seemed justified for the small number of cases; most of them were handled by a global find-and-replace, and the remainder could be easily manually corrected.

### DIFF
--- a/app/services/site_notices_creation.rb
+++ b/app/services/site_notices_creation.rb
@@ -9,7 +9,9 @@ class SiteNoticesCreation
     required
   ].freeze
 
-  def initialize(**params)
+  def initialize(local_authority:, **params)
+    @local_authority = local_authority
+
     ATTRIBUTES.each do |attribute|
       value = params[attribute]
       value = value.is_a?(String) ? value.strip : value
@@ -23,6 +25,7 @@ class SiteNoticesCreation
 
   private
 
+  attr_reader :local_authority
   attr_reader(*ATTRIBUTES)
 
   def importer
@@ -39,7 +42,7 @@ class SiteNoticesCreation
   end
 
   def find_index(reference)
-    PlanningApplication
+    local_authority.planning_applications
       .where("previous_references @> ARRAY[?]::varchar[]", reference)
       .pick(:id)
   end
--- a/lib/bops/cop/safety/no_global_queries.rb
+++ b/lib/bops/cop/safety/no_global_queries.rb
@@ -6,7 +6,7 @@ module Bops
       class NoGlobalQueries < RuboCop::Cop::Base
         def on_send(node)
           return unless node.receiver
-          return unless %w[PlanningApplication CaseRecord User ApiUser].include? node.receiver.source
+          return unless %w[PlanningApplication CaseRecord User ApiUser].include?(node.receiver.source) || node.receiver.source.match(/^(::)?(PlanningApplication|CaseRecord|User|ApiUser)\./)
           return unless %i[find_by find_by!
             find_or_create_by find_or_create_by!
             find_or_initialize_by
--- a/lib/bops/cop/safety/no_global_queries.rb
+++ b/lib/bops/cop/safety/no_global_queries.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Bops
+  module Cop
+    module Safety
+      class NoGlobalQueries < RuboCop::Cop::Base
+        def on_send(node)
+          return unless node.receiver
+          return unless node.receiver.source == "PlanningApplication"
+          return unless %i[find_by find_by!
+            find_or_create_by find_or_create_by!
+            find_or_initialize_by
+            find where count first last exists? any? all? none?].include? node.method_name
+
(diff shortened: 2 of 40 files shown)


app/services/site_notices_creation.rb
```diff
@@ -9,7 +9,9 @@ class SiteNoticesCreation
     required
   ].freeze
 
-  def initialize(**params)
+  def initialize(local_authority:, **params)
```
benjamineskola, line 12: It looks like in practice this parameter does get passed, but adding it explicitly here in order that the error message is clear when it's missed out.

lib/bops/cop/safety/no_global_queries.rb
```diff
@@ -0,0 +1,19 @@
...
+module Bops
+  module Cop
+    module Safety
+      class NoGlobalQueries < RuboCop::Cop::Base
+        extend ::RuboCop::Cop::AutoCorrector
+
+        def on_send(node)
+          return unless node.receiver
+          return unless node.receiver.source == "PlanningApplication" || node.receiver.source == "CaseRecord"
+          return unless %i[find_by find_by! find_or_create_by find_or_create_by! find_or_initialize_by where].include? node.method_name
```
benjamineskola: Not sure if there are any other methods we need to consider?

spec/jobs/import_site_notice_job_spec.rb
```diff
@@ -3,10 +3,10 @@
 require "rails_helper"
 
 RSpec.describe ImportSiteHistoryJob do
-  let!(:local_authority) { create(:local_authority, :default) }
+  let!(:local_authority) { create(:local_authority) }
```
benjamineskola, line 6: This was creating a local authority with the wrong name, and so wasn't finding the existing planning applications in the lookup and failing as a result.

lib/bops/cop/safety/no_global_queries.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Bops
+  module Cop
+    module Safety
+      class NoGlobalQueries < RuboCop::Cop::Base
+        extend ::RuboCop::Cop::AutoCorrector
```
benbaumann95: Do we need this?

lib/bops/cop/safety/no_global_queries.rb
```diff
@@ -0,0 +1,19 @@
...
+
+        def on_send(node)
+          return unless node.receiver
+          return unless %w[PlanningApplication CaseRecord User ApiUser].include? node.receiver.source
```
benbaumann95: Not a huge risk but things like `PlanningApplication.all.find_by(...), PlanningApplication.kept.find_by(...), or ::PlanningApplication.find_by(...)` make this bypassable.
Perhaps worth matching against node.receiver.const_name and walking up the receiver chain so chained scopes are caught too.

lib/bops/cop/safety/no_global_queries.rb
```diff
@@ -0,0 +1,19 @@
...
+          return unless %i[find_by find_by! find_or_create_by find_or_create_by! find_or_initialize_by where].include? node.method_name
```
benbaumann95: `find, first, last, exists?, count..`

But let's just include `find` as a priority.

benjamineskola: I think in practice we don't actually use `find` directly, because we never just have an ID, we're always going from a reference. But it costs nothing to add a few more to the check.

Similarly, I think the others are less likely to be used in a way that's a real risk but it can't hurt to check for them too.

lib/bops/cop/safety/no_global_queries.rb
benjamineskola: Thought this was going to involve some annoying recursive logic but it seems like `receiver` is actually just the entire method chain up to that point, which makes it much simpler to just look for any method chain starting with one of those constants.

benjamineskola: There are still ways around this, though.

There's no way for static analysis like rubocop to catch stuff like:

```ruby
if some_condition
  scope = PlanningApplication.some_scope
else
  scope = PlanningApplication.all
end

# ... some time later ...

scope.find_by(...)
```

It would required evaluating the code to determine what every variable equalled at any given time.

But I think this now gets close enough that we can worry much less about it.

benbaumann95 (approved):

PR #3090, opened 2026-05-26
Add local authority conditions

rebeccaoneill:
### Description of change

Add ability to set standard conditions per local authority in local admin. These are them shown as a drop-down when user starts typing in the 'Add conditions' task.

### Story Link

https://trello.com/c/KuFiEVQt/1945-local-admin-allow-standard-conditions-to-be-added

### Screenshots

image

### DIFF
--- a/app/assets/stylesheets/components/_table.scss
+++ b/app/assets/stylesheets/components/_table.scss
@@ -2,6 +2,13 @@
   overflow-wrap: anywhere;
 }
 
+.line-clamp-3 {
+  display: -webkit-box;
+  -webkit-line-clamp: 3;
+  -webkit-box-orient: vertical;
+  overflow: hidden;
+}
+
 .redacted-document-table {
   .govuk-form-group {
     margin-bottom: 0;
--- a/app/controllers/conditions_controller.rb
+++ b/app/controllers/conditions_controller.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+class ConditionsController < AuthenticationController
+  before_action :set_conditions
+
+  rescue_from Pagy::OverflowError do
+    redirect_to conditions_path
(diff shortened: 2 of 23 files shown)


app/assets/stylesheets/components/_table.scss
```diff
@@ -2,6 +2,13 @@
   overflow-wrap: anywhere;
 }
 
+.line-clamp-3 {
+  display: -webkit-box;
+  -webkit-line-clamp: 3;
+  -webkit-box-orient: vertical;
+  overflow: hidden;
+}
```
benjamineskola, line 10: Had to look this one up. We can probably use the non-webkit version of this now:


```suggestion
.line-clamp-3 {
  line-clamp: 3;
  overflow: hidden;
}
```

But that's untested.

rebeccaoneill: Sure thing, will take a look. I wanted to make sure we limit the line length as table is quite small and the descriptions can be quite verbose

benjamineskola: The `-webkit-` prefix means it's vendor-specific, in this case because it's experimental, and the final/standard name will just be `line-clamp`.

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/line-clamp) the `display: -webkit-box` and `-webkit-box-orient: vertical` are only needed on the prefix version of the property, so _if_ we can switch to plain `line-clamp` we can also drop those.

But actually, according to [caniuse.com](https://caniuse.com/?search=line-clamp) it seems like it's only so far supported with the prefix, so I'm mistaken. We could probably add the un-prefixed version for futureproofing but that's not critical.

benjamineskola (approved):

PR #3091, opened 2026-05-27
Clean up old unused consultation pages

rebeccaoneill:
### Description of change
Redirect consultation tasks page to first task; update or remove all tests to match.

Remove unused routes, controllers, and views.

### Story Link

https://trello.com/c/ehKr2ega/1868-remove-the-old-consultation-controllers-once-sidebar-is-deployed-to-production


benbaumann95:
---

##  Blocking bugs — will crash on first request

### 1. `consultations#edit` / `#update` will raise `NoMethodError`
- [consultations_controller.rb:10](app/controllers/planning_applications/consultations_controller.rb#L10)

```ruby
before_action :show_sidebar
```
But the same commit (`Remove task redirect for consultation`) deleted the `show_sidebar` method body from this controller, and it isn't defined on `AuthenticationController` or any ancestor — the four other `def show_sidebar` definitions in the codebase live on unrelated sibling controllers. So **every** request to `/planning_applications/REF/consultation/edit` (still routed) will raise `NoMethodError: undefined method 'show_sidebar'` from the `before_action` dispatch.

Fix: drop the `before_action :show_sidebar` line too.

### 2. `consultees#index` will raise `AbstractController::DoubleRenderError`
- [consultees_controller.rb:28-34](app/controllers/planning_applications/consultees_controller.rb#L28)

```ruby
def index
  respond_to do |format|
    format.html
    redirect_to planning_application_consultation_path(@planning_application)
  end
end
```

The `format.html` (no block) registers a default-render responder, then `redirect_to` performs the response inside the same block — Rails will then invoke the registered HTML render and raise `DoubleRenderError`. The intent appears to be "just redirect to the consultation page"; in that case the action should simply be:

```ruby
def index
  redirect_to planning_application_consultation_path(@planning_application)
end
```

### 3. Three controllers redirect to a route that no longer exists
- [consultees_controller.rb:81](app/controllers/planning_applications/consultees_controller.rb#L81)
- [consultee/responses_controller.rb:49](app/controllers/planning_applications/consultee/responses_controller.rb#L49)
- [consultee/emails_controller.rb:89](app/controllers/planning_applications/consultee/emails_controller.rb#L89)

All three `ensure_consultation_required` guards do:

```ruby
redirect_to edit_planning_application_consultation_requirement_path(@planning_application), ...
```

`resource :consultation_requirement` and the entire `ConsultationRequirementsController` were deleted in this branch ([routes diff](config/routes/bops.rb), commit `Remove consultation requirement pages`), so the helper no longer exists. Any pre-application without `consultation_required?` true will hit one of these guards → `NoMethodError`. Either remove the guards (if the requirement is now set elsewhere) or point them at the new task / page that manages this.

### 4. `check_publicity` assessment-details views reference 8 deleted helpers
- [assessment_details/check_publicity/edit.html.erb](app/views/planning_applications/assessment/assessment_details/check_publicity/edit.html.erb)
- [assessment_details/check_publicity/new.html.erb](app/views/planning_applications/assessment/assessment_details/check_publicity/new.html.erb)
- [assessment_details/check_publicity/show.html.erb](app/views/planning_applications/assessment/assessment_details/check_publicity/show.html.erb)

These templates are still rendered by [assessment_details/edit.html.erb:6-8](app/views/planning_applications/assessment/assessment_details/edit.html.erb#L6) (and `new.html.erb` / `show.html.erb`), which dispatch on `@category`. The route `resources :assessment_details, except: %i[destroy index]` is unchanged, so `?category=check_publicity` is still reachable by URL — and the templates call `planning_application_press_notice_path`, `new_planning_application_press_notice_path`, `planning_application_press_notice_confirmation_path`, `planning_application_press_notice_confirmation_requests_path`, `planning_application_site_notice_path`, `new_planning_application_site_notice_path`, `edit_planning_application_site_notice_path`, and `planning_application_site_notice_confirmation_requests_path` — **all of which were removed in this branch**.

I couldn't find a live UI link to category=`check_publicity` (the new design uses the publicity task views), so these may be dead in practice. Either delete the `check_publicity/*` templates and remove the `when "check_publicity"` arms from the dispatcher, or restore enough state to keep them functional.

---

## Dead code 

- [config/routes/bops.rb:175](config/routes/bops.rb#L175) — `# resources :responses, controller: "consultee/responses", ...` is commented out rather than deleted. Just remove the line.
- Orphaned views with no controller/route behind them (delete with the rest):
  - [press_notices/_form.html.erb](app/views/planning_applications/press_notices/_form.html.erb)
  - [press_notices/confirmations/show.html.erb](app/views/planning_applications/press_notices/confirmations/show.html.erb), [edit.html.erb](app/views/planning_applications/press_notices/confirmations/edit.html.erb)
  - [consultation_requirements/_form.html.erb](app/views/planning_applications/consultation_requirements/_form.html.erb)

---

## Nits / confirm

- [planning_application_helper.rb:63-69](app/helpers/planning_application_helper.rb#L63) — `confirm_site_notice_path` / `confirm_press_notice_path` lost their fallback and now unconditionally return `task_path(planning_application, slug: CONFIRM_*_NOTICE_SLUG)`. If a planning application doesn't have that task (e.g. an application type whose workflow doesn't include it), the helper silently emits a URL that 404s on click. Worth confirming the task is guaranteed wherever these helpers are called.
- [authentication_controller.rb:24](app/controllers/authentication_controller.rb#L24) — `# to be removed after merge of assessment removal`. Fine, just track it so it actually gets removed.
- The new singular-resource pattern `resource :consultation, only: %i[edit update] do; get "/", to: redirect(...); end` works because `update` already generates the `planning_application_consultation_path` helper. Worth a one-line comment noting this if you want it to be obvious.

---

## Recommendation

Fix the four main issues before merging — items 1–3 are immediate crashes on still-routed pages, and item 4 is a one-direct-URL-away crash. Items 1, 2, 5 are one-line changes. Item 3 is a small behaviour decision (what to do when consultation isn't required). Item 4 is most cleanly resolved by deleting the check_publicity assessment-details templates along with the rest.

app/controllers/authentication_controller.rb
```diff
@@ -21,6 +21,7 @@ def ensure_planning_application_is_not_preapp
     end
   end
 
+  # to be removed after merge of assessment removal
```
benjamineskola: This should be safe to remove now.

benjamineskola (approved):
I think a few of those consultee controllers could probably be deleted entirely (EmailsController, ResponsesController, maybe others), but I think it'd make sense to merge this as-is, doing the bare minimum to enable the sidebar and redirect to the first task, and then we can clean up anything else bit by bit later.

PR #3094, opened 2026-05-29
Avoid unnecessary use of `route_for`

benjamineskola:
### Description of change

`route_for` probably isn't needed if the first parameter is a literal symbol, since in that case there's an equivalent helper method (e.g. `route_for(:task, ...)` is the same as `task_url(...)` or `task_path(...)`).

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -82,7 +82,7 @@ class Task < ApplicationRecord
   end
 
   def url(**args)
-    @url ||= case_record.url_helpers.task_path(case_record.caseable, self, **args)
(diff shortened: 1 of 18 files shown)


app/models/task.rb
```diff
@@ -81,7 +81,7 @@ def first_child
   end
 
   def url(**args)
-    @url ||= case_record.url_helpers.task_path(case_record.caseable, self, **args)
+    case_record.url_helpers.task_path(case_record.caseable, self, **args)
```
benjamineskola, line 85: Memoising here fails if the method is called more than once with different options, which sometimes happens when there are multiple links on the same page each with different parameters.

PR #3152, opened 2026-08-25
Replace erb_lint with herb

benjamineskola:
### Description of change

`erb_lint` is deprecated, use herb instead.

Linter fixes applied if they're either automatic or couldn't reasonably be ignored (e.g. parser errors) or were otherwise straightforward to fix.

### Story Link

https://trello.com/c/wMAKRkTA/19-migrate-from-erblint-to-herb-lint


app/views/application/_header.html.erb
```diff
@@ -28,8 +28,7 @@
                   t(".log_out"),
                   main_app.destroy_user_session_path,
                   method: :delete,
-                  class: "govuk-header__link",
-                  rel: "nofollow"
```
benjamineskola, line 32: nb. this still seems to include `rel="nofollow"` in the output, so it isn't necessary to add explicitly — I assume the helper is doing it? and then herb flags it up as a duplicate because of that.

pixeltrix (approved):

PR #3166, opened 2026-09-08
Scope application reference number to current year

benjamineskola:
### Description of change

This will mean it resets every year rather than continuing to increment. e.g., a 26-00100-HAPP, 27-00100-???, etc; the previous behaviour was that although the year would increment, it would not reset the application number.

The application number is padded to 5 digits but not truncated, so 6-digit numbers will work fine, but we'd rather limit them nevertheless.

### Known issues [OPTIONAL]

Calculating the number like this also has a race condition.


db/migrate/20260916134834_add_year_to_planning_application.rb
```diff
@@ -0,0 +1,15 @@
...
+class AddYearToPlanningApplication < ActiveRecord::Migration[8.1]
+  def change
+    add_column :planning_applications, :year, :integer
+
+    up_only do
+      safety_assured { execute("UPDATE planning_applications SET year = date_part('year', created_at) WHERE year IS NULL") }
+    end
+
+    safety_assured do
+      change_column_null :planning_applications, :year, false
```
benjamineskola, line 12: strong_migrations wants this added in a separate migration, but I'm not sure if that makes sense: it says 'on an existing column', but this is a column being added in the same migration. I'd set it in the `add_column`, but that's invalid because it'll be null until it's set (a couple of lines later).

Ideally I'd like to set `null: false` but not have that constraint checked until the end of the migration.

app/models/planning_application.rb
```diff
@@ -1164,9 +1165,13 @@ def reset_published_at
...
   end
 
+  def set_year
+    self.year = created_at.year
+  end
+
   def set_reference
     self.reference = [
-      Date.current.strftime("%y"),
+      year.to_s[-2, 2],
```
pixeltrix: Can we change the following:

``` ruby
def application_number
  self[:application_number].to_s.rjust(5, "0")
end
```

to

``` ruby
def application_number
  format("%05d", super)
end
```

Then add this method:

``` ruby
def year
  format("%02d", super % 100)
end
```

and finally change this to:

``` ruby
def set_reference
  self.reference = [
    year,
    application_number,
    application_type_suffix
  ].join("-")
```

db/migrate/20260916134834_add_year_to_planning_application.rb
pixeltrix: To do it properly you need to avoid locking the whole table in a transaction, which is what your `UPDATE` query is doing.

The steps are:

1. Add column first and add a fallback for `year` to calculate dynamically
2. Deploy all the way to production
3. Backfill the missing data using a non-blocking method, e.g. `disable_ddl_transaction!` and `find_each`
4. Use `add_check_constraint` to add a constraint to the `year` column but set `validate: false`
5. In one transaction use `validate_check_constraint` to ensure all rows are `NOT NULL`, then use `change_column_null` to set the column option and finally use `remove_check_constraint` to delete the constraint added in the previous migration.

Of course this is probably overkill but it's what I'd do on Petitions - you're probably fine to do it in a single migration and merge/deploy first thing.

To make the year method work with a null column you'd change it to this:

``` ruby
def year
  format("%02d", (super || created_at.year) % 100)
end
```

benjamineskola: Ah, I used the raw-SQL approach because I thought that might be faster than looping over every record.

I suppose it is faster but the expense of having to be locked for the entire query, whereas a loop can be a lot of individual queries that don't need locking. (At the expense of extra migration steps.)

db/migrate/20260916135319_make_planning_application_number_unique_by_year.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class MakePlanningApplicationNumberUniqueByYear < ActiveRecord::Migration[8.1]
+  disable_ddl_transaction!
+
+  def change
+    remove_index :planning_applications, %i[application_number local_authority_id], unique: true
+    add_index :planning_applications, %i[year application_number local_authority_id], unique: true, algorithm: :concurrently
```
benjamineskola, line 8: @pixeltrix It's just occurred to me that this constraint is going to be much more complicated if we want to move the `local_authority` relationship into `CaseRecord` instead.

We might not get around to that for planning applications in the immediate future, but it's already true for enforcements (which I'm adding this to next).

So it looks like it might be necessary to use the additional table approach to fix the race condition, and then we don't have to worry about a unique index?

